PR #6, opened 2020-07-01
Infinite Tracing - PartitionedBlockingCollection

jifeingo:
* Creates a new `PartitionedBlockingCollection<T>` that allows the unification of multiple blocking collections into a single structure.  The rationale is that multiple smaller collections use less memory that a single larger collection.
* Modifies Infinite Tracing `SpanEventAggregatorInfiniteTracing` and the `SpanDataStreamingService` to use this collection.


nrcventura:
jenkins retest this

PR #10, opened 2020-07-01
Infinite Tracing - Manage Response Streams with a single thread

jifeingo:
Manages Response Streams on a single thread
* Creates `ResponseStreamInfo` class 
    * Hold the Stream and the Task for obtaining the next response
    * Exposes GetAwaiter which will use existing task or create new one as needed
    * Exposes IsInvalid flag identifying failure conditions
* Modifies `GrpcWrapper` to return both Request and Response Streams as `out` parameters to the CreateChannel call
* Modifies `DataStreamingService` to manage response streams

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
@@ -707,7 +707,7 @@ namespace NewRelic.Agent.Core.DataTransport
 
                 _hasAnyStreamStarted = true;
 
-                _responseStreamsDic[consumerId] = new ResponseStreamWrapper<TResponse>(consumerId, responseStream, streamCancellationTokenSource.Token);
+                _responseStreamsDic[consumerId] = new ResponseStreamWrapper<TResponse>(consumerId, responseStream, serviceAndStreamCancellationTokenSource.Token);
 
                 while (!serviceCancellationToken.IsCancellationRequested && _grpcWrapper.IsConnected)
                 {
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
@@ -22,7 +22,7 @@ namespace NewRelic.Agent.Core.DataTransport
         public readonly int ConsumerID;
         
         private bool _isInvalid = false;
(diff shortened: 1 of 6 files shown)


src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
```diff
@@ -425,10 +549,13 @@ public void Shutdown(bool withRestart)
             _grpcWrapper.Shutdown();
         }
 
-        private IClientStreamWriter<TRequest> GetRequestStreamWithRetry(int consumerId, CancellationToken cancellationToken)
+        private bool GetRequestStreamWithRetry(int consumerId, CancellationToken cancellationToken, out IClientStreamWriter<TRequest> requestStream, out IAsyncStreamReader<TResponse> responseStream)
```
nrcventura, line 552: Should we just return `AsyncDuplexStreamingCall<TRequest, TResponse>` instead of the two out parameters instead? `AsyncDuplexStreamingCall` implements IDisposable so it may also be useful so that we can call `Dispose` on that object too.

src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
```diff
@@ -3,20 +3,72 @@
...
+        {
+            var success = false;
+
+            try
+            {
+                success = await _responseStream.MoveNext(_streamCancellationToken);
+            }
+            catch (Exception ex)
+            {
+                success = false;
```
nrcventura, line 50: This will likely prevent us from knowing what error occurred when dealing with the response stream. I don't know if this will hurt us in the long run, but so far these errors have lead to us either catching problems with our usage of the grpc library or when the response stream is cancelled. Are we intentionally suppressing this level of detail?

src/Agent/NewRelic/Agent/Core/DataTransport/SpanStreamingService.cs
```diff
@@ -25,9 +25,15 @@ public SpanStreamingService(IGrpcWrapper<Span, RecordStatus> grpcWrapper, IDelay
 
         protected override void HandleServerResponse(RecordStatus responseModel, int consumerId)
         {
+            if (responseModel.MessagesSeen == 0)
```
nrcventura, line 28: Is the log message that this if block prevents from happening still useful when MessagesSeen is 0? If we still have it, it could serve as a heartbeat to let us know that things are still connected. With that being said, an application with a steady load should almost always see a non-zero value here.

PR #11, opened 2020-07-02
Fixing SpanEventsEnabled and DistributedTracingEnabled checks for Infinite Tracing

nrcventura:
Updates the Infinite Tracing version of the span event aggregator to respect the SpanEventsEnabled and DistributedTracingEnabled configuration settings, and adds some caching to the SpanEventsEnabled configuration setting.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Aggregators/SpanEventAggregatorInfiniteTracing.cs
+++ b/src/Agent/NewRelic/Agent/Core/Aggregators/SpanEventAggregatorInfiniteTracing.cs
@@ -127,7 +127,9 @@ namespace NewRelic.Agent.Core.Aggregators
             Log.Info($"SpanEventAggregatorInfiniteTracing: Configuration Setting - Queue Partitions - {_configuration.InfiniteTracingPartitionCountSpans}");
         }
 
-        public bool IsServiceEnabled => _spanStreamingService.IsServiceEnabled;
+        public bool IsServiceEnabled => _configuration.SpanEventsEnabled
+            && _configuration.DistributedTracingEnabled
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/Aggregators/SpanEventAggregatorInfiniteTracing.cs
```diff
@@ -127,7 +127,9 @@ private void LogConfiguration()
             Log.Info($"SpanEventAggregatorInfiniteTracing: Configuration Setting - Queue Partitions - {_configuration.InfiniteTracingPartitionCountSpans}");
         }
 
-        public bool IsServiceEnabled => _spanStreamingService.IsServiceEnabled;
+        public bool IsServiceEnabled => _configuration.SpanEventsEnabled
+            && _configuration.DistributedTracingEnabled
```
jifeingo, line 131: Part of me wanted to see this in the Streaming Service making it the single point for this.   Not overly concerned about this b/c the streaming service doesn't really do anything until aggregator tells it to start consuming the collection.

I suppose it's good.

nrcventura: The main reason I put it in SpanEventAggregatorInfiniteTracing was for symmetry with the SpanEventAggregator class. I also considered moving it into the service, but the logic for that service was already complicated.

nrcventura: Jason and I talked about this and decided to keep the logic here for now.

PR #12, opened 2020-07-02
Infinite Tracing - Batch Endpoint

jifeingo:
* Modifies DataStreamingService to distinguish between `TRequest` and `TRequestBatch`
    * Invokes gRPCWrapper of type `TRequestBatch` instead of `TRequest`
    * Abstract Method to translate from x items of type `TRequest` to a `TRequestBatch`
    * Implements method to pull multiple items from the collection as they are available up to the batch size.

* Tests
    * Config Setting for Partition Count
    * Config Setting for Batch Size
    * Batching is correct


src/Agent/NewRelic/Agent/Core/Aggregators/SpanEventAggregatorInfiniteTracing.cs
```diff
@@ -76,6 +76,8 @@ private void AgentConnected(AgentConnectedEvent _)
 
                 if(IsServiceEnabled)
                 {
+
+                    Log.Info($"SpanEventAggregatorInfiniteTracing: Configuration is invalid - Infinite Tracing will NOT be enabled.");
```
nrcventura, line 80: We may want to change this log message to something else, because `IsServiceEnabled` could be false even if the configuration is valid (such as when DT is disabled).

jifeingo: hmmmm, I am open here.  But felt that write log messages if they aren't using the feature.

src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
```diff
@@ -692,6 +702,48 @@ private void ExecuteConsumer(PartitionedBlockingCollection<TRequest> collection)
...
+            {
+                return false;
+            }
+
+            if(cancellationToken.IsCancellationRequested)
+            {
+                return false;
+            }
+
+            items = new List<TRequest>();
```
nrcventura, line 718: Should we preallocate this list to the `maxBatchSize` to avoid possible extra allocations and copying?

src/Agent/NewRelic/Agent/Core/Aggregators/SpanEventAggregatorInfiniteTracing.cs
nrcventura: I just looked at the rest of the code in this method and you are right, that log message is appropriate.

src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
jifeingo: I don't think we should until we know more about how many items are being pulled using this model.   Our queue size is 500, but if things are working well, each of the workers should be pulling far fewer items than that.  So, if we preallocated, we may cause more memory consumption.

* Let's pull a supportability metric to see how many items we are (on average) pulling for each worker.

PR #20, opened 2020-07-08
Infinite Tracing - Supportability Metric for Queue Size

jifeingo:
* Added supportability metric for Infinite Tracing Span Queue Size (number of items waiting to send)
* Associated Tests

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
@@ -92,6 +92,13 @@ namespace NewRelic.Agent.Core.AgentHealth
             TrySend(metric);
         }
 
+
+        public void ReportSupportabilityGaugeMetric(string metricName, float value)
+        {
+            var metric = _metricBuilder.TryBuildSupportabilityGaugeMetric(metricName, value);
+            TrySend(metric);
+        }
+
         public void ReportDotnetVersion()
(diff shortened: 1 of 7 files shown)


src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -92,6 +92,13 @@ public void ReportSupportabilitySummaryMetric(string metricName, float totalSize
             TrySend(metric);
         }
 
+
+        public void ReportSupportabilityGaugeMetric(string metricName, float value)
```
nrcventura, line 96: This doesn't need to be public yet so we should probably keep this as private for now and also remove it from the interface.

src/Agent/NewRelic/Agent/Core/AgentHealth/IAgentHealthReporter.cs
```diff
@@ -131,6 +131,7 @@ public interface IAgentHealthReporter : IOutOfBandMetricSource
 
         void ReportSupportabilityCountMetric(string metricName, long count = 1);
         void ReportSupportabilitySummaryMetric(string metricName, float totalSize, int countSamples, float minValue, float maxValue);
+        void ReportSupportabilityGaugeMetric(string metricName, float value);
```
nrcventura, line 134: This probably doesn't need to be on the interface yet.

PR #23, opened 2020-07-08
Update vs version for core proj

lspangler584:
Allows build with VS 2019 (VisualStudioVersion 16), by indicating where to find the TextTransform.exe dll dependencies.
This may prevent building with older versions of VS.


lspangler584:
Unintended changes. Closing.

PR #26, opened 2020-07-09
Update memory physical metric source for linux

nrcventura:
Fixes issue where applications running on Linux were either reporting no physical memory usage or using VmData to report the physical memory usage of the application. The agent now uses VmRSS through a call to [`Process.WorkingSet64`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.workingset64) to report physical memory usage. See the [dotnet runtime discussion](https://github.com/dotnet/runtime/issues/28990) and the [proc man pages](https://man7.org/linux/man-pages/man5/proc.5.html) for more details about this change.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
@@ -91,6 +91,13 @@ namespace NewRelic.Agent.Core
             AgentInfo = GetAgentInfo();
         }
 
+        //This method and delegate is here to allow dependency injecting the IsWindows logic via a typed delegate
+        public delegate bool IsWindowsDelegate();
+        public static bool GetIsWindows()
+        {
+            return IsWindows;
+        }
+
         private static string GetAgentVersion()
         {
             try
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### New Features
 ### Fixes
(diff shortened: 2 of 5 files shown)


src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
```diff
@@ -91,6 +91,13 @@ static AgentInstallConfiguration()
             AgentInfo = GetAgentInfo();
         }
 
+        //This method and delegate is here to allow dependency injecting the IsWindows logic via a typed delegate
+        public delegate bool IsWindowsDelegate();
```
alanwest, line 95: Does it make sense to invest in solving this a different way? This pattern of using a delegate seems to be a shortcut to avoid the work of making this a non-static class and introducing an interface. If we run with the delegate here for now, I'd like us to consider coming back soon and doing this differently.

Another thought that may be less effort than refactoring this entire class at the moment would be to just hang `IsWindows` off of `IConfiguration`.

nrcventura: `IConfiguration` doesn't feel like a good workaround either, because it's not really a configuration. I considered making an interface and then only exposing IsWindows from it, and then using the same pattern that we are using for the Process class. 

However, all of this seems overkill for what we are trying to do. I really just want to dependency inject a boolean value into the .ctor, but I did not see a good way to handle that with the dependency injection containers that we are using. Part of me would rather just take the time to remove our reliance on these dependency injection containers because of all of the subtle and not-so-subtle we've run into with things not getting created or started when we expected them to. If we do that, then we could simply pass in the `AgentInstallConfiguration.IsWindows` value directly into the .ctor.

alanwest: True, I've always been a fan of eliminating the overly complex auto dependency injection we're using. Also, I do agree trying to deal with this at the moment is overkill for the task at hand. My primary concern is the propagating this pattern of using a delegate to other things further complicating something that should be far easier.

I'm ok moving forward with this solution for now. Perhaps we should open an issue for the removal of the DI framework. It has led to a stupid amount of issues over the years.

PR #28, opened 2020-07-10
SetILFunctionBody from GetReJITParameters

alanwest:
This PR resurrects the initial implementation of ReJIT support. We did not ship this implementation due to issues with our ability to debug managed agent code - the debugger would crash. We worked around this crash by calling `SetILFunctionBody` from `ReJITCompilationStarted` which is technically not supposed to be supported. This "workaround" has led to a number of issues over time:
* Our ability to support tiered compilation
* An incompatibility with the CLR Instrumentation Engine. See https://github.com/microsoft/CLRInstrumentationEngine/issues/226
* An incompatibility with .NET 5

`SetILFunctionBody` is now invoked upon `GetReJITParameters` as it is supposed to be.

I have performed a quick validation and have found that the debugger crash no longer occurs - perhaps there was a bug in the debugger which has since been resolved.

Further validation needs to be done to see if this PR also resolves the following issues:
- [x] Agent code can be debugged when attached to a .NET Framework application
- [x] Agent code can be debugged when attached to a .NET Core application
- [x] ReJIT integration test suite passes against .NET 5 (separate [branch](https://github.com/newrelic/newrelic-dotnet-agent/compare/alanwest/net50-rejit-test) w/ test app updated to .NET 5)
- [x] Test compatibility with CLR Instrumentation Engine RawProfilerHook
  * Here's a test harness I created containing a build of the CLR IE provided to me by Microsoft: https://github.com/alanwest/clrie-newrelic


src/Agent/NewRelic/Profiler/Profiler/Function.h
```diff
@@ -112,13 +114,13 @@ namespace NewRelic { namespace Profiler
             }
 
             uint32_t tracerFlags = 0;
-            bool hasTransactionOrTraceAttribute = false;
```
nrcventura, line 115: We might want to keep this variable name as `hasTransactionOrTraceAttribute`.

src/Agent/NewRelic/Profiler/Profiler/Function.h
```diff
@@ -202,16 +204,13 @@ namespace NewRelic { namespace Profiler
...
         // Check for the API Transaction and Trace attributes.
-        static bool HasTransactionOrTraceAttribute(CComPtr<IMetaDataImport2> metaDataImport, mdToken metaDataToken, uint32_t& tracerFlags)
+        static bool HasTraceAttribute(CComPtr<IMetaDataImport2> metaDataImport, mdToken metaDataToken, uint32_t& tracerFlags)
         {
             const BYTE *pVal = NULL;
             ULONG cbVal = 0;
 
             HRESULT result = metaDataImport->GetCustomAttributeByName(metaDataToken, _X("NewRelic.Api.Agent.TransactionAttribute"), (const void**)&pVal, &cbVal);
-            // It is not safe to use the SUCCEEDED() macro to check result in this case. Contrary to the documentation, GetCustomAttributeByName sometimes
-            // returns S_FALSE (1), and we don't want to consider that a successful result in this case.
```
nrcventura, line 212: I found this comment to be useful. We might want to keep this comment.

src/Agent/NewRelic/Profiler/Profiler/Function.h
```diff
@@ -310,15 +312,20 @@ namespace NewRelic { namespace Profiler
             // create the token resolver that will be used to get strings from tokens
             _tokenResolver.reset(new CorTokenResolver(_metaDataImport));
 
-            _moduleName = ToStdWString(moduleName.get());
+            // REVIEW : Why do we get the function bytes upfront before we know that we want to instrument the function? SMD
+            // get the bytes that make up this method
```
nrcventura, line 316: Should we keep this comment?

src/Agent/NewRelic/Profiler/Profiler/Function.h
```diff
@@ -310,15 +312,20 @@ namespace NewRelic { namespace Profiler
...
+            _method->assign(method, method + methodSize);
             _signature->assign(signature, signature + signatureSize);
 
+            _functionHeaderInfo = CreateFunctionHeaderInfo(_method);
+
             const BYTE *pVal = NULL;
             ULONG cbVal = 0;
 
             HRESULT attributeResult = _metaDataImport->GetCustomAttributeByName(_metaDataToken, _X("System.Runtime.CompilerServices.AsyncStateMachineAttribute"), (const void**)&pVal, &cbVal);
-            // It is not safe for us to use the SUCCEEDED macro on the result returned from GetCustomAttributeByName
```
nrcventura, line 321: Maybe we should keep this comment too?

src/Agent/NewRelic/Profiler/Profiler/FunctionResolver.h
```diff
@@ -0,0 +1,200 @@
...
+    // info to our map and request another reJIT.  When that triggers, GetReJITParameters will be called once again 
+    // and we will have the function id.
+    // 
+    // Note that the bytecode for a given generic class / method *should* be the same regardless of the specific generic
+    // type as noted by Broman (https://blogs.msdn.microsoft.com/davbr/2011/10/12/rejit-a-how-to-guide/).  For this reason
+    // we only map a single function id to a module/method.  This is okay because when we look up the class and method details
+    // for a function id we use the apis that don't return the type specifics of the generic method.  That means that while 
+    // multiple function ids may map to a single generic method, the AgentShim will see invocations for all of those function ids 
+    // use a single function id.  For our purposes, at least for the current agent functionality we support, this is fine.
+    class FunctionResolver
```
nrcventura, line 58: We should probably go through this class and determine if any of the method parameters should be made into `const &` params.

nrcventura: @alanwest and I talked about this and the parameters are basically ints so we're not concerned with any extra allocations caused by the parameters not being references.

PR #38, opened 2020-07-16
Introduces configurations for expected error

vuqtran88:
- Introduces expectedClasses, expectedMessages and expectedStatusCodes configurations for expected error.
- Adds unit tests.


src/Agent/NewRelic/Agent/Core/Config/Configuration.cs
```diff
@@ -1,8 +1,3 @@
-/*
-* Copyright 2020 New Relic Corporation. All rights reserved.
-* SPDX-License-Identifier: Apache-2.0
-*/
-
```
nrcventura, line 5: We need to restore this comment.

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1243,6 +1243,76 @@
                 </xs:complexType>
               </xs:element>
 
+              <xs:element name="expectedClasses">
+                <xs:complexType>
+                  <xs:annotation>
+                    <xs:documentation>
+                      To prevent specific exceptions reporting to New Relic, list them here.
```
nrcventura, line 1250: Is this the same language that other agents are using?

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1243,6 +1243,76 @@
...
+                          <xs:documentation>
+                              To prevent specific error codes from affecting error rate, list them here.
+                          </xs:documentation>
+                      </xs:annotation>
+  
+                      <xs:sequence minOccurs="0" maxOccurs="unbounded">
+                          <xs:element name="code" type="xs:float">
+                              <xs:annotation>
+                                  <xs:documentation>
+                                      You may use standard integral HTTP errorcodes, such as just 401, or may use Microsoft FullStatusCodes, with decimal points, such as 401.4 or 403.18
```
nrcventura, line 1307: For consistency we should end this sentence with a punctuation.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,39 @@ public TimeSpan ErrorEventsHarvestCycle
         public virtual uint ErrorsMaximumPerPeriod { get { return 20; } }
         public virtual IEnumerable<string> ExceptionsToIgnore { get { return ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorErrorsToIgnore, _localConfiguration.errorCollector.ignoreErrors.exception); } }
 
+        public virtual IEnumerable<string> ExpectedClasses { get { return ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedClasses, _localConfiguration.errorCollector.expectedClasses.errorClass); } }
```
nrcventura, line 1141: Should we consider combining this property with the ExpectedMessages property so that our error handling logic only needs to look at one of the them? This is just something to think about, and we'll see which approach we prefer when we work on using these config values.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,39 @@ public TimeSpan ErrorEventsHarvestCycle
...
+
+        public virtual IEnumerable<KeyValuePair<string, IEnumerable<string>>> ExpectedMessages
```
nrcventura, line 1143: Should this be a ReadOnlyDictionary<string, IEnumerable<string>> to make lookups easier?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,39 @@ public TimeSpan ErrorEventsHarvestCycle
...
+
+                foreach(var errorClass in _localConfiguration.errorCollector.expectedMessages)
+                {
+                    var messages = errorClass.message;
+                    if (messages != null)
+                    {
+                        expectedMessages.Add(new KeyValuePair<string, IEnumerable<string>>(errorClass.name, messages));
+                    }
+                }
+                return ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedMessages, expectedMessages);
```
nrcventura, line 1157: We should probably cache the results of this computation so that we don't re-evaluate this logic with every error on every transaction.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,39 @@ public TimeSpan ErrorEventsHarvestCycle
...
+        public virtual IEnumerable<string> ExpectedStatusCodes
+        {
+            get
+            {
+                var localStatusCodesToIgnore = new List<string>();
+                foreach (var localCode in _localConfiguration.errorCollector.expectedStatusCodes.code)
+                {
+                    localStatusCodesToIgnore.Add(localCode.ToString(CultureInfo.InvariantCulture));
+                }
+                return ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedStatusCodes, localStatusCodesToIgnore);
```
nrcventura, line 1170: We should cache this logic too.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
nrcventura: We should also cache this logic.

vuqtran88: I think we might be able to combine them together.

src/Agent/NewRelic/Agent/Core/Config/Configuration.cs
```diff
@@ -3,12 +3,6 @@
...
-//  </auto-generated>
-// ------------------------------------------------------------------------------
```
nrcventura, line 11: We should probably restore these comments.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,61 @@ public TimeSpan ErrorEventsHarvestCycle
...
+
+            var expectedMessages =  ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedMessages, localExpectedMessages).ToDictionary();
+
+            var expectedClasses = ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedClasses, _localConfiguration.errorCollector.expectedClasses.errorClass);
+
+            foreach (var className in expectedClasses)
+            {
+                if (expectedClasses.Contains(className))
+                {
+                    expectedMessages[className] = null;
```
nrcventura, line 1162: Let's think about whether it's better to use `null` or an empty enumerable in this situation. I'm leaning towards an empty enumerable so that it's one less thing to deal with if we try to take advantage of the nullable reference types in the future.

nrcventura: We should also log a warning message in this case.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,61 @@ public TimeSpan ErrorEventsHarvestCycle
...
+                }
+            }
+
+            var expectedMessages =  ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedMessages, localExpectedMessages).ToDictionary();
+
+            var expectedClasses = ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedClasses, _localConfiguration.errorCollector.expectedClasses.errorClass);
+
+            foreach (var className in expectedClasses)
+            {
+                if (expectedClasses.Contains(className))
```
nrcventura, line 1160: I think that you intended this check to be `expectedMessages.ContainsKey` instead of `expectedClasses.Contains`.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,61 @@ public TimeSpan ErrorEventsHarvestCycle
...
+                {
+                    expectedMessages[className] = null;
+                }
```
nrcventura, line 1163: You'll also need an else clause here in case the class is only defined in `expectedClasses`.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,61 @@ public TimeSpan ErrorEventsHarvestCycle
...
+            }
+
+            return new ReadOnlyDictionary<string, IEnumerable<string>>(expectedMessages);
+        }
+
+        private ReadOnlyDictionary<string, IEnumerable<string>> _expectedMessages;
+        public virtual ReadOnlyDictionary<string, IEnumerable<string>> ExpectedMessages
```
nrcventura, line 1170: Should we call this something like `ExpectedErrors` or `ExpectedErrorMessages` instead so that it's clear that we are dealing with an expected error configuration?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1138,6 +1138,61 @@ public TimeSpan ErrorEventsHarvestCycle
...
+                if (_expectedMessages == null)
+                {
+                    _expectedMessages = ParseExpectedErrorConfigurations();
+                }
+
+                return _expectedMessages;
+            }
+        }
+
+        public virtual IEnumerable<string> ExpectedStatusCodes
```
nrcventura, line 1183: Similarly to the `ExpectedMessages` config item, we should consider changing this name so that it's clear that it's used for an expected error configuration.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
vuqtran88: I swear it was `expectedMessages.ContainsKey`

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
vuqtran88: Ultimately the `ExpectedErrorInfo` will also contains expected error status code, but I like address this in a different PR so that I can make sure it is the right thing to do.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
nrcventura: This will likely be addressed in an upcoming PR.

PR #44, opened 2020-07-17
Error collector enabled setting updates

nrcventura:
With this PR we should now match the behavior of other New Relic agents with how the error collector enabled setting is used.

When the error collector is disabled the following should occur:
- Error metrics should not be generated
- Error Events and Error Traces should not be generated
- Transaction Events should not have error attributes generated
- Span events should not have error attributes generated
- "Frustrated" Apdex metrics should still be generated if a transaction has an error, because errors usually result in "frustrated" users
- apdex perf zone attribute should still be generated


nrcventura:
Because of the apdex requirements, we still need to "collect" errors even though we don't report them. This is also why the ErrorCollectorEnabled check is in so many places.

PR #55, opened 2020-07-22
Expected error classes and error messages implementation.

vuqtran88:
## Summary

Expected error classes and error messages implementation.

## Details
- When an error is expected, Frustrated Apdex metric should not be created automatically for the error transaction.
- When an error is expected, increments count for "ErrorsExpected/all" metric, but MUST NOT increment any of the Errors/* metrics.
- When an error is expected, still increments count for "Supportability/Events/TransactionError/Seen" and "Supportability/Events/TransactionError/Sent".
- When an error is expected, traced errors and error events, span event error attributes are still generated.
- When an error is expected, error trace should include `error.expected` attribute, but this will be addressed in a different PR.
- If an error is set to be ignored and expected, it will be ignored.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Errors/ErrorData.cs
+++ b/src/Agent/NewRelic/Agent/Core/Errors/ErrorData.cs
@@ -14,6 +14,7 @@ namespace NewRelic.Agent.Core.Errors
         public DateTime NoticedAt { get; }
         public string Path { get; set; }
         public ReadOnlyDictionary<string, object> CustomAttributes { get; }
+        public bool IsExpected { get;}
 
         public const string StripExceptionMessagesMessage = "Message removed by New Relic based on your currently enabled security settings.";
 
@@ -26,5 +27,12 @@ namespace NewRelic.Agent.Core.Errors
             Path = null;
             CustomAttributes = customAttributes;
         }
+
+        public ErrorData(string errorMessage, string errorTypeName, string stackTrace, DateTime noticedAt, ReadOnlyDictionary<string, object> customAttributes, bool isExpected)
+        : this(errorMessage, errorTypeName, stackTrace, noticedAt, customAttributes)
+        {
+            IsExpected = isExpected;
+        }
+
     }
 }
--- a/src/Agent/NewRelic/Agent/Core/Errors/ErrorService.cs
+++ b/src/Agent/NewRelic/Agent/Core/Errors/ErrorService.cs
@@ -54,6 +54,50 @@ namespace NewRelic.Agent.Core.Errors
             return ShouldIgnoreError(GetFormattedHttpStatusCode(statusCode, subStatusCode));
         }
 
+        private bool IsErrorExpected(Exception exception)
+        {
+            var isExpected = IsExceptionExpected(exception);
+
+            if (!isExpected)
+            {
+                var baseException = exception.GetBaseException();
+                return IsExceptionExpected(baseException);
+            }
+            return isExpected;
+        }
+
+        private bool IsExceptionExpected(Exception exception)
+        {
+            var exceptionTypeName = GetFriendlyExceptionTypeName(exception);
+            var expectedErrorInfo = _configurationService.Configuration.ExpectedErrorsInfo;
+            if (expectedErrorInfo.ContainsKey(exceptionTypeName))
+            {
+                if(expectedErrorInfo[exceptionTypeName] != Enumerable.Empty<string>())
+                {
+                    return ContainsSubstring(expectedErrorInfo[exceptionTypeName], exception.Message) ? true : false;
+                }
+                else
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private bool ContainsSubstring(IEnumerable<string> list, string subString)
+        {
+            foreach (var item in list)
+            {
+                if (item.Contains(subString))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
         public ErrorData FromException(Exception exception)
         {
             return FromExceptionInternal(exception, _emptyCustomAttributes);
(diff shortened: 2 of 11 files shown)


src/Agent/NewRelic/Agent/Core/Errors/ErrorData.cs
```diff
@@ -26,5 +27,12 @@ public ErrorData(string errorMessage, string errorTypeName, string stackTrace, D
             Path = null;
             CustomAttributes = customAttributes;
         }
+
+        public ErrorData(string errorMessage, string errorTypeName, string stackTrace, DateTime noticedAt, ReadOnlyDictionary<string, object> customAttributes, bool isExpected)
```
alanwest, line 31: Let's come back to this and go with one constructor. I understand you added this one so as to avoid refactoring all the tests as part of this PR. Can you simplify this and do the refactor of the tests in a follow up PR?

src/Agent/NewRelic/Agent/Core/Errors/ErrorService.cs
```diff
@@ -54,6 +54,50 @@ public bool ShouldIgnoreHttpStatusCode(int statusCode, int? subStatusCode)
             return ShouldIgnoreError(GetFormattedHttpStatusCode(statusCode, subStatusCode));
         }
 
+        private bool IsErrorExpected(Exception exception)
```
alanwest, line 57: nit: let's move private methods below the public contract.

Ideally we should have something like stylecop prevent this.

alanwest:
What happens when in the course of a transaction:

1. Expected error occurs, then
2. Ignored error occurs

IIRC - I think we do not send up an TransactionError

---

What happens when in the course of a transaction:

1. Expected error occurs, then
2. Unexpected error occurs

It feels to me that the unexpected error should take precedence, though I'm curious what other agents do.

src/Agent/NewRelic/Agent/Core/Errors/ErrorService.cs
```diff
@@ -54,6 +54,50 @@ public bool ShouldIgnoreHttpStatusCode(int statusCode, int? subStatusCode)
...
+                return IsExceptionExpected(baseException);
+            }
+            return isExpected;
+        }
+
+        private bool IsExceptionExpected(Exception exception)
+        {
+            var exceptionTypeName = GetFriendlyExceptionTypeName(exception);
+            var expectedErrorInfo = _configurationService.Configuration.ExpectedErrorsInfo;
+            if (expectedErrorInfo.ContainsKey(exceptionTypeName))
```
alanwest, line 73:
```csharp
var expectedErrorConfig = _configurationService.Configuration.ExpectedErrorsInfo;
if (expectedErrorConfig.TryGet(exceptionTypeName, out var expectedErrorInfo))
{
    ...
}
```

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Configuration/IConfiguration.cs
```diff
@@ -91,6 +91,7 @@ public interface IConfiguration
         IEnumerable<string> ExpectedErrorClassesForAgentSettings { get; }
         IDictionary<string, IEnumerable<string>> ExpectedErrorMessagesForAgentSettings { get; }
         string ExpectedErrorStatusCodesForAgentSettings { get; }
+        IDictionary<string, IEnumerable<string>> ExpectedErrorsInfo { get; }
```
alanwest, line 94: What do you think about renaming this to `ExpectedErrors` or `ExpectedErrorConfiguration`

alanwest, line 94: I feel like there may be an opportunity to encapsulate all of this into a single `ExpectedErrorsConfiguration` class. The class could maintain the notion of:
* The original config specified by the customer for reporting to `agent_settings`.
* The expected class/message config, and
* The expected status code config

This is just an idea to explore in a follow up PR, maybe when you do the work on dealing with expected status codes.

tests/Agent/UnitTests/Core.UnitTest/Errors/ErrorServiceTests.cs
```diff
@@ -148,5 +150,117 @@ private void SetupConfiguration(List<string> exceptionsToIgnore, List<float> sta
...
+                    config.errorCollector.expectedMessages.Add(x);
+                }
+            }
+
+            EventBus<ConfigurationDeserializedEvent>.Publish(new ConfigurationDeserializedEvent(config));
+        }
+
+        [TestCase(true)]
+        [TestCase(false)]
+        public void FromException_MarkErrorDataAsExpected_ExpectedErrorClasses(bool hasExpectedError)
```
alanwest, line 182: nit: move the tests above the private helper methods

tests/Agent/UnitTests/Core.UnitTest/Errors/ErrorServiceTests.cs
```diff
@@ -148,5 +150,117 @@ private void SetupConfiguration(List<string> exceptionsToIgnore, List<float> sta
             config.stripExceptionMessages.enabled = stripExceptionMessages;
             EventBus<ConfigurationDeserializedEvent>.Publish(new ConfigurationDeserializedEvent(config));
         }
+
+        private void SetupConfiguration(List<string> errorClassesToBeExpected, IEnumerable<KeyValuePair<string, IEnumerable<string>>> errorMessagesToBeExpected, bool errorCollectorEnabled)
```
alanwest, line 154: Does it make sense to combine the two `SetupConfiguration` methods?

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Configuration/IConfiguration.cs
vuqtran88: I make sense to encapsulate all of these configurations into a single one, but I would like keep it that way for now as I want to see how the expected status code implementation work will impact this.

src/Agent/NewRelic/Agent/Core/Errors/ErrorData.cs
vuqtran88: Yes, I will combine these constructors in a different PR.

vuqtran88:
jenkins retest this

Issue #60, opened 2020-07-24
GC Performance Counters for Windows - Unable to obtain Perf Counter Instance Name in IIS until GC has occurred


jifeingo:
When collecting Garbage Collection performance counters for Windows, the agent, in part, uses the application's Instance Name to disambiguate different instances of the same running application. For example, two instances of an application named "Test" may appear as Test and Test#1.

To identify which Instance name belongs to the currently executing process, a reverse lookup is made to the performance counter Process ID and is matched to the Process ID (PID) of the current process. With this instance name, Garbage Collection Performance Counters are created and a reference to each of them is maintained in memory.

**Description**
In IIS Apps, unless a garbage collection has occurred, the reverse lookup does not work.  The PID reported

```
/// <summary>
/// The instance name of a performance counter is a combination of the process Name and an index id to 
/// uniquely identify it when multiple of the same process are running.  For example "TestApp #1" pid=82, "TestApp #2" pid=134
/// are two instances of the same executable, TestApp.  In order to determine if an instance matches a process,
/// an additional performance counter "Process ID" must be used which returns the PID of the process.
/// </summary>
/// <param name="perfCategoryName"></param>
/// <param name="processName"></param>
/// <param name="pid"></param>
/// <returns>The instance name that will be used to collect performance counter data for the process</returns>
private string GetInstanceNameForProcessAndCategory(string perfCategoryName, string processName, int pid)
{
	var performanceCategory = _createPerformanceCounterCategory(perfCategoryName);

	var instanceNames = performanceCategory.GetInstanceNames()
		.Where(x => x.StartsWith(processName, StringComparison.OrdinalIgnoreCase));

	foreach (var instanceName in instanceNames)
	{
		try
		{
			using (var pcPid = _createPerformanceCounter(perfCategoryName, ProcessIdCounterName, instanceName))
			{
				if ((int)pcPid.NextValue() == pid)
				{
					return instanceName;
				}
			}
		}
		catch (Exception ex)
		{
			//Log a message here and continue because this instance may not be relevant to the process that we are looking for.
			Log.Finest($"Error determining if instance '{instanceName}' is process '{processName}'({pid}). Error: {ex}");
		}
	}

        //At this point a match was not found. The caller should handle accordingly
	return null;
}
```

In the above code which performs this reverse lookup.  Unless a GC Has occurred, `(int)pcPid.NextValue()` reports 0 as the PID.

**Expected Behavior**
[...]

Issue #56, opened 2020-07-23
GC Performance Counters for Windows improperly assumes Instance Name Can't change 


jifeingo:
When collecting Garbage Collection performance counters for Windows, the agent, in part, uses the application's Instance Name to disambiguate different instances of the same running application.  For example, two instances of an application named "Test" may appear as `Test` and `Test#1`.

To identify which Instance name belongs to the currently executing process, a reverse lookup is made to the performance counter `Process ID` and is matched to the Process ID (PID) of the current process.  With this instance name, Garbage Collection Performance Counters are created and a reference to each of them is maintained in memory.

On regular intervals, these performance counter references are used to read the current values.

**Description**
* The Instance Name is dynamic and can change throughout the life cycle of an application.  If the instance name changes, two  one of two things can occur:  (A)  Most likely, if the instance name no longer exists, an error will occur; (B) The perf counters for the wrong instance of the application will be read.

**Expected Behavior**
* Agent always collects the perf counters for the current app and does not ever fail when multiple instances of the same app are running.

**Steps to Reproduce**
* Start two instances of this program.
    * You should get `TestProgram` and `TestProgram #1` as instance names.  
* Terminate instance called `TestProgram` 
* Notice `TestProgram#1` will experience an error because the performance counter can no longer be fetched.

```
class Program
{
	static void Main(string[] args)
	{
		while (true)
		{
			var perfCounters = CreateCounters();
			perfCounters = perfCounters.ToList();
			while (true)
			{
				ReadValues(perfCounters);
				Thread.Sleep(TimeSpan.FromSeconds(2));
			}
		}
	}

	private static readonly string[] _perfCounterNames = new[]
	{
		"Gen 0 heap size",			
		"Promoted Memory from Gen 0",
		"Gen 1 heap size",			
		"Promoted Memory from Gen 1",
		"Gen 2 heap size",			
		"Large Object Heap size",	
		"# GC Handles",				
		"# Induced GC",				
		"% Time in GC",				
		"# Gen 0 Collections",		
		"# Gen 1 Collections",		
		"# Gen 2 Collections"		
	};

	private const string perfCatName = ".NET CLR Memory";

	private static IEnumerable<PerformanceCounter> CreateCounters()
	{
		var process = Process.GetCurrentProcess();
		var processName = process.ProcessName;
		var pid = process.Id;
		var perfCat = new PerformanceCounterCategory(perfCatName);

[...]

PR #62, opened 2020-07-24
Garbage Collection Performance Counters Fixes

jifeingo:
## Summary
Redesigns the GC Sampler for Windows to account for missing performance counters and performance counter instance name changes.

## Details
* **Performance Counter Instance Name can change during the lifetime of the application**
    * Removes caching of this value and pulls it once for every sampling cycle.
    * When Performance Counter Instance Name Changes, recreate performance counters and dispose the old ones.
    * Because a race condition can occur where the instance name changes between the time it is captured and the time the values are collected, the GC Sampler is more resilient when failures occur.  Allows up to 5 consecutive failures before shutting down.

* **Reverse lookup using PID to obtain Performance Counter Instance Name will not work until a GC has been reported**
    * Attempts to obtain the instance name every sample cycle (1x/min).
    * If not able to obtain a perf counter instance name, abort the sampling cycle, but do not shut down sampler.  Allows a 60-second delay before trying again.
    * Exception:  If an instance name had been previously captured and now the name is not available, shut down the sampler, something else is going wrong.

* Modifies Logging to be more consistent.
* Ensures that all possible `UnauthorizedAccessExceptions` will shut down the sampler.

## Links
Issue #56 - GC Perf Counters assumes Instance Name is static.
Issue #60 - GC Perf Counters Unable to obtain Instance Name until 1st GC has occurred.

## Proposed Release Notes
Fixes
* **Garbage Collection Performance Metrics for Windows** <br/>
Fixes an issue where Garbage Collection Performance Metrics may not be reported for Windows Applications

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Samplers/GcSampler.cs
+++ b/src/Agent/NewRelic/Agent/Core/Samplers/GcSampler.cs
@@ -33,21 +33,24 @@ namespace NewRelic.Agent.Core.Samplers
 
 #if NETFRAMEWORK
 
-	public class GcSampler : AbstractSampler
-	{
-		private const string GCPerfCounterCategoryName = ".NET CLR Memory";
+    public class GcSampler : AbstractSampler
+    {
+        private const string GCPerfCounterCategoryName = ".NET CLR Memory";
+        private const int MaxConsecutiveFailuresBeforeDisable = 5;
 
-		private readonly IGcSampleTransformer _transformer;
-		private const int GcSampleIntervalSeconds = 60;
-		private IPerformanceCounterProxyFactory _pcProxyFactory;
-		private readonly object _lockObj = new object();
+        private readonly IGcSampleTransformer _transformer;
+        private const int GcSampleIntervalSeconds = 60;
+        private readonly IPerformanceCounterProxyFactory _pcProxyFactory;
 
-		/// <summary>
-		/// Translates the sample type enum to the name of the windows performance counter
-		/// </summary>
-		private readonly Dictionary<GCSampleType, string> _perfCounterNames = new Dictionary<GCSampleType, string>
-		{
-			{ GCSampleType.Gen0Size , "Gen 0 heap size"},					// BuildByteData (MB)
+        private int _countConsecutiveSamplingFailures = 0;
+        private string _perfCounterInstanceName;
+
+        /// <summary>
+        /// Translates the sample type enum to the name of the windows performance counter
+        /// </summary>
+        private readonly Dictionary<GCSampleType, string> _perfCounterNames = new Dictionary<GCSampleType, string>
+        {
+            { GCSampleType.Gen0Size , "Gen 0 heap size"},					// BuildByteData (MB)
 			{ GCSampleType.Gen0Promoted , "Promoted Memory from Gen 0"},	// BuildByteData
 			{ GCSampleType.Gen1Size , "Gen 1 heap size"},					// BuildByteData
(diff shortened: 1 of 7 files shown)


src/Agent/NewRelic/Agent/Core/Samplers/GcSampler.cs
```diff
@@ -33,21 +33,24 @@ public enum GCSampleType
...
 #if NETFRAMEWORK
 
-	public class GcSampler : AbstractSampler
-	{
-		private const string GCPerfCounterCategoryName = ".NET CLR Memory";
-
-		private readonly IGcSampleTransformer _transformer;
-		private const int GcSampleIntervalSeconds = 60;
-		private IPerformanceCounterProxyFactory _pcProxyFactory;
-		private readonly object _lockObj = new object();
```
jifeingo, line 43: No longer needed, Perf Counter Dictionary is created locally and is exchanged

src/NewRelic.Core/NewRelic.SystemInterfaces/PerformanceCounterProxy.cs
```diff
@@ -11,178 +11,179 @@
...
+                        if ((int)pcPid.NextValue() == pid)
+                        {
+                            return instanceName;
+                        }
+                    }
+                }
+                catch (UnauthorizedAccessException)
+                {
+                    //The caller should handle this specificially.  It indicates that the
+                    //current user does not have access to pull performance counter values.
```
nrcventura, line 174: Have you tried to validate this statement when dealing with 2 processes of the same app? Will one app get an unauthorized exception when trying to access the instance name of the other process?

src/Agent/NewRelic/Agent/Core/Samplers/GcSampler.cs
```diff
@@ -61,186 +64,304 @@ public class GcSampler : AbstractSampler
...
+                Stop();
+            }
+        }
+
+        public override void Sample()
+        {
+            var currentPerfCounterInstName = default(string);
+            try
+            {
+                currentPerfCounterInstName = _pcProxyFactory.GetCurrentProcessInstanceNameForCategory(GCPerfCounterCategoryName);
```
nrcventura, line 309: If I'm remembering correctly, getting the instance name can be an expensive operation if there are many processes running on a system. Will this cause any problems with the Timer such that we'll have a few samplers queued up to run back to back? Or will we have a sample every couple of minutes? I'm just wondering if we should still implement some sort of caching for this value, and then only attempt to recheck it when we run into an error?

jifeingo: * I don't think we can cache the value b/c you can't guarantee that the instance name you used to have isn't in use by something else.  
* I don't think that  the caller is limited to accessing other processes performance counters based on the permission level required to do this.

What if we do this.....
* When we pull the list of names, let's put our last known instance name at the top of the list to make it the first one being checked.

nrcventura: Jason did some testing using 200 instances of the same application and noticed the slowdown:
1. The initial loading of the perf counters took a few seconds
2. Subsequent loading of the perf counters took was in the 100-200 ms range

The scheduler that we're using will not queue up a bunch of sampling requests.

Since we're only sampling these values once per minute we don't think this will be a problem for the agent or the application.

nrcventura:
Things are looking good in this PR, if you can resolve the merge conflicts I'll mark this PR as approved.

PR #69, opened 2020-07-27
Generates `error.expected` attribute for Error Event, Error Trace and Span Event

vuqtran88:
## Summary

Generates `error.expected` attribute for Error Events, Error Traces and Span Events

## Details
Generates `error.expected` intrinsic attribute for Error Events and Error Traces if error is expected.
Generates `error.expected` agent attribute for Span Events if error is expected.

### DIFF
--- a/tests/Agent/UnitTests/Core.UnitTest/Spans/SpanEventMakerTests.cs
+++ b/tests/Agent/UnitTests/Core.UnitTest/Spans/SpanEventMakerTests.cs
@@ -80,10 +80,8 @@ namespace NewRelic.Agent.Core.Spans.UnitTest
         private IMetricNameService _metricNameSvc;
         private ITransactionMetricNameMaker _transactionMetricNameMaker;
         private ITransactionAttributeMaker _transactionAttribMaker;
-        private TransactionMetricStatsCollection _metricStatsCollection;
 
         private IConfiguration _config;
-        private int _configVersion = 1;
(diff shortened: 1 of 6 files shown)


tests/Agent/UnitTests/Core.UnitTest/Spans/SpanEventMakerTests.cs
```diff
@@ -377,6 +375,37 @@ public void GetSpanEvent_DoesNotIncludesErrorAttributes_WhenThereIsAnError_IfErr
             );
         }
 
+        [TestCase(true)]
+        [TestCase(true)]
```
nrcventura, line 379: I think that you intended one of these `TestCase` attributes to be `TestCase(false)`.

nrcventura:
Right now these changes are ensuring that the `error.expected` attribute is applied to the root span, but it's not getting applied to the other spans. https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Segments/Segment.cs#L223-L239 is where we are currently adding the child span error attributes for the span specific errors.

tests/Agent/UnitTests/Core.UnitTest/Spans/SpanEventMakerTests.cs
vuqtran88: copy-paste nightmare

PR #73, opened 2020-07-29
Add/update license in nuspec files to be Apache-2.0

nr-ahemsath:
Update the `license` element of .nuspec files on the `net35/main` branch to be Apache-2.0.


lspangler584:
What about NewRelic.Net.Agent.x64.nuspec and NewRelic.Net.Agent.nuspec?

nr-ahemsath:
> What about NewRelic.Net.Agent.x64.nuspec and NewRelic.Net.Agent.nuspec?

I missed them because they were hiding one directory level deeper than I was looking.  Good catch.

PR #77, opened 2020-07-30
Expected status code parser

vuqtran88:
## Summary

Parse the `expectedStatusCodes` configuration into a list of matching status code rules. This list includes rules that match either exact status code or a range of status codes.


vuqtran88:
jenkins retest this

PR #78, opened 2020-07-30
Adds back xsd2code with doc on running it

jaffinito:
Adds in xsd2code to allow creation of Configuration.cs or Extension.cs when either of the following are changed:

- src\Agent\NewRelic\Agent\Core\Config\Configuration.xsd
- src\Agent\NewRelic\Agent\Core\NewRelic.Agent.Core.Extension\Extension.xsd

Includes a minimal readme with the commands to run.


build/Tools/xsd2code/Xsd2Code.exe.config
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0"?>
+<configuration>
+<startup><supportedRuntime version="v2.0.50727"/></startup></configuration>
```
alanwest, line 3: @jaffinito I know this PR has merged already, though I'm fairly confident that Xsd2Code will still run fine if we change this to `v4.0.30319`. That way people don't need to install .NET 3.5 to run it.

PR #79, opened 2020-07-30
Mark error data as expected if response status code is listed as expected 

vuqtran88:
## Summary

- Wiring up to mark error with expected status code as expected.
- Limit expected classes and expected messages configuration entries to 50 entries.
- Refactoring.


vuqtran88:
jenkins retest this

nrcventura:
Did you already verify that the frameworks that we capture "Full" status codes for, work with the int status code comparisons that we're doing?

nrcventura:
I talked to Vu about the full status codes, and I missed the test case that covers that scenario, so these changes look good to me.

PR #99, opened 2020-08-07
Ignore errors new configurations

vuqtran88:


vuqtran88:
jenkins retest this

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1,7 +1,50 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- The documentation in here must be kept up to date with that embedded in comments in newrelic.config -->
-<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" targetNamespace="urn:newrelic-config" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:nr="Configuration.xsd">
-  <xs:element name="configuration">
+<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:tns="urn:newrelic-config" targetNamespace="urn:newrelic-config" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:nr="Configuration.xsd">
+
+    <xs:complexType name="ErrorClassType">
```
nrcventura, line 5: Should we consider a different type name here? How about `ErrorClassCollection` to more closely represent that this type represents a collection of error classes?

src/Agent/NewRelic/Agent/Core/Config/Configuration.cs
```diff
@@ -4089,80 +4167,57 @@ public virtual configurationErrorCollectorIgnoreErrors Clone()
...
+        private List<string> messageField;
+        
+        private string nameField;
         
         /// <summary>
-        /// configurationErrorCollectorIgnoreStatusCodes class constructor
+        /// ErrorMessagesTypeErrorClass class constructor
         /// </summary>
-        public configurationErrorCollectorIgnoreStatusCodes()
+        public ErrorMessagesTypeErrorClass()
```
nrcventura, line 4180: This name reads funny to me, and is harder for me to understand. I think that if we change the name of some of the types in the xsd, this generated name should become easier to understand? See the comments in the xsd file for some suggestions.

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1,7 +1,50 @@
...
+                <xs:annotation>
+                    <xs:documentation>
+                        Fully qualified class name of an exception, such as "System.IO.FileNotFoundException".
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="ErrorMessagesType">
```
nrcventura, line 22: Should this be something like `ErrorMessagesMap` or `ErrorMessagesCollection` to better represent that this type defines a collection of error messages.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1135,53 +1137,25 @@ public TimeSpan ErrorEventsHarvestCycle
         }
 
         public virtual uint ErrorsMaximumPerPeriod { get { return 20; } }
-        public virtual IEnumerable<string> ExceptionsToIgnore { get { return ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorErrorsToIgnore, _localConfiguration.errorCollector.ignoreErrors.exception); } }
 
-        private void ParseExpectedErrorConfigurations()
+        private IEnumerable<string> _exceptionsToIgnore;
+        public virtual IEnumerable<string> ExceptionsToIgnore
```
nrcventura, line 1142: We should consider parsing this config setting in the same place that we parse the other configuration related to ignored errors.

vuqtran88: I'm planning to remove this configuration and replace its usages with the `IgnoreErrorConfiguration` in the next work. I still keep it here because I don't want to modify existing tests that use it.

nrcventura: This setting uses a different local and server-side config setting than the other ignore settings that were just added. My assumption is that we still need to fall back and use these settings if the other settings are not provided?

vuqtran88: Yea you are right, even though I did combine the old config with new config in [here](https://github.com/newrelic/newrelic-dotnet-agent/blob/fed4fe4f720e17a884f40968e0f688aaed1d34b6/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs#L2077), but I forgot that we still have to parse it.

vuqtran88: I just added to code to parse the old config in the same place of the new config parser. I still keep the `ExceptionToIgnore` coz I don’t want to modify some tests just yet. I will remove it in the next work.

PR #109, opened 2020-08-18
Create stale.yml

lspangler584:
## Summary
Adds stale.yml GitHub action to check for abandoned/stale PRs and Issues.

## Details
Current configuration adds `stale` label after 30 days, closes 5 days after that if no action taken.

## Links
Same behavior as in [Ruby Agent](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/.github/workflows/stale.yml).
Uses [this GitHub action](https://github.com/actions/stale).

## Testing
Currently testing the yaml in a test repo, but implementation `days-before-stale` and `days-before-close` smallest period is 1 day, so waiting 1 day for the label to appear, then another day for the close to occur.

### DIFF
--- a/.github/workflow/stale.yml
+++ b/.github/workflow/stale.yml
@@ -12,8 +12,8 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'This has been marked stale after 30 days with no activity. It will be closed in 5 days if there is no activity.'
         stale-pr-message: 'Stale pull request message'
-        days-before-stale: 1
-        days-before-close: 2
+        days-before-stale: 30
+        days-before-close: 5
         stale-issue-label: 'stale'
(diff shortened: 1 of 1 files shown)


.github/workflow/stale.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This has been marked stale after 30 days with no activity. It will be closed in 5 days if there is no activity.'
+        stale-pr-message: 'Stale pull request message'
+        days-before-stale: 1
```
nr-ahemsath, line 15: Will you change this to 30 and 5 once your testing is done?

lspangler584: done now. testing is on a different repo, so this should be what we really want.

PR #138, opened 2020-08-21
Agent decrypted encrypted proxy password

vuqtran88:
### Description

Agent supports obscured proxy password in the proxy configuration. The following changes have been made:
1. Adds new `encryptedPassword` and `obscuringKey` configuration.
2. Obfuscation/Deobfuscation logic to the agent so that the agent can deobfuscate encrypted provided password.


src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -55,6 +55,13 @@
               </xs:documentation>
             </xs:annotation>
             <xs:sequence>
+              <xs:element name="obscuringKey" type="xs:string" minOccurs="0" maxOccurs="1">
+                <xs:annotation>
+                  <xs:documentation>
+                    The key used to deobfuscate the proxy password (optional).
```
jifeingo, line 61: We should remove the reference to proxy password.  This is a key for any configuration value.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1306,7 +1313,44 @@ public virtual string Labels
...
+        }
+
+        private string _proxyPassword;
+        public virtual string ProxyPassword
+        {
+            get
+            {
+                if (!_proxyPasswordEvaluated)
+                {
+                    if (_hasObscuringKey && _hasEncryptedPassword)
```
jifeingo, line 1338: I don't think we need _hasObscuringKey and _hasEncryptedPassword.  It seems like the only time we use these is in this function, so we should just evaluate them here.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -58,6 +59,9 @@ public class DefaultConfiguration : IConfiguration
         private readonly RunTimeConfiguration _runTimeConfiguration = new RunTimeConfiguration();
         private readonly SecurityPoliciesConfiguration _securityPoliciesConfiguration = new SecurityPoliciesConfiguration();
         private Dictionary<string, string> _newRelicAppSettings { get; }
+        private bool _hasObscuringKey;
+        private bool _hasEncryptedPassword;
```
jifeingo, line 63: See below, I don't think we need these.  We only use them in a getter only one time.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -111,6 +115,9 @@ protected DefaultConfiguration(IEnvironment environment, configuration localConf
 
             ParseExpectedErrorConfigurations();
             ParseIgnoreErrorConfigurations();
+
+            _hasObscuringKey = !string.IsNullOrEmpty(ObscuringKey);
+            _hasEncryptedPassword = !string.IsNullOrEmpty(_localConfiguration.service.proxy.encryptedPassword);
```
jifeingo, line 120: I don't think we need this.  Probably can move this logic into the if statements

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1306,7 +1313,44 @@ public virtual string Labels
...
         public virtual int ProxyPort { get { return _localConfiguration.service.proxy.port; } }
         public virtual string ProxyUsername { get { return _localConfiguration.service.proxy.user; } }
-        public virtual string ProxyPassword { get { return _localConfiguration.service.proxy.password; } }
+
+        private string _obscuringKey;
+        public string ObscuringKey
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(_obscuringKey))
```
jifeingo, line 1322: Need to add check variable here.  Otherwise, if an obscuring key is not supplied, this getter will evaluate every time.

One could say that we are only using the ObscuringKey in one place, so this would only happen in one place.  But, since this is an overall obscuring key, it may be used in other places later on.

src/NewRelic.Core/Obfuscator.cs
```diff
@@ -0,0 +1,35 @@
...
+        private static byte[] Encode(byte[] bytes, byte[] keyBytes)
+        {
+            for (int i = 0; i < bytes.Length; i++)
+            {
+                bytes[i] = (byte)(bytes[i] ^ keyBytes[i % keyBytes.Length]);
+            }
+            return bytes;
+        }
+    }
+}
```
jifeingo, line 35: This should no longer exist.   We have functions that can already be used for this purpose in New Relic Core.

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -1063,6 +1063,28 @@ public string IgnoreAndExpectedMessagesSetFromLocalAndServerOverrides(bool serve
...
+        [TestCase("ABCD", "bGxs", null, null, ExpectedResult = "ABCD")]
+        [TestCase("ABCD", null, "123", null, ExpectedResult = "ABCD")]
+
+        public string Encrypting_Decrypting_ProxyPassword_Tests(string password, string encryptedPassword, string localConfigObscuringKey, string envObscuringKey)
+        {
+            Mock.Arrange(() => _environment.GetEnvironmentVariable("OBSCURING_KEY")).Returns(envObscuringKey);
+
+            _localConfig.service.proxy.password = password;
+            _localConfig.service.obscuringKey = localConfigObscuringKey;
+            _localConfig.service.proxy.encryptedPassword = $"{encryptedPassword}";
```
jifeingo, line 1081: We shouldn't need to do string interpolation here.  If there are failures when this is null, something else is wrong.

src/NewRelic.Core/Obfuscator.cs
jifeingo:
```
namespace NewRelic.Core
{
    public static class Strings
    {
        private static readonly UTF8Encoding _encoding = new UTF8Encoding();

        /// <summary>
        /// Sanitize the given file name, replacing illegal characters with _.
        /// </summary>
        /// <param name="name">The file name to sanitize.</param>
        /// <returns>The sanitized file name.</returns>
        public static string SafeFileName(string name)
        {
            foreach (var c in Path.GetInvalidPathChars())
            {
                name = name.Replace(c, '_');
            }
            foreach (var c in Path.GetInvalidFileNameChars())
            {
                name = name.Replace(c, '_');
            }
            return name;
        }

        public static string TryBase64Decode(string val, string encodingKey = null)
        {
            if (val == null)
                return null;

            try
            {
                return Base64Decode(val, encodingKey);
            }
            catch
            {
                return null;
            }
        }

        public static string Base64Decode(string val, string encodingKey = null)
        {
            var bytes = Convert.FromBase64String(val);

            if (!string.IsNullOrEmpty(encodingKey))
                bytes = EncodeWithKey(bytes, encodingKey);

            return Encoding.UTF8.GetString(bytes);
        }

        public static string Base64Encode(string val, string encodingKey = null)
        {
            var encodedBytes = Encoding.UTF8.GetBytes(val);

            if (!string.IsNullOrEmpty(encodingKey))
                encodedBytes = EncodeWithKey(encodedBytes, encodingKey);

            return Convert.ToBase64String(encodedBytes);
        }

        private static byte[] EncodeWithKey(byte[] bytes, string key)
        {
            var keyBytes = Encoding.UTF8.GetBytes(key);

            var keyIdx = 0;
            for (var i = 0; i < bytes.Length; i++)
            {
                if (keyIdx == keyBytes.Length)
                    keyIdx = 0;

                bytes[i] = (byte)(bytes[i] ^ keyBytes[keyIdx]);
                keyIdx++;
            }

            return bytes;
        }

        public static string ObfuscateStringWithKey(string val, string key)
        {
            if (val == null || string.IsNullOrEmpty(key))
            {
                return null;
            }

            var bytes = _encoding.GetBytes(val);

            var maxKeyLength = Math.Min(13, key.Length); // we don't really need this, but it helps for unit testing

            var keyIdx = 0;
            for (var i = 0; i < bytes.Length; i++)
            {

                if (keyIdx == maxKeyLength)
                {
                    keyIdx = 0;
                }

                bytes[i] = (byte)(bytes[i] ^ key[keyIdx]);
                keyIdx++;
            }

            return Convert.ToBase64String(bytes);
        }

        public static string ToString(System.Collections.IEnumerable enumerable, char separator = ',')
        {
            var builder = new StringBuilder();
            var first = true;
            foreach (object obj in enumerable)
            {
                if (first)
                {
                    first = false;
                }
                else
                {
                    builder.Append(separator);
                }
                builder.Append(obj.ToString());
            }
            return builder.ToString();
        }

        public static string Replace(this string originalString, string oldValue, string newValue, StringComparison comparisonType, int count)
        {
            int startIndex = 0;
            int numberReplaced = 0;

            while (numberReplaced < count)
            {
                startIndex = originalString.IndexOf(oldValue, startIndex, comparisonType);
                if (startIndex == -1)
                    break;

                originalString = originalString.Substring(0, startIndex) + newValue + originalString.Substring(startIndex + oldValue.Length);
                numberReplaced++;

                startIndex += newValue.Length;
            }

            return originalString;
        }

        // Must use Encoder/Decoder and not Encoding.GetString.  See: http://msdn.microsoft.com/en-us/library/ms404377(v=vs.110).aspx
        public static string GetStringBufferFromBytes(Decoder decoder, byte[] buffer, int offset, int count)
        {
            var length = decoder.GetCharCount(buffer, offset, count);
            var chars = new char[length];
            decoder.GetChars(buffer, offset, count, chars, 0);
            return new string(chars);
        }

    }
}
```

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
jifeingo: In fact, we should test for both Empty String and NULL values.

vuqtran88: thanks for pointing this out, this was the leftover code from using "!obscured" prefix. This shouldn't be here.

Issue #125, opened 2020-08-20
Modify MongoDB_2.6 unbounded integration tests to use a local Docker-hosted service


nr-ahemsath:
Currently, some of [these tests](https://github.com/newrelic/newrelic-dotnet-agent/tree/main/tests/Agent/IntegrationTests/UnboundedIntegrationTests) rely on a MongoDB server hosted inside New Relic's network.

In order to enable community contributors to run these tests, the plan is to provide a Dockerfile/docker-compose solution for spinning up the necessary external dependencies on a developer's system before running the tests.

Internally we already run the necessary service with:
`docker run -d --restart always --name MongoDBServer -p 27017:27017 mongo:3.6.4`

Note: we have two different groups of MongoDB integration tests, which use different services - because we have two different instrumentation wrappers for MongoDB for different client libraries.  If we have time let's investigate whether we can combine the services.

PR #148, opened 2020-08-25
Add dockerized services for both MongoDB test groups

nr-ahemsath:
### Description

This PR adds dockerized MongoDB services for both groups of MongoDB integration tests (one set uses the 1.10 version of the MongoDB .NET client driver and the other uses the 2.6 version).  This is for issues #123 and #125.

### Testing

A change was needed in the assertions for the `Datastore/instance` metrics because the agent has some special-case logic to use the local hostname (from Dns.GetHostname()) if the server's address is "localhost" or "127.0.0.1".

### Changelog

No changelog entry required.

### DIFF
--- a/tests/Agent/IntegrationTests/UnboundedIntegrationTests/MongoDB/MongoDB2_6_AsyncCursorTests.cs
+++ b/tests/Agent/IntegrationTests/UnboundedIntegrationTests/MongoDB/MongoDB2_6_AsyncCursorTests.cs
@@ -1,11 +1,6 @@
 // Copyright 2020 New Relic, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-
-using System;
-using System.Linq;
-using System.Net;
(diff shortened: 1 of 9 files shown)


tests/Agent/IntegrationTests/UnboundedIntegrationTests/MongoDB/MongoDB2_6_AsyncCursorTests.cs
```diff
@@ -2,6 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 
+using System;
+using System.Linq;
+using System.Net;
+using System.Text.RegularExpressions;
```
lspangler584, line 8: these aren't needed after moving helper functions to commonutils.

nr-ahemsath: Thanks for catching that!

PR #149, opened 2020-08-26
Script changes for dockerized unbounded services

nr-ahemsath:
### Description

This PR introduces a script for starting and stopping the dockerized unbounded services, `unbounded-services-control.ps1`.  At the moment it is essentially a wrapper around `docker-compose up/down`.  However, we anticipate needing to add some setup commands for some services after they are started; they should be added to this new script.

This PR also modifies `run-integration-tests.ps1` to call this new script to start the services before running the unbounded tests and then stop the services after the tests finish.

### Testing

I tested the scripts locally; our CI for the unbounded tests calls the `run-integration-tests.ps1` script so it will get tested on PR merge.  There are no agent behavior changes in this PR.

### Changelog

No changelog entry needed.


nr-ahemsath:
jenkins retest this

PR #150, opened 2020-08-26
Fix assembly version mismatch

lspangler584:
### Description
Removes dependency on `System.Web.Mvc` in the Agent's Mvc3 instrumentation to avoid exceptions such as `System.InvalidCastException: Expected argument 0 to be of type System.Web.Mvc.ControllerContext (v3.0.0.1), but was of type System.Web.Mvc.ControllerContext (v4.0.0.1)` .
This was accomplished using `<dynamic>` instead of specific `System.Web.Mvc` type `ContextController`.

### Testing
Use of `ControllerContext` in `System.Web.Mvc` was checked via decompilervfor use by restricted methods (internal, private) and none were found, so use of it in the Agent should not pose a problem.
Attempts were made to create an Mvc4 test app, as in the reporting customer's case, but the exception was not reproducible.
Removal of Agent instrumentation dependency on the dll should prevent the assembly version mismatch error encountered by the customer.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/MvcRouteNamingHelper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/MvcRouteNamingHelper.cs
@@ -1,9 +1,8 @@
 // Copyright 2020 New Relic, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//using System.Collections.Generic;
-//using System.Linq;
-//using System.Threading;
+using System.Collections.Generic;
+using System.Linq;
 using System.Web.Routing;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 using NewRelic.SystemExtensions;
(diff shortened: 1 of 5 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/SyncInvokeActionWrapper.cs
```diff
@@ -23,7 +22,7 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
         {
             transaction.AttachToAsync();
 
-            var controllerContext = instrumentedMethodCall.MethodCall.MethodArguments.ExtractNotNullAs<ControllerContext>(0);
+            var controllerContext = instrumentedMethodCall.MethodCall.MethodArguments.ExtractNotNullAs<dynamic>(0);
```
nrcventura, line 25: Do you want to apply the same `!= null` check to the controllerContext that you did in the AsyncBeginInvokeActionWrapper?

lspangler584: done

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/MvcRouteNamingHelper.cs
```diff
@@ -23,15 +21,15 @@ public static string TryGetControllerNameFromObject(ControllerContext controller
...
             var actionName = methodCall.MethodArguments.ExtractAs<string>(1);
             if (actionName != null)
                 return actionName;
 
-            var directRouteMatches = routeData.Values.GetValueOrDefault("MS_DirectRouteMatches") as IEnumerable<RouteData> ?? Enumerable.Empty<RouteData>();
-            routeData = directRouteMatches.FirstOrDefault() ?? routeData;
-            actionName = routeData.Values.GetValueOrDefault("action") as string;
+            var directRouteMatches = routeData?.Values?["MS_DirectRouteMatches"] as IEnumerable<RouteData> ?? Enumerable.Empty<RouteData>();
+            routeData = directRouteMatches?.FirstOrDefault() ?? routeData;
+            actionName = routeData?.Values?["action"];
```
lspangler584, line 33: Could have used original code here, but the 'default' of GetValueOrDefault would prevent actionName from being null and the method would return an empty string for the actionName, which doesn't look like the intended behavior.

PR #154, opened 2020-08-27
Changes obscuring key environment variable name.

vuqtran88:
### Description

New Relic specific environment variables should have "NEW_RELIC" prefix. This PR changes the obscuring key environment variable name from "OBSCURING_KEY" to "NEW_RELIC_CONFIG_OBSCURING_KEY".


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -41,6 +41,7 @@ public class DefaultConfiguration : IConfiguration
         private const string ServerConfigSource = "Server Configuration";
         private const int MaxExptectedErrorConfigEntries = 50;
         private const int MaxIgnoreErrorConfigEntries = 50;
+        private const string NewRelicConfigObscuringKey = "NEW_RELIC_CONFIG_OBSCURING_KEY";
```
jifeingo: I don't think that this is necessary since it is only used in one place.

PR #157, opened 2020-08-27
fix regression in RejitMvcApplication

lspangler584:
### Description
Unintended changes were introduced in a prior PR causing Rejit Integration Tests to fail. This PR reverts the RejitMvcApplication to what is was before, then applies a conversion to use PackageReferences instead of packages.config.

### Testing
Rejit Integration Tests should all pass locally and on the server.

### Changelog
None required.


tests/Agent/IntegrationTests/Applications/RejitMvcApplication/Properties/AssemblyInfo.cs
```diff
@@ -13,7 +12,7 @@
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("")]
 [assembly: AssemblyProduct("RejitMvcApplication")]
-[assembly: AssemblyCopyright("Copyright ©  2020")]
+[assembly: AssemblyCopyright("Copyright ©  2017")]
```
nr-ahemsath, line 15: Was this intentional?

tests/Agent/IntegrationTests/Applications/RejitMvcApplication/RejitMvcApplication.csproj
```diff
@@ -42,6 +44,9 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <!-- <Reference Include="Microsoft.CodeDom.Providers.DotNetCompilerPlatform">
```
nr-ahemsath, line 47: Can these comments be cleaned up?

tests/Agent/IntegrationTests/Applications/RejitMvcApplication/Properties/AssemblyInfo.cs
lspangler584: that's when it was originally created. we have other apps with that too.

Issue #37, opened 2020-07-15
Nullability Support for APIs [.NET 5]


elucus:
Microsoft's guidance is to have nullability support put into all APIs by the GA release of .NET 5.

This will apply to our Agent API, Logs in Context, Telemetry SDK, and Lambda products.

Aha! Link: https://newrelic.aha.io/features/DOTNET-236

PR #174, opened 2020-09-02
Nullable Reference Type Support for API

jifeingo:
### Description
* FIxes Issue #37
* Enables C# version 8.0 language features in support of Nullable Reference types.
* Updates the signatures of API Methods, Interfaces, and their Implementations to indicate nullable reference types.

### Testing
* No specific tests need to change
* Old and New API with Old and New Agent.  API Functions bound properly and, when C# v9.0 was enabled in the test app, the warnings for nullable types were recognized.

### Changelog
* Updated, No docs site changes required.


jifeingo:
jenkins test this

src/Agent/NewRelic.Api.Agent/NewRelic.Api.Agent.csproj
```diff
@@ -7,6 +7,8 @@
     <DocumentationFile>$(RootProjectDirectory)\src\_build\$(Platform)-$(Configuration)\$(AssemblyName)\NewRelic.Api.Agent.xml</DocumentationFile>
     <IntermediateOutputPath>$(RootProjectDirectory)\src\_build\$(Platform)-$(Configuration)\$(AssemblyName)\</IntermediateOutputPath>
     <OutputPath>$(RootProjectDirectory)\src\_build\$(Platform)-$(Configuration)\$(AssemblyName)\</OutputPath>
+    <Nullable>enable</Nullable>
+    <LangVersion>8.0</LangVersion>
```
alanwest, line 11: I believe we can declare `LangVersion` in one spot. Change this to 8.0 here:

https://github.com/newrelic/newrelic-dotnet-agent/blob/0182016ebc837f56dc61829e99abe67633e38c95/Directory.Build.props#L3

Then you should be able to remove `LangVersion` from individual csproj files.

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
```diff
@@ -508,6 +515,7 @@ public string GetBrowserTimingHeader()
                 // The transaction's name must be frozen if we're going to generate a RUM script
                 transaction.CandidateTransactionName.Freeze(TransactionNameFreezeReason.ManualBrowserScriptInjection);
 
+                //this can return null
```
alanwest, line 518: Should it be able to return null? What would be the impact if we changed it to always return something?

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/IAgentApi.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2020 New Relic, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+#nullable enable
```
alanwest, line 4: Would it be a lot more effort to enable this for the entire NewRelic.Agent.Extensions project?

alanwest: Alternatively, is it possible to just not touch NewRelic.Agent.Extensions for now? We do not currently publish it to nuget.

src/Agent/NewRelic.Api.Agent/NewRelic.cs
```diff
@@ -642,7 +642,7 @@ public static void StartAgent()
         /// </code>
         /// </example>
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
-        public static void SetApplicationName(string applicationName, string applicationName2 = null, string applicationName3 = null)
+        public static void SetApplicationName(string? applicationName, string? applicationName2 = null, string? applicationName3 = null)
```
alanwest, line 645: Makes sense the that the last two parameters are nullable, but why do we want to enable the first one to be nullable?

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/IAgentApi.cs
jifeingo: It is a lot more work to update the entire project.  It involves the `VisibilityBypasser` and Delegate functions.  I started down the path, and almost every file needed changes.

I can look into further, but not doing it for this project will cause warnings to appear where those items are used.  They'll be treated as Not Nullable for any caller.  If the caller is `<Nullable>enabled</Nullable>`, this will show warnings.

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
jifeingo: Actually, my mistake, this actually doesn't return null

src/Agent/NewRelic.Api.Agent/NewRelic.cs
```diff
@@ -666,7 +666,7 @@ public static void SetApplicationName(string applicationName, string application
         /// </code>
         /// </example>
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
-        public static IEnumerable<KeyValuePair<string, string>> GetRequestMetadata()
+        public static IEnumerable<KeyValuePair<string, string>>? GetRequestMetadata()
```
alanwest, line 669: Let's make this not nullable

src/Agent/NewRelic.Api.Agent/Directory.Build.props
```diff
@@ -1,4 +1,8 @@
 <Project>
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)..'))" />
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'FullAgent.sln'))\build\StyleCop.props" />
+    <PropertyGroup>
```
alanwest, line 4: nit: indentation

src/Agent/NewRelic.Api.Agent/NewRelic.cs
```diff
@@ -583,7 +584,7 @@ public static void DisableBrowserMonitoring(bool overrideManual = false)
         /// </code>
         /// </example>
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
-        public static void SetUserParameters(string userName, string accountName, string productName)
+        public static void SetUserParameters(string userName, string? accountName, string? productName)
```
alanwest, line 587: I thought `userName` could be null.

src/Agent/NewRelic.Api.Agent/NewRelic.cs
alanwest: I unresolved this conversation because I wanted to double check... did you intentionally leave the first `applicationName` as nullable?

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
```diff
@@ -381,16 +384,21 @@ private void AddUserAttributeToCurrentTransaction(string key, object value)
...
 
             using (new IgnoreWork())
             {
+                category = category != null
+                    ? category.Trim()
+                    : string.Empty;
+
                 // Default to "Custom" category if none provided
-                if (string.IsNullOrEmpty(category?.Trim()))
+                if (string.IsNullOrEmpty(category))
```
alanwest, line 398: Just use `IsNullOrWhitespace`. Probably avoids a string allocation. If you still need to check for null that's fine.

src/Agent/NewRelic/Agent/Core/WireModels/MetricWireModel.cs
```diff
@@ -112,6 +112,14 @@ public override bool Equals(object obj)
             return false;
         }
 
+        public override int GetHashCode()
```
alanwest, line 115: What did this do?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/SystemBindingTypes.cs
```diff
@@ -21,7 +21,7 @@ public static class SystemBindingTypes
             typeof(NetTcpBinding),
             typeof(NetNamedPipeBinding),
             typeof(NetMsmqBinding),
-            typeof(NetPeerTcpBinding),
+            //typeof(NetPeerTcpBinding),
```
alanwest, line 24: ??

src/Agent/NewRelic/Agent/Core/WireModels/MetricWireModel.cs
alanwest: Spoke with @jifeingo. Theory is that we implemented equality support for the purposes of tests... for use with `Assert.Contains` for example. Apparently the support was not fully implemented. Adding the `GetHashCode` method resolved a build warning.

alanwest (approved):
👍

PR #181, opened 2020-09-08
Integration test fixes

nr-ahemsath:
### Description

This PR reduces the number of build warnings when building the integration tests solution, primarily by modifying package references in .csproj files and some assembly binding redirections in web.config files.  These warnings were recently introduced by project file changes to support StyleCop.

It also fixes some test failures caused by a) asserting on the value of the ApdexPerfZone attribute rather than just asserting on its presence and b) fixing a copyright header format issue in one file.

### Testing

The integration tests solution builds with fewer warnings, and the tests pass for me locally (as reliably as usual, meaning that out of 900+ tests I typically get 5-15 test flickers (failures which pass upon re-test)).

### Changelog

N/A because no agent functionality changed.


tests/Agent/IntegrationTests/Applications/BasicMvcApplication/BasicMvcApplication.csproj
```diff
@@ -122,7 +108,7 @@
       <Version>5.2.3</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.AspNet.Webpages">
-      <Version>3.2.0</Version>
+      <Version>3.2.7</Version>
```
jaffinito, line 111: Was there a reason we chose to update the versions here?

jaffinito (approved):
Other than my question about versions, this looks good.

tests/Agent/IntegrationTests/Applications/BasicMvcApplication/BasicMvcApplication.csproj
nr-ahemsath: One of the warnings was about a package version downgrade; this change eliminated the warning and the tests still build and pass.

PR #183, opened 2020-09-08
Modify project files to reduce warnings when building

nr-ahemsath:
### Description

Modifies project files in the unbounded int. tests solution to reduce build warnings.

### Testing

The solution builds cleanly with fewer warnings, and the tests still pass.

### Changelog

N/A

### DIFF
--- a/tests/Agent/IntegrationTests/UnboundedApplications/MongoDB2_6Application/MongoDB2_6Application.csproj
+++ b/tests/Agent/IntegrationTests/UnboundedApplications/MongoDB2_6Application/MongoDB2_6Application.csproj
@@ -63,12 +63,9 @@
     <Reference Include="System.Configuration" />
     <Reference Include="System.Web.Services" />
     <Reference Include="System.EnterpriseServices" />
-    <Reference Include="System.Net.Http">
-    </Reference>
     <Reference Include="System.Net.Http.WebRequest">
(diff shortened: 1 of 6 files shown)


tests/Agent/IntegrationTests/UnboundedApplications/MongoDB2_6Application/MongoDB2_6Application.csproj
```diff
@@ -104,11 +101,10 @@
       <Version>4.3.0</Version>
     </PackageReference>
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation">
-      <Version>4.0.0</Version>
+      <Version>4.3.0</Version>
```
jaffinito, line 104: Why the change in version here?

jaffinito (approved):
Other than my question about versions, this is good.

tests/Agent/IntegrationTests/UnboundedApplications/MongoDB2_6Application/MongoDB2_6Application.csproj
nr-ahemsath: One class of build warnings was about package versions being downgraded.  Making this change eliminated the warning, and the tests still build and pass.

Issue #85, opened 2020-08-03
Support MySqlConnector


bgrainger:
[MySqlConnector](https://github.com/mysql-net/MySqlConnector) is an advanced ADO.NET library for MySQL that supports asynchronous operations. (The "standard" MySQL Connector/NET package [does not](https://bugs.mysql.com/bug.php?id=70111).)

### Feature Description

Supporting this library out-of-the-box is a [popular feature request](https://discuss.newrelic.com/t/feature-idea-support-mysqlconnector-driver-for-db-instrumentation/63414) for New Relic customers. 

There is [custom instrumentation XML](https://github.com/mysql-net/MySqlConnector/tree/master/contrib/NewRelic) for MySqlConnector, but as discussed in in [NR Support case 373822](https://support.newrelic.com/tickets/373822/edit) (**NOTE:** private link) and demonstrated [in this repo](https://github.com/bgrainger/DatabaseTimingTest), when it's used, database timings are doubled:

> This trace shows a definite problem with the database timings. Specifically, almost all database calls are logged as extending to the end of the transaction, e.g., 164.66ms then 162.23ms, then 160.97ms, then 158.94ms, etc. (They should be something more like 2.4ms, 1.3ms, 2.0ms, etc.)
> 
> Summing up all these incorrect timings gives an overall incorrect picture of how long the API is spending in database transactions.

According to that support ticket:

> So, after diving more deeply into the code and your repro our dev team has determined the agent code behind the SqlCommandTracerAsync tracer does not work correctly with the MySqlConnector driver and proper support will require additional investigation, development, testing, etc. There isn't a quick fix here :-(

Now that newrelic-dotnet-agent is Open Source, is it possible the community could help develop this fix? What are the specifics of how the "tracer does not work correctly", and what would be the correct approach to fixing it?

### Priority

This is a "Must Have" for us. We use MySqlConnector for its async I/O capabilities, but all database timings are greatly inflated in the NR UI, making it useless for actually diagnosing DB issues.

elucus:
Hi @bgrainger, thanks for your interest in the .NET agent. We believe that support for the MySQLConnector would be a valuable addition. However, at this point, we are unable to commit to a timeline to resolution for this feature.  From your comments, it sounds like perhaps you’d be interested in contributing. If this is the case, please let us know, and we can provide some direction to get you started. Additionally, please let us know what database you’re using the MySQLConnector with.

bgrainger:
[...]

PR #197, opened 2020-09-10
Change how EndSegment is called for Tasks

bgrainger:
### Description

If the `Task` has already completed, we call `EndSegment` (synchronously) immediately. Otherwise, we always continue with `TaskContinuationOptions.ExecuteSynchronously` on the default `TaskScheduler`, which runs our continuation as soon as possible (on the thread that completed the task).

This avoids task continuations being delayed until much later in the ASP.NET pipeline (due to when `AspNetSynchronizationContext` posts them), which causes inaccurate timings.

### Testing

Tested in prod as per https://github.com/newrelic/newrelic-dotnet-agent/issues/85#issuecomment-690669280.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -14,6 +14,7 @@ Enables nullable reference types that are part of the C# 8.0 language specificat
 
 ### Fixes
 * Fixes an issue that may cause `InvalidCastException` due to an assembly version mismatch in Mvc3 instrumentation.
+* Fixes an async timing issue that can cause the end time of `Task`-returning methods to be determined incorrectly.
 
 ## [8.31] - 2020-08-17
 ### New Features
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Providers/Wrapper/Delegates.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Providers/Wrapper/Delegates.cs
@@ -152,22 +152,18 @@ namespace NewRelic.Agent.Extensions.Providers.Wrapper
                 transaction.Hold();
             }
 
-            if (options == TaskContinueWithOption.None)
+            if (task.IsCompleted)
+            {
+                EndSegment(task);
+            }
+            else if (options == TaskContinueWithOption.None)
             {
                 if (!continuationOptions.HasValue) task.ContinueWith(EndSegment);
                 else task.ContinueWith(EndSegment, continuationOptions.Value);
             }
             else
             {
-                var context = SynchronizationContext.Current;
-                if (context != null)
-                {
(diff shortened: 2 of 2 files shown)


bgrainger:
Should this PR also revert this code?
https://github.com/newrelic/newrelic-dotnet-agent/blob/03d6ca84f0dd79fcfda8c735079964d4f80f4514/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis/ExecuteAsyncImplWrapper.cs#L41-L49

nrcventura:
> 
> 
> Should this PR also revert this code?
> 
> https://github.com/newrelic/newrelic-dotnet-agent/blob/03d6ca84f0dd79fcfda8c735079964d4f80f4514/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis/ExecuteAsyncImplWrapper.cs#L41-L49

That's a good question. I think that I still have the test app that I used to verify that fix so I can include that in the additional timing tests we want to do to try to confirm that some previously problematic scenarios are still working as expected. Unfortunately, we haven't come up with a good way to automate these types of timing tests yet.

nrcventura:
I'm still working my way through testing the timing of some of our previously problematic async scenarios, but I was able to test that the Redis instrumentation, with the changes in this PR, can go back to using `return Delegates.GetAsyncDelegateFor<Task>(agent, segment);`. Although, my preference is to have that done in a separate clean up PR, because I suspect that there are a few other changes that can be made to better standardize the way we handle these continuations. I think that we might eventually be able to get rid of the `else if (options == TaskContinueWithOption.None)` block, and then reduce the number of `GetAsyncDelegateFor` overloads that we have.

nrcventura:
ci test this please

nrcventura (approved):
The timings for many of our async scenarios look good. Not only will this fix a few problems, but it helps pave the way forward for simplifying/standardizing our async timing. Great job identifying this!

bgrainger:
This affects our DB timings in production, so I'm anxious to get this fix. However, I'd prefer to run the "official" NR Agent on our infrastructure rather than a version I built locally.

Do you have a rough expectation of when this updated code might ship? (I'm not asking for a firm promise of a release date, but just want to understand when it's likely we'd be able to upgrade to this. I don't even know if you've shipped an agent built from this OSS code base yet, so I understand that things might take longer than expected.)

nrcventura:
We have a few things queued up for a release, and were waiting to kick off our release process until this was merged to better support your use case. We plan on possibly kicking off our release process tomorrow (9/16). This is definitely an area where we can improve our transparency, and can learn from other open source projects.

nrcventura:
We had a minor issue come up during our .NET 5 RC 1 validation that's delaying the release a little, but we still plan on running our release process soon.

PR #198, opened 2020-09-11
Dockerize mssql test dependencies

### LINKED ISSUE #127: Modify MS SQL Server integration tests to use a local Docker-hosted service
Currently, some of [these tests](https://github.com/newrelic/newrelic-dotnet-agent/tree/main/tests/Agent/IntegrationTests/UnboundedIntegrationTests) rely on a MS SQL server hosted inside New Relic's network.

In order to enable community contributors to run these tests, the plan is to provide a Dockerfile/docker-compose solution for spinning up the necessary external dependencies on a developer's system before running the tests.

Note: I think our current MS SQL Server unbounded integration tests expect certain data to already exist in the DB (unlike a lot of the other tests, which set up any data needed to run queries against when the test fixture is initialized) so part of this story will involve modifying the tests to perform their own setup.

https://hub.docker.com/_/microsoft-mssql-server

jaffinito:
### Description
Resolves #127.  Adds a docker image for MSSQL with some minor changes to the test assertions to get the tests to pass, mostly due to names in DB due to cleaning up some old data. These changes don't change the logic of the tests.

Things to note:
- The scripting around starting MSSQL and restore the database is required since the base image from MS does not start MSSQL until the very end blocking the restore attempts.
- As mentioned in entrypoint.cs, the order of the command on line 5 is very important.  MSSQL must be the last thing started.
- Since the original DB came from a Windows server and because the SQL instance in the image is empty, we have to use the MOVE option to setup the DB restore.

This PR also cleans and freshens the DB.

### Testing

The unbounded integration tests pass locally and we'll make sure they pass in our internal CI before merging this PR.

### Changelog

N/A


nr-ahemsath (approved):
Looks good to me.  All the MSSQL-related integration tests passed for me locally using the container. 👍

PR #206, opened 2020-09-11
Refactor connection string parsing

nr-ahemsath:
### Description

This PR improves the robustness of the configuration classes for the unbounded integration tests, by replacing positional parsing of connection strings to extract values needed for assertions (e.g. server host) with use of `System.Data.Common.DbConnectionStringBuilder` (thanks to @bgrainger for the suggestion to use this class).

### Testing

I verified that the unbounded integration tests all pass for me locally with these changes.

### Changelog

N/A


bgrainger (approved):

alanwest:
Sorry futzed with things on mobile device and accidentally hit close... Don't mind me :)

tests/Agent/IntegrationTests/Shared/MsSqlOdbcConfiguration.cs
```diff
@@ -3,6 +3,7 @@
 
 
 using System;
+using System.Collections.Generic;
```
jaffinito, line 6: This can be removed since its not used.

jaffinito (approved):
Other than the using comment, this looks great!

tests/Agent/IntegrationTests/Shared/MsSqlOdbcConfiguration.cs
nr-ahemsath: Thanks for noticing, my last push removed that.

PR #213, opened 2020-09-14
Refactor connection string parsing (#206)

nr-ahemsath:
### Description

Refactor connection string/uri parsing in unbounded integration test configuration classes to be more robust by use of standard classes:
* DbConnectionStringBuilder
* UriBuilder

### Testing

Verified that unbounded integration tests passed locally.

### Changelog

N/A


jaffinito (approved):
Matches up with the main branch. Love seeing these two branches moving closer together!

Issue #209, opened 2020-09-14
Support for net:tcp transport for WCF


pluppplupp:
### Is your feature request related to a problem? Please describe.
We have a web server running ASP.NET framework. It makes WCF calls with a net.tcp transport to .NET services hosted in IIS . 

From what we understand, reading the manual and asking support, if you use net.tcp you need to transfer the transaction information in a custom header or in the payload.

What we’ve done is that we’ve used the NewRelic API and CreateDistributedTracePayload/AcceptDistributedTracePayload and send the tracepayload information on a custom header.

We have a problem with the receiving side. We've implemented the accept method in a WCFMessageInspector (AfterReceiveRequest). But this doesn't work, we've contacted support and a new relic .net engineer has confirmed that the new relic transaction doesn't get created at that point in the lifecycle, and is only available in the actual endpoint code.

### Feature Description
Make the "built-in" distributed tracing work for WCF with net:tcp transport.

### Priority
We really need this to be able to debug outages in production between our web application and core backend systems.

jifeingo:
Hello pluppplupp!

Thank you for taking the time reach out.  The team has evaluated this request and we have incorporated CAT and DT support into our instrumentation for the NetTCP Binding.  After this release, you should no longer need to use the API to Create/Accept Distributed Tracing payloads, it should happen automatically.

This is scheduled to be released as part of our v8.32 release, which should occur sometime within the next week.

Once this is released, please confirm that it is working as you have expected and let us know if you have any questions.

Thanks,
Jason

pluppplupp:
We would be able to test it!!

jifeingo:
Hey pluppplupp!

We just pushed a release of the .NET Agent v8.32 which should have the changes to support NetTCP CAT and DT.
Please let us know if you have any questions.

Thanks!
Jason

jifeingo:
Hey There!
I just wanted to check in and see if you were able to use the NetTCP WCF instrumentation to see if it met your needs.

Thanks,
Jason

magnus-hansson:
@jifeingo We have tested it and it seems to work. One thing that we are unsure of is service map where the down stream WCF service shows up as an external service, where we expected it to show up as a Service. Is this what we can expect?

Thanks
Magnus & Mattias

PR #214, opened 2020-09-15
WCF - Support CAT and DT over NetTCP

jifeingo:
Fixes Issue #209 

### Description

Updates the Agent to support CAT and DT when using NetTCP Binding.  When using NetTCP, adds and retrieves the CAT and DT Headers to the request headers, instead of the HTTP Headers.

### Testing
* Extends the Integration Testing framework for WCF to test for DT and CAT for NetTCP in Self Hosting scenarios

### Changelog
* Added note to change log.


tests/Agent/IntegrationTests/IntegrationTests/WCF/WCFTestBase.cs
```diff
@@ -45,6 +45,9 @@ public enum ASPCompatibilityMode
         protected readonly ConsoleDynamicMethodFixtureFW _fixture;
 
         protected readonly WCFBindingType _bindingToTest;
+        protected string ExpectedTransportType => _bindingToTest == WCFBindingType.NetTcp
+            ? "NetTCP"
```
nrcventura, line 49: I don't remember if we're testing NetMSMQ binding, but this PR might actually enable CAT and DT support for that binding type, because I think it's using a SOAP payload over a MSMQ transport.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/MethodInvokerWrapper.cs
```diff
@@ -145,29 +159,57 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
+                                catch
+                                {
+                                    //gulp
+                                }
+                            }
+
+                            if (retrievedHeaders.Any())
+                            {
+                                haveAcceptedTraceHeaders = true;
+                                transaction.AcceptDistributedTraceHeaders(retrievedHeaders, GetHeaderValue, TransportType.NetTCP);
```
nrcventura, line 210: The `AcceptDistributedTraceHeaders` method takes a function that can be used to extract a header value from a collection. Can we leverage this idea to simplify the header extract code here.

For example is it possible to define to 2 local functions: 1. to extract headers from `httpRequestMessage.Headers`, 2. to extract headers from `messageHeaders`? 

If we do that, then we might be able to simplify the code here to something that looks more like:
```
if (messageHeaders != null && messageHeaders.MessageVersion != MessageVersion.None)
{
    IEnumerable<string> GetMessageHeadersFromIncomingMessageHeaders(...)
    {
        ...
    }

    if (retrievedHeaders.Any())
    {
        haveAcceptedTraceHeaders = true;
        transaction.AcceptDistributedTraceHeaders(retrievedHeaders, GetMessageHeadersFromIncomingMessageHeaders, TransportType.Other);
    }
}
```

And then we could do something similar for the Http message headers.

nrcventura, line 210: We should also change the transport type from `NetTCP` to `Other`, because of the internal agent spec, and because I think this case will also be triggered by other transport types that use a SOAP based payload.

nrcventura (changes requested):
I'm glad that the test changes were not very large, and that our investment in writing this WCF testing infrastructure is paying off.

The biggest change that needs to be made is with the transport type, I didn't realize that TCP was not one of the defined options in the spec (so I don't know how the backend will handle dealing with TCP). I think that it will also be inaccurate, because other transport types might communicate headers in the same way.

There are also a few minor typos in some of the comments, but I didn't flag those individually.

tests/Agent/IntegrationTests/IntegrationTests/WCF/WCFTestBase.cs
jifeingo: So, we are not testing these bindings.   But, I suspect they would work if we tested them.  The Transport Type would be OTHER, not MSMQ however.

nrcventura: I don't think we should add tests for those binding types at this point. If something comes up then I think we can add some additional tests. As it stands WCF tests make up the vast majority of the tests in our integration test suite.

nrcventura (approved):

tests/Agent/IntegrationTests/IntegrationTests/WCF/WCFTestBase.cs
jifeingo: I added an issue to track this.

PR #220, opened 2020-09-15
Add integration tests for MySqlConnector

### LINKED ISSUE #200: Add integration tests for MySqlConnector instrumentation
#85 recently added out of the box instrumentation support for the MySqlConnector library. We should apply our existing MySql integration test suite to this driver. This will help to prevent regressions in the support of this library.

### Feature Description
Apply the existing MySql integration test suite to the MySqlConnector library.

### Additional context
#128 enables developers to run the existing MySql integration tests against a local dockerized version of the test MySql database so that engineers will not need access to the internal New Relic network to run the tests.

### Priority
Really Want

bgrainger:
### Description

This adds integration tests for MySqlConnector and fixes #200.

So far, this is just a copy/paste of the existing MySQL tests and only tests one code path; I want to make it more comprehensive. Opening this PR to demonstrate WIP and for any initial feedback.

### DIFF
--- a/tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
+++ b/tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
@@ -186,4 +186,59 @@ namespace NewRelic.Agent.UnboundedIntegrationTests
         {
         }
     }
+
+    [NetFrameworkTest]
+    public class MySqlConnectorDbCommandExecuteReaderTest : MySqlConnectorTestBase
+    {
+        public MySqlConnectorDbCommandExecuteReaderTest(RemoteServiceFixtures.MySqlConnectorBasicMvcFixture fixture, ITestOutputHelper output)
+            : base(fixture, output, x => x.GetDbCommandExecuteReader(), "DbCommandExecuteReader")
+        {
+        }
+    }
+
+    [NetFrameworkTest]
+    public class MySqlConnectorDbCommandExecuteScalarTest : MySqlConnectorTestBase
+    {
+        public MySqlConnectorDbCommandExecuteScalarTest(RemoteServiceFixtures.MySqlConnectorBasicMvcFixture fixture, ITestOutputHelper output)
+            : base(fixture, output, x => x.GetDbCommandExecuteScalar(), "DbCommandExecuteScalar")
+        {
+        }
+    }
+
+    [NetFrameworkTest]
+    public class MySqlConnectorDbCommandExecuteNonQueryTest : MySqlConnectorTestBase
+    {
+        public MySqlConnectorDbCommandExecuteNonQueryTest(RemoteServiceFixtures.MySqlConnectorBasicMvcFixture fixture, ITestOutputHelper output)
+            : base(fixture, output, x => x.GetDbCommandExecuteNonQuery(), "DbCommandExecuteNonQuery")
+        {
+        }
+    }
+
+    [NetFrameworkTest]
+    public class MySqlConnectorDbCommandExecuteReaderAsyncTest : MySqlConnectorTestBase
+    {
+        public MySqlConnectorDbCommandExecuteReaderAsyncTest(RemoteServiceFixtures.MySqlConnectorBasicMvcFixture fixture, ITestOutputHelper output)
+            : base(fixture, output, x => x.GetDbCommandExecuteReaderAsync(), "DbCommandExecuteReaderAsync")
+        {
+        }
+    }
+
+    [NetFrameworkTest]
+    public class MySqlConnectorDbCommandExecuteScalarAsyncTest : MySqlConnectorTestBase
+    {
+        public MySqlConnectorDbCommandExecuteScalarAsyncTest(RemoteServiceFixtures.MySqlConnectorBasicMvcFixture fixture, ITestOutputHelper output)
+            : base(fixture, output, x => x.GetDbCommandExecuteScalarAsync(), "DbCommandExecuteScalarAsync")
+        {
+        }
+    }
+
+    [NetFrameworkTest]
+    public class MySqlConnectorDbCommandExecuteNonQueryAsyncTest : MySqlConnectorTestBase
+    {
+        public MySqlConnectorDbCommandExecuteNonQueryAsyncTest(RemoteServiceFixtures.MySqlConnectorBasicMvcFixture fixture, ITestOutputHelper output)
+            : base(fixture, output, x => x.GetDbCommandExecuteNonQueryAsync(), "DbCommandExecuteNonQueryAsync")
+        {
+        }
+    }
+
 }
--- a/tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
+++ b/tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
@@ -1,0 +1,130 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+
+using System.Collections.Generic;
+using System.Linq;
+using NewRelic.Agent.IntegrationTestHelpers;
+using NewRelic.Agent.IntegrationTestHelpers.Models;
+using NewRelic.Agent.IntegrationTests.Shared;
+using NewRelic.Testing.Assertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NewRelic.Agent.UnboundedIntegrationTests
+{
+    [NetFrameworkTest]
+    public class MySqlConnectorTests : IClassFixture<RemoteServiceFixtures.MySqlConnectorBasicMvcFixture>
+    {
+        private readonly RemoteServiceFixtures.MySqlConnectorBasicMvcFixture _fixture;
+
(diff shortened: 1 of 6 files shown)


nr-ahemsath:
Looks good so far.  I cloned your fork and was able to build and run the test, which passed.

nr-ahemsath:
ci test this please

nrcventura:
Yes, what you have is basically what I had in mind. There are definitely opportunities within the test to reduce code duplication. One of the techniques that we've used in the past is to have a base test class that defines all of the `Fact` methods and a .ctor. Then we create subclasses for the different things that we want to test. In this case it would probably be a subclass for the MySql library and another for the MySqlConnector library. The main differences between those two tests are probably the transaction names, and the fixture class. You could probably use this same pattern for the fixture class as well.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/RemoteServiceFixtures/MySqlConnectorBasicMvcFixture.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+
+        public override void Dispose()
+        {
+            base.Dispose();
+            DropProcedure();
+        }
+
+        private void GetUrl([CallerMemberName] string pathSuffix = null)
```
nrcventura, line 61: The usage of `CallerMemberName` is a neat way to handle the different endpoints.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
```diff
@@ -0,0 +1,201 @@
...
+                () => Assertions.TransactionEventHasAttributes(expectedTransactionEventIntrinsicAttributes, TransactionEventAttributeType.Intrinsic, transactionEvent),
+                () => Assertions.SqlTraceExists(expectedSqlTraces, sqlTraces),
+                () => Assertions.TransactionTraceSegmentParametersExist(expectedTransactionTraceSegmentParameters, transactionSample)
+            );
+        }
+    }
+
+    [NetFrameworkTest]
+    [Collection("MySqlConnector")]
+    public class MySqlConnectorExecuteReaderTest : MySqlConnectorTestBase
```
nrcventura, line 144: It's interesting to see just how similar all of these tests were, and this code structure makes that really clear!

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
```diff
@@ -0,0 +1,201 @@
...
+using NewRelic.Agent.IntegrationTestHelpers;
+using NewRelic.Agent.IntegrationTestHelpers.Models;
+using NewRelic.Agent.IntegrationTests.Shared;
+using NewRelic.Testing.Assertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NewRelic.Agent.UnboundedIntegrationTests
+{
+    [CollectionDefinition("MySqlConnector", DisableParallelization = true)]
```
nrcventura, line 17: Did you run into some problems when these tests where running in parallel? For example, with our .Net core test apps, we can't run multiple `dotnet publish` commands on a single test app concurrently because it results in publishing errors. Was there some shared resource in the database that caused a problem? I'm mostly asking because we have the other MySql tests and are wondering if they will also interfere with these tests?

tests/Agent/IntegrationTests/UnboundedIntegrationTests/RemoteServiceFixtures/MySqlConnectorBasicMvcFixture.cs
bgrainger: At the potential cost of clarity... it may not be immediately clear how lines 24-29 are different. (I do tend to use it with caution only in places where it is likely to be quickly understood.)

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
bgrainger: Yes; this could _probably_ even be pushed down into a base class for all MySQL ADO.NET tests (Connector/NET and MySqlConnector) although I'm wary of introducing dependencies between the two. All MySqlConnector methods should produce the same results, but that's not necessarily the case (long term) with MySql.Data.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
bgrainger: I was running the tests through the R# Unit Test Runner in VS2019. I would get an error from the fixture setup (sorry for light on details, this is from memory) from `StartWebServer`: https://github.com/newrelic/newrelic-dotnet-agent/blob/2b8e8cdc71146b1196e6a455e981be37fe789532/tests/Agent/IntegrationTests/HostedWebCore/HostedWebCore.cs#L70

The HRESULT converted to an Exception was something like "Cannot access the file when it's already in use". I tried using ProcMon to see what file might have a sharing violation but it didn't reveal anything. So I started to wonder if multiple instances of `ClassFixture<RemoteServiceFixtures.MySqlConnectorBasicMvcFixture>` were being instantiated at the same time (one per test class) and stomping on each other.

It seemed like you had multiple test classes that all used the same fixture presumably successfully (e.g., all the Postgres tests) so it may be the R# test runner that's causing the problem? I wanted to try using some different test runners and raising the issue directly if I was still getting blocked on it.

For now, this was a workaround I was using to run and debug the tests, but I'd like to remove it if it's not truly necessary.

bgrainger: Here's the test failure I got:

```
Test Name:	NewRelic.Agent.UnboundedIntegrationTests.MySqlConnectorExecuteNonQueryTest.Test
Test Outcome:	Failed
Result Message:	
Remote process never generated a .pid file!
Expected: True
Actual:   False
Result StandardOutput:	
IntegrationTestAppName
Test HomeC:\IntegrationTestWorkingDirectory\BasicMvcApplication_a921ef30-eca8-4175-86a7-cce2037547bd\newrelichome

====== [RemoteWebApplication]: WaitForHostedWebCoreToStartListening standard output log =====
[9/16/2020 8:19:58 PM 45524-1] HostedWebCore: Firing up...args: --port=54687
[9/16/2020 8:19:58 PM 45524-1] HostedWebCore: Starting directory: C:\IntegrationTestWorkingDirectory\BasicMvcApplication_a921ef30-eca8-4175-86a7-cce2037547bd\HostedWebCore
[9/16/2020 8:19:58 PM 45524-1] HostedWebCore: Environment Variables:   [snip]
[9/16/2020 8:19:58 PM 45524-1] HostedWebCore: Starting server...
[9/16/2020 8:19:59 PM 45524-1] HostedWebCore: HostedWebCore.exe failed: Reason unknown.: System.IO.FileLoadException: The process cannot access the file because it is being used by another process. (Exception from HRESULT: 0x80070020)
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode)
   at HostedWebCore.HostedWebCore.StartWebServer() in C:\newrelic-dotnet-agent\tests\Agent\IntegrationTests\HostedWebCore\HostedWebCore.cs:line 71
   at HostedWebCore.HostedWebCore.Run() in C:\newrelic-dotnet-agent\tests\Agent\IntegrationTests\HostedWebCore\HostedWebCore.cs:line 60
   at HostedWebCore.Program.Main(String[] args) in C:\newrelic-dotnet-agent\tests\Agent\IntegrationTests\HostedWebCore\Program.cs:line 50
```

bgrainger: VS Test Explorer can run these tests in parallel just fine (I see it spinning up six separate copies of HostedWebCore.exe) so I'll delete these attributes and just use Test Explorer instead. (The ReSharper Unit Test Sessions runner seems to have problems running more than one test simultaneously, it seems?)

nrcventura: I have definitely seen that error before, and I have yet to figure out what the cause is. It only seems to happen occasionally for me. I've seen it happen locally when using Test Explorer, and I've seen it when running the tests using our Integration Tests script. I also thought it had something to do with trying to start up multiple instances of the same app at the same time. There might be something we can do in the HostWebCore app to protect some api call(s) that we make so that we only call it for one application at a time, but we haven't figured out how to reliably reproduce that problem yet.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MySqlConnectorTests.cs
nrcventura: Yes, I think that it makes sense to keep the tests for these libraries separate.

nrcventura:
These tests are looking good. Are there more changes that you have in mind before moving this PR out of Draft?

bgrainger:
I never finished the stored procedure tests, but I don't think that needs to hold up the integration of this PR; those could be added separately later.

(The code that this is testing is already merged, so why delay the tests any further?)

nrcventura:
ci test this please

nrcventura (approved):

PR #225, opened 2020-09-16
AddNewRelicStartupFilter should only be added once.

vuqtran88:
### Description
The current agent is not compatible with .Net 5.0 RC1. This is due to an introduction of a new method overload of the ConfigureWebHost method in .Net 5.0 RC1. The agent currently subscribes to any of this method overload to add the custom AddNewRelicStartupFilter startup filter into the application pipeline in order to instrument Asp.net core calls. Since introduction of the new method overload, the agent subscribed to two method overloads and they were both executed. As a result, the agent added more than one AddNewRelicStartupFilter startup filter into the pipeline. It should add only one. This PR fixes this.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -17,13 +17,10 @@ Enables nullable reference types that are part of the C# 8.0 language specificat
 * **Improved Support for NetTCP Binding in WCF Instrumation**
 When the NetTCP Binding type is used in Windows Communication Foundation (WCF), the Agent will now send and receive trace context information in support of Distributed Tracing (DT) or Cross Application Tracing (CAT).  Implements [#209](https://github.com/newrelic/newrelic-dotnet-agent/issues/209).
 
-* **.NET 5 Prerelease Support**<br/>
-Adds support for .NET 5 prerelease. 
-
 ### Fixes
 * Fixes an issue that may cause `InvalidCastException` due to an assembly version mismatch in Mvc3 instrumentation.
 * Fixes an async timing issue that can cause the end time of `Task`-returning methods to be determined incorrectly.
-* Fixes an issue that causes .NET 5 prerealease incompatability.
+* Fixes [#223](https://github.com/newrelic/newrelic-dotnet-agent/issues/223) that causes the agent incompatible with ASP.NET Core 5 RC1 applications.
 
 ## [8.31] - 2020-08-17
 ### New Features
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -20,7 +20,7 @@ When the NetTCP Binding type is used in Windows Communication Foundation (WCF), 
 ### Fixes
 * Fixes an issue that may cause `InvalidCastException` due to an assembly version mismatch in Mvc3 instrumentation.
 * Fixes an async timing issue that can cause the end time of `Task`-returning methods to be determined incorrectly.
-* Fixes [#223](https://github.com/newrelic/newrelic-dotnet-agent/issues/223) that causes the agent incompatible with ASP.NET Core 5 RC1 applications.
+* Fixes [#223](https://github.com/newrelic/newrelic-dotnet-agent/issues/223) so the agent can be compatible with ASP.NET Core 5 RC1.
 
 ## [8.31] - 2020-08-17
 ### New Features
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -17,9 +17,13 @@ Enables nullable reference types that are part of the C# 8.0 language specificat
 * **Improved Support for NetTCP Binding in WCF Instrumation**
 When the NetTCP Binding type is used in Windows Communication Foundation (WCF), the Agent will now send and receive trace context information in support of Distributed Tracing (DT) or Cross Application Tracing (CAT).  Implements [#209](https://github.com/newrelic/newrelic-dotnet-agent/issues/209).
 
+* **.NET 5 Prerelease Support**<br/>
+Adds support for .NET 5 prerelease. 
+
 ### Fixes
 * Fixes an issue that may cause `InvalidCastException` due to an assembly version mismatch in Mvc3 instrumentation.
 * Fixes an async timing issue that can cause the end time of `Task`-returning methods to be determined incorrectly.
+* Fixes an issue that causes .NET 5 prerealease incompatability.
 
 ## [8.31] - 2020-08-17
 ### New Features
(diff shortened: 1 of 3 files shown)


src/Agent/CHANGELOG.md
```diff
@@ -17,9 +17,13 @@ Enables nullable reference types that are part of the C# 8.0 language specificat
 * **Improved Support for NetTCP Binding in WCF Instrumation**
 When the NetTCP Binding type is used in Windows Communication Foundation (WCF), the Agent will now send and receive trace context information in support of Distributed Tracing (DT) or Cross Application Tracing (CAT).  Implements [#209](https://github.com/newrelic/newrelic-dotnet-agent/issues/209).
 
+* **.NET 5 Prerelease Support**<br/>
+Adds support for .NET 5 prerelease.
```
nrcventura, line 21: Should we be more specific and say that we support .NET 5 RC 1?

src/Agent/CHANGELOG.md
```diff
@@ -17,9 +17,13 @@ Enables nullable reference types that are part of the C# 8.0 language specificat
 * **Improved Support for NetTCP Binding in WCF Instrumation**
 When the NetTCP Binding type is used in Windows Communication Foundation (WCF), the Agent will now send and receive trace context information in support of Distributed Tracing (DT) or Cross Application Tracing (CAT).  Implements [#209](https://github.com/newrelic/newrelic-dotnet-agent/issues/209).
 
+* **.NET 5 Prerelease Support**<br/>
+Adds support for .NET 5 prerelease. 
+
 ### Fixes
 * Fixes an issue that may cause `InvalidCastException` due to an assembly version mismatch in Mvc3 instrumentation.
 * Fixes an async timing issue that can cause the end time of `Task`-returning methods to be determined incorrectly.
+* Fixes an issue that causes .NET 5 prerealease incompatability.
```
nrcventura, line 26: We should be more specific and reference the issue in this changelog entry. For example, a .NET 5 app that does not use aspnetcore would have been just fine.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore/GenericHostWebHostBuilderExtensionsWrapper.cs
```diff
@@ -12,6 +12,7 @@ namespace NewRelic.Providers.Wrapper.AspNetCore
     public class GenericHostWebHostBuilderExtensionsWrapper : IWrapper
     {
         public bool IsTransactionRequired => false;
+        private static bool _isNewRelicStartupFilterAdded;
```
nrcventura, line 15: Does this need to be `volatile` or do we need to use a `Interlocked.CompareExchange`? To answer this question we need to know if this wrapper is ever executed concurrently.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore/GenericHostWebHostBuilderExtensionsWrapper.cs
```diff
@@ -32,20 +33,24 @@ private static class AspNetCore21Types
         {
             public static void AddStartupFilterToHostBuilder(object hostBuilder, IAgent agent)
             {
-                var typedHostBuilder = (Microsoft.Extensions.Hosting.IHostBuilder)hostBuilder;
-                typedHostBuilder.ConfigureServices(AddStartupFilter);
-
-                void AddStartupFilter(Microsoft.Extensions.Hosting.HostBuilderContext hostBuilderContext, IServiceCollection services)
+                if (!_isNewRelicStartupFilterAdded)
```
nrcventura, line 36: See the comment I added to this field to determine if this check is "safe".

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Owin/ResolveAppWrapper.cs
```diff
@@ -11,7 +11,7 @@ namespace NewRelic.Providers.Wrapper.Owin
     public class ResolveAppWrapper : IWrapper
     {
         public bool IsTransactionRequired => false;
-
+        private static bool _isNewRelicMiddlewareAdded;
```
nrcventura, line 14: The same "safety" check concerns that apply to aspnetcore apply to owin as well. We need to see if this wrapper is ever executed concurrently.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore/GenericHostWebHostBuilderExtensionsWrapper.cs
vuqtran88: I don't know for sure if this wrapper will be ever executed concurrently as both of the `ConfigureWebHost` method overloads are public. At the current state of the aspnetcore repo, the 2 overloads are executed one after another as the result of one calling another.

https://github.com/Tratcher/aspnetcore/blob/master/src/Hosting/Hosting/src/GenericHostWebHostBuilderExtensions.cs

nrcventura: I'm fine with keeping the solution simpler for now.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore/GenericHostWebHostBuilderExtensionsWrapper.cs
nrcventura, line 36: After thinking about this more, I'd prefer to treat this block of code as a critical section. If we did something like:
```
if (0 == Interlocked.CompareExchange(ref _isNewRelicStartupFilterAdded, 1, 0))
```
That should help ensure that even in a concurrent scenario we will only ever call `ConfigureServices` once.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Owin/ResolveAppWrapper.cs
```diff
@@ -23,16 +23,21 @@ public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
 
         public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
         {
-            var context = instrumentedMethodCall.MethodCall.MethodArguments[0];
+            if (!_isNewRelicMiddlewareAdded)
```
nrcventura, line 26: See the CompareExchange comment in the aspnetcore wrapper.

nrcventura (approved):

PR #230, opened 2020-09-18
Stop using Exceptions for flow control newrelic#131

geekexplorer:
Thank you for submitting a pull request.  Please review our [contributing guidelines](/CONTRIBUTING.md) and [code of conduct](/CODE_OF_CONDUCT.md).

### Description

- Updated RenameUsingRegexRules to no no longer throw an IgnoreTransactionException and instead return 'null' after logging message that would have previously been logged in the caller's catch block.
- Updated calling code to remove related try/catch blocks.
- Removed a Timer instance in the MetricNameService.NormalizeUrl() method that was being initialized and then later stopped but otherwise wasn't being used.
- Removed IgnoreTransactionException class as this is the only way it was used.
- Removed unused usings in MetricNameService.cs.

### Testing

(in MetricNameServiceTest.cs )
- Updated NormalizeUrl_Thrws_IfIgnoreRuleMatchesInput by renaming it to NormalizeUrl_ReturnsNull_IfIgnoreRuleMatchesInput and updating the test criteria to check for 'null' if the ignore rule matches the input.
- Removed try/catch from NormalizeUrl_PassesAllCrossAgentUrlTests as it is no longer needed.

### Changelog

- Due to the nature of this work no modifications to the Changelog are required.

Please remember to update our [changelog](/src/Agent/CHANGELOG.md) if applicable (or [this changelog](/src/AwsLambda/CHANGELOG.md) for Lambda agent changes),


src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
```diff
@@ -138,8 +121,11 @@ private static string RenameUsingRegexRules(string input, IEnumerable<RegexRule>
...
                 if (rule.Ignore)
-                    throw new IgnoreTransactionException($"Ignoring \"{input}\" because it matched pattern \"{rule.MatchExpression}\"", input);
-
+                {
+                    Log.Debug($"Ignoring \"{input}\" because it matched pattern \"{rule.MatchExpression}\"");
+                    return null;
+                }
+                    
                 if (ruleResult.Replacement == null)
                     throw new Exception("RuleResult matched but returned null replacement string");
```
nrcventura, line 130: Even though the previous code let this exception bubble up, and it doesn't seem like this error is likely to occur, it would be safer to not throw this exception here, and instead log the error using something like the following.
```
Log.Error("RuleResult matched but returned null replacement string");
return null;
```

src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
```diff
@@ -30,18 +28,10 @@ public class MetricNameService : ConfigurationBasedService, IMetricNameService
 
         public string NormalizeUrl(string url)
```
nrcventura, line 29: Previously this method would bubble up any exceptions, and prevent the "problematic" url from affecting the transaction name and other things. Now, this method will return `null` and the callers of this method will need to handle the `null` value appropriately.

nrcventura (changes requested):
@geekexplorer, thank you for helping us out and submitting this PR!

src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
geekexplorer: In digging into this it appears that the MetricNameSerivce.NormalizeUrl() method is called in two different cases (eventually):

- The first is by AgentApi.SetTransactionUri() method. Here a thrown exception is logged and and then swallowed.
- The second is by both ExecuteStepWrapper.TryCreateTransaction() and CreateEventExecutionStepsWrapper.BeforeEvent(). Again, in both cases it's called via an action passed into another methods.  This time by Agent.CreateTransaction().  The results of an exception being thrown here is that a wrapper error is logged. If this error occurs enough times then the wrapper is terminated.

Is this is the correct understanding of the existing behavior and is it the behavior you're looking for moving forward? If so, it seems like a significant change and I'd appreciate a nudge in the right direction. If this isn't the case, guidance on the expected behavior you're looking for would be appreciated.

Thank you.

nr-ahemsath: So, I did a little digging into the "using exceptions for flow control is an anti-pattern" issue, and it seems that the more complete idea is that you should not use exceptions for flow control in _non-exceptional_ cases.  In this case, attempting to rename a transaction according to regex rules and having it "fail" because the rule is set to be ignored, or has a null replacement string, doesn't seem like a truly exceptional case.

Regarding the fact that we want the wrappers to shut down after a number of exceptions to prevent further harm to the instrumented application, it doesn't seem like this is a case where we would want that to happen.  So, I think this PR is heading in the right direction.

One additional thought we had:
Consider replacing:
`private static string RenameUsingRegexRules(string input, IEnumerable<RegexRule> rules)`

with:
` private static bool TryRenameUsingRegexRules(string input, IEnumerable<RegexRule> rules, out string result)`

This follows a pretty common C# pattern for methods which return a result but which might not be able to return a valid result due to non-exceptional circumstances.

geekexplorer: I've updated the code based on the suggestion above to rename RenameUsingRegexRules() to TryRenameUsingRegexRules() and adjust it's related logic.  These changes do not address the issue where NormalizeUrl() is now returning null instead of bubbling up an exception if the result of TryRenameUsingRegexRules() is null.

Having discussed this with @nr-hemsath.  I'm waiting on confirmation as to whether the pre-existing behavior is actually desirable or not.  There is internal discussion that this specific case isn't one that should result in the termination of a wrapper.

geekexplorer:
Thanks for the reminder.  I'll update the code accordingly and re-commit.

src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
```diff
@@ -129,27 +121,38 @@ private static string StripParameters(string url)
             return url;
         }
 
-        private static string RenameUsingRegexRules(string input, IEnumerable<RegexRule> rules)
+        private static bool TryRenameUsingRegexRules(string input, IEnumerable<RegexRule> rules, out string result)
         {
+            result = null;
```
nr-ahemsath, line 126: Initializing `result` to null will cause this method to return null if no rules match (or if there are no rules in the `IEnumerable`).  This is causing a fair number of unit test failures.  I tested initializing `result` to `input` and all the unit tests passed.  I tried committing this change to a new branch on your fork, but I got a permissions error when I tried pushing it up, so I'll have to continue being a "backseat coder" and ask you to make this change yourself.

geekexplorer: I'm not sure why you were unable to commit.  I can check into it...  Can you send along the specific error message?

As for the error you mentioned... for some reason when I ran my tests after making the initial changes they succeeded even though, as you note, they shouldn't have.  Later, they started failing for the reason you mentioned above. After setting result = input they continued to fail.  I'm going to do a clean -dfx and give it another go to see if I can clear things out.

src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
```diff
@@ -30,18 +28,18 @@ public class MetricNameService : ConfigurationBasedService, IMetricNameService
 
         public string NormalizeUrl(string url)
         {
-            ITimer timer = new Timer();
```
nr-ahemsath, line 33: Side note: I'm not sure what this `timer` was doing in here in the first place.  It does seem safe to remove.

src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
nr-ahemsath: Okay, sorry for taking so long to respond on this.  Part of what was confusing me is that the original `RenameUsingRegexRules()` method could return two different exception types: the `IgnoreTransactionException` which we have been discussing in the case where a transaction should be ignored, and a generic `Exception()` in the case where a rule was matched with an null replacement string.  I wanted to walk back the call chains leading to the `RenameUsingRegexRules()` method and make sure both cases were handled as expected.  However, it does not appear that the second case with the generic `Exception()` was ever handled on the path branching backwards from `RenameMetric() -> RenameUsingRegexRules()`.  I confirmed this by modifying `RenameMetric()` to always just throw a generic `Exception()` and saw what broke, which ended up being most unit tests and the one integration test I tried (`BasicMvcApplication()`).  What happened was the agent shut down almost immediately.  This leads me to conclude that we never have hit this case in the wild, or we would have heard about it by now.  So, I think we can stop worrying about that case (`RenameMetric()` handles having a null returned to it by simply returning the original input so the behavior has not changed, other than instead of the agent shutting down it will log an error and continue on).

Moving back to what happens on the `NormalizeUrl()` path:
1. I am okay with no longer having a wrapper shut down because it is producing transaction names which are being ignored due to regex rules.
2. However, it isn't clear to me (even after looking at things for a good while) how returning null from `NormalizeUrl()`  if `RenameUsingRegexRules()` matched an "Ignore" rule causes the transaction to be ignored.  I tried to modify the `Rules` integration test to prove this out, by modifying the `RulesWebApi` fixture to have an `UrlIgnoreRule` method, making a request to an `api/UrlIgnoreRule` endpoint in the `Owin2WebApi` test application.  However, even without having applied any rules to match that pattern, I don't see the expected transaction with `UrlIgnoreRule` in the name, so I'm scratching my head about that.  Anyway, the point remains that we need to convince ourselves that returning null from `NormalizeUrl()` has the desired effect of ignoring a transaction before approving and merging these changes.

nr-ahemsath:
We are still planning on working on this, but it hasn't taken priority over other work in progress.

geekexplorer:
Alex and I are closing this PR and creating an new on as we're getting back to this after a few months and want to start from a clean slate.

PR #235, opened 2020-09-23
Fix HttpClient instrumentation for .NET 5

vuqtran88:
### Description
The current agent is not fully compatible with the new `HttpClient` in .NET 5 RC1. This PR brings support to the new HttpClient.

Starting from .NET 5 RC1, the agent will instrument `System.Net.Http.SocketsHttpHandler.SendAsync` and  `System.Net.Http.SocketsHttpHandler.Send` instead of `System.Net.Http.HttpClient.SendAsync` to cover all of the new `HttpCleint` code path.

The profiler has changed to detect assembly version of the `Syste.Net.Http` assembly, to ignore injecting byte codes in the `HttpClient.SendAsync` method for .Net 5 or `SocketsHttpHandler.SendAsync`/`SocketsHttpHandler.Send` for older .Net framework. With this change, the agent will still continue to instrument `System.Net.Http.HttpClient.SendAsync` for application that doesn't target `.net5.0` framework. 

### Testing

No .Net 5 RC1 tests added because the test servers aren't ready to test .Net 5 RC1 tests. Manual validations were done.


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/NoOp.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using NewRelic.Agent.Api;
+using NewRelic.Agent.Api.Experimental;
+using NewRelic.Agent.Extensions.Providers.Wrapper;
+using NewRelic.SystemExtensions;
+
+namespace NewRelic.Providers.Wrapper.HttpClient
+{
+    public class NoOp : IWrapper
```
alanwest, line 17: How about we only include this wrapper for .NET Core? So:

```csharp
#if NETSTANDARD2_0

<entire contents of this file.>

#endif
```

alanwest: Ideally, we'd spend some effort on introducing the ability to specify a version in the xml and have the profiler read the version... Then we wouldn't need this wrapper at all.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
```diff
@@ -127,23 +162,29 @@ private static void TryAttachHeadersToRequest(IAgent agent, HttpRequestMessage h
...
             {
-                if (!ValidTaskResponse(response) || (segment == null))
+                if (response == null || segment == null)
                 {
                     return;
                 }
 
-                var result = response?.Result;
+                var httpStatusCode = response.StatusCode;
+                externalSegmentData.SetHttpStatusCode((int)httpStatusCode);
```
alanwest, line 185: Why do you need the cast to `int` here?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
```diff
@@ -95,6 +127,9 @@ private static Uri TryGetAbsoluteUri(HttpRequestMessage httpRequestMessage, Syst
             if (httpRequestMessage.RequestUri?.IsAbsoluteUri == true)
                 return httpRequestMessage.RequestUri;
 
+            if (httpClient == null)
```
alanwest, line 130: Did we test to see what happens when you instantiate an `HttpClient` with a base address and then make a call to a relative URI?

src/Agent/CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ Fixes Issue [#XX](https://github.com/newrelic/newrelic-dotnet-agent/issues/XX)
 
 ### Fixes
 * Fixes [#223](https://github.com/newrelic/newrelic-dotnet-agent/issues/223) so the agent can be compatible with ASP.NET Core 5 RC1.
-
+* Fixes `HttpClient` instrumentation for .NET 5 RC1. See [#227](https://github.com/newrelic/newrelic-dotnet-agent/issues/227).
```
alanwest, line 16: How about:

"Fixes issue in .NET 5 applications where external calls made with `HttpClient` may not get instrumented. For example, calls made with `HttpClient.GetStringAsync` would be missed. ([#235](https://github.com/newrelic/newrelic-dotnet-agent/pull/235))"

I like the practice of noting the PR that fixed the bug rather than the issue in the release notes. The PR itself should reference the issue.

alanwest:
PR name could be describe things better. I suggest "Fix HttpClient instrumentation for .NET 5"

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/NoOp.cs
alanwest, line 17: I think we should spend a little time seeing if we can get away without the need for this wrapper. I'd prefer to avoid needlessly instrumenting things. Ideally, our XML would support the ability to specify a version, but it does not. A step in that direction though would be to see what it would take to check the version of these types in the profiler. If it's easy enough to identify the version at the time of JIT specifically for these methods, then I would be content with special casing these methods in the profiler for now, and putting the XML support in our backlog.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
vuqtran88: This is existed code.

alanwest:
Will you planning on adding integration tests in a follow up?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
alanwest: Ok, but why is the cast needed? The previous code had a `HasValue` check which you removed. We need to understand if this cast is still needed and why.

vuqtran88:
> Are you planning on adding integration tests in a follow up?

.Net 5 tests are currently tested on a different branch now, but yes, I'm planning to add those tests a follow up.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
vuqtran88: Since we instrument `SocketsHttpHandler` methods in .Net 5 to cover every `HttpClient` call paths, we could only get the Uri from the `HttpRequestMessage`. Luckily, my test shows that the absolute uri is already available by the time the SocketsHttpHandler is running. I have tested the following test case:

```
           HttpClient client = new();
            client.BaseAddress = new Uri("https://localhost:5001");
            Forecast[] forecasts = await client.GetFromJsonAsync<Forecast[]>("WeatherForecast");
```

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/NoOp.cs
alanwest: The code in the profiler that decides whether to instrument is here https://github.com/newrelic/newrelic-dotnet-agent/blob/e264383d04c924823ed47760647407de4656f0b1/src/Agent/NewRelic/Profiler/MethodRewriter/Instrumentors.h#L30-L34

You may need to pass a handle to `ICorProfilerInfo` into this method so that you can do the appropriate metadata lookup to get the version from either the assembly or module.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
alanwest: Ok, let's get this test case incorporated into your integration tests that you plan to bring in.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
vuqtran88: The cast seems to be still needed (sooner or later) because the statuscode  attribute is type `long`. https://github.com/newrelic/newrelic-dotnet-agent/blob/2b8e8cdc71146b1196e6a455e981be37fe789532/src/Agent/NewRelic/Agent/Core/Segments/ExternalSegmentData.cs#L95

About the `HasValue` check removal. My refactoring makes sure the `response` object is never `null` when retrieving status code from this object.

src/Agent/NewRelic/Profiler/Profiler/Function.h
```diff
@@ -335,6 +336,15 @@ namespace NewRelic { namespace Profiler
                 _tracerFlags |= NewRelic::Profiler::Configuration::TracerFlags::AsyncMethod;
             }
 
+            mdAssembly mda = 0;
+            _metaDataAssemblyImport->GetAssemblyFromScope(&mda);
```
nrcventura, line 340: We should consider using the ThrowOnError helper method for this api call. At the very least we'll need to check the hresult by using the SUCCEEDED macro or comparing against S_OK.

src/Agent/NewRelic/Profiler/Profiler/Function.h
```diff
@@ -335,6 +336,15 @@ namespace NewRelic { namespace Profiler
...
+
+            LPWSTR szName = new TCHAR[200];
+            ULONG cchName = 200;
+            ULONG* pchName = new ULONG;
+            _assemblyProps = ASSEMBLYMETADATA();
+            _metaDataAssemblyImport->GetAssemblyProps(mda, 0, 0, 0, szName, cchName, pchName, &_assemblyProps, 0);
```
nrcventura, line 346: We also need to use the ThrowOnError helper or at least check the hresult here.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -67,6 +67,12 @@ namespace NewRelic { namespace Profiler { namespace Configuration
             ipToFind->MethodName = function->GetFunctionName();
             ipToFind->Parameters = std::unique_ptr<xstring_t>(new xstring_t(methodSignature->ToString(function->GetTokenResolver())));
 
+            if (Strings::AreEqualCaseInsensitive(function->GetAssemblyName(), _X("System.Net.Http"))
```
nrcventura, line 70: It will be easier for us to understand the conditions in this if statement if the logic was broken out into 2 helper methods: 1. IsHttpClient5OrGreater, and 2. IsSocketsHttpHandlerLessThan5. It would also be helpful to add a comment that these hard-coded exceptions should be temporary and can be removed if version checking is added to the instrumentation xml.

nrcventura (changes requested):

src/Agent/CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ Fixes Issue [#XX](https://github.com/newrelic/newrelic-dotnet-agent/issues/XX)
 
 ### Fixes
 * Fixes [#223](https://github.com/newrelic/newrelic-dotnet-agent/issues/223) so the agent can be compatible with ASP.NET Core 5 RC1.
-
+* Fixes issue in .NET 5 applications where external calls made with HttpClient may not get instrumented. For example, calls made with `HttpClient.GetStringAsync` would be missed. ([#235](https://github.com/newrelic/newrelic-dotnet-agent/issues/235))
```
nrcventura, line 16:
```suggestion
* Fixes issue in .NET 5 applications where external calls made with HttpClient may not get instrumented. For example, calls made with `HttpClient.GetStringAsync` would be missed. ([#235](https://github.com/newrelic/newrelic-dotnet-agent/pull/235))
```

nrcventura (approved):

PR #261, opened 2020-09-30
Profiler - .NET Standard Modules without reference to mscorlib fail to be instrumented in .NET Framework

### LINKED ISSUE #257: Profiler fails to Instrument .NET Standard libraries that do not reference mscorlib in .NET Framework applications
**Description**
The Profiler is responsible for injecting byte-code (MSIL) into libraries that are to be instrumented.  In doing so, it rewrites helper methods in the `CannotUnloadAppDomainException` class that are never used.  It then instruments libraries by injecting IL that references these functions.

When a .NET Standard library that does _not_ have a reference to `mscorlib` is instrumented, the references to mscorlib are not valid and instrumentation fails. (gracefully)

**Expected Behavior**
The helper methods should be injected and references to them work properly.

[NOTE]: # ( Provide any other relevant log data such as the agent and/or profiler logs, preferably at DEBUG level or higher. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

**Steps to Reproduce**
1.  Build a .Net Framework application that uses Azure Cosmos DB.
2.  Build Custom Instrumentation for Cosmos DB
3.  Make a CosmosDB call within a transaction of your application.
4.  Notice it is not instrumented.

**Your Environment**
Windows .Net Framework application

**Additional context**
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

jifeingo:
### Description
Fixes #257.  The profiler instruments .NET Framework applications by injecting helper methods into `System.CannotUnloadAppDomainException` and references to these helper methods into other modules.  All of this assumes that `mscorlib` is referenced by the assembly.

A .NET Standard library may not reference `mscorlib`.  An example of this occurs with `Micorosoft.Azure.DocumentDB.Core` (v2.6).  Prior to this change, the module injection failed due to invalid references to `mscorlib` types.  As a result, the library was not instrumented.

This change:
* Detects a .NET Standard Module that does not have a reference to `mscorlib` and injects it.
* Explicitly identifies types to be in `[mscorlib]` for injecting references to mscorlib.

### Testing
* Relied on existing Integration tests to ensure that the profiler executes similiar to before
* Manual testing of CosmosDB document library custom instrumentation.


src/Agent/NewRelic/Profiler/ModuleInjector/ModuleInjector.h
```diff
@@ -92,5 +117,53 @@ namespace NewRelic { namespace Profiler { namespace ModuleInjector
...
+                ZeroMemory(&amd, sizeof(amd));
+                amd.usMajorVersion = 4;
+                amd.usMinorVersion = 0;
+                amd.usBuildNumber = 0;
+                amd.usRevisionNumber = 0;
+
+                auto metaDataAssemblyEmit = module.GetMetaDataAssemblyEmit();
+                mdAssemblyRef assemblyToken;
+                const BYTE pubToken[] = { 0xB7, 0x7A, 0x5C, 0x56, 0x19, 0x34, 0xE0, 0x89 };
+
```
jifeingo, line 145: This is probably the most controversial item.  Needs to be a strongly referenced assembly with the token.

nrcventura: This will probably be okay so long as mscorlib has been using the same strong name key for .net framework 4.5+.

nrcventura: We confirmed that the token from framework 4.5 was the same as the one for framework 4.8, so this appears to be a safe thing to hard-code at this point.

nrcventura (approved):

PR #263, opened 2020-09-30
Save integration test artifacts

### LINKED ISSUE #236: Configure workflow to gather artifacts Integration Tests Build
### Feature Description
Configure workflow to gather artifacts Integration Tests Build

- From job: build-integration-tests
- Need the test applications: To make setting up the folder structure easier, just grabbing the whole project is one approach.  It should already be built.
- Needs integration tests assemblies: Includes Shared, Helpers, etc.
- Needs to be labelled with the runid.  Example: integrationtests_<runid>
- Avoid spaces in file names

jaffinito:
Resolves #236

### Description

Add a step to archive the integration test artifacts.
- Only save the bin and Deploy folders since the checkout will have the rest of the files
- Relies on the checkout to pull the code (won't be needed most of the time, bit its already getting pulled in)
- Artifacts are currently ~400mb in size and take ~5 minutes to upload (full download should be about the same)
- Artifacts have been built to have the repo root as the root of the package to make overlaying the files easier.  (Was done by adding `test.runsettings` to artifacts -- also noted in yml).

The idea is that after checking out the branch, the artifacts can be pulled down on top of the branch resulting in all the required files being in place.

#### Addendum

Adds `if-no-files-found: error` to all the artifacts uploads so that jobs will fail if the upload has no files.  It defaults to warning.  This will cause PRs with missing artifacts to not be mergable which is correct.

### Testing

The initial test is that the workflow builds AND produces an integration test artifact that is around ~400mb in size. As we being to get the integration test matrix jobs running we will likely need to adjust the artifacts slightly.

### Changelog

n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -65,6 +65,7 @@ jobs:
           ${{ github.workspace }}\src\Agent\newrelichome_x64_coreclr_linux
           ${{ github.workspace }}\src\Agent\newrelichome_x86
           ${{ github.workspace }}\src\Agent\newrelichome_x86_coreclr
+        if-no-files-found: error
 
     - name: Create Self-signed code signing cert
       run: |
@@ -138,6 +139,7 @@ jobs:
       with:
         name: test-results-${{ github.run_id }}
         path: ${{ github.workspace }}\TestResults
+        if-no-files-found: error
 
   publish-test-results:
     needs: build-test-fullagent-msi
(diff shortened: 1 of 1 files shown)


lspangler584 (approved):
the `Archive Artifacts` step completed successfully on this very PR; the workflow is behaving as expected. 

image

nr-ahemsath (approved):
LGTM 👍

PR #270, opened 2020-10-01
Group existing unbounded test docker services as linux

vuqtran88:
### Description
Group existing unbounded test docker services as linux to make way for supporting windows unbounded test docker services.

I have tested the script change locally on my box.


build/Scripts/run-integration-tests.ps1
```diff
@@ -20,7 +22,7 @@ if ($saveWorkingFolders) {
 
 $rootDirectory = Resolve-Path "$(Split-Path -Parent $PSCommandPath)\..\.."
 $xUnitPath = Resolve-Path "$rootDirectory\build\Tools\XUnit-Console\xunit.console.exe"
-$unboundedServicesControlPath = Resolve-Path "$rootDirectory\tests\Agent\IntegrationTests\UnboundedServices\unbounded-services-control.ps1"
+$unboundedServicesControlPath = Resolve-Path "$rootDirectory\tests\Agent\IntegrationTests\UnboundedServices\$platform\unbounded-services-control.ps1"
```
nr-ahemsath, line 25: Rather than move this script into the `linux` subfolder, and presumably duplicate it in the `windows` subfolder, could you leave the script where it is and modify it to handle both cases?

vuqtran88: That's a good idea.

nr-ahemsath (approved):
Looks good.  Thanks for updating the documentation. 👍

PR #272, opened 2020-10-02
Fixes the problem of home folders not using profiler built during workflow run

### LINKED ISSUE #255: Agent artifact didn't get updated profiler.
**Description**
The `libNewRelicProfiler.so` profiler from `linuxprofiler_xxxx` artifact is different from the one in the `homefolder_xxxx` artifact. This causes not testing the correct profiler.

jaffinito:
Resolves #255 

### Description

Reorders workflow
- Moves profiler builds to top of workflow so they occur first
- Profiler builds clear our _profilerBuild to ensure fresh files are use
- FullAgent now needs profiler builds
- Full Agent also clears _profilerBuild and pulls new profiler artifacts
- As a result homefolders now contain only files built during that run

### Testing

- Build should complete.
- The profiler-<runid>  should have 3 folders: linux-release, x64-Release, x86-Release.  This folders should have the profiler assemblies.
- Check that the profiler assemblies in the profiler-<runid> artifact match the one in the homefolder-<runid> artifact.

### Changelog

n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -15,8 +15,107 @@ env:
 
 jobs:
 
+  build-test-windows-profiler:
+    name: Build Windows Profiler
+
+    runs-on: windows-2019
+
+    env:
+      profiler_path: ${{ github.workspace }}\src\Agent\NewRelic\Profiler
+      profiler_solution_path: ${{ github.workspace }}\src\Agent\NewRelic\Profiler\NewRelic.Profiler.sln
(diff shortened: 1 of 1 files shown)


nr-ahemsath (approved):
Logic looks good, and I verified that the CRC32 and SHA256 of the 64-bit Windows profiler was the same for the profiler artifacts and the version in the 64-bit framework home folder artifact.  👍

nr-ahemsath (approved):
Approved again!

PR #273, opened 2020-10-03
Fixes the test results publishing issue with re-running a workflow

### LINKED ISSUE #256: PR rerun fails due to existing testresults folder
**Description**
In the `.NET Agent All Solutions Build` Action, the `Publish Test Results` job should create a unique folder name for testresults or remove the existing testresults folder. 
In the current state, if the Action is rerun on the same PR a `ResourceExists` exception is generated.

**Expected Behavior**
Rerunning a PR build should not fail due to an existing `testresults` folder.

**Steps to Reproduce**
Cause an existing PR to be rerun (for instance, by pushing a new commit).
image

**Your Environment**
GitHub Actions

jaffinito:
Resolves #256

### Description

- The publishing of the test results is failing if the workflow is re-run due to the folder structure already existing.
- Updates the script that moves the results to also check if the run folder exists before attempting to create it.

### Testing

First run passed as expected so I re-ran the workflow and that also ran fine, no publishing issues.

### Changelog
n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -173,13 +173,16 @@ jobs:
           $resultname = "fullagent_msi-${{ github.run_id }}".ToLower().Replace(" ", "")
           Write-Host "::set-env name=results_name::$resultname"
 
+          Write-Host "Check if testresults folder exists, create if not."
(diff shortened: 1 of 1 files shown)


nr-ahemsath (approved):
I also re-ran the workflow for this PR and the test publishing job succeeded.  Approved. 👍

PR #276, opened 2020-10-05
Run integration tests

### LINKED ISSUE #238: Create Integration Tests Matrix Job
### Feature Description
This new job will be designed to create several new jobs on the fly to run smaller parts of the integration tests across many servers automatically.
This will speed up the run and help prevent tests from stepping on each other.
The split can be done by class or by namespace.  Class gives the most granularity while namespace is simpler.  I would lean toward class.

Command details:
```
xunit.console.exe
    <assemblyFile>
    -nologo
    -class "Name"
    -namespace "Name"
    -html <filename>
```

- Job will using the matrix strategy https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy
- Decide how to setup the matix: class or namespace
- Test results need to be in html so that they can be uploaded to gh-pages.
- Will need to store artifacts, but name cannot just be <name>_<runid>. Actions does not have an easier value here so find a good approach.
- Avoid spaces in file names

nr-ahemsath:
### Description

Resolves #238.  Steps to run the integration tests have been added to the main workflow file.

Note: many of the integration tests are failing at this point, but I want to get some :eyes: on what has been done so far.

### Testing

This PR gets the (bounded) integration tests to run on every PR, so that will be an improvement to our testing process.

### Changelog

N/A


jaffinito (approved):
This is a great first pass on getting the matrix setup and running.
Since we are still using the jenkins infra for the full test suite this is fine to merge before it is 100%.

PR #283, opened 2020-10-07
Ms sql windows container

### LINKED ISSUE #244: Migrate mssql image to use a Windows image as a base
### Feature Description
Since GitHub Actions does not support Linux docker containers on Windows Hosts we need to convert our existing unbounded application images to use Windows images as their base instead of Linux.

Where possible, use an officially provided image to save time.  Unfortunately, most of our applications will not have an official Windows image.

Try and keep the version of the application as close to the one already in use as possible.

These new images should live in `tests/Agent/IntegrationTests/UnboundedServicesWin/`and be contained in their own folders (see `tests/Agent/IntegrationTests/UnboundedServices/` for an example).  Please create any folders that may be missing.

Base images can be found here: https://hub.docker.com/_/microsoft-windows-base-os-images

**Important**: The base image must match the version of the OS. We use `windows-2019` to try and limit version changes. The  current version of Windows in Actions is `10.0.17763 Build 1457` and the image versions is `20200920.1`. Server info: https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md

Application: mssql
Location in repo: tests/Agent/IntegrationTests/UnboundedServices/mssql

lspangler584:
Description
Resolves #244. Adds a Windows docker image for MSSQL.

Testing
The mssql unbounded integration tests pass locally, although with occasional flickers. These flickers will be addressed when implementing the Unbounded Test execution in GitHub CI, the intended use case for Windows containers.

Changelog
N/A

### DIFF
--- a/tests/Agent/IntegrationTests/UnboundedServices/Windows/docker-compose.yml
+++ b/tests/Agent/IntegrationTests/UnboundedServices/Windows/docker-compose.yml
@@ -11,7 +11,7 @@ services:
             # complex password than 'password' and that is why this service does
             # not follow the same pattern as the others.
             - SA_PASSWORD=password0TS
-            - SQLCMDPASSWORD=password0TS
+            - ACCEPT_EULA=Y
         container_name: MssqlServer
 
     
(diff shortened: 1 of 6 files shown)


nr-ahemsath (approved):
Approved.  👍

PR #285, opened 2020-10-07
Profiler - Maintain Instrumentation Points in a Map (Dictionary) instead of a Set (List)

### LINKED ISSUE #269: Reduce the performance impact of large amount instrumentation
**Description**
The New Relic Profiler responds to CLR events for module loading and JIT compilation.  It reads `Instrumentation.xml` files which identify the methods that are to be instrumented.  The signatures of these methods are stored in a List (Set).  When a JIT request is made for any function (instrumented or not), the list is consulted to see if the JITting function should have instrumentation instructions injected.

This process is inefficient and causes performance impacts when this list is large.  As a list, every search results in a linear scan of the list.

The problem is confounded by the fact that instrumentation is agent/system wide.  So, it affects all processes on the server.

**Recommendation**
* Instead of using a list can a look-up structure be used (such as Map) that is keyed based on the signature.
* Could a tree structure organized by Assembly -> Class -> Method Signature be used.  This would result in a much quicker exit for any assembly or class that is not instrumented.

**Steps to Reproduce**
* Build a moderately complex application
* Add an instrumentation xml with 12K methods in it.

**Your Environment**
.NET Framework (but suspect it is also in .NET Core)

**Additional context**
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

jifeingo:
### Description
* Modifies the Profiler to use a `map` structure to maintain instrumentation instead of a 'set'.
* Introduces the concept of an instrumentation point signature (a string that is used as a key)
* Modifies matching logic to use this new structure.

Fixes #269 
### Testing
No new testing required.
Integration tests pass


src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -45,52 +45,82 @@ namespace NewRelic { namespace Profiler { namespace Configuration
         }
 
         InstrumentationConfiguration(InstrumentationPointSetPtr instrumentationPoints) :
-            _instrumentationPoints(instrumentationPoints),
+            _instrumentationPointsMap(new InstrumentationPointMap()),
```
nrcventura, line 48: Instead of dynamically allocating the InstrumentationPointMap here, we should probably use a field initializer without dynamic allocation to define the map.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -45,52 +45,82 @@ namespace NewRelic { namespace Profiler { namespace Configuration
...
-            ipToFind->AssemblyName = function->GetAssemblyName();
-            ipToFind->ClassName = function->GetTypeName();
-            ipToFind->MethodName = function->GetFunctionName();
-            ipToFind->Parameters = std::unique_ptr<xstring_t>(new xstring_t(methodSignature->ToString(function->GetTokenResolver())));
+            auto result = InstrumentationPointSetPtr(new InstrumentationPointSet());
+            for (auto instrumentationPointKVP : *_instrumentationPointsMap) {
+                result->insert(instrumentationPointKVP.second);
+            }
+
+            return result;
```
nrcventura, line 69: Maybe we should keep the original `_instrumentationPoints` field in addition to the map, especially if the map is just holding references to items in the `_instrumentationPoints` set? That would avoid the additional dynamic allocation occurring in this method, which I think is never getting deallocated.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -45,52 +45,82 @@ namespace NewRelic { namespace Profiler { namespace Configuration
...
             // Temporarily specifically ignore System.Net.Http.HttpClient instrumentation for .Net 5 and greater, and System.Net.Http.SocketsHttpHandler for framework less than .Net 5
             // until version checking from instrumentation xml is supported.
             if (IsHttpClient5OrGreater(function) || IsSocketsHttpHandlerLessThan5(function))
             {
                 return nullptr;
             }
 
-            auto instPoint = TryGetInstrumentationPoint(ipToFind);
+            auto methodSignature = SignatureParser::SignatureParser::ParseMethodSignature(function->GetSignature()->begin(), function->GetSignature()->end());
+            auto params = new xstring_t(methodSignature->ToString(function->GetTokenResolver()));
```
nrcventura, line 82: Is there a better way to clone a string than dynamically allocating a new one?

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -45,52 +45,82 @@ namespace NewRelic { namespace Profiler { namespace Configuration
...
+
         InstrumentationPointPtr TryGetInstrumentationPoint(const InstrumentationPointPtr ipToFind) const
         {
-            for (auto current : *_instrumentationPoints)
+            auto const key = ipToFind->GetMatchKey();
+
+            return TryGetInstrumentationPoint(key);
+        }
+
+        InstrumentationPointPtr TryGetInstrumentationPoint(const xstring_t key) const
```
nrcventura, line 115: The `key` parameter should probably be a reference.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -45,52 +45,82 @@ namespace NewRelic { namespace Profiler { namespace Configuration
...
+            auto instPoint = TryGetInstrumentationPoint(function->GetAssemblyName(), function->GetTypeName(), function->GetFunctionName(), *params);
             return instPoint;
         }
 
     private:
 
+        InstrumentationPointPtr TryGetInstrumentationPoint(const xstring_t assemblyName,
+            const xstring_t className,
+            const xstring_t methodName,
+            const xstring_t parameters) const
```
nrcventura, line 93: These parameters should probably all be references.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -172,6 +202,12 @@ namespace NewRelic { namespace Profiler { namespace Configuration
             instrumentationPoint->MethodName = GetAttributeOrEmptyString(matcherNode, _X("methodName"));
             instrumentationPoint->Parameters = TryGetAttribute(matcherNode, _X("parameters"));
 
+            // void as the parameters means parameterless method call (not all overloads)
+            if (instrumentationPoint->Parameters != nullptr && Strings::AreEqualCaseInsensitive(*(instrumentationPoint->Parameters), _X("void")))
+            {
+                instrumentationPoint->Parameters = std::unique_ptr<xstring_t>(new xstring_t());
```
nrcventura, line 208: Is there a better way to handle this case so that we don't have to dynamically allocate a string? Are we missing the concept of something similar to C#'s `string.Empty`?

src/Agent/NewRelic/Profiler/Configuration/InstrumentationPoint.h
```diff
@@ -56,6 +57,24 @@ namespace NewRelic { namespace Profiler { namespace Configuration
...
         }
 
+        xstring_t GetMatchKey()
+        {
+            return Parameters == nullptr
+                ? GetMatchKey(AssemblyName, ClassName, MethodName)
+                : GetMatchKey(AssemblyName, ClassName, MethodName, *Parameters);
+        }
+
+        static xstring_t GetMatchKey(xstring_t assemblyName, xstring_t className, xstring_t methodName)
```
nrcventura, line 67: These parameters should probably all be `const &`.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationPoint.h
```diff
@@ -56,6 +57,24 @@ namespace NewRelic { namespace Profiler { namespace Configuration
...
+        {
+            return xstring_t(_X("[")) + assemblyName + _X("]") + className + _X(".") + methodName;
+        }
+
+        static xstring_t GetMatchKey(xstring_t assemblyName, xstring_t className, xstring_t methodName, xstring_t parameters)
```
nrcventura, line 72: These parameters should probably all be `const &`.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
jifeingo: Was able to do this for the map, but reintroduced the set and that could not be done b/c a different constructor can send it in.

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
jifeingo: done

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
jifeingo: I couldn't figure out a better way to do this.  But, I am no C++ master.

nrcventura: I think that we might be able to use something like the following `std::make_unique<SomeType>(constructor, parameters, here);`

nrcventura: I just realized that this might not be necessary at all...My modern C++ skills are not strong either. The Unique_Ptr should automatically handle that deallocation for us. So I don't think that we need to change anything here, or in the other places where we are using `new` with one of the smart pointers.

nrcventura (commented):
I think that everything looks good. Let's get the profiler binaries attached and then we can get this merged.

nrcventura (approved):

PR #286, opened 2020-10-08
Set log level to warn for event log appender

alanwest:
When an exception is thrown while the agent is initializing its log4net file appender, the agent falls back to setting up an event log appender. This is only true for .NET Framework applications - not .NET Core.

For example, one way to cause an exception during the file appender initialization is to use the following agent configuration:

```xml
<log level="info" directory="C:\SomeBogusDirectory" />
```

Since the directory does not exist, this code throws an exception:
https://github.com/newrelic/newrelic-dotnet-agent/blob/2b8e8cdc71146b1196e6a455e981be37fe789532/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs#L390-L399.

The exception is caught and the event log appender is set up here:
https://github.com/newrelic/newrelic-dotnet-agent/blob/2b8e8cdc71146b1196e6a455e981be37fe789532/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs#L349-L353

When this occurs, the event log appender uses the `info` log level defined in our example configuration. This causes an unabated blizzard of info-level log messages to appear in the event log.

This PR hard codes the event log appender's log level to `warn` preventing the blizzard, but still enables us to troubleshoot why the file appender failed to initialize (as well as any other warnings and errors from the agent).

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
+++ b/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
@@ -287,6 +287,7 @@ namespace NewRelic.Agent.Core
 #if NET45
 			var appender = new EventLogAppender();
 			appender.Layout = eventLoggerLayout;
+			appender.Threshold = Level.Warn;
 			appender.Name = EventLogAppenderName;
 			appender.LogName = EventLogName;
 			appender.ApplicationName = EventLogSourceName;
@@ -395,7 +396,7 @@ namespace NewRelic.Agent.Core
             catch (Exception exception)
             {
                 log.ErrorFormat("Unable to write the {0} log to \"{1}\": {2}", appenderName, fileName, exception.Message);
-                throw exception;
+                throw;
             }
 
             try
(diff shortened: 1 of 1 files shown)


nr-ahemsath:
How does changing the log level from `info` to `warn` prevent the blizzard of log messages?  Won't there just be a blizzard of warn-level log messages instead of info-level log messages?

nrcventura (approved):

nrcventura:
> How does changing the log level from `info` to `warn` prevent the blizzard of log messages? Won't there just be a blizzard of warn-level log messages instead of info-level log messages?

The main reason it prevents the "blizzard" is that the " harvest started" log messages that occur during every harvest are written at the `info` level. By increasing the level to `warn` there should be far fewer messages written to the log. 

The "harvest started" log messages have been a useful data point for our support team when troubleshooting the agent, so those messages should still be logged at the `info` level.

nr-ahemsath:
> The main reason it prevents the "blizzard" is that the " harvest started" log messages that occur during every harvest are written at the `info` level. By increasing the level to `warn` there should be far fewer messages written to the log.

Got it, thanks...I was confused and thought we were just changing the log level of a frequently logged message, now I understand that we're changing the logger so it only logs `warn` and higher-level messages. 👍

PR #303, opened 2020-10-14
Integration Test - Console Multifunction Application - Trim non ASCII chars on inputs

jifeingo:
### Description
The ConsoleMultifunctionApplication accepts input and performs various methods.  It is exercised by a remote fixture which manages the process and supplies commands via StdIn.   For some unknown reason, in GHA environments, .Net Core versions of the application have a non ASCII character applied to the beginning of the input.

This PR Removes any non Letter value from the input.


nr-ahemsath (approved):
Makes sense.  Compared to a run from yesterday, only 88 int tests failed as opposed to 200+.  Approved.  👍

PR #308, opened 2020-10-19
Integration Tests - Fix faulty logic in error event parsing from log files

nr-ahemsath:
### Description

The previous logic of the ExpectedErrorsTests tests assumed that all error events would be harvested in the first error event payload.  Due to changes for faster event harvest, this assumption is no longer safe.  These changes ensure that all error events are considered in the test logic.

Additionally, the previous `AgentLogBase.GetErrorEvents()` method was poorly named since it actually returned an `IEnumerable<ErrorEventPayload>`, which had the actual error events in a field.  This PR renames that method to `GetErrorEventPayloads()` and then creates a new `GetErrorEvents()` method that builds on top of it and actually returns the error events, not the payloads.

Thanks to @jifeingo for help with the LINQ query syntax.

### Testing

The integration tests build and pass for me locally.

### Changelog

N/A

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/Api/AgentApiTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/Api/AgentApiTests.cs
@@ -13,12 +13,13 @@ using Xunit.Abstractions;
 namespace NewRelic.Agent.IntegrationTests.Api
 {
     [NetFrameworkTest]
-    public class AgentApiTests : IClassFixture<RemoteServiceFixtures.AgentApiExecutor>
+    public class AgentApiTests : NewRelicIntegrationTest<RemoteServiceFixtures.AgentApiExecutor>
     {
 
         private readonly RemoteServiceFixtures.AgentApiExecutor _fixture;
 
         public AgentApiTests(RemoteServiceFixtures.AgentApiExecutor fixture, ITestOutputHelper output)
+            : base(fixture)
         {
             _fixture = fixture;
             _fixture.TestLogger = output;
--- a/tests/Agent/IntegrationTests/IntegrationTests/Api/AgentApiTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/Api/AgentApiTests.cs
@@ -56,18 +56,13 @@ namespace NewRelic.Agent.IntegrationTests.Api
 
             var actualMetrics = _fixture.AgentLog.GetMetrics();
             var errorTrace = _fixture.AgentLog.GetErrorTraces().FirstOrDefault();
-            var errorEventPayload = _fixture.AgentLog.GetErrorEventPayloads().FirstOrDefault();
-
-            //order events chronologically
-            var errorEvents = errorEventPayload.Events
-                .Where(ev => ev.IntrinsicAttributes != null && ev.IntrinsicAttributes.ContainsKey("timestamp"))
-                .OrderBy(ev => ev.IntrinsicAttributes["timestamp"]).ToList();
+            var errorEvents = _fixture.AgentLog.GetErrorEvents().OrderBy(ev => ev.IntrinsicAttributes["timestamp"]).ToList();
 
             NrAssert.Multiple
                 (
                     () => Assertions.MetricsExist(expectedMetrics, actualMetrics),
                     () => Assert.NotNull(errorTrace),
-                    () => Assert.Equal("2", errorEventPayload.Additions.EventsSeen.ToString()),
+                    () => Assert.Equal(2, errorEvents.Count),
                     () => Assertions.ErrorEventHasAttributes(expectedErrorEventIntrinsicAttributes, EventAttributeType.Intrinsic, errorEvents[0]),
                     () => Assertions.ErrorEventHasAttributes(expectedErrorEventIntrinsicAttributes, EventAttributeType.Intrinsic, errorEvents[1]),
                     () => Assertions.ErrorEventHasAttributes(expectedErrorEventCustomAttributes, EventAttributeType.User, errorEvents[1])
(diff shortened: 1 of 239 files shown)


lspangler584 (approved):
Looks good. Test passes in GHA and locally for me as well.

tests/Agent/IntegrationTests/IntegrationTests/Errors/ExpectedErrorTests.cs
```diff
@@ -77,16 +77,17 @@ public void Test()
             };
 
             var errorTraces = _fixture.AgentLog.GetErrorTraces().ToList();
-            var errorEvents = _fixture.AgentLog.GetErrorEvents().ToList();
+            var errorEventPayloads = _fixture.AgentLog.GetErrorEvents().ToList();
+            var errorEvents = errorEventPayloads.SelectMany(payload => payload.Events).ToList();
```
jifeingo, line 81: Is there a way to build this into the Agent Log feature for all event types?   Seems like this functionality could be useful to many tests in making them less brittle.

nr-ahemsath: It's a fair point.  I did consider doing something like this.  A lot of tests use the `GetErrorEvents()` method so it would require changing a lot of test code.

One thing approach I'd explore if we decided to spend more time on this, is rename `GetErrorEvents` to `GetErrorEventPayloads`, because that's what it's ACTUALLY returning.  Then I'd probably need to go look at each test which uses that method and see what it's actually trying to assert on (for example, if it's just asserting on the absence of error events, then making sure that there are no error event payloads is correct, but if it really wants to know about error events than it might make sense to use a new method which returns the actual events).

tests/Agent/IntegrationTests/IntegrationTests/Owin/OwinMiddlewareExceptionTests.cs
```diff
@@ -83,12 +83,12 @@ public void Test()
             NrAssert.Multiple(
                 () => Assertions.MetricsExist(expectedMetrics, metrics),
                 () => Assert.NotEmpty(errorTraces),
-                () => Assert.NotEmpty(_fixture.AgentLog.GetErrorEvents()),
+                () => Assert.NotEmpty(_fixture.AgentLog.GetErrorEventPayloads()),
```
jifeingo, line 86: probably could use local `errorEvents` here.

tests/Agent/IntegrationTests/IntegrationTests/Owin/OwinWebApiTests.cs
```diff
@@ -103,7 +103,7 @@ public void Test()
                 () => Assertions.MetricsDoNotExist(unexpectedMetrics, metrics),
                 () => Assert.NotNull(transactionEventWithExternal),
                 () => Assert.Empty(_fixture.AgentLog.GetErrorTraces()),
-                () => Assert.Empty(_fixture.AgentLog.GetErrorEvents())
+                () => Assert.Empty(_fixture.AgentLog.GetErrorEventPayloads())
```
jifeingo, line 106: Optional, consider using `GetErrorEvents()` instead.   My thinking is that we can widdle down our usage of the GetErrorEventPayloads to the end of removing that method all together.

tests/Agent/IntegrationTests/IntegrationTests/Api/AgentApiTests.cs
```diff
@@ -56,7 +56,7 @@ public void Test()
 
             var actualMetrics = _fixture.AgentLog.GetMetrics();
             var errorTrace = _fixture.AgentLog.GetErrorTraces().FirstOrDefault();
-            var errorEventPayload = _fixture.AgentLog.GetErrorEvents().FirstOrDefault();
+            var errorEventPayload = _fixture.AgentLog.GetErrorEventPayloads().FirstOrDefault();
```
jifeingo, line 59: Optional, this test may benefit from just using `GetErrorEvents()` and comparing counts of that only.
Is this test specifically testing the `EventsSeen` value correctness or the number of errors recorded?

tests/Agent/IntegrationTests/IntegrationTests/CSP/HighSecurityAndCustomAttributes.cs
```diff
@@ -58,7 +58,7 @@ public void Test()
             Assertions.TransactionTraceDoesNotHaveAttributes(unexpectedTransactionTraceAttributes, TransactionTraceAttributeType.User, transactionSample);
 
             var errorTrace = _fixture.AgentLog.GetErrorTraces().FirstOrDefault();
-            var errorEventPayload = _fixture.AgentLog.GetErrorEvents().FirstOrDefault();
+            var errorEventPayload = _fixture.AgentLog.GetErrorEventPayloads().FirstOrDefault();
```
jifeingo, line 61: Could benefit from just using `GetErrorEvents`.   My goal is removing the usage of `GetErrorEvenPayloads` to the end of removing the method entirely.

jifeingo (approved):
I approve the changes.  There are few places where I had recommendations, but none are required so long as the tests pass.  Thanks, great work!

tests/Agent/IntegrationTests/IntegrationTests/Owin/OwinMiddlewareExceptionTests.cs
nr-ahemsath: Good catch, I missed that.

tests/Agent/IntegrationTests/IntegrationTests/Owin/OwinWebApiTests.cs
nr-ahemsath: Good idea.  I started out the refactoring process by trying to make minimal changes, but I eventually decided to go in the direction you're suggesting.  Best to be consistent.

tests/Agent/IntegrationTests/IntegrationTests/CSP/HighSecurityAndCustomAttributes.cs
nr-ahemsath: Agreed.

PR #310, opened 2020-10-20
Fixes fail to parse expected status codes server config. 

vuqtran88:
### Description
The expected status codes returned with the server side config is returning an array instead of a string. This PR fixes parsing expected status codes server config failure. The agent will handle the following `error_collector.expected_status_codes` server config values:

`"error_collector.expected_status_codes":[401, 404]`
`"error_collector.expected_status_codes":["401"," 404"]`
`"error_collector.expected_status_codes":["401"," 404-410"]`


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1144,11 +1144,15 @@ public TimeSpan ErrorEventsHarvestCycle
...
 
-            var expectedStatusCodeArray = expectedStatusCodesString.Split(StringSeparators.Comma, StringSplitOptions.RemoveEmptyEntries);
+            if(expectedStatusCodeArray == null)
+            {
+                return expectedStatusCodes;
+            }
+
             foreach (var singleCodeOrRange in expectedStatusCodeArray)
             {
                 var index = singleCodeOrRange.IndexOf(HyphenChar);
```
jifeingo, line 1158: line 1162:   can it just be `var upperBoundString = singleCodeOrRange.Substring(index+1).Trim()`

src/Agent/NewRelic/Agent/Core/JsonConverters/ExpectedStatusCodesConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+            return new ExpectedStatusCodes(new[] { jt.ToString() });
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            throw new NotImplementedException();
+        }
+    }
+
+}
```
jifeingo, line 46: Is there a documented use case for us receiving a string with a single value or an array of strings for the same property?  If not, we should not have a custom JSON converter.

src/Agent/NewRelic/Agent/Core/Configuration/ServerConfiguration.cs
```diff
@@ -326,4 +329,15 @@ public void OnError(StreamingContext context, ErrorContext errorContext)
...
+    [JsonConverter(typeof(ExpectedStatusCodesConverter))]
+    public class ExpectedStatusCodes
+    {
+        public ExpectedStatusCodes(IEnumerable<string> expectedStatusCodesArray)
+        {
+            ExpectedStatusCodesArray = expectedStatusCodesArray;
+        }
+
+        public IEnumerable<string> ExpectedStatusCodesArray { get; private set; }
+    }
```
jifeingo, line 342: This feels unnecessary.  It seems that the only use case is an string[] which would make this class unnecessary.

src/Agent/NewRelic/Agent/Core/Configuration/ServerConfiguration.cs
```diff
@@ -211,7 +214,7 @@ public class AgentConfig
             public IEnumerable<KeyValuePair<string, IEnumerable<string>>> ErrorCollectorExpectedMessages { get; set; }
 
             [JsonProperty("error_collector.expected_status_codes")]
-            public string ErrorCollectorExpectedStatusCodes { get; set; }
+            public ExpectedStatusCodes ErrorCollectorExpectedStatusCodes { get; set; }
```
jifeingo, line 217: Why a specific class here and not just an array of strings.

src/Agent/NewRelic/Agent/Core/Configuration/ServerConfiguration.cs
```diff
@@ -211,7 +214,7 @@ public class AgentConfig
             public IEnumerable<KeyValuePair<string, IEnumerable<string>>> ErrorCollectorExpectedMessages { get; set; }
 
             [JsonProperty("error_collector.expected_status_codes")]
-            public string ErrorCollectorExpectedStatusCodes { get; set; }
+            public IEnumerable<object> ErrorCollectorExpectedStatusCodes { get; set; }
```
jifeingo, line 217: This should be a string.

vuqtran88: The server response value for this is actually an array of integers. I am using object here instead of int because there is a possibility that server response value would be be an array of strings when server side config for this starts to support a range of status codes.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1902,7 +1902,6 @@ private T HighSecurityModeOverrides<T>(T overriddenValue, T originalValue)
 
         private static T ServerOverrides<T>(T server, T local) where T : class
         {
-            Debug.Assert(local != null);
```
jifeingo, line 1905: Were you able to track down in GHE why this was put in place?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -2048,18 +2047,24 @@ private void ParseExpectedErrorConfigurations()
 
             foreach (var errorClass in _localConfiguration.errorCollector.expectedMessages)
             {
-                var messages = errorClass.message;
-                if (messages != null)
+                if (errorClass.message != null)
```
jifeingo, line 2050: what happens if errorClass.message is empty?

vuqtran88: I believe it will set any errors from that class to be expected.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -2048,18 +2047,24 @@ private void ParseExpectedErrorConfigurations()
...
                 {
-                    localExpectedErrorMessages.Add(errorClass.name, messages);
+                    localExpectedErrorMessages[errorClass.name] = errorClass.message;
                 }
             }
 
-            var expectedErrorInfo = ServerOverrides(_serverConfiguration.RpmConfig.ErrorCollectorExpectedMessages, localExpectedErrorMessages).ToDictionary();
+            var expectedErrorInfo = localExpectedErrorMessages;
```
jifeingo, line 2056: We should reverse the order of this.
Read the server side messages in first.  If there aren't any or it's empty, then we should process the local config.

This entire block of code should be performed after attempting to read the server config settings.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -2069,45 +2066,48 @@ private void ParseExpectedErrorConfigurations()
                 {
                     expectedErrorInfo[className] = Enumerable.Empty<string>();
                     Log.Warn($"{className} class is specified in both errorCollector.expectedClasses and errorCollector.expectedMessages configurations. Any errors of this class will be marked as expected.");
```
jifeingo, line 2068: add "Expected Errors - " to log message

vuqtran88: 👍

src/Agent/NewRelic/Agent/Core/Configuration/ReportedConfiguration.cs
```diff
@@ -60,7 +60,7 @@ public class ReportedConfiguration
         public IDictionary<string, IEnumerable<string>> ErrorCollectorExpectedMessages { get; set; }
 
         [JsonProperty("error_collector.expected_status_codes")]
-        public string ErrorCollectoryExpectedStatusCodes { get; set; }
+        public IEnumerable<string> ErrorCollectoryExpectedStatusCodes { get; set; }
```
jifeingo, line 63: Can we fix the typo in the name?  Should be `ErrorCollectorExpectedStatusCodes`

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -994,9 +994,12 @@ public string IgnoreErrorsAndIgnoreClassesCombineTests(string[] ignoreClasses, s
...
 
-        [TestCase("401", "405", ExpectedResult = "405")]
-        [TestCase("401", null, ExpectedResult = "401")]
-        public string ExpectedStatusCodesSetFromLocalAndServerOverrides(string local, string server)
+
+
+        [TestCase("401", new[] { "405" }, ExpectedResult = new[] { "405" })]
+        [TestCase("401", new string[0], ExpectedResult = new string[0])]
+        [TestCase("401", null, ExpectedResult = new[] { "401" })]
+        public IEnumerable<object> ExpectedStatusCodesSetFromLocalAndServerOverrides(string local, string[] server)
```
jifeingo, line 1002: This should be object[] as part of the input allowing server to send back either ints or strings?
Please add test 
for server new[]{400}
for server new[]{400,"401"}

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -1014,6 +1017,7 @@ public string ExpectedStatusCodesSetFromLocalAndServerOverrides(string local, st
         [TestCase("401.4,401", new string[] { "400", "401", "402", "403", "404" }, new bool[] { false, true, false, false, false })] //does not support full status codes
         [TestCase("404,401.4", new string[] { "400", "401", "402", "403", "404" }, new bool[] { false, false, false, false, true })]
         [TestCase("401.4-404.5", new string[] { "400", "401", "402", "403", "404" }, new bool[] { false, false, false, false, false })] //does not support full status codes
+        [TestCase("Foo,Bar,X-Y", new string[] { "400", "401" }, new bool[] { false, false })] //does not work with non status code values
         public void ExpectedStatusCodesParserTests(string local, string[] inputs, bool[] expected)
```
jifeingo, line 1021: Is it valid to test error cases where the inputs also have decimals.....   for example, inputs = { "400","401.4","402"}

src/Agent/NewRelic/Agent/Core/Configuration/ReportedConfiguration.cs
vuqtran88: 👍

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
jifeingo: I see the test below, maybe combine them?

vuqtran88: this test is testing server config overriding behavior after deserialization finishes so the value for _serverConfig.RpmConfig.ErrorCollectorExpectedStatusCodes can only be IEnumerable<string>. I need to fix the the return type for the test to be IEnumerable<string> instead of IEnumerable<object> due to recent changes.

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
vuqtran88: it is. There is a test case for it `[TestCase("401.4,401", new string[] { "400", "401", "402", "403", "404" }, new bool[] { false, true, false, false, false })]`

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
jifeingo: ok

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
jifeingo: no, the other way around.
Config = 400, 401.4, 402.1-403.2, 405
Input= 400, 401, 401.5, 402, 402.3, 404,405,405.3

vuqtran88: we don't support full status code as an input. I will create a false test case for this.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1144,47 +1144,48 @@ public TimeSpan ErrorEventsHarvestCycle
...
-            {
-                if (rule != null)
-                {
-                    expectedStatusCodes.Add(rule);
-                }
-                else
+                if(matchRule == null)
                 {
-                    Log.Warn($"Cannot parse {statusCode} status code. This status code format is not supported.");
+                    Log.Warn($"Cannot parse {singleCodeOrRange} status code. This status code format is not supported.");
```
jifeingo, line 1174: Sorry, can we add "Expected Errors - " to the front of this message

jifeingo (approved):

PR #311, opened 2020-10-20
Integration Tests - fix broken retry logic

lspangler584:
### Description
Exceptions thrown from the test runner were being masked by a bug where the retry boolean was being overwritten when the remote test application exited successfully. As a result of this bug, retries never occurred (unless the remote application being instrumented exited with code other than 0).

In the case of tests that intentionally throw exceptions, and where those exceptions were not being caught by the runner, this allowed them to pass. Fixing the bug revealed the need to catch the exceptions in the test runner when they were expected.

Efficiency improvement for Integration Tests AgentLogFile. Adds private backing field to avoid constructing new object each time AgentLog is accessed. Once the AgentLog file name/path is determined, there is no need to re-determine it for the current test. Functions in the AgentLogFile object create a new Filestream from the path, so any updates to the log file as execution progresses will be picked up.

### Details
Refactors `RemoteApplicationFixture.Initialize()`:

- Removes unassigned always-false `DelayKill` bool.
- Introduces a separate boolean (`retryTest`), which will be set when an exception occurs in the test runner for any reason AND if the exit code of the remote application is not 0 (success). Previously, `appIsExercisedNormally` boolean was used for both exceptions in the test runner AND based on the exit code of the remote application. In fact, the bool could be false (`appIsExerciseNormally` == false), then be reset to true if the remote application was shut down due to a test runner failure in the same while-loop iteration.
- Adds a timeout in `BasicWebServiceTest` to wait for the `transaction_sample_data` event in the log. With no timeout, there is no wait, and since that event is on a 1 minute harvest cycle flickers are highly likely.
- Other test fixes to catch intentionally thrown exceptions.

### Testing
This change affects most Integration Tests. Verify that tests run in GHA do not fail due to "No agent log file found." or "No log file found." or - in the case of AspNetCoreDistTraceRequestChainTests - "No log file found for {applicationName}."

Tested manually in debugger to simulate a test failure/retry scenario (which led to the discovery of the `RemoteApplicationFixture.Initialize()` issues).

Tests pass locally.

### Changelog
N/A


tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplication.cs
```diff
@@ -201,11 +202,17 @@ public string UniqueFolderName
...
         public string DestinationNewRelicExtensionsDirectoryPath { get { return Path.Combine(DestinationNewRelicHomeDirectoryPath, "extensions"); } }
 
+        private AgentLogFile _agentLogFile;
+
         public AgentLogFile AgentLog
         {
             get
             {
-                return new AgentLogFile(DestinationNewRelicLogFileDirectoryPath, AgentLogFileName, Timing.TimeToConnect);
+                if (_agentLogFile == null)
```
jifeingo, line 211: consider using newer C# syntax
```return _agnetLogFile ??= new AgentLogFile(estinationNewRelicLogFileDirectoryPath, AgentLogFileName, Timing.TimeToConnect);```

lspangler584: if the _agentLogFile is null and a new one is instantiated, _agentLogFile needs to be assigned that newly instantiated object

vuqtran88: There is a problem with it. In the event where a test app fails to run, the test runner will catch it and retry, the directory that contains the test app and agent for the retry will be in a different directory than what the `_agentLogFile` is referring.

lspangler584: the directory that contains the test app and agent for the retry is defined (partially) by the UniqueFolderName, which is also cached in a backing field, so will not change on the retry. Therefore, the cached AgentLog file name will be accurate.

However, while testing this behavior on a retry, I found some other (legacy) issues in the RemoteApplicationFixture.Initialize() method which caused the retry mechanism to not work. Mostly, that the retry flag called appIsExercisedNormally gets reset based on the exit code of the application, which is often a valid 0 even if there were other exceptions from running the test. I will update the PR description to elaborate.

lspangler584: resolved Jason's syntax request

nr-ahemsath (approved):
This has definitely improved the pass rate of our tests, as well as uncovering some fundamental issues that were previously hidden and are now fixed.  Ship it! 🚢

PR #316, opened 2020-10-22
Add back combined artifacts

nr-ahemsath:
### Description

Add back the workflow job that combines the integration test results.  It also disables enhanced logging since it doesn't appear to be valuable anymore.

### Testing

N/A

### Changelog

N/A

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -454,7 +454,7 @@ jobs:
       xunit_console: ${{ github.workspace }}/build/Tools/XUnit-Console/xunit.console.exe
       integration_tests_dll: ${{ github.workspace }}/tests/Agent/IntegrationTests/IntegrationTests/bin/Release/net461/NewRelic.Agent.IntegrationTests.dll
       # Make this variable true to enable extra data-gathering and logging to help troubleshoot test failures, at the cost of additional time and resources
-      enhanced_logging: true
+      enhanced_logging: false
       NR_DOTNET_TEST_SAVE_WORKING_DIRECTORY : 1
       
     steps:
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -442,12 +442,11 @@ jobs:
     runs-on: windows-2019
     strategy:
       matrix:
-        # namespace: [ AgentFeatures, AgentMetrics, Api, AspNetCore, BasicInstrumentation, CatInbound, CatOutbound, 
-        #   CSP, CustomAttributes, CustomInstrumentation, DataTransmission, DistributedTracing, Errors, 
-        #   HttpClientInstrumentation.NetCore, HttpClientInstrumentation.NetFramework, Logging, Owin, ReJit.NetCore, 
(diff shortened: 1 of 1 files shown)


lspangler584 (approved):

.github/workflows/all_solutions.yml
```diff
@@ -518,3 +518,26 @@ jobs:
...
+    if: ${{ always() }}
+    name: Combine Integration Test Results
+        
+    runs-on: windows-2019
+        
+    steps:
+      - name: Download all workflow run artifacts
+        uses: actions/download-artifact@v2
+        
+      - name: Combine Integration Results
```
jaffinito, line 532: You should be able to have the upload action gather these for you using wildcard, similar to jenkins.  This would avoid an extra step.

.github/workflows/all_solutions.yml
```diff
@@ -442,19 +442,20 @@ jobs:
...
-          CSP, CustomAttributes, CustomInstrumentation, DataTransmission, DistributedTracing, Errors, 
-          HttpClientInstrumentation.NetCore, HttpClientInstrumentation.NetFramework, Logging, Owin, ReJit.NetCore, 
-          ReJit.NetFramework, RemoteServiceFixtures, RestSharp, WCF.Client.IIS.ASPDisabled, WCF.Client.IIS.ASPEnabled, 
-          WCF.Client.Self, WCF.Service.IIS.ASPDisabled, WCF.Service.IIS.ASPEnabled, WCF.Service.Self ]
+        # namespace: [ AgentFeatures, AgentMetrics, Api, AspNetCore, BasicInstrumentation, CatInbound, CatOutbound, 
+        #   CSP, CustomAttributes, CustomInstrumentation, DataTransmission, DistributedTracing, Errors, 
+        #   HttpClientInstrumentation.NetCore, HttpClientInstrumentation.NetFramework, Logging, Owin, ReJit.NetCore, 
+        #   ReJit.NetFramework, RemoteServiceFixtures, RestSharp, WCF.Client.IIS.ASPDisabled, WCF.Client.IIS.ASPEnabled, 
+        #   WCF.Client.Self, WCF.Service.IIS.ASPDisabled, WCF.Service.IIS.ASPEnabled, WCF.Service.Self ]
+        namespace: [ Api, Errors ]
```
jifeingo, line 450: This feels unintended?

.github/workflows/all_solutions.yml
```diff
@@ -518,3 +519,29 @@ jobs:
             C:\IntegrationTestWorkingDirectory\TestResults\**\*TestResults.xml
           if-no-files-found: error
           
+  combine-integration-test-results:
```
jifeingo, line 522: Alex, double checking... this doesn't remove the results still being in the namespace artifacts?  I've found that to be handy.

.github/workflows/all_solutions.yml
nr-ahemsath: My goal is to get the .xml XUnit test results file into a single flattened directory.

.github/workflows/all_solutions.yml
nr-ahemsath: That was just for debugging/testing purposes.  It's back to normal now.

.github/workflows/all_solutions.yml
nr-ahemsath: No, they'll still be there.

PR #317, opened 2020-10-22
Adds Running Unbounded Integration Tests to workflow

### LINKED ISSUE #299: Create matrix job to run unbounded tests
### Feature Description
Create matrix job to run unbounded tests.  Similar to bounded integration tests.  Use the namespace to split the tests into groups.  The namespaces are setup along datastore lines.

jaffinito:
Resolves #299 

### Description

If you have questions about any change please reach out!

 Adds job to run unbounded integration tests as a matrix
- Updated docker-compose.yml to properly map all exposed ports instead of a subset.
- docker-compose.yml and linux container folders are already copied to linux server and running
- This PR does not remove the Windows containers, saving that for a cleanup PR
- Updated build-unbounded-tests to save artifacts
- Added run-unbounded-tests matrix job to run tests
- Unbounded tests havce many dependencies so the steps for these items have been expanded
- Artifacts for both run-unbounded-tests and run-integration-tests have been synced and updated to only save working folders on failure
- Due to how GHA hosts servers we may see some failures due to how the firewall on the linux server is restricted. I am working on a better way.

Workflow Job limit handling
- Removes the "push" event from the list of workflow triggers since everything runs during the PR acceptance process.  This will eliminate a source of extra workflow runs.
- Added a job that calls an action that will cancel other previous runs of the same PR that may be active. 
- "cancel-previous-workflow-runs" job is run before all others
- The Profiler build jobs are set to "need" the cancel job so that it runs first.  The Profiler jobs were the first first jobs that ran so they were the only ones that needed the change.
- Added `if: ${{ always() }}` to Profiler jobs so that "cancel-previous-workflow-runs"  job errors don't block runs.

### Testing

Individual datastores have all be tested and pass.
Cancellation Job tested and passed.

To confirm testing, unbounded tests need to pass without real errors.  There can and likely will some failure due to the ip ranges being blocked.  This will be corrected via server changes in the future (still working out a good way to do this, I have ideas.).

### Changelog

n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -32,13 +32,27 @@ jobs:
       profiler_path: ${{ github.workspace }}\src\Agent\NewRelic\Profiler
       profiler_solution_path: ${{ github.workspace }}\src\Agent\NewRelic\Profiler\NewRelic.Profiler.sln
       output_path: ${{ github.workspace }}\src\Agent\_profilerBuild
-      #<other env vars here>
+      profiler_md5: ""
 
     steps:
     - name: Checkout
       uses: actions/checkout@v2
       with:
         fetch-depth: 0
+
+    - name: MD5 Profiler
+      run: |
+        pip install checksumdir
+        $profsum = checksumdir -a md5 ${{ env.profiler_path }}
+        echo "profiler_md5=${profsum}" >> $GITHUB_ENV
+      shell: powershell
+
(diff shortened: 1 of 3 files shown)


lspangler584 (approved):
Even though some tests are still failing on the server, they pass when run locally. The issues are not with the Agent, there seem to be issues with server access. I will follow up with a characterization of the troublesome tests, which fail consistently, which fail intermittently, etc.

PR #324, opened 2020-10-27
Create assignproj.yml

angelatan2:
### Description
Auto-assign projects field in Github so that issues and PR have this field auto-populated.

### Testing

Test by creating a new issue and checking that the Engineering Board is auto-populated with the issue in the correct column.

### Changelog
n/a


nr-ahemsath (approved):
Looks good.  The one in the logenrichers repo works as expected and this looks identical.  Approved.

Issue #319, opened 2020-10-26
Support Content-Security-Policy


bgrainger:
If a site is using `Content-Security-Policy` without `'unsafe-inline'` then inline scripts (such as the Browser Monitoring injected by the NR APM Agent) will not execute.

If the site is already using `Content-Security-Policy 'nonce-BASE64'`, then it would be desirable for https://github.com/newrelic/newrelic-dotnet-agent/blob/3e4d5158f4abdc045fae4ccffd686b7d05d4c10d/src/Agent/NewRelic/Agent/Core/BrowserMonitoring/BrowserMonitoringScriptMaker.cs#L80 to be able to inject a per-request nonce.

Alternatively, the site could add `script-src 'sha256-...'` to its CSP header if the SHA-256 hash of the script were available from the agent.

As a workaround, the [copy/paste approach](https://docs.newrelic.com/docs/browser/new-relic-browser/installation/install-browser-monitoring-agent#copy-paste-app) can be used (with a nonce dynamically injected into the script by the backend code).

### Priority

Really Want

nr-ahemsath:
Hi @bgrainger, thanks for bringing this one to our attention.  Being fully transparent, this is an area that you appear to understand better than I do.  I'm going to bring this to the attention of our Browser team.  In the meantime, would you be willing to submit a PR to implement the functionality you are looking for?

bgrainger:
The only solutions I can think of will need some interaction between the site and the agent; either a pre-request nonce being transferred from site to agent (some kind of dependency injection?) or a script SHA-256 being pulled from the agent by the site (a static property? but maybe that's per transaction, too?).

Or maybe an overload of [`GetBrowserTimingHeader`](https://docs.newrelic.com/docs/agents/net-agent/net-agent-api/getbrowsertimingheader-net-agent-api) that takes a nonce as a string? That may actually be the simplest approach; I could try to see if that's viable.

PR #332, opened 2020-10-29
Add API proposal for generating a script with a nonce

bgrainger:
### Description

This is a suggested new Agent API that will fix https://github.com/newrelic/newrelic-dotnet-agent/issues/319.

### Discussion

I've started by adding an overload of `GetBrowserTimingHeader` because adding an optional parameter to an existing method is a binary-breaking change.

As per [NewRelic.cs comments](https://github.com/bgrainger/newrelic-dotnet-agent/blob/script-nonce/src/Agent/NewRelic.Api.Agent/NewRelic.cs), it's clear that the method bodies have no implementation, but are hooked up to the "right" methods at runtime. Does this process support overloads? If not, would a unique function name such as `GetBrowserTimingHeaderWithNonce` be better?

I've modified the implementation of [AgentApi.cs](https://github.com/bgrainger/newrelic-dotnet-agent/blob/script-nonce/src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs) to pass the nonce through. Does [`ApiMethod`](https://github.com/bgrainger/newrelic-dotnet-agent/blob/script-nonce/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core.Metric/ApiSupportabilityMetricCounters.cs) need a new enum member? If so, does the order of items in that `enum` matter?


src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
```diff
@@ -468,7 +469,31 @@ public static string GetBrowserTimingHeader()
...
+        ///   &lt;body>
+        ///   ...
+        /// </code>
+        /// </example>
+        /// <returns>An html string to be embedded in a page header.</returns>
+        public static string GetBrowserTimingHeader(string nonce)
+        {
+            const ApiMethod apiMetric = ApiMethod.GetBrowserTimingHeader;
+            const string apiName = nameof(GetBrowserTimingHeader);
+            return TryInvoke(InternalApi.GetBrowserTimingHeader, nonce, apiName, apiMetric) ?? "";
```
ELanning, line 496: Instead of adding `parameter` to `TryInvoke` for this functionality, why not do the same using closures?
`TryInvoke(() => InternalApi.GetBrowserTimingHeader(nonce), apiName, apiMetric)`

small nit- `""` is inconsistent with `string.Empty` in the rest of the file.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/InternalApi.cs
```diff
@@ -263,16 +263,16 @@ public static void IgnoreApdex()
...
-            return _agentApiImplementation?.GetBrowserTimingHeader() ?? string.Empty;
+            return _agentApiImplementation?.GetBrowserTimingHeader(nonce) ?? string.Empty;
         }
 
         /// <summary> (This method is obsolete) gets browser timing footer. </summary>
         ///
         /// <returns> An empty string. </returns>
         [Obsolete]
-        public static string GetBrowserTimingFooter()
+        public static string GetBrowserTimingFooter(string? _)
```
ELanning, line 275: If this is already marked as `Obsolete`, might not be worth bothering to update.

bgrainger: I only updated it to stay in sync with the change above. This could be reverted if that were updated.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
bgrainger: It was to avoid creating another closure and another allocation... but maybe it would be the same either way? I'd have to look at the IL or the output from a memory profiler.

But good point; it'd be simpler that way.

bgrainger: This method:

```
    IL_0000: ldnull
    IL_0001: ldftn        string [NewRelic.Agent.Extensions]NewRelic.Agent.Api.InternalApi::GetBrowserTimingHeader(string)
    IL_0007: newobj       instance void class [mscorlib]System.Func`2<string, string>::.ctor(object, native int)
    IL_000c: ldnull
    IL_000d: ldstr        "GetBrowserTimingHeader"
    IL_0012: ldc.i4.5
    IL_0013: call         !!0/*string*/ NewRelic.Agent.Core.AgentApi::TryInvoke<string>(class [mscorlib]System.Func`2<string, !!0/*string*/>, string, string, valuetype NewRelic.Agent.Core.Metric.ApiMethod)
    IL_0018: dup
    IL_0019: brtrue.s     IL_0021
    IL_001b: pop
    IL_001c: ldstr        ""
    IL_0021: ret
```

Proposed closure.

```
  **IL_0000: newobj       instance void NewRelic.Agent.Core.AgentApi/'<>c__DisplayClass27_0'::.ctor()
    IL_0005: dup
    IL_0006: ldarg.0      // nonce
    IL_0007: stfld        string NewRelic.Agent.Core.AgentApi/'<>c__DisplayClass27_0'::nonce
    IL_000c: ldftn        instance string NewRelic.Agent.Core.AgentApi/'<>c__DisplayClass27_0'::'<GetBrowserTimingHeader>b__0'()
  IL_0012: newobj       instance void class [mscorlib]System.Func`1<string>::.ctor(object, native int)
    IL_0017: ldstr        "GetBrowserTimingHeader"
    IL_001c: ldc.i4.5
    IL_001d: call         !!0/*string*/ NewRelic.Agent.Core.AgentApi::TryInvoke<string>(class [mscorlib]System.Func`1<!!0/*string*/>, string, valuetype NewRelic.Agent.Core.Metric.ApiMethod)
    IL_0022: dup
    IL_0023: brtrue.s     IL_002b
    IL_0025: pop
    IL_0026: ldstr        ""
    IL_002b: ret
```

Note the additional allocation for the closure. This would happen per-request, so I thought it might be worth avoiding.

OTOH, allocations don't appear to be highly optimised in the current code (we could get rid of allocating the `Func` per request if we were worried about that, for example), so perhaps the simpler code is better.

src/Agent/NewRelic/Agent/Core/BrowserMonitoring/BrowserMonitoringScriptMaker.cs
```diff
@@ -73,11 +74,13 @@ public string GetScript(IInternalTransaction transaction)
...
                 transaction.LogFinest("RUM: TryGetBrowserTimingHeader success.");
 
+            var scriptNonce = nonce is null ? "" : $@" nonce=""{HttpUtility.HtmlAttributeEncode(nonce)}""";
+
             // The JavaScript variable NREUMQ stands for New Relic End User Metric "Q". This was the name before marketing renamed "EUM" to "RUM".
             // We can't change the name of the variable since: (a) we have to be consistent across agents, and (b) it has to be in sync with the rum.js file which is downloaded from NR servers.
             var javascriptAgentConfiguration = $"window.NREUM||(NREUM={{}});NREUM.info = {browserConfigurationData.ToJsonString()}";
             var javascriptAgent = _configurationService.Configuration.BrowserMonitoringJavaScriptAgent;
-            return $"<script type=\"text/javascript\">{javascriptAgentConfiguration}</script><script type=\"text/javascript\">{javascriptAgent}</script>";
+            return $"<script type=\"text/javascript\"{scriptNonce}>{javascriptAgentConfiguration}</script><script type=\"text/javascript\">{javascriptAgent}</script>";
```
PatrickNausha, line 83: Looks like there's a second script that would also need a `nonce` to avoid CSP violations.
```suggestion
            return $"<script type=\"text/javascript\"{scriptNonce}>{javascriptAgentConfiguration}</script><script type=\"text/javascript\"{scriptNonce}>{javascriptAgent}</script>";
```

Or is there something I'm missing here?

bgrainger: Nope, you're right; I missed that. Will need to update tests, too.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
ELanning: Interesting question. It would be cool to see what kind of empirical impact it has. I imagine it might not even be detectable without extreme precision.

nr-ahemsath:
@bgrainger: first off, sorry it's been almost two weeks since you submitted this PR with no communication from our side.

Changes to New Relic's agent APIs typically require a fair amount of effort: planning, design, testing, documentation, etc.  We also want to make sure that our APIs remain as consistent as possible among all of our language agents.  This is why we haven't been able to prioritize reviewing and merging this PR.  We may end up closing this PR, but your prototype solution should certainly be helpful when we are able to prioritize this work and it will remain linked to the issue you created.

bgrainger:
That all makes sense; thanks for the update!

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
```diff
@@ -502,7 +502,7 @@ public void IgnoreApdex()
         ///   &lt;body&gt;
         ///   ...
         /// </code></example>
-        public string GetBrowserTimingHeader()
+        public string GetBrowserTimingHeader(string? nonce)
```
vuqtran88, line 505: Instead of using optional parameter, we recommend introducing the `GetBrowserTimingHeader(string nonce)` overloading method. Inside the old `GetBrowserTimingHeader()` method, we just have to call `GetBrowserTimingHeader(null)`. We try to avoid using optional parameter as it would cause future test complication.

tests/Agent/UnitTests/Core.UnitTest/BrowserMonitoring/BrowserMonitoringScriptMakerTests.cs
```diff
@@ -85,12 +85,27 @@ public void GetScript_ReturnsValidScript_UnderNormalConditions()
...
             var transaction = BuildTestTransaction(queueTime: TimeSpan.FromSeconds(1), applicationTime: TimeSpan.FromSeconds(2));
 
-            var script = _browserMonitoringScriptMaker.GetScript(transaction);
+            var script = _browserMonitoringScriptMaker.GetScript(transaction, null);
 
             const string expectedScript = @"<script type=""text/javascript"">window.NREUM||(NREUM={});NREUM.info = {""beacon"":"""",""errorBeacon"":"""",""licenseKey"":"""",""applicationID"":"""",""transactionName"":""HBsGAwcLSlMeAx8FEQ=="",""queueTime"":1000,""applicationTime"":2000,""agent"":"""",""atts"":""""}</script><script type=""text/javascript"">the agent</script>";
             Assert.AreEqual(expectedScript, script);
         }
 
+        [Test]
```
vuqtran88, line 94: 👍 for adding a unit test.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/IAgentApi.cs
```diff
@@ -179,7 +180,7 @@ public interface IAgentApi
         ///   &lt;body&gt;
         ///   ...
         /// </code></example>
-        string GetBrowserTimingHeader();
```
vuqtran88, line 182: As the result of my comment above, we like to keep 2 overloading methods of the `GetBrowserTimingHeader`.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/InternalApi.cs
```diff
@@ -263,9 +263,9 @@ public static void IgnoreApdex()
         ///   &lt;body&gt;
         ///   ...
         /// </code></example>
-        public static string GetBrowserTimingHeader()
```
vuqtran88, line 266: As the result of my comment above, we like to keep 2 overloading methods of the `GetBrowserTimingHeader`.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
```diff
@@ -468,7 +468,31 @@ public static string GetBrowserTimingHeader()
         {
             const ApiMethod apiMetric = ApiMethod.GetBrowserTimingHeader;
             const string apiName = nameof(GetBrowserTimingHeader);
-            return TryInvoke(InternalApi.GetBrowserTimingHeader, apiName, apiMetric) ?? string.Empty;
+            return TryInvoke(() => InternalApi.GetBrowserTimingHeader(null), apiName, apiMetric) ?? string.Empty;
```
vuqtran88, line 471: No code change needed after retaining the `GetBrowserTimingHeader()` overload.

vuqtran88 (changes requested):
@bgrainger I just reopened your PR and took a stab reviewing it. Overall it looks great, there are a few thing I would like to change. Please see my comments.

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
bgrainger: Makes sense; it also prevents a binary-breaking change.

bgrainger: I added an overload for the public API; I can do the same here.

vuqtran88: that would be awesome! 👍

vuqtran88: did u already commit the change ? I have not since any changes.

vuqtran88:
@bgrainger 
> As per NewRelic.cs comments, it's clear that the method bodies have no implementation, but are hooked up to the "right" methods at runtime. Does this process support overloads? If not, would a unique function name such as GetBrowserTimingHeaderWithNonce be better?"
I have no reason thinking it does not support overloads. For agent api methods, the profiler will detect them at runtimes base on the `NewRelic.Api.Agent.NewRelic` type and will call into the right agent managed code. You can find more details on how the profiler does that [here](https://github.com/newrelic/newrelic-dotnet-agent/blob/5b375b3c38517c8326aac0a70a4901f99f1a483c/src/Agent/NewRelic/Profiler/MethodRewriter/Instrumentors.h#L89-L108) and  [here](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Profiler/MethodRewriter/ApiFunctionManipulator.h) 

>Does ApiMethod need a new enum member
Can you elaborate more on the usage of the enum and how different using enum is going make ?

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
bgrainger: Oh, sorry, that was confusing.

I meant that in the initial PR, there was an overload in the public API, just not in the implementation. I can make them both the same.

bgrainger: Pushed now.

bgrainger:
> > Does ApiMethod need a new enum member
> 
> Can you elaborate more on the usage of the enum and how different using enum is going make ?

I just didn't know how this is used by the agent, and if adding a new public API impacts it. Mostly just wanted to make sure I wasn't overlooking anything.

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
```diff
@@ -519,7 +519,42 @@ public string GetBrowserTimingHeader()
                 // The transaction's name must be frozen if we're going to generate a RUM script
                 transaction.CandidateTransactionName.Freeze(TransactionNameFreezeReason.ManualBrowserScriptInjection);
 
-                return _browserMonitoringScriptMaker.GetScript(transaction) ?? string.Empty;
+                return _browserMonitoringScriptMaker.GetScript(transaction, null) ?? string.Empty;
+            }
```
vuqtran88, line 523: Just a small nitpicking, the logic of the `public string GetBrowserTimingHeader()` can be greatly simplified by doing
```
public string GetBrowserTimingHeader()
{
   GetBrowserTimingHeader(null);
}
```
we like remove duplicate code a much as possible.

src/Agent/NewRelic/Agent/Core/BrowserMonitoring/BrowserMonitoringScriptMaker.cs
```diff
@@ -73,11 +74,13 @@ public string GetScript(IInternalTransaction transaction)
             if (Log.IsFinestEnabled)
                 transaction.LogFinest("RUM: TryGetBrowserTimingHeader success.");
 
+            var scriptNonce = nonce is null ? "" : $@" nonce=""{HttpUtility.HtmlAttributeEncode(nonce)}""";
```
vuqtran88, line 77: we probably want to handle `null` or `empty` value for `none`. Something like
```
            var scriptNonce = String.IsNullOrEmpty(nonce) ? String.Empty : $@" nonce=""{HttpUtility.HtmlAttributeEncode(nonce)}""";
```

vuqtran88 (changes requested):
@bgrainger thanks for making some recent changes. I still have small change requests. Please see my comments.

vuqtran88:
Hi @bgrainger . I just want to check in to see if you will be able to make the additional changes as requested. Please let us know if you need any helps. Thanks.

vuqtran88:
Hi @bgrainger,
Out of curiosity, what CSP setting is currently used in your website. I tested the new api with a test site with the following CSP setting and the browser agent script still failed to execute because the inner script (`js-agent.newrelic.com/nr-1197.min.js`) violated the setting.

`Content-Security-Policy: default-src 'self'; script-src 'self'; script-src-elem 'nonce-2726c7f26c'`

I'm not sure what I am testing is relevant so I'm interested to know your CSP setting. Thank you.

bgrainger:
@vuqtran88 Yes, I do plan to come back and review your comments; was just busy the last few days.

Our CSP is: `Content-Security-Policy: script-src 'self'  'unsafe-inline' 'nonce-9R4J0pJNZNgXpsdmyOw/tw==' *.faithlifecdn.com *.faithlife.com www.googletagmanager.com www.google-analytics.com [... some other domains ...] js-agent.newrelic.com bam.nr-data.net bam-cell.nr-data.net www.recaptcha.net; object-src 'none'; base-uri 'none'`

It looks like we already white-listed the domains for third-party scripts we know we need to load.

vuqtran88:
> @vuqtran88 Yes, I do plan to come back and review your comments; was just busy the last few days.
> 
> Our CSP is: `Content-Security-Policy: script-src 'self' 'unsafe-inline' 'nonce-9R4J0pJNZNgXpsdmyOw/tw==' *.faithlifecdn.com *.faithlife.com www.googletagmanager.com www.google-analytics.com [... some other domains ...] js-agent.newrelic.com bam.nr-data.net bam-cell.nr-data.net www.recaptcha.net; object-src 'none'; base-uri 'none'`
> 
> It looks like we already white-listed the domains for third-party scripts we know we need to load.
@bgrainger 
Thanks for the response. It makes sense. I will test it out with something similiar and get back to you.

One more clarification, in the CSP setting you provided, I see `'unsafe-inline'` is specified. I assume it is there because you wanted the current agent script to work, and you are hoping not to specify it after this change is merged. Is that correct ?

vuqtran88:
Whitelisting `https://js-agent.newrelic.com` `https://bam.nr-data.net` and `https://bam-cell.nr-data.net` are actually required according to https://docs.newrelic.com/docs/browser/new-relic-browser/getting-started/compatibility-requirements-browser-monitoring. I have tested the change and it did work.

bgrainger:
> in the CSP setting you provided, I see 'unsafe-inline' is specified

That definitely looks like a mistake. Thanks for pointing that out; I had missed it. I'll have our team look into it.

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
bgrainger: I can do that, but then were' back to `GetBrowserTimingHeader(string? nonce)` (for nullability) which you initially objected to, and was why I introduced an entirely new overload. Maybe you were just asking for the internal API to match the public API?

The method now simply delegates to the other implementation.

bgrainger:
I believe all outstanding comments are now addressed.

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
vuqtran88: @bgrainger as a good practice, we are very careful with using nullable parameters as method arguments. Imagine, for some reasons, we would introduce a new overload like `GetBrowserTimingHeader(NewType? nonce)`. We now have 2 apis as the followings.

`GetBrowserTimingHeader(SomeType? nonce)`
`GetBrowserTimingHeader(string? nonce)`

What happens when a user call `GetBrowserTimingHeader(null)`. The user will get a compile error.

bgrainger: That will happen whether or not it's written `string` or `string?`, since `string` is a reference type and can always accept `null` as an argument value. This change is _purely_ for compile-time nullability flow analysis.

If you like, I can change it back to `string` and use `default!` in the call site; it's somewhat irrelevant from an ABI perspective.

bgrainger: Would you like this renamed to `GetBrowserTimingHeaderWithNonce` to avoid future overloaded API issues?

I'm not at all committed to the current name choice; as indicated in the OP this is just a suggestion for the new API to get the process started.

vuqtran88:
@bgrainger,
I was trying to resolve conflicts as well as merge `main` into your branch, but looks like I might have broken your PR as well. My apologies. I will try to get it back to the working stage. I hope you don't mind me updating your branch directly for this particular reason.

src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs
vuqtran88: > That will happen whether or not it's written `string` or `string?`, since `string` is a reference type and can always accept `null` as an argument value. This change is _purely_ for compile-time nullability flow analysis.
> 
> If you like, I can change it back to `string` and use `default!` in the call site; it's somewhat irrelevant from an ABI perspective.

You are absolutely right. I realized my concerns about optional parameter were totally off. I want to take a step back before I overcomplicate this myself. I just recently updated your branch with changes. Please do let me know if they look ok to you.

vuqtran88:
> I believe all outstanding comments are now addressed.
I'm happy with the current state of the PR now. I will do some additional testing before approving it. Please do let me know if you have any other questions. Thanks again for your contribution.

vuqtran88 (approved):
LGTM. All tests passed.

vuqtran88:
@bgrainger I went ahead and merged your PR. Thanks again.

angelatan2:
@ELanning, @PatrickNausha, @bgrainger Thanks for the reviews and the contribution.  This content should be in the next release, barring anything unforeseen.

PR #336, opened 2020-10-29
Deployment mainyaml

### LINKED ISSUE #325: Update event triggers, setup ArtifactBuilder, and get agent version in main.yaml
### Feature Description

- Add event trigger to include starting the workflow based on tags being pushed.  This is essentially the same as the log enricher, so  we can just copy/paste from there.  See [this](https://github.com/newrelic/newrelic-logenricher-dotnet/blob/main/.github/workflows/main.yml).
- The “build-test-fullagent-msi” job will need to be updated to store a new _global_ env var to store the agent version for use by the linux build and packaging jobs. The powershell for this is in /build/build.ps1 on line 101. See [this doc](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable) for adding/updating environment variables in Actions.  
- The “build-test-fullagent-msi” job will need to produce a new artifact for “\src\_build\” which is needed by the ArtifactBuilder, maybe called “agent-build-folder-artifacts”.
- Add a job, maybe called “run-artifactbuilder”, that depends on the home folder artifact and the new “agent-build-folder-artifacts” artifact.  This job builds the artifacts and creates a new artifact, maybe called “deploy-artifacts”. This job only runs on a release build.

### Additional context
This work is part of a larger effort to get our deployment process moved to GHA.

### Priority
Blocker

jaffinito:
Resolves  #325

### Description

- Added release trigger
- Added step to build-test-fullagent-msi to create agent_version.txt
- Added step to build-test-fullagent-msi to save agent_version.txt as an artifact
- Added step to build-test-fullagent-msi to save /src/_build as an artifact
- Add new Job run-artifactbuilder to run ArtifactBuilder, depends on build-test-fullagent-msi
- Job run-artifactbuilder  creates deploy-artifacts artifact
- Fixed formatting of steps to match yaml spec
- Renamed the profiler jobs to remove "test" from the naame
- minor formatting cleanup

### Testing

- Job build-test-fullagent-msi  should pass
- Workflow should have these new artifacts: build-folder-artifacts, agent-version, deploy-artifacts, NewRelic.OpenTracing.AmazonLambda.Tracer,NewRelic.NuGetHelper,NewRelic.Agent.Extensions
- Job run-artifactbuilder completes and saves artifacts.

### Changelog
n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -201,6 +201,13 @@ jobs:
           name: NewRelic.NuGetHelper
           path: ${{ github.workspace }}\build\NewRelic.NuGetHelper\bin
           if-no-files-found: error
+          
+      - name: Archive NewRelic.Agent.Extensions
+        uses: actions/upload-artifact@v2
+        with:
+          name: NewRelic.Agent.Extensions
+          path: ${{ github.workspace }}\src\Agent\NewRelic\Agent\Extensions\NewRelic.Agent.Extensions\bin\Release
+          if-no-files-found: error
 
       - name: Archive FullAgent Home folders
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


jaffinito (commented):
👍 Approved.

vuqtran88 (approved):
this line `dotnet run --project '$artifactBuilderCsproj' LinuxPackages $configuration` in the `package.ps1` fails due to no linux packages being built yet, but this is currently being worked on.

PR #338, opened 2020-10-30
Remove is metric name allowed

### LINKED ISSUE #337: (Memory Leak) Strange behavior in MetricNameService.RenameTransaction
During a memory leak analysis, I found a list of transactions stored on the `MetricNameService`. The list stored in the `_transactionNames` instance field, increases progressively, even using the SetTransactionName for grouping or IgnoreTransaction API call.

**Description**
`MetricNameService` stores a list of unique transactions in the instance field `_transactionNames`, [here](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs).
The method `IsMetricNameAllowed` stored the transaction but never use the list and return always `true`.
It seems a strange behavior because the `IsMetricNameAllowed` result is used in the method `RenameTransaction`.
Why do you need an `if` in your code, if the operation is always `true`? [here](https://github.com/newrelic/newrelic-dotnet-agent/blob/v8.34.0/src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs#L56).

**Expected Behavior**
I expect for a web request that I call your API methods, NewRelic.SetTransactionName or NewRelic.IgnoreTransaction, the MetricNameService doesn't store the transaction. e.g.

WebRequest: /product/product-seo-description-123.aspx
Call API: NewRelic.SetTransactionName(null, "Product-Detail-Page")
Records:
WebTransaction/ASP/product/product-seo-description-123.aspx
WebTransaction/MVC/Product/Index/{id}
WebTransaction/Custom/Product-Detail-Page
I expect:
WebTransaction/Custom/Product-Detail-Page

WebRequest: /static/image.png
Call API: NewRelic.IgnoreTransaction()
Records:
WebTransaction/ASP/static/image.png
I expect:
No transaction records.

**Note:** In the NewRelic dashboard, Transactions view, all seems right and shows the expected results. Ignored transactions are not showed and custom transactions are shown as expected.

I don't have a log to be shared here, but I already shared other information internally with the NewRelic support team member.

**Steps to Reproduce**
Just do some traffic and you can see the duplicated transaction names on the `MetricServiceName`. I extracted a memory dump to have access to the object stored.

**Your Environment**
I using the NewRelic agent 8.34 version in a ASP.NET Core application hosted by a Windows Server with IIS.

**Additional context**
snippet API calls: https://gist.github.com/tiagodaraujo/c3d9ac425019b26b29c66e7f01015470

nr-ahemsath:
### Description

Fixes #337.  A while back, we removed functionality from the agent that capped the total number of metric names that could be created.  That change made the `IsMetricNameAllowed()` method obsolete, and caused the `_transactionNames` collection to grow without bound.  Since the only reason that collection existed in the first place was to enable the functionality that was removed, it should have been removed at that time as well.

### Testing

All unit tests passed for me locally.  All but two integration tests passed, and those two failures appear to be due to unrelated connectivity issues.

### Changelog

CHANGELOG was updated.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
+++ b/src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
@@ -24,8 +24,6 @@ namespace NewRelic.Agent.Core.Metrics
         private static readonly TransactionMetricName NormalizedWebTransactionMetricName = new TransactionMetricName(MetricNames.WebTransactionPrefix, "Normalized/*");
         private static readonly TransactionMetricName NormalizedOtherTransactionMetricName = new TransactionMetricName(MetricNames.OtherTransactionPrefix, "Normalized/*");
 
-        private readonly Utils.HashSet<string> _transactionNames = new Utils.HashSet<string>();
-
         #region Public API
 
         public string NormalizeUrl(string url)
@@ -71,9 +69,6 @@ namespace NewRelic.Agent.Core.Metrics
 
             // Renaming rules are not allowed to change the first segment of a transaction name
             var newTransactionName = GetTransactionMetricName(newPrefixedTransactionName, proposedTransactionName, shouldIgnore);
-
-            if (!IsMetricNameAllowed(newPrefixedTransactionName))
-                return proposedTransactionName.IsWebTransactionName ? NormalizedWebTransactionMetricName : NormalizedOtherTransactionMetricName;
 
             return newTransactionName;
         }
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Metrics/MetricNameService.cs
```diff
@@ -24,8 +24,6 @@ public class MetricNameService : ConfigurationBasedService, IMetricNameService
         private static readonly TransactionMetricName NormalizedWebTransactionMetricName = new TransactionMetricName(MetricNames.WebTransactionPrefix, "Normalized/*");
```
lspangler584, line 24: These 2 "Normalized..." vars are no longer needed either.

nr-ahemsath: Thanks for catching that.  Yay, even more cleanup!

lspangler584 (approved):
Unit & IntegrationTests pass locally.
Looks good.
Great find @tiagodaraujo !  THANK YOU for making the Agent better.

Issue #334, opened 2020-10-29
Add GPG key to repo as an encrypted file


jaffinito:
### Feature Description
We are going to go with encrypting the files, storing in repo, and decrypting on the fly. There appear to be a few different approaches to this problem.

Some exploration of them will be needed and we need a way to encrypt and decrypt.

Currently, this only needs to work on Linux.

### Priority
Blocker

PR #340, opened 2020-11-03
Package rpm (without GPG signing) and make available to artifactbuilder

### LINKED ISSUE #327: Setup job to package linux artifacts into a RPM package
### Feature Description

- Add a job to build the RPM Linux package.
- This will be run on a Linux runner.
- This is already completely dockerized and composed. See /build/Linux/build/rpm.
- This requires the home folders, deploy-artifacts, and maybe some additional ones.
- This _requires_ our GPG key which will need some special handling. This is being worked on in #334. 
- This job with only run on a release build.
- Generates artifacts

### Additional context
This work is part of a larger effort to get our deployment process moved to GHA.

### Priority
Blocker

jaffinito:
Resolves #327 

### Description

**Very Important Notes**: 
- The jobs and steps added here will not run until a release occurs. 
- Until the Debian packaging is done, the artifact builder will fail since it needs all the packages.  If you look at the error [here](https://github.com/jaffinito/newrelic-dotnet-agent/runs/1349858453?check_suite_focus=true#step:8:64) you will see that it is not missing the RPM package.
- The GPG version of the build command is commented out since it will not work until the GPG work in #334 is completed.  This was left in as a comment to be clear that this not the final version.

Changes:
- Updated rpm/build.sh to make agent version defects clearer (wraps it in pipes '|')
- Updated Linux/docker-compose.yml to chomd 777 the sh files
- Added create-package-rpm to build RPM package
- Updated run-artifactbuilder to need create-package-rpm not build-test-fullagent-msi
- Added if to allow artifactbuilder and rpmbuilder to only run on release

### Testing

Workflow should complete as normal and the `create-package-rpm` and `run-artifactbuilder` jobs should not run.  See [here](https://github.com/jaffinito/newrelic-dotnet-agent/actions/runs/344362598) for a truncated run showing things work as described.

### Changelog

n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -627,7 +627,7 @@ jobs:
           if-no-files-found: error
   
   run-artifactbuilder:
-    needs: build-test-fullagent-msi
+    needs: [ create-package-rpm ]
     name: Run ArtifactBuilder
     runs-on: windows-2019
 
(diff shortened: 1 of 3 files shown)


.github/workflows/all_solutions.yml
```diff
@@ -678,3 +679,52 @@ jobs:
           name: deploy-artifacts
           path: ${{ github.workspace }}\build\BuildArtifacts
           if-no-files-found: error
+
+  create-package-rpm:
```
vuqtran88, line 683: since this job is required for the `run-artifactbuilder`, could you relocate this job the be before the `run-artifactbuilder` ?

build/Linux/docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 build_deb: 
     build: ./build/deb 
-    command: bash -c "dos2unix /deb/build.sh && /deb/build.sh"
+    command: bash -c "dos2unix /deb/build.sh && chmod 777 /deb/build.sh && /deb/build.sh"
```
vuqtran88, line 3: the `dpkg-deb` command that creates the debian package requires permission in the range `[???, 755]` so 777 does not work. This command works for me.

`bash -c "dos2unix /deb/build.sh && chmod -R 755 /deb && /deb/build.sh"`

.github/workflows/all_solutions.yml
jaffinito: yeah, we could,, but its not 100% required  i didn't want to rock the boat too much.  I can do that though.

build/Linux/docker-compose.yml
jaffinito: Ahh.  That was mostly a placeholder just to make sure everything had a permissions check.  I can will updated it based on your note, and we can double  check again when we review your Debian PR.

vuqtran88 (approved):
LGTM

PR #344, opened 2020-11-04
 Add the PFX for code signing into the workflow and remove self-signing

### LINKED ISSUE #343: Add the PFX for code signing into the workflow and remove self-signing
### Feature Description
In order to release the agent we need to sign the MSI files.  We use a PFX for this.  This file needs to brought into the workflow in a secure way and installed so that the WiX installer picks it up and does the signing.

Need to remove the current temporary self-signing step.

### Priority
Blocker

jaffinito:
Resolves #343 

### Description

Note: the secrets needed for these changes have already been added to repo.

- Removes self-signing step
- Add step to convert SIGNING_CERT to PFX
- Adds step to install cert

### Testing

New steps should build successfully and MSI must build successfully.

### Changelog
n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -214,10 +214,16 @@ jobs:
             ${{ github.workspace }}\src\Agent\newrelichome_x86_coreclr
           if-no-files-found: error
 
-      - name: Create Self-signed code signing cert
+      - name: Get Code Signing Certificate
+        id: write_cert
+        uses: timheuer/base64-to-file@v1
+        with:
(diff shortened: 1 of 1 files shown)


lspangler584 (commented):
Want to make sure the Secrets names aren't confusing (even though their use will be obvious if you look at the workflows). 
Is `SIGNING_CERT` specific enough? 
Can we call `CERT_PASSPHRASE` `SIGNING_CERT_PASSPHRASE`?

vuqtran88:
I'm ok with any names listed above. They don't see them confuse on the usage.

lspangler584 (approved):

PR #345, opened 2020-11-04
Build debian package

vuqtran88:
### Description

Updated Linux/docker-compose.yml to chomd 755 the `/deb` directory
Added script to build debian packages.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -626,7 +626,7 @@ jobs:
           path: C:\IntegrationTestWorkingDirectory\TestResults\**\*TestResults.xml
           if-no-files-found: error
 
-  create-package-rpm:
+  create-linux-packages:
     needs: build-test-fullagent-msi
     if: ${{ github.event.release }}
     name: Create RPM Package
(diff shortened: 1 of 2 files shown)


.github/workflows/all_solutions.yml
```diff
@@ -626,7 +626,7 @@ jobs:
           path: C:\IntegrationTestWorkingDirectory\TestResults\**\*TestResults.xml
           if-no-files-found: error
 
-  create-package-rpm:
+  create-linux-packages:
```
jaffinito, line 629: Is there a reason we are combining the building here?   I think it might be better to run each build in its own job so that that they run faster.  What are your thoughts?

vuqtran88: The reason is I wasn't sure if having 2 jobs running in parallels and creating artifacts in the same output directory `_build\CoreArtifacts` would cause a problem. I guess may be not, I can change that.

jaffinito: It shouldn't since it acts like a share

jaffinito (approved):
Looks great!

jaffinito (approved):
Still awesome!

PR #346, opened 2020-11-04
Use Base64 to store key in secrets and then recreate the key as needed.

### LINKED ISSUE #334: Add GPG key to repo as an encrypted file

jaffinito:
Resolves #334 

### Description

 - Adds step to rebuild gpg key from secret
- Passes in the key zip from rebuild step to docker-compose
- Removes previous step that had no gpg
- Creates keys dir
- Copies GPG to keys

### Testing

With the key being provided it should use that to succeed. However, this code will not run since it only runs on release.  See [here ](https://github.com/jaffinito/newrelic-dotnet-agent/actions/runs/344856818)for a run that shows the key being used.  Starting [here ](https://github.com/jaffinito/newrelic-dotnet-agent/runs/1351113936?check_suite_focus=true#step:8:973)is the log you can see the key being applied with the passphrase.

### Changelog

n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -663,6 +663,12 @@ jobs:
           fileName: 'gpg.tar.bz2'
           encodedString: ${{ secrets.GPG_KEY }}
 
+      - name: Copy GPG Key to keys
+        run: |
+          mkdir ${{ github.workspace }}/build/Linux/keys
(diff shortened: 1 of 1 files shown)


lspangler584 (approved):
Nice work!

PR #348, opened 2020-11-06
Guard code in api stub methods

nr-ahemsath:
### Description

A customer reported exceptions being thrown from one of our API methods when called without the agent attached (meaning that the stub methods in `NewRelic.cs` were being executed, rather than the "real" API methods that the stubs get rewritten to when the agent is attached).  This PR attempts to guard against any potential exceptions in two ways:

Add try/catch blocks around System.Diagnostics.Trace.WriteLine calls to prevent any exceptions thrown from use of this method from affecting a customer app.

Put nameof() around parameters passed to string.format to prevent any issues that might arise from trying to .ToString() objects passed in by users.

### Testing

Unit tests pass.  Did a quick smoke test of an app which used the API without the agent attached and it behaved as expected.

### Changelog

Changelog updated

### DIFF
--- a/src/Agent/NewRelic.Api.Agent/NewRelic.cs
+++ b/src/Agent/NewRelic.Api.Agent/NewRelic.cs
@@ -721,7 +721,7 @@ namespace NewRelic.Api.Agent
         {
             try
             {
-                System.Diagnostics.Trace.WriteLine(string.Format("NewRelic.SetUserParameters({0},{1},{2})", nameof(userName), nameof(accountName), nameof(productName)));
+                System.Diagnostics.Trace.WriteLine($"NewRelic.SetUserParameters({nameof(userName)},{nameof(accountName)},{nameof(productName)})");
             }
             catch
             {
@@ -747,7 +747,7 @@ namespace NewRelic.Api.Agent
         {
             try
             {
-                System.Diagnostics.Trace.WriteLine(string.Format("NewRelic.RecordCustomEvent({0},{1})", nameof(eventType), nameof(attributes)));
+                System.Diagnostics.Trace.WriteLine($"NewRelic.RecordCustomEvent({nameof(eventType)},{nameof(attributes)})");
             }
             catch
             {
@@ -801,7 +801,7 @@ namespace NewRelic.Api.Agent
         {
             try
             {
-                System.Diagnostics.Trace.WriteLine(string.Format("NewRelic.SetApplicationName({0},{1},{2})", nameof(applicationName), nameof(applicationName2), nameof(applicationName3)));
+                System.Diagnostics.Trace.WriteLine($"NewRelic.SetApplicationName({nameof(applicationName)},{nameof(applicationName2)},{nameof(applicationName3)})");
             }
             catch
             {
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic.Api.Agent/NewRelic.cs
```diff
@@ -586,7 +719,14 @@ public static void DisableBrowserMonitoring(bool overrideManual = false)
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
         public static void SetUserParameters(string? userName, string? accountName, string? productName)
         {
-            System.Diagnostics.Trace.WriteLine("NewRelic.SetUserParameters(String userName, String accountName, String productName)");
+            try
+            {
+                System.Diagnostics.Trace.WriteLine("NewRelic.SetUserParameters(String userName, String accountName, String productName)");
```
lspangler584, line 724: Wouldn't this fail the compile since the params aren't actually being referenced?

alanwest: We should do `nameof(paramName)` on all params of all methods regardless of whether they are nullable.

src/Agent/NewRelic.Api.Agent/NewRelic.cs
```diff
@@ -605,7 +745,14 @@ public static void SetUserParameters(string? userName, string? accountName, stri
...
         {
-            System.Diagnostics.Trace.WriteLine(string.Format("NewRelic.RecordCustomEvent({0},{1})", eventType, attributes));
+            try
+            {
+                System.Diagnostics.Trace.WriteLine(string.Format("NewRelic.RecordCustomEvent({0},{1})", nameof(eventType), nameof(attributes)));
+            }
+            catch
+            {
+                // Swallow any exception thrown from here
+            }
```
alanwest, line 755:
```suggestion
            System.Diagnostics.Trace.WriteLine(string.Format("NewRelic.RecordCustomEvent({0},{1})", nameof(eventType), nameof(attributes)));
```

alanwest: All the methods should be safe without the `try/catch` now that we're not attempting to `ToString` the parameters.

For example, this `RecordCustomEvent` method can be problematic if the underlying type's `ToString` method is not thread-safe when `attributes.ToString()` is called. The type would be something like:

```csharp
class MyEnumerable : IEnumerable<KeyValuePair<string, object>>
{
    // Performs work that is not thread-safe
    public override string ToString() ...
}
```

`Dictionary<string, object>` is such a type. https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.

nr-ahemsath: I don't think we positively established that the reason the customer (or any customer in general) might experience an exception from these tracing calls can only be a non-threadsafe `.ToString()`.  I'd prefer the belt-and-suspenders approach for maximum safety.

lspangler584:
After discussion, decided to leave in try/catch since we are not certain the customer exception issue is resolved, and try/catch would guard against any future issues.

lspangler584 (approved):

PR #350, opened 2020-11-09
Caching artifacts to reduce job time

jaffinito:
### Description

Add caching support to several jobs in all_solutions.yml.
- build-windows-profiler
- build-linux-profiler
- build-integration-tests
- build-unbounded-tests
- build-platform-tests

Can reduce the build time by at least 8 minutes if there are no profiler changes (pretty common).  If no changes are made to integration tests, unbounded tests and platform tests (all share components so changes in one result in not caching all) it could same up to 10 minutes. Build times are currently running around 45 minutes in most cases.

### Testing

Build steps should complete just normally, faster if they are cached. Artifacts from the cached jobs will still be created so we should not see see any upload errors.

### Changelog

n/a

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -682,7 +682,7 @@ jobs:
           agentVersion=$(head -n 1 ${{ github.workspace }}/agent_version.txt)
           agentVersion=${agentVersion/$'\r\n\t'}
           cd ${{ github.workspace }}/build/Linux
-          docker-compose build
+          docker-compose build build_rpm
           docker-compose run -e AGENT_VERSION=$agentVersion -e GPG_KEYS=/keys/gpg.tar.bz2 build_rpm
         shell: bash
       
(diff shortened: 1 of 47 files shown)


.github/workflows/all_solutions.yml
```diff
@@ -404,13 +464,30 @@ jobs:
...
         uses: microsoft/setup-msbuild@v1.0.1
 
+      - name: Get PlatformTests Hash
+        id: platformtests-hash
+        run: |
+          Write-Host "::set-output name=hash::${{ hashFiles('tests\Agent\IntegrationTests','tests\Agent\PlatformTests') }}"
+        shell: powershell
+
+      - name: Cache Tests
+        id: cache-platformtests # same for integration and unbounded since they are so intertwined.
```
nr-ahemsath, line 474: Don't think we need this comment here

nr-ahemsath (dismissed):
Great work improving the efficiency of our CI!

PR #354, opened 2020-11-10
Downloadsite deploy fixes

jaffinito:
- Removes AWS config step and uses env vars instead
- switches shell over to pwsh from powershell since it handles secrets a bit better.


nr-ahemsath (approved):
👍

PR #356, opened 2020-11-11
Remove profile from download site push

jaffinito:
Removes the --profiler bit from the previous_release aws s3 sync command.  Was already removed from latest_release.


nr-ahemsath (approved):
👍

PR #360, opened 2020-11-11
Output checksum table in markdown for release notes

nr-ahemsath:
### Description

This will help reduce toil in the release process.  The `DownloadSite` target of the ArtifactBuilder now creates a `checksum.md` file in the `BuildArtifacts/DownloadSite/SHA256` folder, which has the table of filenames and SHA256 checksum values that we add to the release notes as part of the release process.  It should be as simple as copying and pasting the contents of this file into the appropriate portion of the release notes.

*Note*: this also adds `build/Linux/keys` to the `.gitignore` file.  Let's not accidentally check in a signing key!

### Testing

To test, run `build/package.ps1` with the `-IncludeDownloadSite` option.  This assumes you have built all the other things that the artifact builder expects to be there.

### Changelog

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,7 +88,6 @@ Agent/NewRelic.Agent.IL/_NewRelic.Agent.IL.tlb
 newrelichome_*
 src/Agent/New Relic Home*
 
-
 # Ignore New Relic Home generated registry files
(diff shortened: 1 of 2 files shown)


lspangler584 (approved):
Walk through with Alex to discuss changes.

PR #367, opened 2020-11-13
build.ps1 should build linux packages with or without the RPM signing key

nr-ahemsath:
### Description

Signing the .rpm package for the core agent is only required on release.  For local builds, it is optional.  Before this PR, the `build.ps1` script would skip building either the .deb or the .rpm if the signing key was not supplied.  This would prevent community contributors from using this script to build the Linux packages because they will not have our signing key.

### Testing

I tested both cases (either supplying the signing key or not) and the Linux install packages were built in both cases.

Note: I am having a problem where I am unable to run the `build.ps1` script multiple times in the same Powershell session.  On running it the second time, it will end abnormally the first time it tries using Docker.  This seems to be due to some Powershell/Docker interaction that is not yet understood.  The workaround is to close the current Powershell session and open a new one in order to run the script a second time.

### Changelog

N/A

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -104,22 +104,27 @@ $agentVersion = (Get-Item "$rootDirectory\src\_build\AnyCPU-$Configuration\NewRe
 # Linux Packaging #
 ###################
 
+Write-Host "===================================="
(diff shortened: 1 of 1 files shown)


jaffinito (approved):
Changes look good!

PR #368, opened 2020-11-16
- Add back self-signed cert step

jaffinito:
### Description

In order to allow forks to build and test, we need to have add back the self-signing cert step.

- Add back self-signed cert step
- Self-signed step only runs on PR
- Update actual cert steps to only run on release

### Testing

Workflow should build using self-signed cert for PR since it is not a release build.

### Changelog

N/A

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -229,8 +229,8 @@ jobs:
           certutil.exe -f -user -p ${{ secrets.CERT_PASSPHRASE }} -importPFX ${{ steps.write_cert.outputs.filePath }} NoRoot
         shell: powershell
       
-      - name: Create Self-signed code signing cert
-        if: ${{ github.event.pull_request }}
+      - name: Create Self-signed code signing cert  
+        if: ${{ github.event.pull_request || github.event.workflow_dispatch }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/all_solutions.yml
```diff
@@ -215,16 +215,26 @@ jobs:
...
           
       - name: Install Code Signing Certificate
+        if: ${{ github.event.release }}
         run: |
+          Write-Host "certutil.exe -f -user -p <passphrase> -importPFX ${{ steps.write_cert.outputs.filePath }} NoRoot"
           certutil.exe -f -user -p ${{ secrets.CERT_PASSPHRASE }} -importPFX ${{ steps.write_cert.outputs.filePath }} NoRoot
         shell: powershell
+      
+      - name: Create Self-signed code signing cert
+        if: ${{ github.event.pull_request }}
```
nr-ahemsath, line 233: This will cover pull requests, and the step that is gated by `github.event.release` takes care of release builds.  What happens for manual runs or merges to `main`?

jaffinito: - Your right, I missed manual runs.  I'll add it.
- Merges to main only happen via the PR process.  We don't run any tests on merge anymore since we run all tests during the PR review phase.

nr-ahemsath: Got it, thanks

nr-ahemsath (approved):
Looks good.  I verified that the self-signed cert was used on the workflow run for this PR.  Approved. 👍

PR #369, opened 2020-11-16
Update or replace actions that use now disabled methods

jaffinito:
### Description

The day has come were GH is blocking any actions that use add-path, set-env, and the like. It appears that some of the actions we use are still using these methods.

- Moved from @v1 for microsoft/setup-msbuild to get updates
- Replaced base64 steps with scripting to fix add-path issue

### Testing

Workflow doesn't have errors related to the disabled methods.

### Changelog
n/a


nr-ahemsath (approved):
The PR build process seems to be going well at this point.  Approved.

Issue #301, opened 2020-10-13
Server Side Config for expected errors causing agent to shutdown


nrcventura:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

**Description**
The expected status codes returned with the server side config is returning an array instead of a string. The agent needs to correctly parse all of the server side config values for expected and ignored errors.

```
2020-10-12 00:20:30,144 NewRelic  ERROR: [pid: 3772, tid: 19] Json serializer context path: agent_config['error_collector.expected_status_codes']. Error message: Unexpected character encountered while parsing value: [. Path 'agent_config['error_collector.expected_status_codes']', line 1, position 472.2020-10-12 00:20:30,160 NewRelic  ERROR: [pid: 3772, tid: 19] Unable to connect to the New Relic service at http://collector-009.newrelic.com:443 : Newtonsoft.Json.JsonReaderException: Unexpected character encountered while parsing value: [. Path 'agent_config['error_collector.expected_status_codes']', line 1, position 472.
```

**Expected Behavior**
The agent should correctly parse the expected and ignored errors server side config and successfully start up.

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results**

**Steps to Reproduce**

**Your Environment**

**Additional context**
See [internal link](https://newrelic.atlassian.net/browse/GTSE-5109) for more details.

nrcventura:
Here's another related [internal link](https://newrelic.atlassian.net/browse/GTSE-5174) for this issue.

nrcventura:
@brsmith080 At this point in time, a setting was disabled on the New Relic backend that should allow the agent to work properly again, but without the ability to configure expected errors through the New Relic UI. To workaround that, you can configure expected error settings using the newrelic.config file. 

If you are still experiencing this problem downgrading the agent to version [8.30](https://docs.newrelic.com/docs/release-notes/agent-release-notes/net-release-notes/net-agent-83000) is an option.

PR #370, opened 2020-11-23
SQL parsing: ignore leading SET statements

### LINKED ISSUE #224: Sql Parsing should deprioritize set statements in multistatement SQL
Our sql parsing logic does not properly prioritize set statements appearing before the real SQL statement being executed. We only attempt to handle it today for SELECT statements.

Here's an example statement from a customer that doesn't work as expected.

SET CONTEXT_INFO = @0; SET NOCOUNT ON; EXEC dbo.spSomeProconOurSystem

This code will result in using the set statement to name the segment, when it should be using the exec statement instead.

As part of this work we should update all of our sql statement parsing to better handle having set statements appear before the real statement. The Cross Agent Test suite should be updated as well.

**MISC**
[New Relic Internal Only] - [Zendesk](https://newrelic.zendesk.com/agent/tickets/389313)

nr-ahemsath:
### Description

Resolves #224:
For a compound SQL statement like `SET @foo = 1; EXEC dbo.someStoredProcedure`, the parsed DB operation will be the `EXEC`, not the `SET`.  (This case *was* previously handled for SELECT, but no other operations.) 

### Testing

Additional unit tests were added to `SqlParserTests.cs` to ensure coverage.

### Changelog

Changelog is updated.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 New Feature Description
 
 ### Fixes
-Fixes Issue [#224](https://github.com/newrelic/newrelic-dotnet-agent/issues/224): leading "SET" commands will be ignored when parsing compound SQL statements.
+Fixes Issue [#224](https://github.com/newrelic/newrelic-dotnet-agent/issues/224): leading "SET" commands will be ignored when parsing compound SQL statements.([#370](https://github.com/newrelic/newrelic-dotnet-agent/pull/370))
 
 
 ## [8.35] - 2020-11-09
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -8,7 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### New Features
 * **New Feature Template** <br/>
 New Feature Description
-Fixes Issue [#XX](https://github.com/newrelic/newrelic-dotnet-agent/issues/XX)
+
+### Fixes
+Fixes Issue [#224](https://github.com/newrelic/newrelic-dotnet-agent/issues/224): leading "SET" commands will be ignored when parsing compound SQL statements.
+
 
 ## [8.35] - 2020-11-09
 
--- a/src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
+++ b/src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
@@ -39,7 +39,7 @@ namespace NewRelic.Parsing
         private const string SqlParamPrefix = "@";
 
         // Regex Phrases
-        private const string SelectPhrase = @"(?=^\bset\b.*;\s*\bselect\b|^\bselect\b).*?\s+";
+        private const string SelectPhrase = @"^\bselect\b.*?\s+";
         private const string InsertPhrase = @"^insert\s+into\s+";
         private const string UpdatePhrase = @"^update\s+";
         private const string DeletePhrase = @"^delete\s+";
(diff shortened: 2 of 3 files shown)


tests/Agent/UnitTests/ParsingTests/SqlParserTests.cs
```diff
@@ -466,11 +466,29 @@ public void SqlParserTest_TestInnerSelect()
...
 
             var parsedDatabaseStatement = SqlParser.GetParsedDatabaseStatement(DatastoreVendor.MSSQL, CommandType.Text, test);
             Assert.IsNotNull(parsedDatabaseStatement);
-            Assert.AreEqual("(subquery)", parsedDatabaseStatement.Model, string.Format($"Expected model (subquery) but was {parsedDatabaseStatement.Model}", "(subquery)", parsedDatabaseStatement.Model));
-            Assert.AreEqual("select", parsedDatabaseStatement.Operation, string.Format($"Expected operation select but was {parsedDatabaseStatement.Operation}", "select", parsedDatabaseStatement.Operation));
+            Assert.AreEqual(expectedModel, parsedDatabaseStatement.Model, $"Expected model {expectedModel} but was {parsedDatabaseStatement.Model}");
+            Assert.AreEqual(expectedOperation, parsedDatabaseStatement.Operation, $"Expected operation {expectedOperation} but was {parsedDatabaseStatement.Operation}");
+        }
+
+        [Test]
```
vuqtran88, line 478: [Test] here is redundant I think.

src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
```diff
@@ -129,6 +131,16 @@ public static ParsedSqlStatement GetParsedDatabaseStatement(DatastoreVendor data
                 // Remove comments.
                 var statement = CommentPattern.Replace(commandText, string.Empty).TrimStart();
 
+                if (!IsSingleSqlStatement(statement))
+                {
+                    // Remove leading SET commands
+
+                    // Trimming any trailing semicolons is necessary to avoid having the LeadingSetPattern
+                    // match a SQL statement that ONLY contains SET commands, which would leave us with nothing
+                    statement = statement.TrimEnd(';');
```
vuqtran88, line 140: If you could make ';' a constant, it would be great.

tests/Agent/UnitTests/ParsingTests/SqlParserTests.cs
nr-ahemsath: It isn't.  If you remove that [Test], after rebuilding one or more times, the Test Explorer will show 5 test cases for the `TestFromGeneratedGibberish` test, which is the next method with the [Test] attributed on it.

src/Agent/CHANGELOG.md
```diff
@@ -8,7 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### New Features
 * **New Feature Template** <br/>
 New Feature Description
-Fixes Issue [#XX](https://github.com/newrelic/newrelic-dotnet-agent/issues/XX)
+
+### Fixes
+Fixes Issue [#224](https://github.com/newrelic/newrelic-dotnet-agent/issues/224): leading "SET" commands will be ignored when parsing compound SQL statements.
```
vuqtran88, line 13: It is a good practice to also reference the PR at the end of the fix message . For example, "Fixes issue #301 where the agent incorrectly parses server-side configuration causing agent to shutdown.(#310)"

src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
nr-ahemsath: Done.

src/Agent/CHANGELOG.md
nr-ahemsath: Good idea, thanks.

vuqtran88 (approved):
LGTM

PR #371, opened 2020-11-25
Restore using msbuild

nr-ahemsath:
### Description

This fixes a build problem that recently started happening due to a mismatch between MSBuild and NuGet versions on the GHA runners, described here: https://developercommunity.visualstudio.com/content/problem/1248649/error-netsdk1005-assets-file-projectassetsjson-doe.html

The fix is to use MSBuild for package restore instead of using NuGet directly.

This PR also fixes an issue with the workflow logic which was gating the creation of a self-signed certificate on manual (workflow_dispatch) runs.

### Testing

If the PR build works as expected (i.e. all build steps pass, but there will be some expected failures in integration tests) this should be good to go.

### Changelog

N/A


vuqtran88 (approved):
👍

PR #372, opened 2020-11-30
Fix agent failed to read settings from appsettings.{environment}.json

vuqtran88:
### Description
Starting from .Net Core 2.1, for AspNetCore applications, the value from the `ASPNETCORE_ENVIRONMENT` environment variable will determine which  `appsettings.{environment}.json` files will be loaded. The agent currently does not use the `ASPNETCORE_ENVIRONMENT` environment variable for environment name; therefore, it ignores any New Relic specific settings in the `appsettings.{environment}.json`.

### Testing
Manually tested. Steps that I did for testing:
1. Create an AspNetCore application targets .Net Core 2.1 or greater.
2. Updates the value of the `ASPNETCORE_ENVIRONMENT` as `Staging` in the `launchSettings.json`
3. Creates an `appsettings.Staging.json` file with the following content:
```
{
  "Logging": {
    "LogLevel": {
      "Default": "Warning"
    }
  },
  "NewRelic.AppName": "Staging",
  "AllowedHosts": "*"
}
``` 
4. Runs the application with the new agent attached.
5. Checked the agent debug level logs to see whether the agent reports the test application name as "Staging".

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
@@ -25,9 +25,11 @@ namespace NewRelic.Agent.Core.Configuration
 		{
 			var env = new SystemInterfaces.Environment();
 			var currentDirectory = Directory.GetCurrentDirectory();
-			var environment = env.GetEnvironmentVariable("EnvironmentName");
+            var environment = env.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
+            if (string.IsNullOrEmpty(environment))
+                environment = env.GetEnvironmentVariable("EnvironmentName");
 
-			var builder = new ConfigurationBuilder()
+            var builder = new ConfigurationBuilder()
 				.SetBasePath(currentDirectory)
 				.AddJsonFile("appsettings.json", optional: true, reloadOnChange: false);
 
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
```diff
@@ -51,8 +53,8 @@ public static string GetAppSetting(string key)
 
 			if (Log.IsDebugEnabled)
 			{
-				if (string.IsNullOrWhiteSpace(value)) Log.Debug($"Reading value from appsettings.json: '{key}' not defined. Searched: {_appSettingsFilePaths}.");
-				else Log.Debug($"Reading value from appsettings.json: '{key}={value}'");
+				if (string.IsNullOrWhiteSpace(value)) Log.Debug($"Reading value from appsettings.json and appsettings.*.json: '{key}' not defined. Searched: {_appSettingsFilePaths}.");
```
lspangler584, line 56: These conditionals should be updated to use our coding conventions.

lspangler584:
I'm wondering if we should also consider env var DOTNET_ENVIRONMENT, as it also seems to be specific to asp net core configurations: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-5.0.

vuqtran88:
> I'm wondering if we should also consider env var DOTNET_ENVIRONMENT, as it also seems to be specific to asp net core configurations: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-5.0.

I think ASPNETCORE_ENVIRONMENT is reliable enough and supported in all versions of .net core. The DOTNET_ENVIRONMENT is only available from .net core 3.1. Moreover, ASPNETCORE_ENVIRONMENT will override DOTNET_ENVIRONMENT.

src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
```diff
@@ -8,55 +8,68 @@
...
+        {
+            get
+            {
+                return _configuration ?? (_configuration = InitializeConfiguration());
+            }
+        }
+
+        private static IConfigurationRoot InitializeConfiguration()
+        {
+            var env = new SystemInterfaces.Environment();
```
lspangler584, line 26: Slight perf improvement: Looks like this var declaration can be moved inside the if on line 29, since that is the only place it is used and may be avoided if ASPNETCORE_ENVIRONMENT is defined.

lspangler584 (approved):
Integration and Unbounded Integration Tests ran locally.

Issue #93, opened 2020-08-05
Update Async Timing Calculations


nr-ahemsath:
**Description**
The agent does not deduct the duration of asynchronous methods from parent span/segment/methods' exclusive duration, which leads to misleading durations reported when those asynchronous child methods are called in a synchronous manner.

One such scenario is ASP.NET Web API. We instrument `ExecuteRequestHandler` that calls `System.Web.Http.Controllers.ApiControllerActionInvoker.InvokeActionAsync` which we also instrument. `InvokeActionAsync` is an async method, so the agent does not deduct its time from the time of `ExecuteRequestHandler`. Since we know `ExecuteRequestHandler` actually does block on `InvokeActionAsync` it probably makes sense to special case these types of scenarios.

**Expected Behavior**
The reported exclusive duration of common ASP.NET middleware pipeline steps like `ExecuteRequestHandler` should not include the time spent in blocking async child methods like `InvokeActionAsync`.

Internal NR only:
[ref1](https://newrelic.atlassian.net/browse/DOTNET-4633)
[ref2](https://newrelic.zendesk.com/agent/tickets/407105)
[ref3](https://docs.google.com/drawings/d/1eSncVm2WvbHHVldSFL3nGeC3IqVuAjpCXMoWzPW0OaI/edit?ts=5d533145)

angelatan2:
This update was resolved in 8.36.

PR #374, opened 2020-12-03
Fix async calculation

lspangler584:
### Description
Resolves Issue #93 
In cases where asynchronous child methods are called in a synchronous manner, the duration of the parent method should not include the duration of the child method. These cases include ASP.NET Web API, ASP.NET Core, OWIN, and some WCF async instrumentation scenarios.
 
The solution provides a property to use in known cases to mark the child segment so that duration calculation of the parent will not include the child's duration.

For example:
- in the case of ASP.NET Core, the call to the `next` pipeline step is `await`ed (blocked). Therefore, the calculation of the parent step should not include the duration of the child (`next`) step.
- in the case of WCF, async operations use either the Asynchronous Programming Model (APM or IAsyncResult) or the Task-based Asynchronous Pattern (TAP). For APM, the agent instruments `AsyncMethodInvoker.InvokeBegin`, which returns an `IAsyncResult` and does not wait or block so the `InvokeBegin`'s segment duration is not included in the parent's (segment or transaction) duration. However, for TAP style methods in WCF the Agent instruments `TaskMethodInvoker.InvokeAsync`, which awaits the invocation, thereby behaving as a synchronous method. In this case, the child segment duration will now be deducted from the parent's duration.

### Testing
Unit test added to verify the new property, when present, will deduct the child's duration from the calling parent.
Integration Test applications were used to observe the desired behavior in the UI.

Example: WebApiAsyncTests runs a WebApi application using the Integrated Pipeline. ExecuteRequestHandler is the parent pipeline step, AsyncAwait.CpuBoundTasksAsync() is the controller method requested (child segment).

Before (ExecuteRequestHandler duration includes the duration of the controller method):
image

With fix (ExecuteRequestHandler duration does not include the duration of the controller method):
image


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Asp35/IntegratedPipeline/ExecuteStepWrapper.cs
```diff
@@ -90,8 +90,9 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
 
             transaction = TryCreateTransaction(agent, httpContext, requestNotification);
             segment = transaction.StartTransactionSegment(instrumentedMethodCall.MethodCall, requestNotification);
+            segment.AlwaysDeductChildDuration = true;
 
-            httpContext.Items[HttpContextActions.HttpContextSegmentKey] = segment;
+        httpContext.Items[HttpContextActions.HttpContextSegmentKey] = segment;
```
vuqtran88, line 95: need indentation.

src/Agent/CHANGELOG.md
```diff
@@ -11,10 +11,14 @@ New Feature Description
 
 ### Fixes
 Fixes Issue [#224](https://github.com/newrelic/newrelic-dotnet-agent/issues/224): leading "SET" commands will be ignored when parsing compound SQL statements. ([#370](https://github.com/newrelic/newrelic-dotnet-agent/pull/370))
+
 Fixes Issue [#226](https://github.com/newrelic/newrelic-dotnet-agent/issues/226): the profiler ignores drive letter in `HOME_EXPANDED` when detecting running in Azure Web Apps. ([#373](https://github.com/newrelic/newrelic-dotnet-agent/pull/373))
 
+Fixes Issue [#93](https://github.com/newrelic/newrelic-dotnet-agent/issues/93): Update Async Timing Calculations ([#370](https://github.com/newrelic/newrelic-dotnet-agent/pull/374))
```
vuqtran88, line 17: Fix Issue [#93](https://github.com/newrelic/newrelic-dotnet-agent/issues/93): When the parent methods are blocked by their asynchronous child methods, the agent deducts the child methods' duration from parent methods' exclusive duration.([#374](https://github.com/newrelic/newrelic-dotnet-agent/pull/374))

vuqtran88 (approved):
LGTM. Great work !

PR #376, opened 2020-12-07
Dont dispose listeners

### LINKED ISSUE #234: Agent should not crash app when performance EventListeners get error
**Description**
Customer ticket reported their app crashed when a Fatal Internal CLR error occurred during [Dispose of the ThreadEventsListener](https://github.com/newrelic/newrelic-dotnet-agent/blob/e88e01afb2821897e4adcaefe8ac058c2bfc1b89/src/Agent/NewRelic/Agent/Core/Samplers/ThreadStatsSampler.cs#L85). At present, this does not appear to be a bug in the Agent, but the Agent should attempt to avoid or manage its occurrence. 

The error from the customer did not show a specific exception, but rather 
`Fatal Error. Internal CLR error. (0x80131506) at System.Diagnostics.Tracing.EventPipeInternal.Disable(UInt64) at
System.Diagnostics.Tracing.EventPipeInternal.Disable(UInt64) at 
System.Diagnostics.Tracing.EventPipeEventDispathcher.CommitDispatchConfiguration() at
System.Diagnostics.Tracing.EventPipeEventDispatch.RemoveEventListener(System.Diagnostics.Tracing.EventListener) at
System.Diagnostics.Tracing.EventListener.RemoveReferencesToListenerInEventSources(System.Diagnostics.Tracing.EventListener) at
System.Diagnostics.Tracing.EventListener.Dispose() at
NewRelic.Agent.Core.Samplers.SampledEventListener...Dispose() at
NewRelic.Agent.Core.Samplers.ThreadStatsSampler.Stop() at
NewRelic.Agent.Core.Configuration.ConfigurationService.OnServerConfigurationUpdated(...)...`

I think `EventPipeInternal.Disable()` is unmanaged code (needs verification), hence the Internal CLR error and not an Exception. It is unlikely this error will generate an exception up the call stack to the managed Agent since it is a Fatal CLR error, but adding exception handling on Dispose may be good practice anyway. Another attempt to avoid this error may be to avoid recycling the listeners OnServerConfigurationUpdated - where a cycle of shutdowns and starts occur in quick succession -  and instead do it on a less frequent event such as OnAgentConnected.

[Internal Only] [Support ticket](https://newrelic.atlassian.net/browse/GTSE-4535)

**Expected Behavior**
Agent should not crash the app.

**Workaround**
Disable the collection of Garbage Collection and Thread Statistics metric collection by setting the `NewRelic.EventListenerSamplersEnabled` `appSetting` to `false` in your `newrelic.config` file. 
```
    <appSettings>
        <add key="NewRelic.EventListenerSamplersEnabled" value="false" />
    </appSettings>
```
The `appSettings` element is a child element of the `configuration` element. Disabling these metrics will still allow the collection of Memory and CPU usage.

**Steps to Reproduce**
No repro with the Agent is available. Problem is intermittent.
A [test app](https://github.com/newrelic/newrelic-dotnet-agent/files/5563239/ConsoleAppNetCore.zip) exists without the Agent that simultaneously creates/disposes an EventListener on multiple threads; this app produces the Fatal CLR error about 2 out of every 3 runs (still intermittent, but more reliably so).

It's possible there is some interference between customer app and Agent use of EventListeners. (e.g. customer code subscribing to or disposing an event source at the same time as the Agent). Customer Environment shows use of `Microsoft.Extensions.Logging.EventSource`, but customer never confirmed they are using EventListeners.

**Your Environment**
Customer reporting the incident was using .NET Core 3.1 in a console (.exe) app.

**Discovery So Far**
It is unnecessary to Create/Dispose the Agent's EventListeners on every ConfigurationUpdatedEvent. It is sufficient to Disable/Enable Events for the EventPipe-based EventListeners in the Agent. This work has been started on this [branch](https://github.com/newrelic/newrelic-dotnet-agent/tree/DontDisposeListeners). By making this change in the test app mentioned above (Disable/Enable instead of Dispose/Create listener), the Fatal CLR error has not occurred. That does not necessarily mean the branch addresses all causes of the problem, but it seems to have reduced the risk of occurrence. Before this branch is approved, there should be some load testing to assure memory is not leaking.

Another scenario that has not yet been explored is that the EventListener Dispose() is occuring while a GC is happening.

lspangler584:
### Description
Resolves #234 .
This PR reduces the likelihood of a Fatal CLR Error due to rapidly Creating/Disposing the Agent's EventListeners for performance metrics, a situation which occurs in the Agent upon startup and Configuration update.

A Fatal CLR Error was generated in a test app by rapidly Creating/Disposing EventListeners,  By Disabling Events, but not Disposing the listener, then Enabling Events on the same listener instead of instantiating a new listener, this Fatal CLR Error was avoided in the test app. Hence, the Agent was modified to use Disable/Enable pattern instead of the Dispose/Instantiate pattern.

### Testing
Testing to verify no memory leaks, due to previous issues with the Agent's performance metrics EventListeners:
A netcoreapp3.1 console [app](https://github.com/newrelic/newrelic-dotnet-agent/files/5655055/GenerateGCThreadActivityFix.zip) was created to generate GC and thread activity. One copy of the app was soak tested on one server with the existing version of the Agent, and another copy of the test app was soak tested with this PR version of the Agent.
In both cases, memory use was stable over time.

Unit and Integration Tests pass.


src/Agent/NewRelic/Agent/Core/Samplers/GCSamplerNetCore.cs
```diff
@@ -110,20 +110,20 @@ public override void Start()
                 }
 
                 _listener = _listener ?? _eventListenerFactory();
+                _listener.StartListening();
             }
             catch (Exception ex)
             {
                 Log.Error($"Unable to start Garbage Collection Event Listener Sample.  Further .NetCore GC metrics will not be captured.  Error: {ex}");
                 Stop();
+                Dispose();
```
vuqtran88, line 119: is there a reason why we call Stop() and Dispose() ? Can we just call Dispose () ?

src/Agent/NewRelic/Agent/Core/Samplers/ThreadStatsSampler.cs
```diff
@@ -50,8 +50,8 @@ public override void Sample()
             catch (Exception ex)
             {
                 Log.Error($"Unable to get Threadpool stats sample.  No .Net Threadpool metrics will be reported.  Error : {ex}");
-
                 Stop();
+                Dispose();
```
vuqtran88, line 54: same with above comment.

src/Agent/NewRelic/Agent/Core/Samplers/GCSamplerNetCore.cs
lspangler584: no good reason, there's a bad reason, but I won't bore you with that :)

nr-ahemsath (approved):
Looks good to me. 👍

vuqtran88 (approved):
LGTM

PR #386, opened 2020-12-08
fix unit test

lspangler584:


vuqtran88:
Are we supposed to test the main use case which is when sampler stops, don't dispose the listener ?

vuqtran88 (approved):
👍

PR #391, opened 2020-12-10
Remove redundant RabbitMq{Legacy}ReceiveTests

nr-ahemsath:
### Description

As part of consolidating RabbitMq testing infrastructure in preparation for expanding tested cases, we found that the RabbitMq{Legacy}ReceiveTests were redundant with other RabbitMq integration tests.  This PR deletes those test files and their associated fixture and test applications.

There was one assertion in the Receive tests, on a transaction trace segment, that did not exist in the remaining tests, but has been ported over.

### Testing

I verified that the remaining RabbitMq tests build and pass locally.

### Changelog

N/A


lspangler584:
It looks like we can remove the CreateQueues... method from RabbitMqUtils as well.

lspangler584 (approved):

PR #398, opened 2020-12-23
Port RabbitMQ integration tests to use ConsoleMultiFunction test apps

### LINKED ISSUE #383: Consolidate existing RabbitMQ tests
### Feature Description
Existing RabbitMQ UnboundedIntegrationTests only target .NET Framework, and 2 versions of RabbitMQ.Client (3.5.2 and 5.1.0). As part of this Milestone, we will be adding testing using .NET Core applications and RabbitMQ.Client 6.2.1. The current testing infrastructure uses separate tests, fixtures, and applications for each RabbitMQ version we test. Now that we are adding an additional .NET framework and an additional RabbitMQ version, using existing infrastructure pattern will (has) become unwieldy. 

### Describe Alternatives
* tests and fixtures
Inheritance may be used for consolidation of tests and fixtures, an example is used in our [Owin](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/tests/Agent/IntegrationTests/IntegrationTests/Owin/OwinWebApiTests.cs) tests.

* applications
  - [PackageReference Conditions](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#adding-a-packagereference-condition) (for applications) may allow consolidation to one application that can conditionally load the version of RabbitMQ being tested. 
  - ~~Alternatively, load RabbitMQ libraries [programmatically](https://www.infosysblogs.com/microsoft/2007/04/loading_multiple_versions_of_s.html).~~

  - Consider ConsoleMultiFunctionApp. 
      - ConsoleMultiFunctionApp(s) are currently only in IntegrationTests, not UnboundedIntegrationTests. To use for Rabbit testing, the app(s) will need to accessible from both - maybe by moving to Shared.
      - ConsoleMFApp will need to be multi-targeted, which means changing to SDK-style .csproj.
      - Rabbit code currently in RabbitMqBasicMvcApplication will need to be ported to an .exe and triggered by ConsoleMFApp.

### Priority
Really want

nr-ahemsath:
### Description

Resolves #383 by modifying the RabbitMQ integration tests to use the ConsoleMultiFunction (FW/Core) test apps instead of web test apps.  By doing this, we are able to test multiple versions of .NET Framework/Core and multiple RabbitMQ client library versions with the same two test apps, by using multitargeting and framework-conditional package references.

Another large PR that should be reviewed commit-by-commit.  Summary of what was done:
* Added the new SharedApplications folder to the UnboundedIntegrationTests solution
* Created a new RabbitMQ test method library in MultiFunctionApplicationHelpers, mirroring the controller methods from the RabbitMqBasicMvcApplication
* Converted existing RabbitMQ tests for .NET Framework to use the ConsoleMultiFunctionApplicationFW test app
  * RabbitMQ.Client 3.5.2 is used in the net461 version of ConsoleMFAppFW
  * RabbitMQ.Client 5.1.0 is used in the net471 version of ConsoleMFAppFW
* Added a .NET Core version of the existing RabbitMQ tests using RabbitMQ.Client 5.1.0
  * There should be a followup PR soon to add tests for RabbitMQ.Client 6.x 

### Testing

Verified RabbitMQ int. tests pass.  Verified integration tests pass (since changes to code in SharedApplications now affects both solutions).

### Changelog

N/A


tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
```diff
@@ -18,6 +18,14 @@
     <PackageReference Include="System.Diagnostics.Process" Version="4.3.0" />
   </ItemGroup>
```
lspangler584, line 20: as discussed in our PR reviews, this would be a good place for a comment explaining that we are mapping FW versions to Rabbit versions.

lspangler584:
holding approval for Profiler build problems (unlikely related to this PR?)

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
nr-ahemsath: Good idea.  Done.

lspangler584 (approved):

Issue #406, opened 2020-12-29
Add NEW_RELIC_LOG environment variable


lspangler584:
### Feature Description
NEW_RELIC_LOG specifies the unqualified filename for the agent's log file.
Include Unit Tests and documentation [here](https://docs.newrelic.com/docs/agents/net-agent/configuration/net-agent-configuration#environment-variables). 

### Additional context
See Java for description: https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file#Environment_Variables
Question: is supporting stdout a requirement?

### Priority
[~~Nice to Have, Really Want,~~ Must Have, ~~Blocker~~]

PR #410, opened 2020-12-31
adds NEW_RELIC_LOG

lspangler584:
### Description
Resolves Issue #406.
Adds NEW_RELIC_LOG environment variable to specify the unqualified filename for the agent's log file.
This environment variable takes precedence over other log file name settings.
Fixes and optimizes SafeFileName method.

### Testing
Updates ChangeLogFilenameTests (Integration Tests) and adds SafeFileName tests (Unit tests).

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
+++ b/src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
@@ -484,7 +484,13 @@ namespace NewRelic.Agent.Core.Config
 
         private string GetLogFileName()
         {
-            string name = fileName;
+            string name = System.Environment.GetEnvironmentVariable("NEW_RELIC_LOG");
+            if (name != null)
+            {
+                return name;
+            }
+
+            name = fileName;
             if (name != null)
             {
                 return Strings.SafeFileName(name);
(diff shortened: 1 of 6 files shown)


src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
```diff
@@ -484,7 +484,13 @@ public string GetFullLogFileName()
...
         private string GetLogFileName()
         {
-            string name = fileName;
+            string name = System.Environment.GetEnvironmentVariable("NEW_RELIC_LOG");
+            if (name != null)
+            {
+                return name;
+            }
+
+            name = fileName;
```
angelatan2, line 493: The first if (name != null) block if true will never hit the following if block (name != null) especially since the conditions are similar.  Is it true that you never want to hit the following if block that returns return Strings.SafeFileName(name)?

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
```diff
@@ -484,7 +484,13 @@ public string GetFullLogFileName()
...
             if (name != null)
             {
                 return Strings.SafeFileName(name);
```
nr-ahemsath, line 496: Should we also pass the value of `name` we get from the env var through this `SafeFileName` method?

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
lspangler584: yes. if the NEW_RELIC_LOG env var is set, that takes precedence over any other naming.

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
lspangler584: yes. working on that. found a bug in `SafeFileName`. Good catch.

nr-ahemsath (approved):
Looks good to me. 👍

PR #418, opened 2021-01-06
Add more RabbitMQ client version/runtime combinations to RabbitMQ integration tests

nr-ahemsath:
### Description

1. Add some additional combinations of RabbitMQ.Client version and .NET runtime/platform version:

RabbitMQ 5.1.0 + .NET Core (netcoreapp2.1)
RabbitMQ 6.2.1 + .NET Framework (net48)

2. Introduce a new naming scheme for RabbitMQ test classes to make it more clear what versions/runtimes are being tested:

RabbitMQ{FW,NetCore}{RabbitMQ.Client Version}Tests

### Testing

RabbitMQ integration tests all pass locally.  All (bounded) integration tests pass locally (except a small number (<3) of known flaky tests).

### Changelog

N/A


nrcventura:
Just something to think about, and it shouldn't block this PR, but netcoreapp2.1 will eventually be unsupported my MSFT in about a year. When that happens you may need to use a different strategy for testing different library versions, or shift the libraries to use different .supported net core versions.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/RabbitMQ.cs
```diff
@@ -176,7 +176,7 @@ private string BasicGetMessage(string queueName)
 
             // See this project's .csproj file for target framework => RabbitMQ.Client mappings
             // 'netcoreapp3.1' is tied to RabbitMQ.Client version 6.x+, making this necessary
```
lspangler584, line 178: may want to add that NET48 is also using 6.2.1, or genericize the language so the comment doesn't need to be updated for each change

nr-ahemsath: I reorganized things so that we `#define` a `RABBIT6PLUS` symbol conditionally based on the target framework at the top of the file, and then use the symbol in the `#if` directives down in the actual code.  Let me know if you think that helps to clarify what is going on.

vuqtran88 (approved):
All RabbitMq tests passed after this revert [PR](https://github.com/newrelic/newrelic-dotnet-agent/pull/424). LGTM!

PR #424, opened 2021-01-12
8t logging improvement

vuqtran88:
### Description

The following improvements have been done to grpc logging:

1. Changes logging level from `FINEST` to `INFO` for the `Creating gRPC channel to endpoint ...` log message. Also removes the `(attempt {attemptId})` from this log message because it was always gonna be `(attempt 0)` which is unnecessary.

2. Changes logging level from `FINEST` to `INFO` for the `gRPC channel to endpoint {EndpointHost}:{EndpointPort} connected.` log message. Also adds `(attempt {attemptId})` to this log message so that we know exactly how many attempts the agent tries to establish the connection.

3. Refines the log message for the expected errors during data transmission. 

### Testing

Manually tested.


nr-ahemsath (approved):
LGTM 👍

PR #425, opened 2021-01-14
Infinite tracing handle failed precondition

### LINKED ISSUE #412: 8T Polish - agent should handle a FAILED_PRECONDITION (HTTP 400) response from the 8T collector by disconnecting and reconnecting
One of the two user stories we are servicing in this milestone reads: "As a user of Infinite Tracing, I expect that I would not lose span data when New Relic redirects my traffic to a different cell."

There is an associated agent spec:
```
A FAILED_PRECONDITION status code from gRPC indicates that the Trace Observer URL now resolves to a new destination and that the agent needs to reconnect. The agent MUST close the TCP connection and connect to the cell identified by the new resolution of the DNS record. For languages that cache DNS resolution, the cache for the Trace Observer URL MUST be cleared. Agents SHOULD retain span events while re-establishing the stream, and SHOULD transmit the cached span events to the new stream.

An agent that continues to recieve FAILED_PRECONDITION responses after attempting to reconnect indicates that the DNS record is cached at a layer other than the application and that the update to the DNS record is not being propagated to the application. Subsequent reconnection attempts after the first SHOULD be attempted with the same backoff strategy as for connect requests, until the connection is established, at which point the backoff strategy should reset. The connection can be considered established when a normal RecordStatus is recieved by the agent or when the agent detects it successfully sent a Span.
```

The is the general area of the agent code that needs to be modified: https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs

nr-ahemsath:
### Description

Resolves #412.  Add logic to handle a "FAILED_PRECONDITION" GRPC status response, which indicates that the agent should shut down and restart the 8T span streaming service in order to handle being moved to a different host (IP address).

### Testing

Two existing 8T unit tests which verified that the span streaming service would be shut down and restarted upon receiving a status code of "UNAVAILABLE" are extended to verify the same behavior when receiving the FAILED_PRECONDITION status.

I did a manual smoke test of infinite tracing and did not see any unexpected behavior changes.  There is another story for manually testing the "traffic being moved to another host" scenario.

### Changelog

Updated changelog.


vuqtran88:
looks like the new unit tests failed.

nr-ahemsath:
> looks like the new unit tests failed. More specifically, this line `var endpointIpAddr = Dns.GetHostAddresses(EndpointHost)[0].ToString() ?? "unknown" `threw exception during the tests due to the mock endpoint `infiniteTracing.net`. Since this call potentially threw exceptions, we might want to handle them.

Good catch, it's fixed now.

vuqtran88 (approved):
I have inspected the logic of handling `FAILED_PRECONDITION` status conforms with the specs which are:
1. The agent MUST close the TCP connection and connect to the cell identified by the new resolution of the DNS record.
2. Agents SHOULD retain span events while re-establishing the stream, and SHOULD transmit the cached span events to the new stream.

LGTM!

PR #430, opened 2021-01-19
use existing string instead of creating a new one

lspangler584:
### Description
GetType() returns a Type object that already has the string needed to compare. No need to create another one.

### Testing
RabbitMQ unbounded tests pass locally.

### Changelog
N/A


nr-ahemsath (approved):
LGTM, rabbit tests pass for me locally too. 👍

PR #431, opened 2021-01-20
8T integration tests

vuqtran88:
### Description

This PR adds initial integration tests for the Agent Infinite Tracing functionality. The scope of this PR is limited to Infinite Tracing smoke tests on different flavors of .NET Frameworks and Core. The set of tests doesn't test all infinite tracing case scenarios.

In order to run the tests, trace observer url need to be configured in the local secrets.json. Adds the line at the root node of your local secret file.
```
{
...
...
"TraceObserverUrl" : "_REPLACE_WITH_A_TRACE_OBSERVER_URL_"
}
```

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingTests.cs
@@ -47,7 +47,7 @@ namespace NewRelic.Agent.IntegrationTests.InfiniteTracing
             var sentCount = _fixture.AgentLog.GetInfiniteTracingAttemptToSendSuccessCount();
             var receivedCount = _fixture.AgentLog.GetInfiniteTracingGrpcServerReceivedCount();
 
-            var senderExpectedMetrics = new List<Assertions.ExpectedMetric>
+            var actualMetrics = new List<Assertions.ExpectedMetric>
             {
                 new Assertions.ExpectedMetric { metricName = @"Supportability/InfiniteTracing/Span/Seen", callCount = sendCount },
                 new Assertions.ExpectedMetric { metricName = @"Supportability/InfiniteTracing/Span/Sent", callCount = sentCount },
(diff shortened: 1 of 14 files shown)


tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                }
+            );
+
+            _fixture.Initialize();
+        }
+
+        [Fact]
+        public void Test()
+        {
+            var expectedSeenCount = 2;
```
nr-ahemsath, line 47: Given that the way the test application is being exercised (one method that creates a transaction is being called only once), what is causing the expected count of span(s) seen, sent and received to be 2?

vuqtran88: https://github.com/newrelic/newrelic-dotnet-agent/blob/5b375b3c38517c8326aac0a70a4901f99f1a483c/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs#L322. According to this logic, we always send an extra count. I believe 1 span is the instrumented method, the extra is for root span.

nr-ahemsath: Cool, thanks.  I think we should add a comment to the test to explain this to the next engineer who looks at it.

vuqtran88: Added.

nr-ahemsath (approved):
LGTM.  New tests pass for me locally.  I added the new `TraceObserverUrl` entry to the `example-secrets.json` file.  👍

vuqtran88:
> LGTM. New tests pass for me locally. I added the new `TraceObserverUrl` entry to the `example-secrets.json` file. 👍

Good deal !

PR #432, opened 2021-01-21
Markdown dead link check

nr-ahemsath:
### Description

This adds a workflow to check for broken links in all our markdown files.  It also fixes a bunch of links that have been broken for a while, which should explain why this is a good idea.

### Testing

Since this action is configured to run on push, I was able to see it run on this branch before PRing it.  The last run was successful.

### Changelog

N/A


vuqtran88 (approved):
👍

PR #433, opened 2021-01-25
8t more failed precondition handling

nr-ahemsath:
### Description

1. Updates the handling of GRPC responses to catch `FAILED_PRECONDITION` and restart the infinite tracing service per the spec.
2. Added support for an additional internal testing parameter which specifies which GRPC status code the backend should randomly throw (the `flaky_code`).
3. Improved handling of the `INTERNAL` GRPC response code, which indicates that the current stream has been reset due to inactivity, and which the agent handles by disposing of the current stream and then recreating it.
4. General tweaks to log message levels and wording.

### Testing

1. Added a unit test to cover handling of the `INTERNAL` status code response when sending spans.
2. Verified that the 8T integration tests pass.

### Changelog

N/A

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -1051,6 +1051,15 @@ namespace NewRelic.Agent.Core.Configuration
                 _infiniteTracingObserverTestFlaky = TryGetAppSettingAsFloat("InfiniteTracingSpanEventsTestFlaky");
             }
 
+            if (int.TryParse(_environment.GetEnvironmentVariable("NEW_RELIC_INFINITE_TRACING_SPAN_EVENTS_TEST_FLAKY_CODE"), out var flakyCodeVal))
+            {
+                _infiniteTracingObserverTestFlakyCode = flakyCodeVal;
+            }
+            else
+            {
+                _infiniteTracingObserverTestFlakyCode = TryGetAppSettingAsInt("InfiniteTracingSpanEventsTestFlakyCode");
+            }
+
             if (int.TryParse(_environment.GetEnvironmentVariable("NEW_RELIC_INFINITE_TRACING_SPAN_EVENTS_TEST_DELAY"), out var delayVal))
             {
                 _infiniteTracingObserverTestDelayMs = delayVal;
@@ -1074,6 +1083,20 @@ namespace NewRelic.Agent.Core.Configuration
                 }
 
                 return _infiniteTracingObserverTestFlaky;
+            }
+        }
+
+        private int? _infiniteTracingObserverTestFlakyCode;
+        public int? InfiniteTracingTraceObserverTestFlakyCode
+        {
+            get
+            {
+                if (!_infiniteTracingObtainedSettingsForTest)
+                {
+                    GetInfiniteTracingFlakyAndDelayTestSettings();
+                }
+
+                return _infiniteTracingObserverTestFlakyCode;
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1051,6 +1051,15 @@ private void GetInfiniteTracingFlakyAndDelayTestSettings()
                 _infiniteTracingObserverTestFlaky = TryGetAppSettingAsFloat("InfiniteTracingSpanEventsTestFlaky");
             }
 
+            if (int.TryParse(_environment.GetEnvironmentVariable("NEW_RELIC_INFINITE_TRACING_SPAN_EVENTS_TEST_FLAKY_CODE"), out var flakyCodeVal))
```
lspangler584, line 1054: Maybe not for this PR, but we should consider adding the NEW_RELIC_INFINITE_TRACING_... env vars to our [docs ](https://docs.newrelic.com/docs/agents/net-agent/configuration/net-agent-configuration)site.
image

src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
```diff
@@ -865,40 +888,38 @@ private TrySendStatus TrySend(int consumerId, IClientStreamWriter<TRequestBatch>
...
+                    default:
+                        var rpcEx = grpcEx.InnerException as RpcException;
+                        LogMessage(LogLevel.Finest, consumerId, $"Attempting to send {items.Count} item(s) - gRPC Exception: {rpcEx?.Status.ToString() ?? grpcEx.Status}");
+                        break;
                 }
             }
             catch (Exception ex)
             {
-                LogMessage(LogLevel.Finest, consumerId, $"Attempting to send {items.Count} item(s)", ex);
+                LogMessage(LogLevel.Debug, consumerId, $"Unknown execption attempting to send {items.Count} item(s)", ex);
```
lspangler584, line 922: misspell exceptoin

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
nr-ahemsath: That's a good idea.  However, we should exclude the env vars that configure the "flaky" test settings because those are only for internal testing.

Some of the configurable settings, like the span batch size, queue size and partition size, probably need to be explained in  enough detail so that customers can understand why they would want to change these settings (or leave them alone).

src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
nr-ahemsath: Good catch, fixed.

src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs
```diff
@@ -40,30 +42,29 @@ public ResponseStreamWrapper(int consumerID, IAsyncStreamReader<TResponse> respo
...
         private async Task<int> WaitForResponse()
         {
-            bool success;
+            var success = false;
             try
             {
                 success = await _responseStream.MoveNext(_streamCancellationToken);
             }
-            catch (Exception ex)
+            catch (RpcException rpcEx)
```
vuqtran88, line 50: Confirmed that `success = await _responseStream.MoveNext(_streamCancellationToken)` will never result an `AggregateException`. The `await` mechanism implicitly unwraps propagated AggregateException, always leaving the underlying exception. 👍 on this catch.
LGTM!

vuqtran88 (approved):
LGTM! 🐑 it.

PR #442, opened 2021-01-28
Add password setting env vars to dockerfile

nr-ahemsath:
### Description

This resolves issues with unbounded integration tests on the `main` branch that require the MS SQL Server service.

It's not clear from git history, because things recently got moved around and deleted when we got rid of the Windows/Linux services split on the main branch, how these two env vars got removed from the Dockerfile for the mssql service.  However, the version on the `net35/main` branch was working, and these two vars were the only difference.  I think that, somehow, not having these vars set prevented the `NewRelic` database from being restored from the backup file.

### Testing

MS SQL unbounded integration tests now pass for me locally when running against the local docker `mssql` service.

### Changelog

N/A


lspangler584 (approved):
hooray!

PR #443, opened 2021-01-28
Fixes for net35/main deploy issues

nr-ahemsath:
### Description

This PR fixes a few issues with the `deploy.yml` workflow on the `net35/main` branch (for deploying 6.x agent versions).

1. The `deploy` tools folder doesn't exist in 6.x, because those tools are for pushing the .NET Core Linux packages to APT and YUM repos.  We don't build any .NET Core assets for 6.x because customers wanting to monitor .NET Core apps can use the latest agent versions.
2. The `NugetAgent` nuget package doesn't exist in 6.x, so it doesn't need to be deployed.

### Testing

I did a test deploy run on the latest version of this branch and it succeeded: https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/518358727

### Changelog

N/A

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -111,20 +111,6 @@ jobs:
           nuget.exe setApiKey ${{ secrets.NUGET_APIKEY }} -Source ${{ env.nuget_source }}
         shell: pwsh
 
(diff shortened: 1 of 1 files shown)


lspangler584 (approved):
sweet

PR #455, opened 2021-02-09
Add dotnet version supportability metrics for net5

vuqtran88:
### Description

Add  `Supportability/Dotnet/NetCore/net5` for .Net 5.

### Testing

Manually tested. Steps I did for testing:
1. Created a vanilla .net core 3.0 app from Visual studio template.
2. Changed the target framework in the app project file to the following target framework and checked for the following supportability metrics.

   `netcoreapp20` expects `Supportability/Dotnet/NetCore/LessThan30`
   `netcoreapp21` expects `Supportability/Dotnet/NetCore/LessThan30`
   `netcoreapp22` expects `Supportability/Dotnet/NetCore/LessThan30`
   `netcoreapp30` expects `Supportability/Dotnet/NetCore/netcoreapp30`
   `netcoreapp31` expects `Supportability/Dotnet/NetCore/netcoreapp31`
   `net50` expects `Supportability/Dotnet/NetCore/net5`

### DIFF
--- a/src/NewRelic.Core/DotnetVersion.cs
+++ b/src/NewRelic.Core/DotnetVersion.cs
@@ -26,6 +26,7 @@ namespace NewRelic.Core
     {
         LessThan30,
         netcoreapp30,
+        netcoreapp31,
         net5,
         Other,
     }
@@ -86,6 +87,11 @@ namespace NewRelic.Core
             if (envVer.Major == 3 && envVer.Minor == 0)
             {
                 return DotnetCoreVersion.netcoreapp30;
+            }
+
+            if (envVer.Major == 3 && envVer.Minor == 1)
+            {
+                return DotnetCoreVersion.netcoreapp31;
             }
 
             if (envVer.Major == 5)
--- a/src/NewRelic.Core/DotnetVersion.cs
+++ b/src/NewRelic.Core/DotnetVersion.cs
@@ -26,7 +26,8 @@ namespace NewRelic.Core
     {
         LessThan30,
         netcoreapp30,
-        GreaterThan30
+        net5,
+        Other,
     }
 
     public static class DotnetVersion
@@ -87,12 +88,17 @@ namespace NewRelic.Core
                 return DotnetCoreVersion.netcoreapp30;
             }
 
+            if (envVer.Major == 5)
+            {
+                return DotnetCoreVersion.net5;
+            }
+
             if (envVer.Major == 4)
             {
                 return DotnetCoreVersion.LessThan30;
             }
 
-            return DotnetCoreVersion.GreaterThan30;
+            return DotnetCoreVersion.Other;
         }
 #endif
     }


src/NewRelic.Core/DotnetVersion.cs
```diff
@@ -26,7 +26,8 @@ public enum DotnetCoreVersion
     {
         LessThan30,
         netcoreapp30,
-        GreaterThan30
```
nr-ahemsath, line 29: Is there any reason why we don't have a `netcoreapp31` version?

vuqtran88: I don't have the historical context why we didn't have that in the past, but it is good that you pointed this out because, net31 would also fall into the `GreaterThan30` category in the current logic together with `net5`. So to keep thing more consistent from now on, I see 2 ways to address this:

1. Still keeping `GreaterThan30` metrics and assume it will represent `netcoreapp31`. This is simplest.
2. Removing `GreaterThan30` metric and creating `netcoreapp31` metric, this would mean `GreaterThan30` and `netcoreapp31` will now represent `netcoreapp31`.

vuqtran88: I'm kinda leaning toward option 1 to keep thing simple.

lspangler584: Using `GreaterThan30` to represent `netcoreapp31` creates a question unless you know the Agent implementation and that it really means 'greater than 3.0 but not 5.0'. Granted, customers don't (currently) have access to Supportability Metrics, but for other (internal) users (Support, FIT, etc.) `GreaterThan30` doesn't tell the full story. I vote for 2.

lspangler584:
Unit tests NewRelic.Core.Tests target .NET Core. That would be a good place to add a unit test (even though it would only test one case). Probably not worth the effort to add Integration Tests since it is a Supportability Metric (not currently visible to customers), but that would be a way to test multiple paths of the method.
Can you add some description about how you manually tested?

src/NewRelic.Core/DotnetVersion.cs
vuqtran88: Discussed with @nr-ahemsath. Decided to go with option 2 as it would reduce the amount the customer has to think about things.

vuqtran88:
> Even though Unit tests NewRelic.Core.Tests target .netcoreapp2.1, the tests run in .net fw (maybe because the test runner is .net fw?). It'd be great if we could figure out a way to run unit tests on .net core too, but not a job for this PR.
> Probably not worth the effort to add Integration Tests since it is a Supportability Metric (not currently visible to customers), but that would be a way to test multiple paths of the method.
> Can you add some description about how you manually tested?
@lspangler584 
The NewRelic.Core.Tests actually targets both .netcoreapp2.1 and net451 because it has a mixture or netcore tests and fw tests (ie. TypeInstantiator and AppDomainExtensions). So tests in this project are ran by both .net core and .net fw runners.

Regarding to the test for supportability metrics, I am leaning toward only manually testing is good enough as the risk is very low. I can add more info on how I manually tested it.

nr-ahemsath (approved):
LGTM 👍

PR #461, opened 2021-02-12
Remove whitespace from instrumentation parameters

### LINKED ISSUE #169: Profiler should be able to match method parameters from XML that contain a space. Trim it !
**Description**
The following XML won't match ONLY because there is a space after the comma in the method parameters.

We should trim the whitespace to be more forgiving to ourselves and, more importantly, our clients doing custom instrumentation.
```
<match assemblyName="RestSharp" className="RestSharp.RestClient">
<exactMethodMatcher methodName="ExecuteTaskAsync" parameters="RestSharp.IRestRequest,System.Threading.CancellationToken" />
</match>
```
**Expected Behavior**
Ignore whitespace for method parameters.

**Steps to Reproduce**

**Your Environment**

**Additional context**
Migrated from the original issue.

nr-ahemsath:
### Description

Resolves #169 by removing whitespace characters from the parameter list in a custom instrumentation XML method matcher.

### Testing

Added unit test for matching multiple parameters with a space between the comma and the second parameter type.

Manually tested custom instrumentation matching a method with multiple parameters, with spaces between the parameters in the list, like this:

Method in code:
`static void doSomething(int aNumber, bool aBool, string aString)`
Method matcher in CI XML:
`<exactMethodMatcher methodName="doSomething" parameters="System.Int32, System.Boolean, System.String" />`

Tested on both Windows and Linux.  With the current released profiler, this CI would not match and this method would not be instrumented.  With the profiler built from this PR branch, the CI matches the method and the method is instrumented.

### Changelog

Updated changelog.


src/Agent/NewRelic/Profiler/ConfigurationTest/InstrumentationConfigurationTest.cpp
```diff
@@ -658,13 +658,42 @@ namespace NewRelic { namespace Profiler { namespace Configuration { namespace Te
             auto function = std::make_shared<MethodRewriter::Test::MockFunction>();
             BYTEVECTOR(signatureBytes,
                 0x00, // default calling convention
-                0x00, // 1 parameter
+                0x00, // 0 parameters
```
vuqtran88, line 661: Should this be "1" ?

src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -394,6 +388,25 @@ namespace NewRelic { namespace Profiler { namespace Configuration
...
 
+        static std::unique_ptr<xstring_t> NormalizeParameters(std::unique_ptr<xstring_t> rawParams)
+        {
+            if (rawParams == nullptr)
+            {
+                return nullptr;
+            }
+
+            // void as the parameters means parameterless method call (not all overloads)
+            if (Strings::AreEqualCaseInsensitive(*(rawParams), _X("void")))
```
vuqtran88, line 399: This is not related to this PR but rather interesting to me. I don't know if `void` parameter is a thing in C#. It is a thing in C++ or C but we are instrumenting C# functions.

vuqtran88: I actually found my answer. Whatever method signatures we declare in the instrumetation xml are IL method signatures as opposed to C# or languages. So I believe `void` parameter is a thing in IL world.

lspangler584:
We should add the newly built NewRelic.Profiler.dll to _profilerBuild folders.

lspangler584:
Ran unit tests, Profiler tests, IntegrationTests. Looks good.

lspangler584 (approved):

PR #466, opened 2021-02-22
Do not cache property setter actions from VisibilityBypasser

### LINKED ISSUE #464: Distributed Tracing over RabbitMQ does not work with RabbitMQ.Client version 6.x
When using the .NET Agent to instrument an application that is sending messages over RabbitMQ using RabbitMQ.Client version 6.x, distributed tracing does not work as expected.  More specifically, only the first message published to the queue will have the correct tracing headers added to the message.  All subsequent messages published by the sending application will not have any tracing headers added to the message, with the following caveat: if the user's application adds their own headers to the `BasicProperties` object, then things work as expected.

The reason has to do with this code in the RabbitMQ instrumentation: https://github.com/newrelic/newrelic-dotnet-agent/blob/65bbe645d26f545dfd6e222d63dc0120f6abb2bd/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/RabbitMqHelper.cs#L30-L36

We're using the VisibilityBypasser to generate a property setting `Action` for the `Headers` property on the `BasicProperties` object.  We need to do this for RabbitMQ.Client 6.x versions because the visibility to set this property changed from public to private.  Note that this action is being cached so we only need to create it once.  However, because the mechanism that generates the action takes a specific `BasicProperties` object instance, the cached action can only work for that specific object, which explains why things only work the first time (unless the application code adds its own headers, in which case this code is never used).

The likely solution will be to generate a new property setting action every time it is needed; I already have a test branch where I've done this and proven it fixes the problem.  The fix PR for this should also add unit tests for the property setter generating method in the VisibilityBypasser and also add an integration test for RabbitMQ DT that makes multiple calls.

nr-ahemsath:
### Description

Fixes #464 by no longer caching the Action returned from the `VisibilityBypasser` that allows the RabbitMQ `BasicProperties.Headers` dictionary to be set by our instrumentation wrapper. 

### Testing

RabbitMQ DT integration tests updated to make multiple pub/sub calls across the queue to verify it works the second time.

Added unit tests for the `VisibilityBypasser.Instance.GeneratePropertySetter` method.

### Changelog

Changelog updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ If the issue has been confirmed as a bug or is a Feature request, please file a 
 **Support Channels**
 
 * [New Relic Documentation](https://docs.newrelic.com/docs/agents/net-agent): Comprehensive guidance for using our agent
-* [New Relic Community](https://discuss.newrelic.com/tags/c/full-stack-observability/agents/.netagent): The best place to engage in troubleshooting questions
+* [New Relic Community](https://discuss.newrelic.com/tags/c/full-stack-observability/agents/466/dotnetagent): The best place to engage in troubleshooting questions
 * [New Relic Developer](https://developer.newrelic.com/): Resources for building a custom observability applications
 * [New Relic University](https://learn.newrelic.com/): A range of online training for New Relic users of every level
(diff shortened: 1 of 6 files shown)


jaffinito (approved):
Changes look good and tests pass!

PR #474, opened 2021-02-26
Create AWS Lambda release build and test workflow

### LINKED ISSUE #469: Create release build workflow for Lambda
### Feature Description
Workflow should use the existing all_solutions.yml as an example as much as possible!

- Has a conditional that prevents it from running during an agent release
- Builds the lambda projects
- Signs the assemblies
- Tests the  lambda projects
- Packages up the assemblies into a nuget
- Stores the nuget an artifact

Conditional: `if: github.event_name == 'release' && startsWith( github.ref, 'refs/tags/AwsLambdaOpenTracer_v' )`

### Additional context
https://github.com/newrelic/newrelic-dotnet-agent/issues/467

jaffinito:
Resolves #469 

### Description

Creates a new workflow for building and testing AWS Lambda releases
- Only runs on a release event with a tag that starts with `refs/tags/AwsLambdaOpenTracer_v` OR manually
- Only builds the Lambda Wrapper, Tracer, and NewRelic.Core
- Runs the Lambda Unit tests and save results
- Runs ArtifactBuilder for the Tracer only and saves results
- All steps are pulled from all_solutions,yml with limited changes to keep things simple

Creates new deploy workflow for AWS Lambda releases
- Use the same pattern as the agent deploy: copy artifacts>same them locally>deploy
- Deploys to nuget only
- Only inputs are run_id and deploy switch

### Testing

Create two fake releases, one with the proper tag and one without it.   It should only fully run the correct release. Note: do not use a tag that starts with `refs/tags/v` since that will start the agent release build.

Note: Cannot test deploy fully until we do a release for lambda.

### DIFF
--- a/.github/workflows/awslambda_release.yml
+++ b/.github/workflows/awslambda_release.yml
@@ -1,0 +1,129 @@
+name: .NET New Relic Lambda Tracer Release Build
+
+on:
+  release:
+    types: [ published ]
+  workflow_dispatch:
+
+env:
+  scripts_path: ${{ github.workspace }}\build\scripts
+  tools_path: ${{ github.workspace }}\build\Tools
+  DOTNET_NOLOGO: true
+
+jobs:
+
+  cancel-previous-workflow-runs:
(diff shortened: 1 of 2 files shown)


vuqtran88 (approved):
LGTM

PR #480, opened 2021-03-04
Removes nuget check from AWS Lambda

jaffinito:
Agent deploy has an extra conditional for deploying that lambda doesn't need.


vuqtran88 (approved):
LGTM

PR #481, opened 2021-03-05
Lambda release fixes

angelatan2:
### Description
Remove always if the condition for Win and Linux profiler
Added path to tracer test project and wrapper test project so that actions can find the unit tests


jaffinito (approved):
Paired on this with Angela.

PR #482, opened 2021-03-05
Fixed artifact builder paths to build properties

angelatan2:
Fixed artifact builder paths to build properties.


jaffinito (approved):
Paired with Angela on this.

PR #483, opened 2021-03-05
Matched the deploy upload path to the download path in row 30

angelatan2:
Matched the deploy upload path to the download path in row 30 so that the script can find the artifacts.


vuqtran88 (approved):
LGTM!

PR #486, opened 2021-03-05
Fix site extension release script

vuqtran88:
Fix site extension release script. Need to build the NewRelic.NuGetHelper for the New Relic Azure Site Extension.


jaffinito (approved):
Changes look good.

PR #487, opened 2021-03-05
Fixes generateBuildProperties.ps1 to work with lightweighted tags

vuqtran88:
### Description

There is problem with the generateBuildProperties.ps1 script to generate release versions for all of our packages. In that script, we use this command git describe --match "$tagPrefix[0-9]*" --abbrev=0 HEAD to pull up recently created tags that match whatever prefixes, however, that command only works with annotated tags. The tags created with Github releases are “lightweighted” tags. This PR fixes it.


jaffinito (approved):
Great catch!

PR #492, opened 2021-03-11
Profiler handling of utf8 with bom

### LINKED ISSUE #267: config files (newrelic.config, app.config, etc.) UTF-8 encoded with Byte Order Mark can't be parsed on Linux
**Description**
The profiler component of the agent has a problem parsing `newrelic.config` or other config files (e.g `app.config` for a profiled application)  if it is encoded in UTF-8 with BOM (Byte Order Mark: https://en.wikipedia.org/wiki/Byte_order_mark)

The symptom of the issue is that a profiler log will only have the following lines:
```
[Info ] 2020-02-19 11:33:36 Logger initialized.
[Trace] 2020-02-19 11:33:36 New Relic .NET CoreCLR Agent
[Info ] 2020-02-19 11:33:36 Found newrelic.config at: /usr/local/newrelic-netcore20-agent/newrelic.config
```

**Expected Behavior**
The agent should start normally if the config file is in UTF-8 with BOM.

**Steps to Reproduce**
1. Get a `newrelic.config` file encoded in UTF-8 with BOM.  You can do this easily with VS Code.  Open the config file in VS Code, click the encoding in the lower right hand corner, choose "Save with encoding" from the dialog that opens, and select "UTF-8 with BOM" to save the file as.  You can confirm that this worked by installing the "Hexdump" extension and viewing the file as a hexdump and confirming that the first three bytes of the file are "EF BB BF".
2. Use this file as the global `newrelic.config` file in `/usr/local/newrelic-netcore20-agent` in a Linux environment, and run any .NET Core application with profiling enabled.  Look at the resulting profiler log and note that it stops after the first three lines.

Initial debugging with extra logging statements shows that the breakdown is occurring somewhere when trying to read the resultant `xstring_t` from the `ReadFile()` method in `ICorProfilerCallbackBase.h`.  `xstring_t` is a custom typedef we use to handle Windows/Linux cross-compatibility, it's defined in `xplat.h`.

For Windows:
```
typedef wchar_t xchar_t;
typedef std::basic_string<xchar_t> xstring_t;
```

For Linux:
```
typedef std::u16string::value_type xchar_t;
typedef std::u16string xstring_t;
```

**Your Environment**
.NET Core agent on Linux

nr-ahemsath:
### Description

Fixes #267 by reading in config files as binary bytes, then detecting the UTF-8 BOM and bypassing it if found, and then converting the read data to our `xstring_t` string type.

I also tweaked the log messaging when catching a `rapidxml::parse_error` exception so that an error message will always be logged safely. 

### Testing

Tested manually on Linux and Windows.  

### Changelog

CHANGELOG is updated.


src/Agent/NewRelic/Profiler/Profiler/ICorProfilerCallbackBase.h
```diff
@@ -797,22 +797,34 @@ namespace Profiler {
 // Ideally we need to be able to handle file paths with multibyte characters in them. Currently, we do not.
 #pragma warning(push)
 #pragma warning(disable : 4244)
-            // attempt to open the file
-            xifstream fileStream(std::string(filePath.begin(), filePath.end()), std::ios::in);
+
+            // Open file to read as a binary byte stream.  This lets us reliably detect and
+            // remove a UTF-8 BOM at the beginning of the file.
+            // See https://github.com/newrelic/newrelic-dotnet-agent/issues/267 for context
+            ifstream inFile(std::string(filePath.begin(), filePath.end()), std::ios::binary);
```
vuqtran88, line 804: Is there a reason why we are no longer using the `x` macro for some of the type here .ie `ifstream`, `ostringstream` ... . I'm not expert in this but, my impression was the `x` macro ensure the cross platform behavior according [this](https://github.com/newrelic/newrelic-dotnet-agent/blob/5b375b3c38517c8326aac0a70a4901f99f1a483c/src/Agent/NewRelic/Profiler/Common/xplat.h#L17)

vuqtran88: Discussed with @nr-ahemsath on this. The reason why this code didn't use `x` macro was UTF-8 BOM can only be detected in a binary byte (char) stream in Linux.

vuqtran88 (approved):
LGTM!

Issue #22, opened 2020-07-08
Exception when distributed tracing is enabled and using ASP.NET Core RequestLocalization middleware


alanwest:
## Description

Agent causes exception when distributed tracing is enabled and ASP.NET Core application uses the `RequestLocalization` middleware. Appears to affect ASP.NET Core apps running on Linux in a docker container.

## Steps to Reproduce

Create new Web API
```
mkdir locale
cd locale
dotnet new webapi
```

Add the RequestLocalization middleware in `Startup.cs`
```csharp
var supportedCultures = new[]
{
    new CultureInfo("en-US"),
    new CultureInfo("ar"),
};

app.UseRequestLocalization(new RequestLocalizationOptions
{
    DefaultRequestCulture = new RequestCulture("en-US"),
    SupportedCultures = supportedCultures
});
```

Add the following controller
```csharp
using System.Threading.Tasks;
using Microsoft.AspNetCore.Mvc;

namespace RequestLocalization.Controllers
{
    [ApiController]
    [Route("[controller]")]
    public class TestController : ControllerBase
    {
        [HttpGet]
        public async Task<string> Get()
        {
            using (var client = new System.Net.Http.HttpClient())
            {
                return await client.GetStringAsync("https://newrelic.com");
            }
        }
    }
}
```

Create a `Dockerfile` and replace `put_license_key_here`
```
FROM mcr.microsoft.com/dotnet/core/aspnet:3.1.0

RUN curl -O https://download.newrelic.com/dot_net_agent/latest_release/newrelic-netcore20-agent_8.29.0.0_amd64.deb
RUN dpkg -i ./newrelic-netcore20-agent*.deb

ENV CORECLR_NEWRELIC_HOME /usr/local/newrelic-netcore20-agent
ENV NEW_RELIC_LICENSE_KEY put_license_key_here

WORKDIR /app
COPY ./bin/Debug/netcoreapp3.1/publish .

EXPOSE 5000

ENTRYPOINT ["/bin/bash", "-c", "dotnet locale.dll"]
```

Create a `docker-compose.yml` file
```
version: "3"

services:
  locale:
    image: locale:latest
    ports:
      - 5000:5000
    environment:
      ASPNETCORE_URLS: http://*:5000
      ASPNETCORE_ENVIRONMENT: Development
      CORECLR_ENABLE_PROFILING: 1
[...]

PR #493, opened 2021-03-11
Language independently convert float number value to string for tracestate Priority value

vuqtran88:

### Description
- This PR is a fix for the issue [#22](https://github.com/newrelic/newrelic-dotnet-agent/issues/22).
- The call `PriorityFormat, transaction.Priority` in the `BuildTracestate` function, under “Arabic” culture, yields two different string interpretations of the same float number in 2 different platforms (Window vs Unix). In windows, the decimal point is converted into an ascii character (.colon). In contrast, in Linux, the decimal point is converted into an unicode character (٫ Arabic decimal separator).
This difference is the real main cause of this exception `System.Net.Http.HttpRequestException: Request headers must contain only ASCII characters when DT was enabled`. The agent basically added a non ascii character in the outgoing payload when Arabic culture is set in Linux. 

### Testing
- Manually tested in Linux.
- Some unit tests are added, but they indirectly provide coverage for Linux. The issue only occurs in Linux and we currently don't have Linux test coverage.

### DIFF
--- a/tests/Agent/UnitTests/Core.UnitTest/Wrapper/AgentWrapperApi/DistributedTracing/DistributedTracePayloadHandlerTests.cs
+++ b/tests/Agent/UnitTests/Core.UnitTest/Wrapper/AgentWrapperApi/DistributedTracing/DistributedTracePayloadHandlerTests.cs
@@ -764,7 +764,7 @@ namespace NewRelic.Agent.Core.Wrapper.AgentWrapperApi.DistributedTracing
             Mock.Assert(() => _agentHealthReporter.ReportSupportabilityTraceContextCreateSuccess(), Occurs.Once());
         }
 
-        //This test makes sure string presentation for Priority is culture independent.
+        //This test makes sure string presentation for Priority in a tracestate is culture independent.
         [TestCase(1.123000f, "1.123")]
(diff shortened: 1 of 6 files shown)


tests/Agent/UnitTests/Core.UnitTest/Wrapper/AgentWrapperApi/DistributedTracing/DistributedTracePayloadHandlerTests.cs
```diff
@@ -762,6 +764,53 @@ public void W3C_InsertDistributedTraceHeaders_ExcludeNewRelicHeader()
             Mock.Assert(() => _agentHealthReporter.ReportSupportabilityTraceContextCreateSuccess(), Occurs.Once());
         }
 
+        //This test makes sure string presentation for Priority in a tracestate is culture independent.
+        [TestCase(1.123000f, "1.123")]
+        public void W3C_InsertDistributedTraceHeaders_CultureIndependent_PriorityInRightFormat(float testPriority, string expectedPriorityString)
```
vuqtran88, line 769: By making sure string presentation for Priority in a tracestate is culture independent, we will never get non ascii character for the string presentation.

nr-ahemsath (approved):
LGTM.  Unit and integration tests pass for me locally.  🚢

PR #501, opened 2021-03-25
Updated TransactionAttributeMaker to always include an error attribute

### LINKED ISSUE #500: Agent should always create the "error" intrinsics attribute
**Description**
In the event of error occurs in a transaction, the agent create the "error" intrinsics attribute and set its value to "true" in it for transaction event. However, It doesn't create this attribute for "normal" transactions.   

**Expected Behavior**
Agent creates the "error" intrinsics attribute for transaction events with/without errors and set the value appropriately.

**Your Environment**
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of the agent and any relevant configurations, such as the .NET framework or core version of your application. )

**Additional context**
Ref code: https://github.com/newrelic/newrelic-dotnet-agent/blob/9d6ae8fad78a2319c0b68b7932495dc1d6ac02e6/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionAttributeMaker.cs#L99

tehbio:
### Description

* Fixes #500
* As per Agent specs, the error attribute is still omitted when the Error Collector is disabled.

### Testing

Both a new unit test was added and an integration test was updated to validate this.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### New Features
 ### Fixes
+* Fixes issue [#500](https://github.com/newrelic/newrelic-dotnet-agent/issues/500): Agent should always create the "error" intrinsics attribute. ([#501](https://github.com/newrelic/newrelic-dotnet-agent/pull/501))
 
 ## [8.39.1] - 2021-03-17
 ### Fixes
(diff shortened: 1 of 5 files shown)


src/Agent/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### New Features
 ### Fixes
+* Fixes issue [#500](https://github.com/newrelic/newrelic-dotnet-agent/issues/500): Agent should always create the "error" intrinsics attribute. ([#501](https://github.com/newrelic/newrelic-dotnet-agent/pull/501))
```
vuqtran88, line 10: It might be worth be a bit clearer on this. How's about ?

"For transactions without errors, Agent still create the `error` intrinsics attribute with its value set to `false`."

vuqtran88 (changes requested):
Everything looks pretty good. I would like the release note message to be a bit clearer.

vuqtran88 (approved):
LGTM!

Issue #504, opened 2021-03-29
Move instrumenation in (MongoDb26) AsyncCursorWrapper.cs to XML


jaffinito:
### Feature Description
Move instrumenation in MongoCollectionDefaultWrapper.cs to XML

Move MethodSignatures to instrumentation.xml.  Replace method.MatchesAny(...) with methodInfo.RequestedWrapperName(...) as in ExecuteStepWrapper.cs (as an example).

### Additional context
Enables future Multiverse work.

Issue #503, opened 2021-03-29
Move instrumenation in MongoCollectionDefaultWrapper.cs to XML


jaffinito:
### Feature Description
Move instrumenation in MongoCollectionDefaultWrapper.cs to XML

MethodSignatures are duplicated in part compared to instrumentation.xml. Replace method.MatchesAny(...) with methodInfo.RequestedWrapperName(...) as in ExecuteStepWrapper.cs (as an example).

### Additional context
Enables future Multiverse work.

PR #509, opened 2021-03-31
Remove hardcoded instrumentation points in MongoDb wrappers.

vuqtran88:
Fix for issue #503 and issue #504 .
Integration tests passed.


jaffinito (approved):
Looks great!  The wrapper files are so much easier to read now.  I wonder if we will see a tiny increase in performance with the simplification?

tehbio (approved):

Issue #506, opened 2021-03-29
Move instrumenation in (RabbitMq) BasicPublishWrapperLegacy.cs to XML


jaffinito:
### Feature Description
Move instrumenation in (RabbitMq) BasicPublishWrapperLegacy.cs to XML

Move MethodSignatures to instrumentation.xml. Replace method.MatchesAny(...) with methodInfo.RequestedWrapperName(...) as in ExecuteStepWrapper.cs (as an example).

### Additional context
Enables future Multiverse work.

PR #510, opened 2021-04-02
Removes hard coded instrumentation points in rabbit wrappers.

vuqtran88:
Fixes for #505 and #506 issues.


jaffinito (approved):
Awesome work!  So much easier to read now.

PR #523, opened 2021-04-07
Update StackTraces.ScrubAndTruncate to handle zero max depth

### LINKED ISSUE #522: maxStackTraceLines="0" is not honored
The `maxStackTraceLines` configuration value (https://docs.newrelic.com/docs/agents/net-agent/configuration/net-agent-configuration/#config-maxStackTraceLines) is not honored if the value is set to "0".

Expected behavior: no stack trace lines will be sent.
Actual behavior: all available stack trace lines are sent.

Our support team brought this to our attention; a customer wants to configure the agent to keep sending error trace data to New Relic, but without any stack trace information included.  Upon investigation, we found that setting `maxStackTraceLines` to 0 had no effect on the number of stack trace lines sent due to a flaw in the logic for handling this config value.  This will be fixed in a PR shortly.

nr-ahemsath:
### Description

This fixes #522 by adjusting the logic of the `StackTraces.ScrubAndTruncate` method to return an empty list when the `maxDepth` parameter is set to 0.

### Testing

Added one unit test to cover this case.  I also manually tested a test app that threw an exception from a stack depth of 3 and verified that no stack trace lines are sent in the `error` data when `maxStackTraceLines` is set to 0 in the agent config file.

### Changelog

Changelog is updated.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### New Features
 ### Fixes
 * Fixes issue [#500](https://github.com/newrelic/newrelic-dotnet-agent/issues/500): For transactions without errors, Agent should still create the `error` intrinsics attribute with its value set to `false`. ([#501](https://github.com/newrelic/newrelic-dotnet-agent/pull/501))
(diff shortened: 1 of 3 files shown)


tehbio (approved):

nr-ahemsath:
Converted to draft while waiting for answers to some questions.  There is a question of whether a config value of "0" should actually mean "0", or if it should mean "infinity".  I checked the agent specs for error traces and I didn't find an answer either way.  I'm working on getting input from other agent teams to see what they do.

nr-ahemsath:
No conclusive feedback either way; all our integration tests pass for me locally with this change, and it also seems unlikely that the backend would like the agent to send them an "infinite" amount of stack frames in any case.  Going ahead with this change.

PR #524, opened 2021-04-07
Adds support for env vars in config.yml

jaffinito:
### Description

Using other YAML solutions as an example (docker, GHA) added ability to add "tokens" to the config.
A token starts with `${{` and ends with `}}`.  Within this wrapper you can add an env var and the tool will pick it up and replace the token with env vars value.

If the env var doesn't exist or the value is empty/whitespace it will throw an error.  An error will also be thrown if the token is not formatted properly.

### Testing

Everything should be build, but we can't test till we get more inplace.  Worked locally.


vuqtran88 (approved):
LGTM

Issue #264, opened 2020-09-30
GCSampler - CountData calculated as negative causes exception


nr-ahemsath:
**Description**

We are unsure of the reason, but the GC Sampler is trying to create a count metric that is a negative value.  As a result, the GC Sampler shuts down (because too many exceptions have been thrown).

Here is where the exception is being thrown from: https://github.com/newrelic/newrelic-dotnet-agent/blob/707a2269b63bd76337597cb439829fed273f1f21/src/Agent/NewRelic/Agent/Core/WireModels/MetricDataWireModel.cs#L117

**Expected Behavior**

Since many of the count values are delta values from the prior value, intercept a negative value and record it as 0 in the event.
When doing so, add a log message (finest) indicating the Metric and requested value that was switched to 0.

**Steps to Reproduce**

Unknown at this point.

JoshuaColeman:
Hey @angelatan2! I have a potential solution for this in [https://github.com/newrelic/newrelic-dotnet-agent/pull/525](url)

PR #525, opened 2021-04-11
Clamps negative count metrics to 0 to prevent GCSampler exceptions

JoshuaColeman:
### Description
Well hello!

I reviewed the bugs you had committed to in the next milestone and #264 seemed like one I could figure it out without much help. 

Essentially: Sometimes random negative counts come back from the GCSampler, and there is a desire to make this code more robust and correct any negative values to 0 so that the GCSampler isn't shut down for throwing too many exceptions.

The description of the ticket makes it sound like a 3 line change, but the requirement to log the name of the metric that is being corrected complicates things since the code that was throwing exceptions is not aware of the metric that it is validating. I tried to satisfy the requirement without too much copy/paste.

### Testing

I added unit tests for the core code that I modified which detects and corrects negative count metrics. I added identity tests to make sure positive values are not being transformed either.

There should probably be at-least one integration test that supplies a negative count value, but I have not cracked that code open and it's already late. If you can point me to a good spot to add this then I would totally add it :)

### Changelog

Updated with everything but this PR#


nr-ahemsath:
Hey, @JoshuaColeman, thanks for taking a look at this and submitting a PR.  Somebody from the team should be able to review this PR and work with you to address any issues we find sometime soon.

JcolemanNR:
Thanks for the feedback @tehbio! 

@nr-ahemsath and I reviewed this issue/PR today and found that there is a GC specific layer that we can implement this validation in. I had implemented this fix at a low level to prevent any exceptions from being thrown, but it doesn't make sense to completely disable exceptions for negative counts when we know the issue is in the GC instrumentation. Since the code and unit tests will be completely different I will cancel this PR.

JcolemanNR:
Closing to create a PR with a different solution

PR #526, opened 2021-04-12
Refactors the Multi Version Scanner

vuqtran88:
### Description
Refactoring the Multi Version Scanner to do the followings:
- One AssemblyAnalyzer instance is responsible to validate one instrumented assembly per target framework at a time.
- The Muti Version Scanner will validate the instrumented assemblies for all target frameworks that the validated nuget packages support.


jaffinito (approved):
Awesome work Vu!  This runs a whole lot better now that it handles assemblies better.

Issue #516, opened 2021-04-05
Update nuget downloader to get latest package version along with the already specified versions


jaffinito:
### Feature Description
Currently nuget packages are downloaded based on a specific list in the config file.  We need this to also pull down the latest version of package automatically.  This will require some work to determine what that is.

Acceptance Criteria
- Determine what the latest version of the package is
- Download the latest version
- Use the latest version in the test matrix
- Should not update the config with the latest verison.

PR #527, opened 2021-04-14
Mvs replace nuget

jaffinito:
Resolves: #516 

### Description
Uses NuGet.Protocol and NuGet.Packaging
Downloads packages in memory and extracts files, no zip
Finds latest non-prerelease version of package
Adds latests version to versions list for download and checking
Version in NugetSet change from sting[] to List<string> to make it easy to append to

### Testing

Should build fine, cannot test since it is not run.

### Changelog
N/A


vuqtran88 (approved):
LGTM

PR #535, opened 2021-04-16
Changed GHA All Solutions Build to trigger on Release creation instead of publish

tehbio:
### Description

This is to support the new release process not having to immediately publish a release before all the necessary steps are actually completed. Publishing a release will be moved to the final step instead of first, so the workflow will need to trigger on creation.

### Testing

Testing will be completed/verified after this merge is completed.


nr-ahemsath (approved):
Seems legit; we can easily revert if testing proves otherwise.

PR #537, opened 2021-04-19
Revert "Changed GHA All Solutions Build to trigger on Release creation instead of publish"

tehbio:
Reverts newrelic/newrelic-dotnet-agent#535 as it proved unworkable due to GHA not being able to trigger on Release drafts.


nr-ahemsath (approved):
🔙 👍

PR #539, opened 2021-04-19
Fixes wierd powershell behaviour

jaffinito:
Powershell appears to require some extra bits.
Added a "dir" check as well to make sure folder looks right.


nr-ahemsath (approved):
Yeah, need to explicitly provide the `.\` to run a command in the current directory.

PR #546, opened 2021-05-01
Update debian container images to latest stable and fix bugs

nr-ahemsath:
### Description

This PR:
1. Updates two `debian:jessie` container images to `debian:stable`.  These containers are used by the build and deploy processes for the agent's Linux packages (.deb and .rpm).
2. Removes a stray debugging early exit from `build.ps1` that got committed accidentally.
3. Fixes a bug in the `build.sh` scripts for both the .deb and .rpm builds, introduced when we restructured the codebase.  What broke was the scripts' ability to automatically determine the agent version.  (FYI, when these scripts are used in our actual CI build process, the agent version is explicitly passed in to the container, so this feature is just for developers' convenience when doing local package builds.)

### Testing

I tested building the containers and running them to build the .rpm and .deb packages after my changes and it worked as expected.

Note: the container image for the rpm/deb deploy process was also updated from `jessie` to `stable`.  I tested building the container and it built as expected.  Since we have no "dry run" option for the deploy process inside the container, I can't test this change all the way to the end.

### Changelog
N/A

### DIFF
--- a/build/Linux/build/deb/Dockerfile
+++ b/build/Linux/build/deb/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:jessie
+FROM debian:stable
 
 RUN apt-get update
-RUN apt-get install --assume-yes dos2unix
+RUN apt-get install --assume-yes dos2unix exiftool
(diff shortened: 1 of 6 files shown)


tehbio (approved):
👍

PR #550, opened 2021-05-07
Clamp negative GC counts to 0

### LINKED ISSUE #264: GCSampler - CountData calculated as negative causes exception

JcolemanNR:
### Description

This PR fixes #264 . Essentially, we occasionally encounter negative GC counts which cause an exception to be thrown. This exception can cause the GC sampler to be shut down after multiple errors are encountered. In order to prevent this from happening, we want to clamp all negative count metrics coming from the GC.

The fix is implemented at the GC layer, and will work for both .NET framework, and .NET standard.

### Testing

A new unit test was added that verifies that negative GC counts are clamped to 0. As we do not have a repro path, an integration test was not written.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### New Features
 ### Fixes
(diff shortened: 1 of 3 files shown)


nr-ahemsath (approved):
Looks good to me.  Ship it! 🚢

PR #554, opened 2021-05-10
Ignore file that keeps showing up in git after I build.

JcolemanNR:
### Description

The following file keeps showing up in git after I have built. I believe this should be ignored:
src/Agent/NewRelic/Profiler/out/build/x64-Debug/VSInheritEnvironments.txt

### Testing

No tests required (.gitignore change)


nr-ahemsath (approved):
LGTM 👍

PR #555, opened 2021-05-14
Reduce integration test solution build warnings

nr-ahemsath:
### Description

This PR reduces the number of build/restore warnings seen in the integration test and unbounded integration test solutions.  There were more than 70 warnings in the integration test solution before this work and the number is now down to 6 (all caused by the same issue which may be addressed in a separate PR, but I believe it will require a slightly more significant change).

Part of the motivation for doing this now is that as we begin to test .NET 6 preview versions, we want to make sure that any new warnings don't get lost in the noise.  However, this should help anybody either internally or externally to notice new warnings that may appear.

The following things were done to reduce the number of warnings:

* ServiceStackApplication was modified to reference 5.5.0 versions of ServiceStack nuget packages, since the version we were previously referencing (5.0.2) was no longer available.  This change is benign since the tests have been using 5.5.0 anyway.
* Suppressed warnings about out-of-date .NET Core target versions (netcoreapp2.2 specifically) - some of our tests need to specifically test these EOLed .NET Core versions
* In the Agent API tests, suppressed warnings about some of the APIs being deprecated - as long as we're still shipping these APIs we need to test them
* Suppressed warnings about `async` methods without `await` operators: we have tests that are testing this specific scenario on purpose.
* In MultiFunctionApplicationHelpers, ignore warnings about NuGet packages not being fully compatible with a given runtime version.  The way this project works, it is multi-targeted to several .NET Framework and .NET Core versions to allow testing multiple runtimes from the same test code.  However, not all targets are used in every test.  The warnings were for package references incompatible with `netcoreapp2.1`, but those packages are not used by any tests that run on `netcoreapp2.1`
* Updated binding redirects for `NewtonSoft.Json` to account for the many versions of this package used throughout the tests solutions.

### Testing

I made sure all of the integration tests and unbounded integration tests passed after making these changes.

### Changelog

N/A

### DIFF
--- a/tests/Agent/IntegrationTests/Applications/AspNetCore3BasicWebApiApplication/AspNetCore3BasicWebApiApplication.csproj
+++ b/tests/Agent/IntegrationTests/Applications/AspNetCore3BasicWebApiApplication/AspNetCore3BasicWebApiApplication.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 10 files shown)


JcolemanNR (approved):
looks good!

PR #569, opened 2021-05-18
High security mode will disable request header capturing

vuqtran88:
### Description:
This PR updates the followings:
1. High security mode will disable request header capturing. This is for spec compliance.
2. When allowAllHeaders config is set to "true", automatically add "request.header.*" to the inclusion list even when users don't explicitly specify it in the attribute element. This is doing similar to the `captureParameters` configuration.

### Testing
Unit tests.


jaffinito (approved):
Looks good to me

PR #572, opened 2021-05-19
Change SetRequestHeaders method signature.

vuqtran88:
### Description

Changes `SetRequestHeaders` method signature to generically handle any header collection implementation.

### Testing
No tests as this is not being used anywhere yet.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
+++ b/src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
@@ -1200,7 +1200,7 @@ namespace NewRelic.Agent.Core.Transactions
             foreach (var key in keysToCapture)
             {
                 var value = getter(headers, key);
-                if (!string.IsNullOrEmpty(value))
+                if (value != null)
                 {
                     var paramAttribute = _attribDefs.GetRequestHeadersAttribute(key);
                     TransactionMetadata.UserAndRequestAttributes.TrySetValue(paramAttribute, value);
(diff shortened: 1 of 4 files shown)


JcolemanNR:
Suggestion: Unit test that calls the newly created API. It looks like the existing unit tests just check the configuration service?

I would expect a change to a public API to have associated unit tests. The unit tests should call the API with multiple different collection types IMHO.

vuqtran88:
> Suggestion: Unit test that calls the newly created API. It looks like the existing unit tests just check the configuration service?
> 
> I would expect a change to a public API to have associated unit tests. The unit tests should call the API with multiple different collection types IMHO.

Good suggestion. My plan was to have integration tests to cover the usage of this new api when there are actual usages (soon in next PR). We currently have unit tests that cover the new configuration and the request header attribute creation. I will see if i can create unit tests that use this new api.

About public api. This new api is actually not public yet, it is still internal. Let me know if u think this api method is useful for customer to use so that we can plan to expose it.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -1180,19 +1180,30 @@ public void InsertDistributedTraceHeaders<T>(T carrier, Action<T, string, string
...
+
+            if (getter == null)
+            {
+                throw new ArgumentNullException(nameof(getter));
+            }
+
+            foreach (var key in keysToCapture)
+            {
+                var value = getter(headers, key);
+                if (!string.IsNullOrEmpty(value))
```
tehbio, line 1203: I think I recall seeing something in the spec about capturing empty parameter values, so we might want to only do a null check here.

vuqtran88: you are right. [spec](https://source.datanerd.us/agents/agent-specs/blob/32153bcabe8389ef03d0c7b18371bbccc72c7989/Agent-Attributes-PORTED.md#nullempty-attribute-values)

tehbio (approved):
Looks good. 👍

PR #585, opened 2021-05-27
Updates conditional to allow scheduled ruhns

jaffinito:
Allows scheduled runs to occur.


angelatan2 (approved):
Thanks for addressing this! Approved.

PR #587, opened 2021-05-27
Console logger honors configured log level

### LINKED ISSUE #584: Console log output threshold is hardcoded to WARN level
When setting `console="true"` in `newrelic.config` (either globally or app-locally), the expectation is that the same agent log messages that would be sent to the agent log file (usually in `$NEWRELIC_HOME/logs`) are sent to the console output of the instrumented application.  However, only WARN and higher-level agent log messages are being sent to the console.

It looks like the issue is that the `ConsoleLogAppender` is being set at start time with a hardcoded WARN-level threshold and not updated based on agent configuration: https://github.com/newrelic/newrelic-dotnet-agent/blob/7c45757033d1b24835c7fbec29df4ae2ee0998db/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs#L326

nr-ahemsath:
Fixes #584.  The console logger (if enabled by setting `console="true"`  in the `<log />` element of `newrelic.config`) will now output agent log messages at the same configured level as the agent log file.  This also fixes a related problem where the console logger was outputting all "audit" level messages that should only go to the audit log file.

This PR also contains changes to the profiler project file to fix a profiler build issue caused by the latest versions of MS build tools and libraries in GitHub Actions.  Thanks to @JcolemanNR for figuring out the fix, and also for encouraging us to implement this fix in a way that would allow the console logging level to be dynamically updated in the same way that the agent's file logging is.

Unit tests added.

Changelog updated.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
+++ b/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
@@ -64,7 +64,6 @@ namespace NewRelic.Agent.Core
         private static string AuditLogName = "Audit";
 
         // Watch out!  If you change the time format that the agent puts into its log files, other log parsers may fail.
-        // See, specifically, the orion IIS QA package, file lib/qa-tools-utils/dotnet_agent_log_parser.rb
         private static ILayout AuditLogLayout = new PatternLayout("%utcdate{yyyy-MM-dd HH:mm:ss,fff} NewRelic %level: %message\r\n");
         private static ILayout FileLogLayout = new PatternLayout("%utcdate{yyyy-MM-dd HH:mm:ss,fff} NewRelic %6level: [pid: %property{pid}, tid: %thread] %message\r\n");
 
@@ -85,7 +84,7 @@ namespace NewRelic.Agent.Core
             GlobalContext.Properties["pid"] = new ProcessStatic().GetCurrentProcess().Id;
 
             SetupStartupLogAppender(logger);
-            SetupConsoleLogAppender(logger);
+            SetupConsoleLogAppender(logger, Level.Warn);
             SetupTemporaryEventLogAppender(logger);
         }
 
@@ -115,7 +114,16 @@ namespace NewRelic.Agent.Core
             SetupAuditLogger(logger, config);
             SetupDebugLogAppender(logger);
             logger.RemoveAppender(TemporaryEventLogAppenderName);
-            if (!config.Console) logger.RemoveAppender(ConsoleLogAppenderName);
+
+            // The console logging was initially set up with a fixed log level of "WARN"
+            // Remove the existing console log appender, and if console logging is
+            // enabled by config, add it back with the configured logging level
+            logger.RemoveAppender(ConsoleLogAppenderName);
+            if (config.Console)
+            {
+                var consoleLogLevel = logger.Hierarchy.LevelMap[config.LogLevel];
+                SetupConsoleLogAppender(logger, consoleLogLevel);
+            }
 
             logger.Repository.Configured = true;
 
@@ -318,12 +326,12 @@ namespace NewRelic.Agent.Core
         /// </summary>
         /// <param name="logger">The logger you want to attach the console log appender to.</param>
         /// <param name="config">The configuration for the appender.</param>
-        private static void SetupConsoleLogAppender(log4netLogger logger)
+        private static void SetupConsoleLogAppender(log4netLogger logger, Level level)
         {
             var appender = new ConsoleAppender();
             appender.Name = ConsoleLogAppenderName;
             appender.Layout = FileLogLayout;
-            appender.Threshold = Level.Warn;
+            appender.Threshold = level;
             appender.AddFilter(GetNoAuditFilter());
             appender.ActivateOptions();
             logger.AddAppender(appender);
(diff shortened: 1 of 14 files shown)


src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
```diff
@@ -318,12 +326,12 @@ private static void SetupDebugLogAppender(log4netLogger logger)
         /// </summary>
         /// <param name="logger">The logger you want to attach the console log appender to.</param>
         /// <param name="config">The configuration for the appender.</param>
-        private static void SetupConsoleLogAppender(log4netLogger logger)
+        private static void SetupConsoleLogAppender(log4netLogger logger, Level level)
```
JcolemanNR, line 329: update documentation?

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
```diff
@@ -85,7 +84,7 @@ public static void Initialize()
             GlobalContext.Properties["pid"] = new ProcessStatic().GetCurrentProcess().Id;
 
             SetupStartupLogAppender(logger);
-            SetupConsoleLogAppender(logger);
+            SetupConsoleLogAppender(logger, Level.Warn);
```
JcolemanNR, line 87: After thinking about the use case (containerized process with no FS), I think if the ConsoleLog config is enabled then we should log at the specified config level from the start. If we only log warn events in this case, we could miss some startup/initialization messages of interest before the switch happens.

tests/Agent/UnitTests/Core.UnitTest/NewRelic.Agent.Core.FromLegacy/AgentLoggerTest.cs
```diff
@@ -110,5 +154,29 @@ static private ILogConfig LogConfigFixtureWithAuditLogEnabled(string logLevel)
             var configuration = ConfigurationLoader.InitializeFromXml(xml);
             return configuration.LogConfig;
         }
+
+        static private ILogConfig LogConfigFixtureWithConsoleLogEnabled(string logLevel)
+        {
+            var xml = string.Format(
+                "<configuration xmlns=\"urn:newrelic-config\">" +
+                "   <service licenseKey=\"dude\"/>" +
```
JcolemanNR, line 162: dude 💯

JcolemanNR (changes requested):
I think if console logging is enabled, we should log at that level from application start. The customer use case is to get information out of the application in a containerized environment with no FS access... we may want to consider allowing them to disable the FS logging as well?

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
```diff
@@ -64,7 +64,6 @@ public static class LoggerBootstrapper
         private static string AuditLogName = "Audit";
 
         // Watch out!  If you change the time format that the agent puts into its log files, other log parsers may fail.
-        // See, specifically, the orion IIS QA package, file lib/qa-tools-utils/dotnet_agent_log_parser.rb
```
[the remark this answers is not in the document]
nr-ahemsath: Yes, Orion hasn't been a thing since before I joined the team.

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
JcolemanNR: Oh, I see. This is a console logger that is setup before we even read the logging config.

JcolemanNR (approved):
Thanks for explaining how the logging worked in this scenario @nr-ahemsath! Looks good to me!

The only concern I have is that the way the level is set on the console appender may prevent the log level from updating dynamically? Asking for a friend.

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
nr-ahemsath: Good catch, done.

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
nr-ahemsath: I was going for minimal changes to existing agent behavior.  If the console logger is enabled in the agent config, the level will get updated to the configured level as soon as the config is parsed, which happens very early on in the startup process.

I'm assuming that whoever wanted the initial console logger threshold set to WARN had a good reason to do so. 🤷‍♂️

nr-ahemsath:
> The only concern I have is that the way the level is set on the console appender may prevent the log level from updating dynamically? Asking for a friend.

Yes, that is true.  If we want to support having the console logging update it's level dynamically like the main agent file logging, it would probably be simplest to let the console appender inherit its threshold from the top-level logger's level, since it looks like that's how the file logging level gets set and, I assume (dangerous I know!) updated dynamically.  That would end up making things simpler anyway.

JcolemanNR:
Looks good to me, and works on my box. I made a minor change so we don't need to delete/recreate the console logger which I think simplifies things. The console logger was trying to setup an audit log exclusion filter during initial startup that ended up not working since the audit level hadn't been created at that point

PR #594, opened 2021-06-07
Removes query string in Referer, Location and Refresh request header attributes.

vuqtran88:
### Description
This PR makes sure that the agent will trip query string from the Referer, Location and Refresh request headers when capturing them as attributes.

https://source.datanerd.us/agents/agent-specs/blob/master/Agent-Security.md#request--response-headers

### Testing
unit tests.


tehbio (approved):
👍

PR #595, opened 2021-06-08
Updated changelog for Header capturing

tehbio:
Updated changelog with better detail about what's supported with the http request header capturing.


vuqtran88 (approved):
LGTM

PR #608, opened 2021-06-15
Update Mvc3 instrumentation to better handle Controller and Action misses

### LINKED ISSUE #155: MVC Invalid Action for valid Controller can cause MGI

**Description**
If an invalid method (controller action) is specified in a route, but the controller is valid, the system records a transaction.   This can be problematic when there are phishing attempts and/or security audits of sites looking for vulnerabilities.

**Expected Behavior**
This is something that still needs to be decided. Should we just not apply the MVC controller-action naming, or should we provide some sort of "invalid" controller-action name that allows these types of requests to be rolled up. I do not think that these types of requests should be marked as ignored.

N/A

**Steps to Reproduce**
Simply make a request that is routed to a valid controller but doesn't have a method for the action that's requested.

**Your Environment**
N/A

**Additional context**
See [internal link](https://newrelic.atlassian.net/browse/DOTNET-4087) for the original issue.

For example:
Account/Edit/32, Valid controller, valid action

Account/XYZABDCDRE, Valid controller, invalid action  → still

CONT1234/Edit, invalid controller  →  Should double check, but likely covered by current agent behavior

**Impact**

While this does not happen very often, there isn’t really a workaround to help customer experiencing issue.

tehbio:
### Description

Resolves #155 by adding some extra `Mvc3` instrumentation to handle missing Controllers or Actions already targeted by the Route.

These request transactions now get appropriately named `WebTransaction/StatusCode/404`, mirroring `AspNetCore` instrumentation.

### Testing

Included is an update to the `BasicMvcApplication` integration test to include these types of 404 requests, and validate that they are no longer generating MGI.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New Features
 ### Fixes
+* Fixes issue [#155](https://github.com/newrelic/newrelic-dotnet-agent/issues/155): MVC invalid Action for valid Controller can cause MGI. ([#608](https://github.com/newrelic/newrelic-dotnet-agent/pull/608))
 
 ## [8.40.0] - 2021-06-08
 ### New Features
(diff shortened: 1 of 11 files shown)


tehbio:
I decided to instead move the added validation logic to a new `BasicMvcNotFoundTests` integration test to avoid the hacky `Delay()` call needed when they are grouped together to trigger the proper transaction trace.

vuqtran88 (approved):
This is a pretty clever approach. 🚢  it !

JcolemanNR (approved):
Looks good to me!

Issue #551, opened 2021-05-07
Missing external calls in WCF Service.


vuqtran88:
**Description**
It is reported that external calls made by HttpClient inside a WCF IIS Hosted service endpoint are not tracked by the agent. I am able to repro the behavior by calling the the HttpClient multiple times in async/await fashion in side a test WCF service hosted in IIS (ASP compatible mode enabled).

```
    [ServiceBehavior(AddressFilterMode = AddressFilterMode.Any)]
    public class Service1 : IService1
    {
		public async Task<string> MakeSomeExternalCall()
		{
            var httpClient = new HttpClient();

            var a = await httpClient.GetAsync("https://google.com");

            var b = await httpClient.GetAsync("https://bing.com");

            var c = await httpClient.GetAsync("https://yahoo.com");

            return $"{a.IsSuccessStatusCode} + {b.IsSuccessStatusCode} + {c.IsSuccessStatusCode}";
		}
	}
```

While investigating the issue, I notice the following messages in the agent log:
```
2021-05-07 20:51:05,238 NewRelic FINEST: [pid: 13500, tid: 15] Trx 8043d0d7e18ed004: Attempting to execute NewRelic.Providers.Wrapper.Wcf3.MethodInvokerWrapper found from InstrumentedMethodInfo: Method: System.ServiceModel.Dispatcher.TaskMethodInvoker, System.ServiceModel, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089:InvokeAsync(System.Object,System.Object[]), RequestedWrapperName: NewRelic.Agent.Core.Tracer.Factories.DefaultTracerFactory, IsAsync: True, RequestedMetricName: , RequestedTransactionNamePriority: , StartWebTransaction: False
2021-05-07 20:51:05,255 NewRelic FINEST: [pid: 13500, tid: 15] Trx 8043d0d7e18ed004: Setting transaction name to TransactionName{"IsWeb":true,"Category":"WCF","Name":"WcfService1.IService1.MakeSomeExternalCall","UnprefixedName":"WCF/WcfService1.IService1.MakeSomeExternalCall"} (priority 6, FrameworkHigh) from TransactionName{"IsWeb":true,"Category":"ASP","Name":"ServiceHttpHandler","UnprefixedName":"ASP/ServiceHttpHandler"} (priority 3, Handler)
[...]

PR #610, opened 2021-06-18
Fix missing external call instrumentation.

vuqtran88:
This is a fix for #551 .

### DIFF
--- a/tests/Agent/IntegrationTests/Shared/WCF/WcfClient.cs
+++ b/tests/Agent/IntegrationTests/Shared/WCF/WcfClient.cs
@@ -189,6 +189,12 @@ namespace NewRelic.Agent.IntegrationTests.Shared.Wcf
 
         }
 
+        public Task<string> TAP_TAPMakeExternalCalls()
+        {
+            return Channel.TAP_TAPMakeExternalCalls();
+
+        }
+
 
         public string Sync_TAPThrowException()
         {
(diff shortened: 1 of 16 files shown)


tests/Agent/IntegrationTests/Shared/WCF/WcfService.cs
```diff
@@ -62,6 +63,18 @@ public async Task<string> TAPGetData(int value)
             return await DoWork(value.ToString(), false, false);
         }
 
+        public async Task<string> TAPMakeExternalCalls()
+        {
+            using (var httpClient = new HttpClient())
+            {
+                _ = await httpClient.GetAsync("https://google.com");
+                _ = await httpClient.GetAsync("https://bing.com");
+                _ = await httpClient.GetAsync("https://yahoo.com");
```
JcolemanNR, line 72: Do we care about success or failure of these requests? or is that unimportant?

tests/Agent/IntegrationTests/Shared/WCF/WcfClient.cs
```diff
@@ -189,6 +189,12 @@ public Task<string> TAP_TAPGetData(int value)
 
         }
 
+        public Task<string> TAP_TAPMakeExternalCalls()
+        {
+            return Channel.TAP_TAPMakeExternalCalls();
+
```
JcolemanNR, line 195: nitpicky: extra newline

JcolemanNR (dismissed):
needs a changelog update

tests/Agent/IntegrationTests/Shared/WCF/WcfService.cs
tehbio: It's unimportant - we just need to validate that the instrumentation of the calls happened, even if the requests failed it's not significant for this test.

JcolemanNR (commented):
Looks good!

JcolemanNR (approved):
Looks good!

PR #614, opened 2021-06-23
Explain plan generation fixes

### LINKED ISSUE #168: When transactionTracer.enabled = false, we generate explain plans w/ no data
**Description**
When transactionTracer.enabled = false, we grab a SqlObfuscator that returns null for all values. This results in time being spent generating explain plans but then we null out everything. See TransactionTransformer.TryGenerateExplainPlans() and DataStoreSegmentBuilder.ExecuteExplainPlan()

This behavior seems like a bug. 

**Expected Behavior**
although explain plans are put into transaction traces, the main purpose is to put them into slow-sqls (aka slow query traces).  So if your agent creates slow-sqls when `transactionTracer.enabled = false`, I think you should still generate an explain plan.  (the relationship between slow-sql configuration and transaction-trace configuration is murky:  some agents have all slow-sql configuration under the transactionTracer namespace (I think))

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag]

**Steps to Reproduce**

**Your Environment**

**Additional context**
Migrated from an original issue.

jaffinito:
Fixes #168 

### Description

Summary:
Explain plans are only capture if TransactionTracer.explainPlan is true and either TransactionTracer.enabled and/or slowSql.enabled  are true.  This prevents an explain plan from being capture when it is never going to be sent up.  It prevent a capture explain plan from being dropped (very expensive to capture) and replaced using a "null" SqlObfuscator if TransactionTracer.enabled is false, but slowSql.enabled  is true. 

Note: By default, the agent will not capture explain plans, but it will capture transaction traces and slow sql traces. This should not affect most customers.  Customers with the following configuration will start seeing explain plans and the performance hit to generate them:
- TransactionTracer.explainPlan is **true**
- TransactionTracer.enabled is **false**
- slowSql.enabled is **true**

Context:
- TransactionTracer determines if transaction traces are collected
- SlowSQL determines if SQL traces are collected
- Explain p-lans are optional for both trace types.
- Explain plans are generated in TransactionTransformer before either trace is created.
- Explain plans are only generated if TransactionTracer.explainPlan is true.
    (src\Agent\NewRelic\Agent\Core\Transformers\TransactionTransformer\TransactionTransformer.cs:396)
    (src\Agent\NewRelic\Agent\Core\Agent.cs#171)
- A "nulling" SqlObfuscator will be used if TransactionTracer.enabled is false despite a real explain plan being collected.
During *TraceMaker workflow, ExplainPlan types are serialized into ExplainPlanWireModels independently.

Fix 1
 Changes when explain plans run.  Now explain Plans now run only when:
- TransactionTracer.explainPlan is true AND one or more of the following is also true:
- TransactionTracer.enabled is true
- slowSql.enabled is true

Otherwise, no explain plans are run.

Fix 2
Changes required config for explain plans. 
- Removes the need for TransactionTracer.enabled to be true from SqlObfuscator since SlowSql also can use explain plans.
- SqlObfuscator now will return a proper obfuscator depending on TransactionTracer.recordSql which was always passed into the the getter,
- Updated dependent classes to not pass in TransactionTracer.enabled anymore and fixed up tests for the same

### Testing

Unit tests should mass as normal.

### Changelog

We will need to add a note to the change log for the above, depending if it is accepted.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Database/SqlObfuscator.cs
+++ b/src/Agent/NewRelic/Agent/Core/Database/SqlObfuscator.cs
@@ -25,13 +25,8 @@ namespace NewRelic.Agent.Core.Database
             return ObfuscatingSqlObfuscatorInstanceUsingExplicit;
         }
 
-        public static SqlObfuscator GetSqlObfuscator(bool transactionTracerEnabled, string recordSqlValue)
+        public static SqlObfuscator GetSqlObfuscator(string recordSqlValue)
         {
-            if (!transactionTracerEnabled)
-            {
-                return new NoSqlObfuscator();
-            }
-
             if (string.IsNullOrEmpty(recordSqlValue))
             {
                 return GetObfuscatingSqlObfuscator();
--- a/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
+++ b/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
@@ -406,7 +406,7 @@ namespace NewRelic.Agent.Core.Transformers.TransactionTransformer
                     short count = 0;
                     var sqlExplainPlansMax = _configurationService.Configuration.SqlExplainPlansMax;
                     var threshold = _configurationService.Configuration.SqlExplainPlanThreshold;
-                    var obfuscator = SqlObfuscator.GetSqlObfuscator(_configurationService.Configuration.TransactionTracerEnabled, _configurationService.Configuration.TransactionTracerRecordSql);
+                    var obfuscator = SqlObfuscator.GetSqlObfuscator(_configurationService.Configuration.TransactionTracerRecordSql);
                     foreach (var segment in segments.Where(s => s.Data is DatastoreSegmentData))
                     {
                         if (segment.Duration > threshold)
--- a/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
+++ b/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
@@ -393,6 +393,8 @@ namespace NewRelic.Agent.Core.Transformers.TransactionTransformer
 
         private void TryGenerateExplainPlans(IEnumerable<Segment> segments)
         {
+            // First, check if explainPlans are disabled and return if they are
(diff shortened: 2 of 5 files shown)


src/Agent/NewRelic/Agent/Core/Database/SqlObfuscator.cs
```diff
@@ -25,13 +25,8 @@ public static SqlObfuscator GetObfuscatingSqlObfuscator()
             return ObfuscatingSqlObfuscatorInstanceUsingExplicit;
         }
 
-        public static SqlObfuscator GetSqlObfuscator(bool transactionTracerEnabled, string recordSqlValue)
+        public static SqlObfuscator GetSqlObfuscator(string recordSqlValue)
         {
-            if (!transactionTracerEnabled)
-            {
-                return new NoSqlObfuscator();
```
JcolemanNR, line 32: NoSql like MongoDB? I kid :)

src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
```diff
@@ -393,18 +393,20 @@ private void GenerateAndCollectSqlTrace(ImmutableTransaction immutableTransactio
 
         private void TryGenerateExplainPlans(IEnumerable<Segment> segments)
         {
-            if (!_configurationService.Configuration.SqlExplainPlansEnabled)
+            if (!_configurationService.Configuration.SqlExplainPlansEnabled
+                || (!_configurationService.Configuration.TransactionTracerEnabled && !_configurationService.Configuration.SlowSqlEnabled))
```
JcolemanNR, line 397: This logic is a little convoluted to read. Suggestion: write a comment on the intention

JcolemanNR (approved):
Looks good. A few suggestions, but no required changes on my end.

Also, are we are testing all the different configuration options for the intended behavior? I think there are 3 configuration options that can have 2 options, which I think would be 6 permutations? I only saw 5 tests updated though.

src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
jaffinito: Added more comments.

PR #617, opened 2021-06-25
Updated AgentManager to still try to configure logging when configuration loading fails

### LINKED ISSUE #266: Agent fails to initialize and provides no logs when configured with capitalized booleans
**Description**
When using application allow-listing to explicitly configure an app pool to be instrumented, and `true` is not set to all lowercase, we get a message in the profiler log that the app pool should be instrumented, but no agent log is created, no data is reported, and there is no message indicating an XML validation issue.

**Expected Behavior**
Either document casing requirements and log an error when they are violated, or accommodate multiple casing options. 
Regardless, log XML validation errors from the Unmanaged Agent (profiler).

**Steps to Reproduce**
Add this to `newrelic.config` to enable allow-listing for app pool `foo`:
  ```
  <applicationPools>
      <defaultBehavior instrument="false" />
      <applicationPool name="foo" instrument="True" />
  </applicationPools>
  ```
Send traffic to the app and note this message in the profiler log:
`This application pool (foo) is explicitly configured to be instrumented.`

We see these DLLs loaded into the process (see image) but no other NR dlls are loaded:
  - NewRelic.Profiler.dll
  - NewRelic.Agent.Core.dll
  - NewRelic.Agent.Extensions.dll

Screen Shot 2019-11-21 at 9 26 00 AM

No agent log is created and no errors indicate an issue with the config file.

When attempting to manually validate the config file we get:

`element applicationPool: Schemas validity error : Element ‘{urn:newrelic-config}applicationPool’, attribute ‘instrument’: ‘TruE’ is not a valid value of the atomic type ‘xs:boolean’.
test.config fails to validate`

tehbio:
### Description

Resolves #266 by adding an attempt to build a default logger when we fail to initialize configuration.

A more complete solution would be an updated Xml (de)serializer that can better handle these values, but this fix at least helps provide visibility to the user when an issue arises.

### Testing

Played around with building some tests, but couldn't find a good way to do it with our current setup.


JcolemanNR (approved):
Looks good

PR #618, opened 2021-06-25
Updated Kudu check string and added new command to unit test

### LINKED ISSUE #616: Linux Kudu not accessible when .NET agent presents.
**Description**
When installing the latest agent on an app specifically targeting .NET 5 on linux azure, Kudu is no longer accessible. When removing the agent, kudu works again. It appears that Kudu was instrumented by the agent and it is unclear why the agent broke Kudu.

This only happens in linux azure. It didn't happen on Windows azure, because the profiler was hardcoded to not instrument the .Net Framework version of Kudu.

This is how the profiler is currently detecting Kudu so that it can ignore Kudu.

https://github.com/newrelic/newrelic-dotnet-agent/blob/2ba99ea91ac20f79af40259e485949c6b3b1aac3/src/Agent/NewRelic/Profiler/Configuration/Configuration.h#L171 . This detection would fail when Kudu is invoked with this command `$: /opt/Kudu/Kudu.Services.Web`.
I suspect Microsoft changed how they invoke Kudu which made our detection no longer works.

A fix would be updating the profiler to detect the kudu process in linux and don't do anything.

We may consider a feature where we could specify name of apps that the agent should not instrument.

**Expected Behavior**
Kudu shouldn't break.

tehbio:
### Description

Resolves #616 with an updated Kudu check string for the Profiler.

### Testing

Added a new unit test to validate this new Kudu command is ignored.


vuqtran88 (approved):
🚀  💥

PR #622, opened 2021-06-30
Fix sqlparser word matching

### LINKED ISSUE #476: When generating and explain plan MS SQL parsing is matching parts of words instead of whole words
**Description**
When generating and explain plan MS SQL parsing is matching parts of words instead of whole words.  This results in broken explain plans.

Example:
Param values:
- @someParameters = someValuesGoHere

This snippet shows the query as provided by the code:
`DECLARE @someParametersCount INT = (SELECT Count(1) FROM @someParameters);`

This is what it looks like after the parser has finished:
`DECLARE someValuesGoHereCount INT = (SELECT Count(1) FROM someValuesGoHere);`

As you can see that "Count" parameter was replaced with a string which is not correct. 

**Expected Behavior**
MS SQL parsing should only look at whole words not parts.
Using the above example it should look like:
`DECLARE @someParametersCount INT = (SELECT Count(1) FROM someValuesGoHere);`

**Steps to Reproduce**
Create an application that performs parameterized MS SQL queries and have two parameters share start with the same value as the example able.

**Additional context**
https://newrelic.atlassian.net/browse/GTSE-7194
https://newrelic.zendesk.com/agent/tickets/442602

jaffinito:
Resolves #476 

### Description
Replaces string.replace with regex.replace
Regex is: @\b{paramName}\b
This looks for a parameter with a word-boundary on either side with an @ before the first boundary
This is needed to handle the @ symbol not being a word character
Old method used the length of the string to get around not looking for whole words
This needed a sorted list, a custom comparer, and other code, all of this was removed
Added new SqlParser.FixParameterizedSql to cover different scenarios.
No integration tests since unit test are pretty thorough
Tested performance and there was not appreciable difference between the two

### Testing

Added new SqlParser.FixParameterizedSql to cover different scenarios.
No integration tests since unit test are pretty thorough
Tested performance and there was not appreciable difference between the two

### Changelog

Updated.


tehbio (approved):
👍

PR #624, opened 2021-06-30
Add Linux package deploy test mode

### LINKED ISSUE #606: Linux package deploy to APT/YUM process needs a "test mode"
Recently, we tried updating the base container image used by the [dockerized process](https://github.com/newrelic/newrelic-dotnet-agent/tree/main/deploy/linux) that deploys new .deb and .rpm install packages to New Relic's public APT and YUM repos.  See the PR for this change: https://github.com/newrelic/newrelic-dotnet-agent/pull/546

Note my words from that PR's description: "Since we have no "dry run" option for the deploy process inside the container, I can't test this change all the way to the end."  Well, that turned out to be prescient because when we ran the deploy process for agent version 8.40.0, we ended up breaking the GPG signing of the production APT and YUM repos due to [this issue](https://github.com/newrelic/newrelic-dotnet-agent/issues/605).

We need a way of testing this deploy process without running a real deploy to production.  For now, the overall idea is:
1. Create a private mirror of the production APT and YUM repos we can test against without worrying about breaking anything.
2. Add an option to the GHA workflow for deploying the agent to target the test mirror instead of production.

Internal Record:  13161

nr-ahemsath:
### Description

Resolves #606.

1. Modified `deploy.yml` to add a new input called `linux-deploy-to-production` with a default value of `false`.  The workflow adds values from the secrets store for a) the S3 bucket to operate on for the deployment and b) the AWS access keys to use to operate on that bucket.  When the new value is false, it will operate on an S3 bucket that the team owns and is set up to simulate the "real" S3 bucket that backs `apt.newrelic.com` and `yum.newrelic.com`.  When set to true (as in a real release deploy), it will operate on the "real" S3 bucket.
2. Updated the README.md file for the deploy system to be more understandable and fix some inaccuracies leftover from old versions.
3. Added some progress/troubleshooting output to the `repoman_rebuild.bash` script.

### Testing

I tested the changes to the deploy process script locally against the test S3 bucket.  I have not tested the `deploy.yml` functionality yet since it will need to work with a test "release" based on these code changes.

### Changelog

N/A since these changes are for an internal NR process and do not affect agent behavior.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -270,14 +270,18 @@ jobs:
           touch docker.env
           echo "AGENT_VERSION=$AGENT_VERSION" >> docker.env
           echo "ACTION=release" >> docker.env
-          echo "S3_BUCKET=${{ secrets.PROD_MAIN_S3 }}" >> docker.env
-          echo "AWS_ACCESS_KEY_ID=${{ secrets.LINUX_AWS_ACCESS_KEY_ID }}" >> docker.env
-          echo "AWS_SECRET_ACCESS_KEY=${{ secrets.LINUX_AWS_SECRET_ACCESS_KEY }}" >> docker.env
+          if [ "${{ github.event.inputs.linux-deploy-to-production }}" == "true" ] ; then
+            # We're actually deploying to production (apt.newrelic.com and yum.newrelic.com)           
+            echo "S3_BUCKET=${{ secrets.PROD_MAIN_S3 }}" >> docker.env
+            echo "AWS_ACCESS_KEY_ID=${{ secrets.LINUX_AWS_ACCESS_KEY_ID }}" >> docker.env
+            echo "AWS_SECRET_ACCESS_KEY=${{ secrets.LINUX_AWS_SECRET_ACCESS_KEY }}" >> docker.env
+          else
+            # Deploy to the test bucket that mirrors apt/yum.newrelic.com
+            echo "S3_BUCKET=${{ secrets.TEST_S3_BUCKET }}" >> docker.env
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -266,14 +270,18 @@ jobs:
...
+            echo "AWS_ACCESS_KEY_ID=${{ secrets.LINUX_AWS_ACCESS_KEY_ID }}" >> docker.env
+            echo "AWS_SECRET_ACCESS_KEY=${{ secrets.LINUX_AWS_SECRET_ACCESS_KEY }}" >> docker.env
+          else
+            # Deploy to the test bucket that mirrors apt/yum.newrelic.com
+            echo "S3_BUCKET=${{ secrets.TEST_S3_BUCKET }}" >> docker.env
+            echo "AWS_ACCESS_KEY_ID=${{ secrets.TEST_BUCKET_AWS_ACCESS_KEY_ID }}" >> docker.env
+            echo "AWS_SECRET_ACCESS_KEY=${{ secrets.TEST_BUCKET_AWS_SECRET_ACCESS_KEY }}" >> docker.env
+          fi
           echo "GPG_KEYS=/deployscripts/gpg.tar.bz2" >> docker.env
-          echo "PROD_MAIN_S3=${{ secrets.PROD_MAIN_S3 }}" >> docker.env
```
nr-ahemsath, line 273: These values are no longer required by the deploy process scripts.  They are a holdover from our initial borrowing of these scripts from the PHP agent team, which has a more complex process.  Our version of the scripts has been simplified.

tehbio (approved):
Test deployment run worked for me, much better documentation. 👍

PR #625, opened 2021-07-01
Fix async await in lambda tracer

williamdenton:

### Description

The Lambda Tracing Wrapper uses Async over Sync and Reflection to call `.Result` on the task returned from lambda execution. Calling `.Result` can result in deadlocks and is considered bad practise.

This PR adds overloads to `LambdaWrapper` that support tasks correctly

### Testing

What unit or integration tests were added or modified, or why no testing changes were required.

### Changelog

Please remember to update our [changelog](/newrelic/newrelic-dotnet-agent/src/Agent/CHANGELOG.md) if applicable (or [this changelog](/newrelic/newrelic-dotnet-agent/blob/main/src/AwsLambda/CHANGELOG.md) for Lambda agent changes),

### DIFF
--- a/src/AwsLambda/AwsLambdaWrapper/LambdaWrapper.cs
+++ b/src/AwsLambda/AwsLambdaWrapper/LambdaWrapper.cs
@@ -83,6 +83,7 @@ namespace NewRelic.OpenTracing.AmazonLambda
             var scope = BeforeWrappedMethod(handler.Method.Name, context, distributedTraceContext, input);
             TOut output = default(TOut);
             object realOutput = null;
+
             try
             {
                 output = handler(input, context);
@@ -99,10 +100,11 @@ namespace NewRelic.OpenTracing.AmazonLambda
             return output;
         }
 
-        public async Task<TOut> LambdaWrapper<TInput, TOut>(Func<TInput, ILambdaContext, Task<TOut>> handler, TInput input, ILambdaContext context, ISpanContext? distributedTraceContext = null)
+        public async Task<TOut> LambdaWrapper<TInput, TOut>(Func<TInput, ILambdaContext, Task<TOut>> handler, TInput input, ILambdaContext context, ISpanContext distributedTraceContext = null)
         {
             var scope = BeforeWrappedMethod(handler.Method.Name, context, distributedTraceContext, input);
-            TOut? output;
+            TOut output = default(TOut);
+
             try
             {
                 output = await handler(input, context).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


angelatan2:
Hello @williamdenton, thank you for submitting a PR to our repository. We appreciate that you are calling out a potential deficiency in our agent and suggesting a solution.  We will review your PR shortly.

vuqtran88:
@williamdenton
Thanks for your contribution. I have some comments/thought about this change.
1.  I think adding 2 new `Task` friendly overloads makes a lot sense and love to have these. So 👍  on this.
2. I'm not so sure removing Async over Sync handling logic is a good idea at this moment. Even though your suggestion is great (elimination potential deadlock and stuffs), this api is already being used by customers for async calls. Removing this logic will surprise existing customers (async calls no longer correctly tracked). To minimize the impact, the right thing to do is probably still keep this logic.

Please let me know your thought.

williamdenton:
@vuqtran88 agree on removing the old method, that is probably too aggressive given this library probably needs to retain backwards binary compatibility. Clients recompiling will get the correct overloads, ,and others will continue to function as they did previously.

To retain blame correctly on the code, i have reverted that change and force pushed a new commit that doesn't touch the existing code.

Also i believe that in the case of a `Task` (rather than `Task<>`) this code would previously not handled the span correctly, essentially disposing the scope before the task had finished. There is provision for handling `Task<>` but that only applies to lambdas that return something (like APiGateway) but not others like SQS. This will fix that.

williamdenton:
https://docs.newrelic.com/docs/serverless-function-monitoring/aws-lambda-monitoring/enable-lambda-monitoring/enable-serverless-monitoring-aws-lambda-legacy/

image

this PR addresses this issue with "blocking"

vuqtran88:
@williamdenton 
Thanks for the update. I just came from a short vacation. We are still reviewing your change.

williamdenton:
these failing tests don't appear to be related to this PR - is there anything i missed that's causing this? Looks like something to do with S3 credentials for uploading build logs?

tehbio (approved):
Hey @williamdenton!

Don't worry about the test results here on GH - since this PR is coming from a fork it cannot access our secrets so they fail. I've pulled down your changes and have been testing them myself.

Your changes work well! I pushed a small update to your branch to drop the null annotations to keep things consistent and since the project doesn't actually target C# 8, I also updated the changelog with some credit to you for this change, and I will be updating our async handler documentation to reflect this fix as well.

Thanks again for your contribution and helping to improve our Agent!

JcolemanNR (approved):
💯 🥇

PR #626, opened 2021-07-01
Always use thread id in agent logs

### LINKED ISSUE #600: In .NET 6, thread ID in agent log shows generic name instead of numeric thread ID
When running the agent attached to a .NET 6 Preview application, the agent's log file shows messages like this:

`“NewRelic FINEST: [pid: 24300, tid: .NET ThreadPool Worker]”` instead of having the `tid` field be a numeric thread ID as it currently is in all other .NET runtimes.  We're assuming this is because the .NET threadpool is re-implemented in managed code in .NET 6.

More specifically, our Log4Net logging template uses the `%thread` formatting directive.  Under the covers, Log4Net first tries to get the thread's *name* and then falls back to getting the numeric thread ID if no name is set.  Apparently pre-.NET 6 the threadpool threads were not named by default, and in .NET 6 they will be.  See this SO discussion for more context and a possible solution: https://stackoverflow.com/questions/42675024/log4net-printing-threadid-and-thread-name-in-the-log/44112461

JcolemanNR:
### Description

While testing out .NET 6, we noticed that thread names were being logged for .NET managed threads instead of the actually desired thread id in agent logfiles. This happened because Log4Net prefers to log a thread name versus a thread id, and Microsoft began setting thread names for their managed threads in .NET 6. The behavior is reproducible in previous versions of the product, but only if a user explicitly set the name of a thread.

The resolution is to check/set a thread id in the log4net thread context object, and update the logging layout to log this property. This is the same approach we use to log the process id, with the exception of only ever needing to set the process id once.

Resolves #600. 

### Testing

Unit tests were added to verify that the thread id property is set in log4net by every possible agent log binding. The .net core mvc async integration test application was updated to set a main thread name, and assertions were added in the text fixture to verify that this thread name, and the new .net background worker thread name does not show up in the log file instead of a thread id.


nr-ahemsath (approved):
Looks good to me.  I did the following testing on this branch:
1. Built the agent and ran the unit tests, which all passed except for one unrelated unit test which has been flickering (failing intermittently).
2. Built the integration tests and ran them.  All but 9 passed; the 9 failures were either related to package publishing issues caused by my having the .NET 6 preview 4 SDK installed, or are known bad tests (e.g. ThreadProfileStressTest).
3. Manually tested running a .NET 6-targeted console app with the agent attached and examined the log file, and saw that it is now logging with the numeric thread id rather than the ".NET Threadpool Worker Thread" name.

:sheep: it!

PR #628, opened 2021-07-01
Update Linux package deploy process to use Debian 10 (Buster) and gnupg1

### LINKED ISSUE #605: Update GPG signing in linux package deploy scripts to handle GPG 2.x+
Summary: the APT/YUM repo signing command used in the scripts that deploy our .deb/.rpm install packages for Linux fails when run on a `debian:stable` system.  We believe the issue is caused by `debian:stable` having a 2.x version of GPG, whereas `debian:jessie` has a 1.x version.

More context: we [recently tried updating](https://github.com/newrelic/newrelic-dotnet-agent/pull/546) the base container image used by the APT/YUM repo deploy process from `debian:jessie` to `debian:stable`; however, the first time we tried deploying the agent to `apt.newrelic.com` and `yum.newrelic.com` with the updated process, the repo signing process failed, causing anybody trying to run `apt-get update` with our repo configured to get an error about an unsigned repo.

The specific error in the output from the [signing command](https://github.com/newrelic/newrelic-dotnet-agent/blob/5a808b374c385659ca834632d094e89017e682a2/deploy/linux/deploy_scripts/libexec/repoman-rebuild.bash#L83) was:
```
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
```

We've reverted the base container image for this process back to `debian:jessie` for now, but we want to upgrade eventually to mitigate CVEs found in our repo by security scan tools due to using an old, out-of-date version of debian for this process (although note that this container is only used for a deploy process and isn't shipped to any customers with the agent).  In order to upgrade we will have to find a fix for this signing error.

nr-ahemsath:
### Description

Resolves #605.  This PR:

1.  Updates the base container image for the Linux package deploy process to a version of Debian Buster, which is supported through 2024.
2. Installs the `gnupg1` package in the container and updates the part of the script which signs the repo to use `gpg1` instead of `gpg`.  This allows us to continue to sign the repo in the same way we have been signing it since 2017.

Note: this PR doesn't address a separate concern in #605, which is that the deploy process in GitHub Actions should fail if the repo signing step (or any other step) has errors.  More investigation is needed to address that issue and I will create a follow-on issue to track that work.

### Testing

Tested locally against the test S3 bucket.  The deploy process succeeded and I was able to configure an Ubuntu 18.04 system to use the test bucket as a package source and successfully install the version of the agent I had just deployed, without any key/signing warnings or errors.

### Changelog

N/A since this is an internal process that doesn't affect agent behavior.


JcolemanNR:
Looks good to me! I'd like to learn how to test locally against the dev bucket tomorrow

tehbio (approved):
Managed to deploy a new version to test bucket and validated that it worked, then performed a rollback and confirmed that was successful as well. 👍

JcolemanNR (approved):
Looks good. I ran through the instructions and was able to publish and rollback a release to the test s3 bucket

PR #630, opened 2021-07-01
Improve sqlparsing tostring in FixParameterizedSql

### LINKED ISSUE #477: SQL Explain plans MS SQL parser needs to be able to ToString an object to work with parameterized queries
**Description**
SQL Explain plans MS SQL parser needs to be able to ToString an object to work with parameterized queries.  If the type is a custom one, the call to ToString returns the type details instead of the expected string data. 

[Code](https://github.com/newrelic/newrelic-dotnet-agent/blob/ddc5335fc643b9e64cc02527ddb96559c7dccb97/src/Agent/NewRelic/Agent/Parsing/SqlParser.cs#L402):
`sql = sql.Replace(paramName, value.ToString());`

Example SQL snippet prior to parsing:
`INNER JOIN @someParameters fp ON p.[Name] = fp.[Key]`

Example SQL snippet post parsing:
`INNER JOIN Custom.Special.Stuff.CustomParameter+CustomParameterValues`1[System.String] fp ON p.[Name] = fp.[Key]`

Here is a snippet of an example customer type (full example is in JIRA):
`private sealed class CustomParameterValues<T> : IEnumerable<SqlDataRecord>`

Here is an example error:
```
2021-01-01 13:07:05,271 NewRelic  DEBUG: [pid: 3410, tid: 4] Unable to execute explain plan: System.Data.SqlClient.SqlException (0x80131904): Incorrect syntax near '.'.
The object name 'Custom.Special.Stuff.CustomParameter' contains more than the maximum number of prefixes. The maximum is 3.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds)
   at System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior)
   at NewRelic.Parsing.SqlServerExplainPlanActions.GenerateExplainPlan(Object resources)
   at NewRelic.Agent.Core.Segments.DatastoreSegmentData.ExecuteExplainPlan(SqlObfuscator obfuscator)
ClientConnectionId:xxxxxxx
Error Number:102,State:1,Class:15
```

**Expected Behavior**
Agent should be able get access to values OR not parse the SQL statement.

**Steps to Reproduce**
Create an application that performs parameterized MS SQL queries and have a custom type that is not a string as one of the parameters.

**Your Environment**
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of the agent and any relevant configurations, such as the .NET framework or core version of your application. )

**Additional context**
https://newrelic.atlassian.net/browse/GTSE-7194
https://newrelic.zendesk.com/agent/tickets/442602

jaffinito:
Resolves #477

### Description

Big change tl;dr: If DbType.Binary and DbType.Object are detected while parsing, we will exit and not collect an explain plan.  There is a debug message indicating as such.  This prevent any errors with custom objects or objects that are not easily serialized.

Prevents explain plans from being created when Binary or Object parameters are detected
- SqlParser.FixParameterizedSqlnow returns a bool to indicate if we should try to run the plan
- MySqlExplainPlanActions and SqlServerExplainPlanActions hasve been update to honor this boolean
- Without a large amount of overhead, we are not able to determine how to write the underlying objects as strings.  With the SqlParser we want to keep it very lightweight since it is already a large performance hit.  This should mean less errors for the customer while not reducing the amount of plans we can create since we would error out before
- Logs a Debug message when we block explain plans for Binary or Object
- Uses DbType enum for detecting Boolean
- Reorganizes SqlParser.FixParameterizedSql

### Testing

Expanded the SqlParser.FixParameterizedSql tests to cover more of the options include failures

### Changelog

Updated, but please check it for verbage!

### DIFF
--- a/src/Agent/NewRelic/Agent/Parsing/MySqlExplainPlanActions.cs
+++ b/src/Agent/NewRelic/Agent/Parsing/MySqlExplainPlanActions.cs
@@ -39,15 +39,14 @@ namespace NewRelic.Parsing
         public static ExplainPlan GenerateExplainPlan(object resources)
         {
             if (!(resources is IDbCommand))
+            {
                 return null;
-
-            var dbCommand = (IDbCommand)resources;
-            if (dbCommand.Connection.State != ConnectionState.Open)
-                dbCommand.Connection.Open();
+            }
 
             ExplainPlan explainPlan = null;
 
             //KILL THE CONNECTION NO MATTER WHAT HAPPENS DURING EXECUTION TO AVOID CONNECTION LEAKING
+            var dbCommand = (IDbCommand)resources;
             using (dbCommand)
             using (dbCommand.Connection)
             {
@@ -55,6 +54,11 @@ namespace NewRelic.Parsing
                 if (!shouldGeneratePlan)
                 {
                     return explainPlan;
+                }
+
+                if (dbCommand.Connection.State != ConnectionState.Open)
+                {
+                    dbCommand.Connection.Open();
                 }
 
                 dbCommand.Transaction = null;
@@ -77,9 +81,7 @@ namespace NewRelic.Parsing
                     }
 
                     explainPlan.ExplainPlanDatas = explainPlanDatas;
-
                 }
-
             }
 
             return explainPlan;
--- a/src/Agent/NewRelic/Agent/Parsing/MySqlExplainPlanActions.cs
+++ b/src/Agent/NewRelic/Agent/Parsing/MySqlExplainPlanActions.cs
@@ -51,9 +51,13 @@ namespace NewRelic.Parsing
             using (dbCommand)
             using (dbCommand.Connection)
             {
+                var shouldGeneratePlan = SqlParser.FixParameterizedSql(dbCommand);
+                if (!shouldGeneratePlan)
+                {
+                    return explainPlan;
+                }
 
                 dbCommand.Transaction = null;
-                SqlParser.FixParameterizedSql(dbCommand);
                 dbCommand.CommandText = "EXPLAIN " + dbCommand.CommandText;
 
 
--- a/src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
+++ b/src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
@@ -3,6 +3,7 @@
 
 using NewRelic.Agent.Extensions.Parsing;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
+using NewRelic.Core.Logging;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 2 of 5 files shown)


src/Agent/NewRelic/Agent/Parsing/MySqlExplainPlanActions.cs
```diff
@@ -51,9 +51,13 @@ public static ExplainPlan GenerateExplainPlan(object resources)
             using (dbCommand)
             using (dbCommand.Connection)
```
JcolemanNR, line 52: Not related to your changes, but should the 'dbcommand.connection.Open()' on line 46 be INSIDE this using block?

src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
```diff
@@ -357,31 +358,33 @@ protected override ParsedSqlStatement CreateParsedDatabaseStatement(DatastoreVen
...
+                // Binary: We cannot determine exactly what is in the binary so the out could be invalid for SQL.
+                // Object: Object could be almost anything so there is no easy way ToString it that just works without trying to detect all the differnt objects
+                // It is safer, less error prone, and more efficient to short-circuit on these types than to try and parse them out.
                 var type = dbParam.DbType;
-                object value = dbParam.Value;
-                if (quotableTypes.Contains(type.ToString()))  // the TypeCode for Strings is Int32 for some reason
+                if (type == DbType.Binary || type == DbType.Object)
                 {
-                    value = QuoteString(value.ToString());
+                    Log.DebugFormat("Not executing explain plan since DbType is either Binary or Object."); // not getting the enum string to save resources
```
JcolemanNR, line 384: you could get the type and log it by using two if statements with static log messages.

JcolemanNR (approved):
Looks good. I'm wondering if it would be a good idea to log one message per agent session, or rate limiting a log message to one per minute at the warn level to inform users that we tried to generate an explain plan, but were unable to?

src/Agent/NewRelic/Agent/Parsing/MySqlExplainPlanActions.cs
jaffinito: You know that does look a little odd.... let me read through it some more...

src/Agent/NewRelic/Agent/Parsing/SqlParser.cs
jaffinito: I guess we could. I don't really think it needed a separate statement.

jaffinito: Split things up.

src/Agent/NewRelic/Agent/Parsing/MySqlExplainPlanActions.cs
jaffinito: Great catch Josh! I moved the connection.open calls into the using, after the parameter check.  If the connection is closed and we don't run a plan, we will now save a DB connection.

JcolemanNR (approved):
Ship it!

PR #636, opened 2021-07-07
Prevent inlining in integration test app methods

nr-ahemsath:
### Description

This PR improves integration test reliability by adding the `MethodImplOptions.NoInlining` attribute to methods in test applications also decorated with `Transaction` or `Trace` attributes.

At least one integration test, `DistributedTracingApiTests_Legacy`, has been failing when the tests are built and run in `Release` mode (as they are when run in GHA as part of our CI) because some test methods in the underlying test application were being inlined by the compiler.  These methods were decorated with `[Transaction]` and were expected to create transactions in the agent to make the test work.  Preventing these methods from being inlined causes the test to pass reliably when built and run in `Release` mode.

I found other examples of transaction- or trace-decorated methods similarly unprotected from inlining, and added the MethodImpl attribute to those methods as well, just to be safe.

### Testing

The integration test all pass (with the exception of four tests known to have unrelated issues) for me on my system when built and run in release mode.

### Changelog

N/A

### DIFF
--- a/tests/Agent/IntegrationTests/Applications/AspNetCoreMvcAsyncApplication/CustomMiddleware.cs
+++ b/tests/Agent/IntegrationTests/Applications/AspNetCoreMvcAsyncApplication/CustomMiddleware.cs
@@ -19,6 +19,7 @@ namespace AspNetCoreMvcAsyncApplication
         }
 
         [Trace]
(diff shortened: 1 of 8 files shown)


tehbio (approved):
Looks good, slowly turning everything green. 👍

PR #637, opened 2021-07-07
Improve integration test reliability

JcolemanNR:
### Description

Issues have been identified with differences in ASP pipeline events generated between the release and debug runs of our WCF distributed tracing events. Additionally, issues have been identified where we think a port is available for use because no one is listening, but it is actually prevented from being used by the OS. 

The first issue was resolved by adjusting the static number of expected pipeline events between debug and release builds. I left some commented code in the PR which was useful in comparing the span contents of debug/release runs of the WCF DT tests.

The second issue was resolved by starting a TcpListener instead of using a TcpClient to test if a port is available. The TcpListener will get an error if the port is unavailable.

### Testing

I ran integration tests locally in debug and release and did not see any new failures

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/WCF/WCFServiceTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/WCF/WCFServiceTests.cs
@@ -9,6 +9,7 @@ using NewRelic.Agent.IntegrationTestHelpers.Models;
 using NewRelic.Testing.Assertions;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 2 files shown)


tests/Agent/IntegrationTests/IntegrationTests/WCF/WCFServiceTests.cs
```diff
@@ -361,7 +362,14 @@ public override void DistributedTracing_SpanEvents()
             var unexpectedMethodNames = new List<string>();
 
             var includeASPPipelineEvents = _hostingModelOption == HostingModel.IIS && _aspCompatibilityOption == ASPCompatibilityMode.Enabled;
-            const int COUNT_ASP_PIPELINE_EVENTS = 10;
+
+            // In debug we consistently get 10 extra ASP pipeline events, in Release we consistently get 9.
```
nr-ahemsath, line 366: This is great and if it gets us a consistently passing integration test I'm all for it.  However, I think @tehbio had an alternative solution in mind where we make `PreExecuteRequestHandler` do something non-trivial in the test application so that it always runs regardless of build mode.  Thoughts?

nr-ahemsath: Spoke with folks offline, we're okay with the way this PR handles this debug/release difference for this test.

nr-ahemsath (approved):
Approving this PR. 👍 

Some caveats: I did a run of all the tests in the IntegrationTests.sln on my system.  There were 23 failures out of 1024 tests.  Some of the failures were still caused by a test application being unable to bind to the port it had been assigned.  It's hard to know why.  22 of the 23 failures passed when re-run (the one continuing failure being the ThreadProfileStressTest, which is a known issue).  However, 22/2^10 test flickers seems better than I've typically experienced so this seems like an improvement.  It's also worth noting that all of the integration test jobs passed in GHA on the CI run for this PR, which seems 👏-worthy.  So, this seems to be an improvement we should merge even if we still have port-in-use-related flickers on our dev systems.

PR #638, opened 2021-07-08
Separate MSI/Debian/RPM build artifacts

JcolemanNR:
### Description

A common failure in our GitHub Actions all solutions build is in the debian/rpm artifact upload steps. It is common to get 500 errors during artifact upload, and then 400 errors causing the deb or rpm artifacts to be missed. I believe this is due to the use/re-use of the 'build-folder-artifact' that is generated in the MSI build step. The deb/rpm builds were both downloading this artifact, adding 1-3 files to it, then simultaneously trying to upload the entire build folder again in to the same artifact name, including the already downloaded files from the MSI build step. This was causing simultaneous uploads of the same artifact name/file, which I believe is why we were encountering intermittent errors. GitHub warns against doing what we were doing here:
https://github.com/actions/upload-artifact#uploading-to-the-same-artifact

I made the debian/rpm builds upload only the new packages they generated as separate artifacts, and re-assembled everything in the ArtifactBuilder step to not require changes to expected paths. I can see a few ways to refactor this, but this should improve the reliability of our builds by avoiding trying to upload the same file(s) from concurrently running build steps.

### Testing

There is no code level change, I extensively tested the changes by manually kicking off the all solutions workflow in GHA.

I was hoping the RPM/Deb build steps would be INSANELY faster because they are only uploading about a 3rd of the data as artifacts, but the impact was pretty minimal.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -251,431 +251,431 @@ jobs:
           path: ${{ github.workspace }}\src\AwsLambda\AwsLambdaOpenTracer\bin\Release\netstandard2.0-ILRepacked
           if-no-files-found: error
 
-      - name: Unit Tests
-        run: |
-          # Write-Host ${{ env.scripts_path }}\DotNet-Agent-CI-UnitTests.ps1
-          # ${{ env.scripts_path }}\DotNet-Agent-CI-UnitTests.ps1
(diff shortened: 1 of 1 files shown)


nr-ahemsath (approved):
Looks good.  Good catch on GitHub's advice on not uploading to the same artifact multiple times. 🥇

Issue #545, opened 2021-04-30
Support MySql version 8.0.25


angelatan2:
### Is your feature request related to a problem? Please describe.
Currently, the .NET agent only supports MySql.Data version 6.10.4. or the MySQL Connector: https://github.com/mysql-net/MySqlConnector.

### Feature Description
Please support version 8.0.22 or the latest version.
Support .NET Framework and .NET Core.

### Additional context
[Internal](https://drive.google.com/file/d/1mgsO988raWu-aLyjrDQSjEBmuYIZQ6_R/view?usp=sharing) 
[Internal FR](https://newrelic.aha.io/ideas/ideas/DOTNET-I-1170)

angelatan2:
Supported in released version 8.41.0.

PR #645, opened 2021-07-13
Update integration tests to use MySql.Data 8.0.25

tehbio:
### Description

Addresses #545 by updating the unbounded integration tests to use the latest `MySql.Data` version of `8.0.25`, also added `MySql.Data` package to MVS.

Versions after `6.10.9` require .NET Framework 4.5.2 so I also had to bump up one of the projects to use it - this project does host a few things for other tests so we may want to break out the `MySqlController` into its own project, but so far I have not seen any issues with other tests because of this.

### Testing

With this version update `Network Address` is not supported in connection strings, so any existing secrets will need to be updated to use `Server` instead - GH secrets have been updated.


JcolemanNR (approved):
Looks good! The tests pass too!

PR #648, opened 2021-07-14
RabbitMQ: Use generic IDictionary interface when manipulating headers

### LINKED ISSUE #639: RabbitMQ: NR agent version 8.39.2 overwrites message headers
**Description**
After upgrading NewRelic agent **8.37.0 to 8.39.2** our services started to fail due to missing headers in RabbitMQ messages.

**Expected Behavior**
New Relic should not overwrite application's headers with it's own.

**Steps to Reproduce**
Application written in F# 5, publishing messages looks like this (simplified)

    let ch = conn.CreateModel()
    let props = ch.CreateBasicProperties()
    props.Persistent <- true
    props.ContentType <- "application/json"
    props.Headers <-
        Map.ofList [ "trace", trace |> Array.map (fun it -> it.ToString()) :> obj
                     "publisher", appName :> obj
                     "ts", now |> Utils.instantToISOString :> obj ]
    ch.BasicPublish(exchange, routingKey, props, ReadOnlyMemory(Encoding.UTF8.GetBytes body))

**Your Environment**
- Linux 5.4
- dotnet 5.0
- RabbitMQ 3.6 server
- Nuget: "RabbitMQ.Client" Version="6.2.1"
- newrelic-netcore20-agent_8.39.2.0_amd64.tar.gz

**Additional context**
Nothing yet, work in progress…

nr-ahemsath:
### Description

Fixes #639 by modifying the RabbitMQ instrumentation to read and write RabbitMQ message headers via the generic `IDictionary<string,object>` interface rather than casting everything to a concrete Dictionary.  This will prevent the instrumentation from accidentally overwriting existing message headers if they were set using an IDictionary collection other than Dictionary. 

### Testing

Updated the test application used by the RabbitMQ integration tests to set some user headers in a `SortedDictionary` when sending messages and then verify they still exist and are correct when receiving messages.  Verified the tests pass locally.

### Changelog

Changelog updated.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New Features
 ### Fixes
-* Fixes issue [#639](https://github.com/newrelic/newrelic-dotnet-agent/issues/639): RabbitMQ instrumentation can delete user headers from messages. ([#648](https://github.com/newrelic/newrelic-dotnet-agent/pull/648))
+* Fixes issue [#639](https://github.com/newrelic/newrelic-dotnet-agent/issues/639): RabbitMQ instrumentation can delete user headers from messages. Thank you @witoldsz for finding and reporting this bug. ([#648](https://github.com/newrelic/newrelic-dotnet-agent/pull/648))
 
 ## [8.40.1] - 2021-07-08
 ### Fixes
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New Features
 ### Fixes
+* Fixes issue [#639](https://github.com/newrelic/newrelic-dotnet-agent/issues/639): RabbitMQ instrumentation can delete user headers from messages. ([#648](https://github.com/newrelic/newrelic-dotnet-agent/pull/648))
+
+## [8.40.1] - 2021-07-08
+### Fixes
 * Fixes issue [#485](https://github.com/newrelic/newrelic-dotnet-agent/issues/485): `SendDataOnExit` configuration setting will prevent Infinite Traces data sending interuption on application exit. ([#550](https://github.com/newrelic/newrelic-dotnet-agent/pull/609))
 * Fixes issue [#155](https://github.com/newrelic/newrelic-dotnet-agent/issues/155): MVC invalid Action for valid Controller can cause MGI. ([#608](https://github.com/newrelic/newrelic-dotnet-agent/pull/608))
 * Fixes issue [#186](https://github.com/newrelic/newrelic-dotnet-agent/issues/186): Attribute based Routing (ex WebAPI) can cause transaction naming issues. ([#612](https://github.com/newrelic/newrelic-dotnet-agent/pull/612))
@@ -289,7 +293,8 @@ Fixes issue where updating custom instrumentation while application is running c
 ### Fixes
 * New Relic distributed tracing relies on propagating trace and span identifiers in the headers of external calls (e.g., an HTTP call). These identifiers now only contain lowercase alphanumeric characters. Previous versions of the .NET agent used uppercase alphanumeric characters. The usage of uppercase alphanumeric characters can break traces when calling downstream services also monitored by a New Relic agent that supports W3C trace context (New Relic's .NET agent does not currently support W3C trace context. Support for W3C trace context for .NET will be in an upcoming release). This is only a problem if a .NET application is the originator of the trace.
 
-[Unreleased]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.40.0...HEAD
+[Unreleased]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.40.1...HEAD
+[8.40.1]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.40.0...v8.40.1
 [8.40.0]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.39.2...v8.40.0
 [8.39.2]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.39.1...v8.39.2
 [8.39.1]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.39.0...v8.39.1
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/RabbitMqHelper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/RabbitMqHelper.cs
@@ -20,16 +20,16 @@ namespace NewRelic.Providers.Wrapper.RabbitMq
         public const string TypeName = "RabbitMQ.Client.Framing.Impl.Model";
 
         private static Func<object, object> _getHeadersFunc;
-        public static IDictionary<string, object> GetHeaders(object Properties)
+        public static IDictionary<string, object> GetHeaders(object properties)
         {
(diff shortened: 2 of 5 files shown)


src/Agent/CHANGELOG.md
```diff
@@ -289,7 +293,8 @@ Fixes issue where updating custom instrumentation while application is running c
 ### Fixes
 * New Relic distributed tracing relies on propagating trace and span identifiers in the headers of external calls (e.g., an HTTP call). These identifiers now only contain lowercase alphanumeric characters. Previous versions of the .NET agent used uppercase alphanumeric characters. The usage of uppercase alphanumeric characters can break traces when calling downstream services also monitored by a New Relic agent that supports W3C trace context (New Relic's .NET agent does not currently support W3C trace context. Support for W3C trace context for .NET will be in an upcoming release). This is only a problem if a .NET application is the originator of the trace.
 
-[Unreleased]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.40.0...HEAD
+[Unreleased]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.40.1...HEAD
+[8.40.1]: https://github.com/newrelic/newrelic-dotnet-agent/compare/v8.40.0...v8.40.1
```
tehbio, line 297: Thanks for this, I forgot that part of the process... 🤦‍♂️

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/RabbitMQ.cs
```diff
@@ -206,10 +221,47 @@ void handler(object ch, BasicDeliverEventArgs basicDeliverEventArgs)
...
                     receivedMessage = Encoding.UTF8.GetString(basicDeliverEventArgs.Body);
 #endif
+                    VerifyHeaders(basicDeliverEventArgs.BasicProperties.Headers);
 
                     manualResetEvent.Set();
                 }
             }
         }
+
+        private void VerifyHeaders(IDictionary<string, object> headers)
```
nr-ahemsath, line 231: I realize this code is pretty gross and I'd appreciate any ideas for how to make it cleaner.

tehbio: Looks like this is because the header strings get encoded as bytes, so you might be able to simplify a little by only checking that case and falling back to `.Equals()` otherwise.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/RabbitMQ.cs
```diff
@@ -206,10 +221,47 @@ void handler(object ch, BasicDeliverEventArgs basicDeliverEventArgs)
...
+                        }
+                    }
+                    else
+                    {
+                        throw new Exception("Unknown user header value type");
+                    }
+                }
+                else
+                {
+                    throw new Exception("Did not find expected user header value in recieved message.");
```
JcolemanNR, line 256: typo: received. Not sure why my brain spots these things from 10,000 feet. 🤷‍♂️

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/RabbitMQ.cs
JcolemanNR: I feel like the code is complicated because it attempts to future-proof additions to the headers collection instead of just using known keys/values test cases. I'm not sure I would refactor it to not be future proof though. 👍

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/RabbitMqHelper.cs
```diff
@@ -20,16 +20,16 @@ public class RabbitMqHelper
         public const string TypeName = "RabbitMQ.Client.Framing.Impl.Model";
 
         private static Func<object, object> _getHeadersFunc;
-        public static Dictionary<string, object> GetHeaders(object Properties)
+        public static IDictionary<string, object> GetHeaders(object Properties)
```
JcolemanNR, line 23: input parameter should be named `properties` instead of `Properties`

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/RabbitMqHelper.cs
```diff
@@ -20,16 +20,16 @@ public class RabbitMqHelper
...
         {
             var func = _getHeadersFunc ?? (_getHeadersFunc = VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(Properties.GetType(), "Headers"));
-            return func(Properties) as Dictionary<string, object>;
+            return func(Properties) as IDictionary<string, object>;
         }
 
-        public static void SetHeaders(object Properties, Dictionary<string, object> headers)
+        public static void SetHeaders(object Properties, IDictionary<string, object> headers)
```
JcolemanNR, line 29: `properties` instead of `Properties`

JcolemanNR (changes requested):
Looks good! Fix the typo's and style points and you win my 🥇

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/RabbitMQ.cs
nr-ahemsath: I took @tehbio's advice and that simplified things a bit, take a look.

JcolemanNR (approved):
Ship it!

witoldsz:
I am pretty sure this fix will make NewRelic library crash our services, because you assume the `IDictionary` is mutable, but it is not guaranteed by the `IDictionary` interface and F#'s `Map` implements `IDictionary` and is not mutable.

Please, take a look:
```
$ dotnet fsi

Microsoft (R) F# Interactive version 11.4.2.0 for F# 5.0
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;
> let m = Map.ofList(["a",1]);;
val m : Map<string,int> = map [("a", 1)]

> open System.Collections.Generic;;
> let mm = m :> IDictionary<string,int>;;
val mm : IDictionary<string,int> = map [("a", 1)]

> mm.Clear();;
System.NotSupportedException: Map values cannot be mutated.
   at Microsoft.FSharp.Collections.FSharpMap`2.System.Collections.Generic.ICollection<System.Collections.Generic.KeyValuePair<'Key, 'Value>>.Clear()

> mm.Add("b", 2);;
System.NotSupportedException: Map values cannot be mutated.
   at Microsoft.FSharp.Collections.FSharpMap`2.System.Collections.Generic.IDictionary<'Key, 'Value>.Add(TKey k, TValue v)
```

nr-ahemsath:
@witoldsz Thanks for bringing this to our attention.  A few thoughts:

1. If we can't add our tracing headers to the RabbitMQ message's header collection, our distributed tracing feature will not work.  Is there any way to use a mutable header collection in your application?  (I realize the answer is probably "no" given that F# is a functional language, but I figured I'd ask.)
2. As you may know from looking at our repo, our automated integration tests are all written in C#.  We do not regularly test F#.  We will probably need to think a bit about how we want to deal with the possibility of immutable header collections in our instrumentation libraries that need to add headers to requests/messages in order to support all NR features.

witoldsz:
@nr-ahemsath It is not F# vs C#. You can write C# code and use immutable or persistent collections. I would argue against mutating the objects of intercepted methods, especially when you interfere with 3rd party code (the case of NR agent), but replace it with new one with extra items in it instead.

PR #651, opened 2021-07-15
Create stack traces on every segment and send them up in transaction traces

### LINKED ISSUE #643: Stack Traces: Capture stack trace at end of each instrumented method
- Capture stack trace at end of each instrumented method
- Requires data from spike to determine how best to implement
- Needs to be store in Agent Attributes (Agent specifically)
- Evaluate Performance
- Can this be enabled by default?
- Does performance allow this to be enabled default?
- Add or update integration test to verify
- Stack trace is added as an attribute
- Validate size is within limits set in config
-

jaffinito:
Resolves #643 

### Description

Creates stack traces on every segment and send them up in transaction traces

- Updated configuration.xsd (xsd2code was run)
  - maxStackTrace has been changed from 30 to **0** to disable capturing stack traces in transaction traces by default.   This setting was not being used prior to this change.

- Updates Segment.Finish() in Segment.cs
  - StackTrace is created in this method since all segments go through here at somepoint unless ended prematurely.
  - Changes the logic to no longer default to a static empty KeyValue<> array
  - Uses the configuration service to check if we should create stack traces.  Does this in a using since we want to release our subscription to this as soon as we are done with it.
  - Only ExternalSegmentData.Finish actually returns something, a Dictionary, so we need to support that use case in the if-then
  - Will truncate and clean up stack traces to keep under the maxcStackFrames config value - this was done while serializing the stack trace in the past.  We are doing it earlier to reduce memory pressure of large traces.
  - After truncating, the stackframe list is parsed into a list<string>.  This removes the need for special JsonConverters farther down. 

- StackTraces.cs
  - Updates the portion of the code used in transaction traces to use for loops instead of for-each loops to improve speed and efficiency.
  - Fixed a small bug in the logic for setting up the frames list
  - Cleaned up file to match our current code standards

- StackTraceTransactionTraceTests.cs 
  - Adds composite tests to ensure that we are getting the stacktrace when we need to.

- StackTracesTest.cs 
  - Updates a test address changes in parsing

### performance

Here are some performance numbers from my laptop.   They don't look too bad:

```
With StackTraces										
		# Samples	Average	Min	Max	Std. Dev.	Error %	Throughput	Received KB/sec	Sent KB/sec	Avg. Bytes
HTTP Request	4996		98	59	328	24.75		0.000%	16.64939	10.75		2.13		661.2
TOTAL		4996		98	59	328	24.75		0.000%	16.64939	10.75		2.13		661.2
										
Without StackTraces										
		# Samples	Average	Min	Max	Std. Dev.	Error %	Throughput	Received KB/sec	Sent KB/sec	Avg. Bytes
HTTP Request	4998		84	52	261	17.99		0.000%	16.65856	10.76		2.13		661.3
TOTAL		4998		84	52	261	17.99		0.000%	16.65856	10.76		2.13		661.3
										
Diff				14	7	67						

```

### Testing
Unit tests cover the basics.
I suggest manually setting up an end-to-end test as a sanity check.

### Changelog

Please remember to update our [changelog](/src/Agent/CHANGELOG.md) if applicable (or [this changelog](/src/AwsLambda/CHANGELOG.md) for Lambda agent changes),

Updated!


src/Agent/NewRelic/Agent/Core/JsonConverters/EventAttributesJsonConverter.cs
```diff
@@ -83,6 +86,15 @@ public override void WriteJson(JsonWriter writer, IEnumerable<KeyValuePair<strin
                 {
                     writer.WriteValue((byte)kvp.Value);
                 }
+                else if (kvp.Value is StackTrace)
+                {
+                    using (var configurationSubscription = new ConfigurationSubscriber())
```
JcolemanNR, line 91: I'm curious what would happen if we made a single static config subscriber.... It would be a better pattern vs creating/destroying a subscriber whenever access is needed

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core.Utils/StackTraces.cs
```diff
@@ -99,7 +99,7 @@ public static string[] ParseStackTrace(string stackTrace)
 
         public static ICollection<StackFrame> ScrubAndTruncate(StackFrame[] frames, int maxDepth)
         {
-            List<StackFrame> list = new List<StackFrame>(frames.Length);
+            List<StackFrame> list = new List<StackFrame>(Math.Min(frames.Length, maxDepth));
```
JcolemanNR, line 102: var list = ...

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
```diff
@@ -219,7 +220,33 @@ private void Finish()
         {
             var endTime = _transactionSegmentState.GetRelativeTime();
             RelativeEndTime = endTime;
-            _parameters = Data.Finish() ?? EmptyImmutableParameters;
+            _parameters = Data.Finish();
+            using (var configurationSubscription = new ConfigurationSubscriber())
```
JcolemanNR, line 224: I'm thinking each class should just instantiate one static instance, if config access is needed

JcolemanNR (changes requested):
I think we should go with one static config subscriber if it is thread safe... please test and refactor if this provides the same performance benefit

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core.Utils/StackTraces.cs
jaffinito: Fixed entire class!

JcolemanNR (approved):
Looks good, I like that you took the var transformation to the limit! 🥇

jaffinito:
Integration test failures were flickers.  The tests passed fine locally.

jaffinito:
Reran performance tests and I am satisfied that we are within an acceptable range especially given the feature is disabled by default.

PR #653, opened 2021-07-16
Add App Name and Source + NR and App configs to Environment

### LINKED ISSUE #580: Send app name and source in environment data
One of our support team's largest sources of customer tickets is application naming.  We have a separate and specific public doc for how to name your applications: https://docs.newrelic.com/docs/agents/net-agent/configuration/name-your-net-application/, which is an indication that it's somewhat complicated.

One of the things that makes this issue challenging is that neither a user of the agent nor our support team can tell where an application name is coming from (main `newrelic.config`, app-local `newrelic.config`, app.config, environment variable) without examining the agent log file at a non-default level (DEBUG level at least).  It would be helpful for everybody involved if information about where an application's name came from was always available in a convenient place.

### Feature Description
The source of an application's name should be available in the NR user interface.  An initial proposal will be to put this data in the `Environment` page for each application.  This will mean modifying this part of the agent: https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/Environment.cs  It would also mean having to modify [this code](https://github.com/newrelic/newrelic-dotnet-agent/blob/6271f650dd0e035633a2bc4ef36c272368e8b25a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs#L219) to return the source of the application name as well as the name itself.

Potential concern: will the configuration parsing be finished before the environment data is sent to NR?

### Priority
[Really Want]

tehbio:
### Description

Resolves #580 by adding `Application Names`, `Application Names Source`, `New Relic Conffig`, and `Application Config` to the `Environment` data sent up in the connect payload.

For .Net Core I leveraged the [existing code](https://github.com/newrelic/newrelic-dotnet-agent/blob/2379aace42dc1ddc7ce6bc6a1cd105b380c4f842/src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs#L58..L59) in the Agent to try to grab the application config, however this approach is not ideal as it currently only supports default (`appsettings.json` and `appsettings.{environment}.json`) files. I tried finding a better way to do this, but I believe for a complete solution we would need to be instrumenting the `ConfigurationBuilder` so we can intercept all the files it's utilizing to build the application settings.

### Testing

Included some new unit tests to validate `Application Names Source`, and I broke out the `EnvironmentTests` into framework and core versions to validate the `New Relic Config` and `Application Config` file paths.

### DIFF
--- a/tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
+++ b/tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
@@ -1816,8 +1816,9 @@ namespace NewRelic.Agent.Core.Configuration.UnitTest
             NrAssert.Multiple(
                 () => Assert.AreEqual(2, _defaultConfig.ApplicationNames.Count()),
                 () => Assert.AreEqual("MyAppName1", _defaultConfig.ApplicationNames.FirstOrDefault()),
-                () => Assert.AreEqual("MyAppName2", _defaultConfig.ApplicationNames.ElementAtOrDefault(1))
-                );
+                () => Assert.AreEqual("MyAppName2", _defaultConfig.ApplicationNames.ElementAtOrDefault(1)),
+                () => Assert.AreEqual("API", _defaultConfig.ApplicationNamesSource)
+            );
         }
 
         [Test]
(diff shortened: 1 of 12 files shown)


tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -1871,8 +1874,9 @@ public void ApplicationNamesPullsSingleNameFromIisExpressSitenameEnvironmentVari
 
             NrAssert.Multiple(
```
JcolemanNR, line 1875: Not related to your code: I'm not sure I like this assert helper... It makes it difficult to see the exact line that fails, and is more code than just a list of asserts 🤷‍♂️

jaffinito (approved):
This looks good to me.

JcolemanNR (approved):
Looks good!

nr-ahemsath:
Manual testing looks good, except in the case of an app name set via the agent API, in which case the app does get the name set by the API, but the environment data shows the name source as being the NR config file (and the app name as being the one from the config file).  I'm guessing this is because the environment data is sent up once, almost immediately when the agent first starts, before the API is called by the instrumented application.

PR #663, opened 2021-07-26
Fix gRPC license key header issue

### LINKED ISSUE #627: Grpc channel shutdown can cause `license_key is required` error message.
**Description**
When something causes the grpc channel disconnect, the agent tries to shutdown and reconnect. It also resets the metadata object which has the license key header in it. This line of code causes the issue:

https://github.com/newrelic/newrelic-dotnet-agent/blob/c75eeb0f6175c5d61a310e18cd7053001562c6ce/src/Agent/NewRelic/Agent/Core/DataTransport/DataStreamingService.cs#L626.

Even though when the agent reconnects, the MetadataHeaders will get set again. The problem is, during the connection shutdown, the agent had a bunch of streaming workers that were still streaming data with the cleared MetadataHeaders.

A potential fix is here https://github.com/newrelic/newrelic-dotnet-agent/tree/fix-license-key-header-issue.

**Expected Behavior**
Streaming worker should not use `null` value MetadataHeaders to send spans when the channel connection is restarting.

tehbio:
### Description

Resolves #627 with some updates to not clear headers.

### Testing

Played around with adding some tests for this but couldn't find a consistent way to do it, though this change is relatively low impact.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### New Features
 ### Fixes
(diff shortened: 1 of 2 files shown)


jaffinito (approved):
Looks great and thank you for updating the change log!

PR #665, opened 2021-07-30
Adding some UI to the MultiverseScanner

jaffinito:
### Description

Adds a new utility to the MultiverseScanner called ReportBuilder.
ReportBuilder takes a report.yml from the MultiverseScanner and transforms it into collection of static HTML pages.
The HTML pages display tables showing how our XML instrumentation matches up with the various versions of libraries we support.

These page are meant to be updated to [GitHub Pages](https://docs.github.com/en/pages), which is meant to display static HTML pages.

The existing multiverse_run.yml file has been rebuilt to only run on a release using the workflow `workflow_run` event.  This event is triggered when the all_solutions.yml completes its run (success or not sadly).  he first job in multiverse_run.yml  now checks to see if the all_solutions.yml was run for release or not before continuing.  The workflow pulls down the S3 artifacts for MVS, gets a copy of the agent_vesion.txt and uses that to build the HTML.

In addition, this job no longer pushes the reports.yml files to S3, instead they are included along with the HTML pages with a link to the relevant one on the version home page.

Once the HTML report is created, it is pushed automatically to the gh-pages branch.  GH then takes over and push it out to the .io page: , which should be: https://newrelic.github.io/newrelic-dotnet-agent/mvs/

Feedback is most welcome!

### Testing

You can run the ReportBuilder locally and see if the output is valid HTML

### Changelog

N/A


tehbio (approved):
Looks nice!

tests/Agent/MultiverseTesting/ReportBuilder/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "ReportBuilder": {
+      "commandName": "Project",
+      "commandLineArgs": "\"1.4.0\" \".\\reports.yml\" \".\\wikitemp\""
+    }
+  }
+}
```
JcolemanNR, line 8: missing newline

JcolemanNR (approved):
Maybe add the newline, but I wouldn't convert to a full logging library right now. Otherwise, ship it! :)

tests/Agent/MultiverseTesting/ConsoleScanner/Program.cs
```diff
@@ -18,6 +18,8 @@ namespace NewRelic.Agent.ConsoleScanner
 {
     public class Program
     {
+        private static bool _writeLogs = true;
```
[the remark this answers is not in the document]
jaffinito: Yeah, this was meant to be quick and easy.  It definitely could use a cleanup pass at some point.

tests/Agent/MultiverseTesting/ReportBuilder/Properties/launchSettings.json
jaffinito: Oddly enough, that is straight from MS.

JcolemanNR (approved):
Merge it! :)

PR #678, opened 2021-08-05
Update config docs link in default config

nr-ahemsath:
The link to our agent config documentation in the default `newrelic.config` was incorrect, this PR updates it to be current.


jaffinito (approved):
Cool!

Issue #683, opened 2021-08-17
Configured maximum stack trace depth is not always honored


JcolemanNR:
**Description**
While running the unit tests in debug (not an intended configuration), I noticed a unit test failure with how stack depth is managed. While switching to release configuration does allow the unit tests to pass, there is actually some erroneous logic in how we reduce stack traces.

For example, even the release configuration test asks for a max stack depth of 3, but is satisfied when only 2 are returned despite there being over 40 stack frames that aren't from 'NewRelic' assemblies.

**Expected Behavior**
If there are enough non 'NewRelic' stack frames, then we should return the requested maximum stack frames.

PR #684, opened 2021-08-17
Fix issue with how stack trace depth is managed

JcolemanNR:
Resolves: https://github.com/newrelic/newrelic-dotnet-agent/issues/683

Basically if any of the stack frames within the requested depth are from NewRelic, then we don't save the requested number of frames.

### Testing

Updated an existing unit test, and added one to capture an edge case where more stack frames are requested than available


tehbio (approved):
👍

PR #689, opened 2021-08-19
enable ci builds for 9x-feature-branch (#688)

JcolemanNR:
Perhaps I need to merge this in to the actual feature branch, instead of main....


JcolemanNR:
I thought it would do a fast-forward merge, but it created a competing commit with main, so the result of this PR has been destroyed with a force push

PR #690, opened 2021-08-19
Sort versions from newest to oldest

jaffinito:
Sorts the index for the Multiverse Testing Report from newest oldest to make it more readable.


nr-ahemsath (approved):
Looks legit. 👍

PR #700, opened 2021-08-26
Update default config to use DT over CAT

### LINKED ISSUE #672: Set default value of  distributedTracing.Enabled to “true” to enable DT by default
- Flip default value of  distributedTracing.Enabled to `true` to enable DT by default.
  1. Explicitly set distributedTracing.Enabled to `true` in the default newrelic.config.

- Updates existing unit tests and integration tests for this change.
  1. For DT related integration tests, remove any codes that explicitly enable DT.
  2. For CAT related integration tests, explicitly disable DT.

#### Acceptance Criterion:
- All tests must pass.
- All integration tests must pass.

tehbio:
### Description

Resolves #672 with the updated settings and necessary tests.

### Testing

DT and CAT related tests have been updated to correctly use these settings or override with their own.

### DIFF
--- a/src/Agent/Configuration/newrelic.config
+++ b/src/Agent/Configuration/newrelic.config
@@ -6,7 +6,7 @@
 	<application>
 		<name>My Application</name>
 	</application>
-	<log level="info"/>
+	<log level="info" />
 	<allowAllHeaders enabled="true" />
 	<attributes enabled="true">
 		<exclude>request.headers.cookie</exclude>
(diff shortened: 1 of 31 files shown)


nr-ahemsath (approved):
Changes look good.  All unit tests are passing for me locally.  All regular integration tests are passing for me locally.  I ran the RabbitMQ unbounded integration tests locally (since RabbitMq is the only wrapper that does automatic DT that is covered by unbounded tests) and they all passed.

👍

PR #701, opened 2021-08-27
Make maximum Span Events configurable, and use new cycle and maximums sent from server

### LINKED ISSUE #671: Make span event reservoir max reservoir size configurable
- Introduces `maximumSamplesStored` attribute with default value of 1000 for the `spanEvents` element.
- Introduces the `NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED` environment variable.
- Wires up the agent to use the `maximumSamplesStored` and `NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED`.
       1. `NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED` will override `maximumSamplesStored` if set and has value.
       2. Updates the `ReporteConfiguration` with the new `span_events.maximum_samples_stored` config like everything else. https://github.com/newrelic/newrelic-dotnet-agent/blob/acf267f986f55110b8b91dbf910dbb2bd09057a5/src/Agent/NewRelic/Agent/Core/Configuration/ReportedConfiguration.cs#L12 
       3. Upon sending the Connect request, the agent will use the value from the `maximumSamplesStored` or `NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED` for `event_harvest_config.harvest_limit.span_event_data` in the Connect payload.
       4. After the agent receives the Connect Service response, it will use the final span reservoir size and report period from the Connect Service replies payload. The field needs to be used in the Connect Service payload for this is
       
```
{
    {
         "span_event_harvest_config": {
         "report_period_ms": 60000,
         "harvest_limit": 1000
    }
  ...
}
```

Refs: https://source.datanerd.us/agents/agent-specs/blob/master/Connect-LEGACY.md#span-event-harvest-config

#### Acceptance Criterias
- NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED will override `maximumSamplesStored` if set and has value.
- Agent will use the `span_event_harvest_config` setting from Connect Service response.

tehbio:
### Description

Resolves #671 by adding a new `spanEvents.maximumSamplesStored` local setting, and implementing the new `span_event_harvest_config` values sent down by the server.

Note - somehow the new setting already got pulled into this feature branch, so I removed them from this PR to not have any conflicts.

### Testing

Two related unit tests were updated to check these configurations and validate they are being set correctly.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -28,6 +28,8 @@ namespace NewRelic.Agent.Core.Configuration
     {
         private const int DefaultSslPort = 443;
         private const int DefaultSqlStatementCacheCapacity = 1000;
+        private const int SPAN_EVENTS_MIN_SAMPLES_STORED = 1000;
+
 
         public static readonly string RawStringValue = Enum.GetName(typeof(configurationTransactionTracerRecordSql), configurationTransactionTracerRecordSql.raw);
         public static readonly string ObfuscatedStringValue = Enum.GetName(typeof(configurationTransactionTracerRecordSql), configurationTransactionTracerRecordSql.obfuscated);
@@ -882,8 +884,17 @@ namespace NewRelic.Agent.Core.Configuration
 
         public int? SamplingTarget => _serverConfiguration.SamplingTarget;
 
-        public int SpanEventsMaxSamplesStored => ServerOverrides(_serverConfiguration.SpanEventHarvestConfig?.HarvestLimit,
-           EnvironmentOverrides(_localConfiguration.spanEvents.maximumSamplesStored, "NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED").GetValueOrDefault());
+        public int SpanEventsMaxSamplesStored
+        {
+            get
+            {
+                var local = EnvironmentOverrides(_localConfiguration.spanEvents.maximumSamplesStored, "NEW_RELIC_SPAN_EVENTS_MAX_SAMPLES_STORED").GetValueOrDefault();
+
+                local = local > SPAN_EVENTS_MIN_SAMPLES_STORED ? local : SPAN_EVENTS_MIN_SAMPLES_STORED;
+
+                return ServerOverrides(_serverConfiguration.SpanEventHarvestConfig?.HarvestLimit, local);
+            }
+        }
 
         public int? SamplingTargetPeriodInSeconds => _serverConfiguration.SamplingTargetPeriodInSeconds;
 
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -28,8 +28,6 @@ namespace NewRelic.Agent.Core.Configuration
     {
         private const int DefaultSslPort = 443;
         private const int DefaultSqlStatementCacheCapacity = 1000;
-        private const int SPAN_EVENTS_MIN_SAMPLES_STORED = 1000;
-
 
         public static readonly string RawStringValue = Enum.GetName(typeof(configurationTransactionTracerRecordSql), configurationTransactionTracerRecordSql.raw);
         public static readonly string ObfuscatedStringValue = Enum.GetName(typeof(configurationTransactionTracerRecordSql), configurationTransactionTracerRecordSql.obfuscated);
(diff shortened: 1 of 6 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -888,7 +882,10 @@ private bool IsDistributedTracingEnabled()
 
         public int? SamplingTarget => _serverConfiguration.SamplingTarget;
 
-        public int SpanEventsMaxSamplesStored => ServerOverrides(_serverConfiguration.EventHarvestConfig?.SpanEventHarvestLimit(), EnvironmentOverrides(_localConfiguration.spanEvents.maximumSamplesStored, "SPAN_EVENTS_MAX_SAMPLES_STORED").GetValueOrDefault()); public int? SamplingTargetPeriodInSeconds => _serverConfiguration.SamplingTargetPeriodInSeconds;
+        public int SpanEventsMaxSamplesStored => ServerOverrides(_serverConfiguration.SpanEventHarvestConfig?.HarvestLimit,
```
vuqtran88, line 885: This part is fine, but it might be worth to make sure that we always want to send up the non-negative user specified harvest limit for span events on the Connect.

https://github.com/newrelic/newrelic-dotnet-agent/blob/2ba99ea91ac20f79af40259e485949c6b3b1aac3/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectModel.cs#L182

vuqtran88 (changes requested):
I have some comments/requests. It was weird that the `maximumSamplesStored` appeared in another PR. It could be that I mistakenly merged some of my work by trying to sync up branches while reviewing an older PR  😞.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
tehbio: Unfortunately there's no good way of accessing specifically the local configuration at this level, however there is no issue here - the `_serverConfiguration` used before Connect has a null `SpanEventHarvestConfig` so it always falls to the local config or default, and is then replaced by the server provided after Connect.

src/Agent/NewRelic/Agent/Core/Configuration/SingleEventHarvestConfig.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class SingleEventHarvestConfig
+    {
+        [JsonProperty("report_period_ms")]
+        public int? ReportPeriodMs { get; set; }
+
+        [JsonProperty("harvest_limit")]
+        public int HarvestLimit { get; set; }
+
+        public TimeSpan? HarvestCycle => ReportPeriodMs.HasValue ? TimeSpan.FromMilliseconds(ReportPeriodMs.Value) : (TimeSpan?)null;
```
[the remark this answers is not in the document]
tehbio: When I built this I copied the [existing implementation](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Configuration/EventHarvestConfig.cs#L19) thinking there was a good reason to handle these different potential server responses, but if you're confident there's no reason for them I can instead assume valid values are always present.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
vuqtran88: You are correct. However, I'm specifically curious about what happens if users accidentally set the limit to `0` or a negative number. On startup, the agent will report this limit to the server then the server will response with the final limit. The server might already have a logic to handle this number but I'm not 100% sure. It is a rare case but it might be good if we could as well protect this from happen on the agent side.

src/Agent/NewRelic/Agent/Core/Configuration/SingleEventHarvestConfig.cs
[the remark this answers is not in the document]
vuqtran88: Yup, i'm pretty confident that it will always be available.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
tehbio: Just tested and negative values are kept and sent back by the server - that definitely doesn't make sense, so I can clamp the value here as you suggested.

Is zero not a valid value for this, though? Seems like it could be valid for some use cases, but I guess it could potentially break other things...

vuqtran88: per my conversation with other agent teams, we we seem to agree to enforce the minimum of 1000 for the following reasons:
1. Making sure DT is operated with minimal desired behavior. 1000 seems adequate, and historically customers wanted more.
2. Don’t have to deal with incorrect input .ie negative numbers.

vuqtran88 (approved):
LGTM 👍

JcolemanNR (approved):
🚢 :shipit:

Issue #669, opened 2021-08-01
Removing `requestParameter` configuration setting


vuqtran88:
Considers the following:
- Removes the `requestParameter` config code usages in the agent.
- Logs out "This {oldConfiguration} is obsoleted. Please use this {newConfiguration}." message when these configs are used in the agent v9.0. This will help Support with troubleshooting.

Notes:
We don't want to remove these deprecated configs from the xml schema because it will cause configuration parsing exception when upgrade which will crash the agent.

#### Acceptance Criterion:
- Unit tests pass.
- Integration tests pass.
- Expects warning log messages when old configs are used.

PR #711, opened 2021-09-03
Disable RequestParameters config

JcolemanNR:
### Description

Disables the RequestParameters.enabled configuration option in favor of the newer attributes api that allows "request.parameters.*" to be added to an include list. When RequestParameters.enabled is specified, a warning will be logged. 

Resolves: #669 

### Testing

Unit and Integration tests were removed or updated as appropriate, unbounded integration tests pass as well.

### Changelog

Updated


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -753,10 +753,6 @@ public virtual bool CaptureRequestParameters
             get
             {
                 var localAttributeValue = false;
-                if (_localConfiguration.requestParameters.enabledSpecified)
-                {
-                    localAttributeValue = _localConfiguration.requestParameters.enabled;
-                }
                 var serverAttributeValue = _serverConfiguration.RpmConfig.CaptureParametersEnabled;
```
JcolemanNR, line 756: Do we need to disable server side config for this?

tests/Agent/UnitTests/Core.UnitTest/WireModels/ErrorTraceWireModelTests.cs
```diff
@@ -77,7 +77,8 @@ private static IConfiguration CreateMockConfiguration()
                 Mock.Arrange(() => configuration.CaptureAttributes).Returns(true);
                 Mock.Arrange(() => configuration.CaptureAttributesExcludes)
                     .Returns(new List<string>() { "identity.*", "request.headers.*", "response.headers.*" });
-                Mock.Arrange(() => configuration.CaptureRequestParameters).Returns(true);
+                Mock.Arrange(() => configuration.CaptureAttributesIncludes).Returns(new string[] { "request.parameters.*" });
```
vuqtran88, line 80: Is there a reason why we need to set this  ? Tests in this files seems to pass without this line of code.

tests/Agent/UnitTests/Core.UnitTest/WireModels/TransactionEventWireModelTests.cs
```diff
@@ -30,7 +30,8 @@ private static IConfiguration CreateMockConfiguration()
                 Mock.Arrange(() => configuration.CrossApplicationTracingEnabled).Returns(true);
                 //Mock.Arrange(() => configuration.CaptureAttributesExcludes)
                 //    .Returns(new List<string>() { "identity.*", "request.headers.*", "response.headers.*" });
-                Mock.Arrange(() => configuration.CaptureRequestParameters).Returns(true);
+                Mock.Arrange(() => configuration.CaptureAttributesIncludes).Returns(new string[] { "request.parameters.*" });
```
vuqtran88, line 33: Is there a reason why we need to set this ? Tests in this files seems to pass without this line of code.

vuqtran88 (changes requested):
Overall this PR looks good. I only some minor questions.
Removing the local `requestParameters` config indirectly removes the usage of the `capture_params` server config but this is expected.

tests/Agent/UnitTests/Core.UnitTest/WireModels/ErrorTraceWireModelTests.cs
tehbio: I think we removed this because the `CaptureRequestParameters` boolean no longer exists (and doesn't really make sense anymore since parameters being captured is no longer an all-or-nothing thing).

vuqtran88: oh ic, I misunderstood the change. The change was to replace the existing line of code with the equivalence one. The old line of code didn't seem to really do anything to the tests.

vuqtran88 (approved):
👍

JcolemanNR:
> Overall this PR looks good. I only some minor questions.
> Removing the local `requestParameters` config indirectly removes the usage of the `capture_params` server config but this is expected.

We weren't sure if we should leave the server side config in place, but we couldn't find anywhere this was exposed via the Server Side Config ui, so figured it was safe to remove

Issue #679, opened 2021-08-10
Considering upgrading Grpc and Grpc.Core packages dependencies to newer version


vuqtran88:
### Feature Description
The libgrpc_csharp_ext.x64.so used by the new relic agent is not stripped of symbols and is extremely large at 109mb. This makes the customer’s docker containers twice the size. This is fixed in the Grpc.Core nuget package v2.37.

We need to update the agent to use the Grpc and Grpc.Core nuget packages version v2.37+.

We need to make sure the upgrade will work in both Windows, Linux.

We need to make sure the upgrade will work with different versions of .NET Core and .NET Framework.

PR #721, opened 2021-09-15
Update gRPC libraries to v2.40.0

JcolemanNR:
### Description

The version of gRPC that we are currently using in the agent has extra debug symbols included which greatly increases the size of our deliverable. By upgrading to a newer version of gRPC we can reduce the size of our install files.

Resolves: #679

### Testing

Unit/Integration/Unbounded tests all pass

### Changelog

Updated


jaffinito (approved):
Paired on the changes.  Looks great!

PR #723, opened 2021-09-15
Better unbounded service security

nr-ahemsath:
### Description

The changes in this PR help enable better security for the Docker-hosted external services needed by the unbounded integration tests.  Summary of changes:
1. Enable the use of a username and password for some services that were previously set up to use no authentication at all (e.g. redis)
2. Allow (almost) all services to have credentials specified from the environment in `docker-compose.yml`, although default values are still specified to make it easy to use these containers for testing on a local machine.
3. Remove unnecessary port mappings from the `docker-compose.yml` file.
4. In some cases, the version of the container image for the service was updated to a more recent version.

### Testing

I've verified that all unbounded integration tests still pass for me locally using this version of the services setup and the `example-secrets.json` file with appropriate modifications (e.g. NR license key).

### Changelog

N/A

### DIFF
--- a/tests/Agent/IntegrationTests/Shared/RabbitMqConfiguration.cs
+++ b/tests/Agent/IntegrationTests/Shared/RabbitMqConfiguration.cs
@@ -8,6 +8,8 @@ namespace NewRelic.Agent.IntegrationTests.Shared
     public class RabbitMqConfiguration
     {
         private static string _rabbitMqServerIp;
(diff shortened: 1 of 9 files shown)


JcolemanNR (approved):
Looks good!

JcolemanNR:
I think we should merge this, but I'm curious about how we re-enable the unbounded integrations tests in GHA?

PR #804, opened 2021-11-10
Fix .Net 6 HttpClient header thread safety issue

### LINKED ISSUE #803: [.NET 6] Thread safety issue in access to HTTP headers collection in HttpClient on .Net 6
We've been informed by Microsoft that New Relic is causing exceptions in some customer applications due to our access to the HTTP headers collection in HttpClient: https://github.com/microsoft/reverse-proxy/issues/1337.

We will want to fix this one way or another before announcing support for .NET 6 in the .NET agent.

tehbio:
## Description

Resolves #803 by synchronizing the Agent's access of the HttpClient headers.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
@@ -97,9 +97,8 @@ namespace NewRelic.Providers.Wrapper.HttpClient
             }
             else
             {
-                // With .Net 6 the HttpClient headers are not thread safe, so we need our continuation to happen synchronously.
+                // With .Net 6 the HttpClient headers are especially not thread safe (and weren't guaranteed to be before), so we need our continuation to happen synchronously.
                 // This does mean any work done by TryProcessResponse will be included in the instrumented method segment.
-
                 return Delegates.GetAsyncDelegateFor<Task<HttpResponseMessage>>(agent, segment, true, InvokeTryProcessResponse, TaskContinuationOptions.ExecuteSynchronously);
 
                 void InvokeTryProcessResponse(Task<HttpResponseMessage> httpResponseMessage)
@@ -173,7 +172,7 @@ namespace NewRelic.Providers.Wrapper.HttpClient
                 externalSegmentData.SetHttpStatusCode((int)httpStatusCode);
 
                 // Everything after this is for CAT, so bail if we're not using it
-                if (!agent.Configuration.CrossApplicationTracingEnabled)
+                if (agent.Configuration.DistributedTracingEnabled || !agent.Configuration.CrossApplicationTracingEnabled)
                     return;
 
                 var flattenedHeaders = response.Headers?.Select(Flatten);
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
@@ -97,13 +97,10 @@ namespace NewRelic.Providers.Wrapper.HttpClient
             }
             else
             {
-                // 1.  Since this finishes on a background thread, it is possible it will race the end of
-                //     the transaction. Using holdTransactionOpen = true to prevent the transaction from ending early.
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
```diff
@@ -175,11 +172,13 @@ private static void TryProcessResponse(IAgent agent, HttpResponseMessage respons
                 var httpStatusCode = response.StatusCode;
                 externalSegmentData.SetHttpStatusCode((int)httpStatusCode);
 
-                var headers = response.Headers?.ToList();
-                if (headers == null)
+                // Everything after this is for CAT, so bail if we're not using it
+                if (!agent.Configuration.CrossApplicationTracingEnabled)
```
JcolemanNR, line 176: Do we need to check that DT is also not enabled? I think both can be enabled at the same time

tehbio: DT doesn't rely on response headers. 🙂

JcolemanNR: Yeah, my initial comment had the logic wrong so I changed it. 🧇 

The concern was that CrossApplicationTracingEnabled may be set to true in the actual configuration at the same time as DistributedTracing/8T. I looked at the Configuration class and it does look like it checks that the CrossApplicationTracer is actually initialized before returning true here.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
```diff
@@ -97,13 +97,10 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
             else
             {
-                // 1.  Since this finishes on a background thread, it is possible it will race the end of
-                //     the transaction. Using holdTransactionOpen = true to prevent the transaction from ending early.
-                // 2.  Do not want to post to the sync context as this library is commonly used with the
-                //     blocking TPL pattern of .Result or .Wait(). Posting to the sync context will result
-                //     in recording time waiting for the current unit of work on the sync context to finish.
-                //     This overload GetAsyncDelegateFor does not use the synchronization context's task scheduler.
-                return Delegates.GetAsyncDelegateFor<Task<HttpResponseMessage>>(agent, segment, true, InvokeTryProcessResponse);
+                // With .Net 6 the HttpClient headers are not thread safe, so we need our continuation to happen synchronously.
```
MihaZupan, line 100: Note that the headers were never thread safe, the race conditions are just easier to hit on 6.0.
Even on 5.0, you may end up corrupting the headers or throwing during enumeration.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
vuqtran88: @JcolemanNR  good point, i think the right check would be the following, because the current behavior is that DT setting will trump CAT setting.

`if (agent.Configuration.DistributedTracingEnabled  ||  !agent.Configuration.CrossApplicationTracingEnabled)`

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
tehbio: Good to know. We couldn't reproduce this race condition on 5.0, but this fix does ensure synchronization across all versions. I've updated the comment with a bit more detail.

MihaZupan: It would only happen in very specific situations - the destination would have to respond with an invalid header that would then be removed during enumeration. You could hit a race there where one thread was removing a value from a dictionary while another attempted to read from it.

vuqtran88 (approved):
LGTM. Just a suggestion, we might want to call out the timing change in the CHANGELOG.md  for transparency.

Issue #808, opened 2021-11-15
Tests stored procedure calls 


vuqtran88:
Tests stored procedure calls in CosmosDb client. Calls need to be tested:
`Scripts.CreateStoredProcedureAsync()`
`Scripts.ExecuteStoredProcedureAsync()`

PR #810, opened 2021-11-15
Cosmosdb stored procedure tests

vuqtran88:
Implemetation for https://github.com/newrelic/newrelic-dotnet-agent/issues/808

At level, this PR tests:
- Creates a simple stored procedure and stores it in the test cosmosdb database. CosmosDB stored procedures are basically javascript scripts that can be executed from the server side. The created test stored  procedure doesn’t do database related things much. All it does is returning “Hello, World” string to the client call.

- Executes the stored procedure.


nr-ahemsath (approved):
Looks good, tests pass locally 💯

PR #814, opened 2021-11-19
Add support for disabling Server-Side Configuration from the Agent.

browe:
* Adds the NEW_RELIC_DISABLE_SERVER_CONFIG environment variable to
allow local configuration override to ignore 'agent_config'.

## Description

Adds the ability to disable Server-Side Configuration of the .NET APM Agent through local configuration of the agent. Setting `NEW_RELIC_DISABLE_SERVER_CONFIG = true` will cause the agent to ignore any server-side configuration pushed to it.

Originally this was implemented by simply removing `agent_config` from the dictionary returned by the server connect in `ConnectionHandler`, which was very straightforward but not very testable and seemed a bit leaky.

Instead, this change adds a `ServerConfigurationCreationOptions` parameter to the factory methods in `ServerConfiguration`. Currently only the `IgnoreServerConfiguration` option is supported which, when set, will ensure we overwrite whatever server-side agent configuration we got from the server in `RpmConfig` with unconfigured defaults instead.

This change is motivated by compliance requirements for disabling server-side control of agent configuration.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


javierfigueroa:
Great job @browe ! ;)

vuqtran88:
Hi @browe

Thanks for submitting this PR. Our of curiosity, could you expand on why this is necessary ? You can contact [New Relic Support](https://docs.newrelic.com/docs/apm/agents/manage-apm-agents/configuration/server-side-agent-configuration/#disable-ssc) to disable server side configuration. The main reason why I am asking this is our cross agent spec has an agreement that server-side config must always override any agent configurations. This change would violate this rule.

browe:
Hi @vuqtran88 ,

The change is motivated by a compliance requirement where the trust level of the application is actually higher than the New Relic SaaS offering, and the regulating body will not accept New Relic if it is able to assert control over the agent configuration. For example, there is a requirement to ensure SQL traces are never transmitted from the application to the New Relic servers and the agents are configured to set `slowSql = false` but then an actor in the untrusted boundary of the New Relic servers could push Server-Side Configuration that re-enables it.

I can give you more context and detail about the motivation including the particulars of the certification requirements if you contact me directly at `bradley.rowe@citrix.com` . There is a thread open with New Relic that I can rope you into.

vuqtran88:
Hi @browe, I'm sorry for the late response as I just got back from a week of vacation. From what you described, it is reasonable use case, but I’m not sure disabling SSC from the agent using an environment variable is good enough because the New Relic server controls the agent behaviors from more than just the `agent_config` field in the server reply. The `agent_config` field in the reply contains configurations that can be configured from the UI. The server also controls the agent behavior from the fields as well .ie:

```
...
...
 collect_error_events
 collect_analytics_events
 collect_errors
 collect_traces
 event_harvest_config
 ...
 ... 
``` 

In other words, if you want to completely control the agent, ignoring the `agent_config` field isn’t enough. Ignoring other fields in the server reply isn’t recommended as well because it is a contract between all agents with the New Relic server.  Ignoring server request might lead to unexpected behavior, ie: the NR server want to limit how event rate agent can send up to NR to protect NR from handling large amount of data.
 
While disabling all sever configs isn't a good idea for us, I wonder disabling certain server configs is a good enough solution for you.

Thanks,

nr-ahemsath:
Our support team had a good idea if we do decide to implement this: add an property to the environment payload so that you can see in the UI if this feature has been enabled.

browe:
Hi @vuqtran88, no problem, I was out for much of last week as well. Hope you had a good vacation. Thanks for the feedback and pointing out these other settings.

Compliance requirements tend to take a stricter view on the content of the data that flows **out** of the agent, and that's why there are already quite a few configuration options in the agents around that -- not to mention High Security Mode. The extra use case this change addresses is if the compliance regime also does not trust the New Relic servers to meddle with that configuration.

The `collect_*` properties do not allow the server to *enable* additional data exports that were otherwise (purposefully) disabled in the local agent configuration. This is not concerning because they cannot be used to undo data restrictions. For `event_harvest_config`, `span_event_harvest_config`, et. al. - those set a rate at which data may be collected but not what data is collected so they are also not concerning.

@nr-ahemsath , thanks, I'll look into adding that to the PR.

src/Agent/NewRelic/Agent/Core/Configuration/ServerConfiguration.cs
```diff
@@ -332,4 +342,15 @@ public void OnError(StreamingContext context, ErrorContext errorContext)
             Log.ErrorFormat("Json serializer context path: {0}. Error message: {1}", errorContext.Path, errorContext.Error.Message);
         }
     }
+
+    /// <summary>
+    /// Provides options controlling the creation of <see cref="ServerConfiguration"/> instances.
+    /// </summary>
+    public class ServerConfigurationCreationOptions
```
vuqtran88, line 349: I wonder if it is necessary to create a class for this. I could be as simple as a bool.

vuqtran88 (changes requested):

src/Agent/NewRelic/Agent/Core/Configuration/ServerConfiguration.cs
browe: Debated that myself - this could be used to support more options in the future without needing to tack on more flag params and is more explicit in the code since you need to spell out the param name where as `call(param1, true)` lacks a certain amount of context unless you use named params for the bool. However, if this will be the only param for a long while an extra bool would suffice.

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -239,6 +239,20 @@ public void TransactionEventsMaxSamplesOf0ShouldDisableTransactionEvents()
...
+        public void DisableServerConfigIsFalseByDefault()
+        {
+            Assert.IsFalse(_defaultConfig.DisableServerConfiguration);
+        }
+
+        [TestCase("true", ExpectedResult = true)]
+        [TestCase("false", ExpectedResult = false)]
+        public bool DisableServerConfigSetFromEnvironment(string environment)
+        {
+            Mock.Arrange(() => _environment.GetEnvironmentVariable("NEW_RELIC_DISABLE_SERVER_CONFIG")).Returns(environment);
```
vuqtran88, line 252: Could you change the env variable name to `NEW_RELIC_IGNORE_SERVER_SIDE_CONFIG` to better reflect what is going on ?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -378,6 +378,21 @@ public string ProcessHostDisplayName
             }
         }
 
+        private bool? _disableServerConfiguration;
```
vuqtran88, line 381: we probably want to call it `_ignoreServerSideConfiguration` instead.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -378,6 +378,21 @@ public string ProcessHostDisplayName
...
+
+        public bool DisableServerConfiguration
```
vuqtran88, line 383: we probably want to call it `IgnoreServerSideConfiguration` instead.

src/Agent/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] changes
 ### New Features
+* Add ability to disable agent support for Server-Configuration with `NEW_RELIC_DISABLE_SERVER_CONFIG` environment variable. ([#814](https://github.com/newrelic/newrelic-dotnet-agent/pull/814))
```
vuqtran88, line 9: we probably want to call it `NEW_RELIC_IGNORE_SERVER_SIDE_CONFIG` instead.

vuqtran88 (changes requested):
Hi @browe, I took another pass reviewing your PR. I have some additional changes request.  Other than that, it looks good.

browe:
> @browe I wonder if you need any help making the requested change. If you prefer, I could help making the change for you.

@vuqtran88, I started working on them today. Testing a new build with the changes.

browe:
@vuqtran88 - Ready for review. The installer CI test is consistently failing server-side due to a problem loading the code signing cert and is blocking the rest from running.
```
certutil.exe -f -user -p <passphrase> -importPFX D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\newrelic_code_sign_cert.pfx NoRoot
Expected no more than 1 args, received 2
CertUtil: Too many arguments
```

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/Environment.cs
```diff
@@ -76,6 +76,8 @@ public Environment(ISystemInfo systemInfo, IProcessStatic processStatic, IConfig
                 if (!String.IsNullOrEmpty(configurationService.Configuration.AppSettingsConfigFilePath))
                     AddVariable("Application Config", () => configurationService.Configuration.AppSettingsConfigFilePath);
 
+                AddVariable("Server-Config Ignored", () => configurationService.Configuration.IgnoreServerSideConfiguration);
```
vuqtran88, line 79: Hi @browe
I'm sorry for the back and forth. I'm not sure whether this is the right place to report this config at this moment, so it would be great if you could remove this logic.

After you remove this logic, I will approve this PR and handle the rest.

vuqtran88 (changes requested):

vuqtran88:
> @vuqtran88 - Ready for review. The installer CI test is consistently failing server-side due to a problem loading the code signing cert and is blocking the rest from running.
> 
> ```
> certutil.exe -f -user -p <passphrase> -importPFX D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\newrelic_code_sign_cert.pfx NoRoot
> Expected no more than 1 args, received 2
> CertUtil: Too many arguments
> ```

@browe , it is expected as the CI ran from your fork which doesn't have our secrets. While I'm looking for a way, could you make the change that I just suggested ? Thanks.

vuqtran88 (approved):
Approved. There are a few items to follow up internally such as:
1. Creating the equivalent local config setting.
2. Report the new config setting in the Connect Request payload so that the backend knows the agent ignores SSC.

browe:
Thanks @vuqtran88 , FYI I don't have permissions to merge even after approve and it looks like a conflict was recently introduced on the CHANGELOG. Would you like me to resolve?

vuqtran88:
> Thanks @vuqtran88 , FYI I don't have permissions to merge even after approve and it looks like a conflict was recently introduced on the CHANGELOG. Would you like me to resolve?

Hi @browe, I resolved the conflict. Thank you so much for your contribution. We will take it from there.

PR #835, opened 2021-12-06
Guard against agent config file schema validation errors preventing agent startup; unify framework/core behavior

nr-ahemsath:
## Description

A recent support ticket ([NR internal-only link](https://newrelic.atlassian.net/browse/GTSE-12296)) had to do with an exception being thrown during `newrelic.config` schema validation.  They were using the .NET Framework agent and were getting an error like this:

```
NewRelic  ERROR: [pid: 1748, tid: 10] There was an error initializing the agent: NewRelic.Agent.Core.Config.ConfigurationLoaderException: Unable to find the New Relic Agent configuration file C:\ProgramData\New Relic\.NET Agent\newrelic.config ---> System.IO.FileNotFoundException: Could not load file or assembly 'NewRelic.Agent.Core.resources, Version=9.0.0.0, Culture=en-US, PublicKeyToken=06552fced0b33d87' or one of its dependencies. The system cannot find the file specified. ---> System.IO.FileNotFoundException: Could not load file or assembly 'NewRelic.Agent.Core.resources' or one of its dependencies. The system cannot find the file specified. ---> System.IO.FileNotFoundException: Could not load file or assembly 'NewRelic.Agent.Core.resources' or one of its dependencies. The system cannot find the file specified.
```

The root cause of why this customer was having the problem was never found.  However, in looking at the code for reading the config schema .xsd, we found that the framework agent and core agent had different behaviors: the framework agent was depending on reading the schema as an embedded resource from one of the agent's DLLs, whereas the core agent just read the file from disk.  Since the `newrelic.xsd` file is present for both agents, we decided to just use the core agent behavior for both.

**Additionally, and perhaps more importantly**, we put exception handling around the schema validation so that any error in this process will not prevent agent startup.

Unit tests pass locally.  Unit tests were refactored to no longer depend on the config schema being an embedded resource in the agent's core DLL, and the embedded resource was removed.  Added a new unit test to cover the expected behavior if the config schema file is missing or empty.
 
I don't think we have or need any integration tests around this, nor is performance testing warranted.  I also don't think a changelog entry is needed since this change should be completely transparent to users.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 New Relic, Inc. All rights reserved.
+    // Copyright 2020 New Relic, Inc. All rights reserved.
```
jaffinito, line 1: Looks like a tab snuck in

jaffinito (approved):

jaffinito (approved):
👍

tehbio (approved):
Good change to unify a needless difference. 👍

PR #855, opened 2021-12-15
Report total physical system memory on Linux

### LINKED ISSUE #36: Linux systems do not report total physical memory on agent connect
On agent connect the following `utilization` hash (example below) is included in the connect payload. On linux systems `total_ram_mib` is always null. This is because the code that pulls this information only works on Windows.

See:
* https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Utilities/SystemInfo.cs#L29-L46
* https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Utilities/SystemInfo.cs#L123-L124

I have not reviewed this discussion very closely, yet, though it appears there is an open discussion around an API to retrieve this data: https://github.com/dotnet/runtime/issues/22948. To my knowledge, there is nothing on `System.Diagnostics.Process` that gets us physical memory, though this should be confirmed.

However, even if there was an .NET API to retrieve this information, containerized environments may cause us additional problems.

Example utilization hash:
```json
"utilization": {
    "metadata_version": 5,
    "logical_processors": 1,
    "total_ram_mib": null,
    "hostname": "...",
    "full_hostname": "...",
    "ip_address": [
        "..."
    ],
    "boot_id": "...",
    "vendors": {
        "azure": {
            "location": "...",
            "name": "...",
            "vmId": "...",
            "vmSize": "Standard_D1_v2"
        },
        "docker": {
            "id": "..."
        },
        "kubernetes": {
            "kubernetes_service_host": "..."
        }
    }
},
```

nr-ahemsath:
## Description

Resolves #36.

Total system memory will now be reported by .NET core agents running on Linux systems.  The agent reads `/proc/meminfo` and pulls out the `MemTotal` value.

Regarding testing: I considered adding a unit test for this, but the only real logic is the regex parsing of the `MemTotal: 12345 kB` line, and so a unit test using mock/fake `/proc/meminfo` data would just be testing that the regex works or doesn't.  I can still add a unit test like this if desired; some refactoring would be required to make the method in question testable.  We don't have any existing integration tests for this feature of the agent, and an integration test seems too heavyweight for this.

I did test this change manually on Ubuntu 20.04 and verified that the total physical memory was being sent in the utilization payload.

# Author Checklist
- [X] Unit tests, Integration tests, and Unbounded tests completed - see above statement
- [X] Performance testing completed with satisfactory results (if required) - N/A

# Reviewer Checklist
- [x] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Utilities/SystemInfo.cs
+++ b/src/Agent/NewRelic/Agent/Core/Utilities/SystemInfo.cs
@@ -8,6 +8,7 @@ using System;
 using System.ComponentModel;
 using System.IO;
 using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
 
 namespace NewRelic.Agent.Core.Utilities
 {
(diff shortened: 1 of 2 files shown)


tehbio (approved):
Looks good! Don't forget to update the changelog.

src/Agent/NewRelic/Agent/Core/Utilities/SystemInfo.cs
```diff
@@ -27,20 +28,49 @@ public SystemInfo(IDnsStatic dnsStatic)
...
                 {
-                    int err = Marshal.GetLastWin32Error();
-                    throw new Win32Exception(err);
+                    var memInfo = File.ReadAllText("/proc/meminfo");
+                    var memTotalRegex = new Regex(@"MemTotal\:\s*(\d+)\ kB");
+                    var match = memTotalRegex.Match(memInfo);
+                    if (match.Success)
+                    {
+                        // Despite the text of meminfo showing everything as 'kB' which should be
+                        // 'kilobytes', i.e. 1000 bytes, in fact it is reporting 'kibibytes' (KiB, 1024 bytes)
```
jaffinito, line 45: Great comment!

jaffinito (approved):

PR #856, opened 2021-12-15
NServiceBus updates, clean up and working pass

jaffinito:
## Description

- Split out the various calls to the VisibiltyBypasser from the NSB wrappers into their own functions
- Updated tests to account for names now appearing
- Added support for noticing errors from the LoadHandlersConnectorWrapper with test
- Minor formatting and naming changes

# Author Checklist

# Reviewer Checklist
- [x] Perform code review
- [x] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NServiceBus/LoadHandlersConnectorWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NServiceBus/LoadHandlersConnectorWrapper.cs
@@ -29,15 +29,12 @@ namespace NewRelic.Providers.Wrapper.NServiceBus
             var incomingLogicalMessageContext = instrumentedMethodCall.MethodCall.MethodArguments[0];
 
             var message = NServiceBusHelpers.GetMessageFromIncomingLogicalMessageContext(incomingLogicalMessageContext);
-
             if (message == null)
             {
                 throw new NullReferenceException("logicalMessage");
             }
 
-            var headers = NServiceBusHelpers.GetHeadersFromIncomingLogicalMessageContext(incomingLogicalMessageContext);
-
-            var queueName = NServiceBusHelpers.TryGetQueueName(message);
+            var queueName = NServiceBusHelpers.TryGetQueueNameLoadHandlersConnector(message);
 
             //If the transaction does not exist.
             if (!transaction.IsValid)
@@ -53,6 +50,7 @@ namespace NewRelic.Providers.Wrapper.NServiceBus
 
             var segment = transaction.StartMessageBrokerSegment(instrumentedMethodCall.MethodCall, MessageBrokerDestinationType.Queue, MessageBrokerAction.Consume, BrokerVendorName, queueName);
 
+            var headers = NServiceBusHelpers.GetHeadersFromIncomingLogicalMessageContext(incomingLogicalMessageContext);
             NServiceBusHelpers.ProcessHeaders(headers, agent);
 
             void OnComplete(Task task)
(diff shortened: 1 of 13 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NServiceBus/LoadHandlersConnectorWrapper.cs
```diff
@@ -61,6 +59,10 @@ void OnComplete(Task task)
                 {
                     transaction.End();
                 }
+                else if (task?.Status == TaskStatus.Faulted)
+                {
+                    transaction.NoticeError(task.Exception);
```
JcolemanNR, line 64: Do we still want to end the transaction after noticing the error, or does that happen somewhere else? 

I reviewed some of the other handlers and it looks like the default delegates, in general, set segment.ErrorData instead of calling NoticeError. I'm wondering if calling NoticeError is correct?

tests/Agent/IntegrationTests/UnboundedIntegrationTests/NServiceBus/NsbAsyncCmdHandlerTests.cs
```diff
@@ -13,12 +13,12 @@
 
 namespace NewRelic.Agent.UnboundedIntegrationTests.NServiceBus
 {
-    public abstract class NsbAsyncCmdHandlerTests<TFixture> : NewRelicIntegrationTest<TFixture>
+    public abstract class NsbAsyncCmdHandlerTestsBase<TFixture> : NewRelicIntegrationTest<TFixture>
```
JcolemanNR, line 16: Let the naming wars begin! :) For posterity, I was trying to reduce the length of the filenames/folders written by the learning transport, but this is not one of the typenames that ended up being written to disk.

I agree the way you have it in this PR is now consistent with the rest of our tests.

JcolemanNR (approved):
I have a question about how instrumentation should handle errors in general, looks good though!

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NServiceBus/LoadHandlersConnectorWrapper.cs
jaffinito: Nice find and good question!  I don't think it will end on its own.  I copied how the ReceiveMessageWrapper captured the error.  Let me take a look and see what other wrappers are doing.

jaffinito: Okay. 

ending the transaction: looking at the GetDelegateFor method, it will call all three actions (onSuccess, onFailure, onComplete) in that order.  onComplete is always called if the Action is not null. Since we handle ending the the onComplete for this wrapper, we are good.

NoticeError vs ErrorData:  I checked out a few other wrappers with RabbitMQ being the closest and it uses the same pattern - NoticeError.  I saw a few others that did things a bit different, but this seems to be a pattern we are using in places.

Thoughts?

JcolemanNR: Thanks for the information. This satisfies my curiosity.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/NServiceBus/NsbAsyncCmdHandlerTests.cs
jaffinito: Yeah, I was trying to keep things in line the existing name.  I agree that it is probably time for another re-org of the names.

PR #858, opened 2021-12-15
NSB: Fix issue with Handler transactions being ended by GC

JcolemanNR:
## Description

After reviewing some of the CI failures, the throwingHandler tests were relying on GC to close the transactions.

This updates the Handler wrapper to close the segment and transaction if the task is cancelled/failed/complete. I also added a check to all the handler tests to make sure we aren't relying on GC to close transactions.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NServiceBus/LoadHandlersConnectorWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NServiceBus/LoadHandlersConnectorWrapper.cs
@@ -55,13 +55,22 @@ namespace NewRelic.Providers.Wrapper.NServiceBus
 
             void OnComplete(Task task)
             {
-                if (task?.Status == TaskStatus.RanToCompletion)
+                if (task == null)
                 {
-                    transaction.End();
+                    return;
                 }
-                else if (task?.Status == TaskStatus.Faulted)
+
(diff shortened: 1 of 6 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NServiceBus/LoadHandlersConnectorWrapper.cs
```diff
@@ -55,14 +55,23 @@ public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
 
             void OnComplete(Task task)
             {
-                if (task?.Status == TaskStatus.RanToCompletion)
+                if (task == null)
                 {
-                    transaction.End();
```
jaffinito, line 60: Do we want to end the segment to?

JcolemanNR: I'm not sure when/how OnComplete is called with a null Task, but doing nothing if it is null was copied from some default delegate implementations. Unless you're sure we need to end the segment, then I think we should leave it out here

jaffinito: nah, that makes sense.

jaffinito (approved):
👍

PR #862, opened 2021-12-22
Use pipefail option in Linux package deploy scripts

### LINKED ISSUE #633: Deploy processes must fail the GHA deploy job if they have errors
See #605 for initial context.  We have observed that at least one deploy process (that of deploying the Linux .rpm/.deb install packages for the agent to our APT and YUM repositories) could have an error internally without causing the process to a) stop execution immediately before causing further issues downstream and b) fail the GHA job so that the team would be aware of the issue.

A few initial thoughts:
1. Make sure shell scripts are run with the `set -e` option so they halt execution on any error (unless this would leave things in a bad state, in which case there are ways to trap errors and handle them appropriately in the script).
2. In the particular case of the Linux package deploy, there are at least three levels of script invocation: `deploy.bash` invokes `deploy_packages.bash`, which invokes a number of additional sub-scripts to do its work.  Need to do an experiment to see if an error in a subscript will bubble up as expected.
3. Need to verify what needs to happen in a GHA run to make sure the job halts on any script error.

nr-ahemsath:
## Description

Fixes #633.

Testing purposefully causing an error in gpg signing demonstrated that the reason such an error was being masked was because, in `repoman-rebuild.bash`, the output from the `rebuild_apt` subroutine (where the signing actually happens) is piped to a `sed` command for pretty-printing purposes.  Unfortunately, by default, `set -e` (the bash option which causes a script to exit immediately if any error happens) only considers errors in the *last* command in a pipeline.  However, by additionally setting the `pipefail` option, we get the behavior we want - an error in any pipeline step will cause the script to exit with an error code.  I tested to make sure that errors generated in nested scripts bubble up to the top level, and ultimately cause the `docker-compose run deploy_packages` command to have a non-zero exit code, which will cause the GHA deploy action to fail as well.

I added the `pipefail` option to all of the scripts which are relevant to our use of these scripts (i.e. I didn't consider the `jenkins-*` scripts) and which have pipelines.

Other testing of how errors in various bash constructs (for/while/case statements, for example) are handled by the `set -e` option demonstrated that we get the expected behavior already. 

To test this, you can modify the `gpg1` command in `repoman-rebuild.bash` to have some error (e.g. change one of the arguments to something nonsensical).  Test run the deploy process on current main and notice that the ultimate exit code from the `docker-compose run` command is 0.  Do the same test on this branch and see that the exit code is nonzero.

Additionally, this PR modifies the way this containerized process works to eliminate sharing of files between the host and container - all assets required are copied into the container image at container build time.  See my comment in the linked issue for more context.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


tehbio (approved):
Nice fix!

PR #873, opened 2022-01-12
CHANGELOG updates

nr-ahemsath:
Two changelog updates:
1. Added shoutout to community contributor for the unreleased database segment naming feature.
2. Retroactively added fix for config file schema validation guard code to 9.3 fixes.


JcolemanNR (approved):
Looks good!

PR #894, opened 2022-01-26
Log metrics

jaffinito:
## Description

Adds support for instrumenting log4net messages to capture the number of lines and the size of messages, at both the individual severity level and in total.

Work:
- Instrumentation point is the second to last method called prior to the message reach the appenders.  This method is called once vs. the other option being called inside a foreach.  The method chosen should be more efficient.
- Method will also work for log forwarding.
- Uses bypasser to capture the properties on LogEvent
- I chose to hang the collections for the lines and sizes off of the AgentHealthReporter since it has all the tooling to store and harvest the metrics.  I didn't want to role a new service just for these metrics.  Eventually there will be supportability metrics living here as well.
- I am using the experimental API instead of the main one since it is much easier to extend and doesn't require changes to the main customer facing APIs.

Testing:
- Added unit tests to the AgentHealthReported for the collections
- Added a new set of integration tests using the consoleMF app that covers adding metrics and counts.
- Integration tests are setup to test FW and Core.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


tests/Agent/IntegrationTests/IntegrationTests/LogInstrumentation/Log4netMetricsTests.cs
```diff
@@ -0,0 +1,138 @@
...
+
+
+using System;
+using System.Collections.Generic;
+using MultiFunctionApplicationHelpers;
+using NewRelic.Agent.IntegrationTestHelpers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NewRelic.Agent.IntegrationTests.LogInstrumentation
```
nr-ahemsath, line 12: So, the instrumentation wrapper project is called "logging", but we already have an integration testing namespace called "Logging", which tests the agent's own logging features.  Personally, it would make more sense for us to call this new test namespace "Logging" since it will match the instrumentation wrapper project name, and rename the other namespace "AgentLogs" or something like that.  Thoughts?

jaffinito: Sounds fine to me,

JcolemanNR (approved):
Looks good enough to merge in to the feature branch

PR #900, opened 2022-01-28
Uses log4net 2.0.14

vuqtran88:
Log4net 2.0.8 has a security vulnerability. Updates the agent to use the later version even though our usage of log4net isn't impacted by this security thread.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Core.csproj
+++ b/src/Agent/NewRelic/Agent/Core/Core.csproj
@@ -33,7 +33,7 @@
     </PackageReference>
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
     <PackageReference Include="ILRepack" Version="2.0.16" />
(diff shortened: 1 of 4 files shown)


nr-ahemsath:
There is one integration test that appears to have been broken by this change: https://github.com/newrelic/newrelic-dotnet-agent/blob/821d16a7a05d104a39843151bf6f7d97dfd42c3a/tests/Agent/IntegrationTests/IntegrationTests/AgentFeatures/InstrumentationLoaderTests.cs#L62 

It is consistently failing for me locally as well as in CI.  Not sure what the issue is.

nr-ahemsath (changes requested):
As I mentioned in a comment, the `InstrumentationLoaderTestsCore` integration test appears to have been broken by this change.  Otherwise I think the change is good.  If we can get that test to pass again we should be good to go.

PR #922, opened 2022-02-02
Remove unneeded package references from WebApiAsync test app

nr-ahemsath:
Cleaning up unnecessary package references from one of our test application projects.


jaffinito (approved):

JcolemanNR:
Hmm.. was this causing issues? I mean, I'm all for cleaning up references, but I'm curious about the tale of how this ended up under the microscope :)

JcolemanNR (approved):

nr-ahemsath:
I should have done a `git clean -dfx` before testing these changes locally; if I had, I would have caught the fact that the test app was completely broken by the changes.  Closing this PR.

PR #933, opened 2022-02-04
Sets up LogEventAggregator to capture logs from inside and outside transactions

### LINKED ISSUE #908: Add LogEventAggregator to TransactionTransformer
### Feature Description

----

**IMPORTANT**
Until PR #894 is merged, base branches off of `log-metrics` branch.
Merge any PRs into `logsisnt-feature-branch` branch

----

- In TransactionTransformer
  - Add [LogEventAggregator ](https://github.com/newrelic/newrelic-dotnet-agent/blob/06a3a22acd1f3b4490f263b581e357d4a8ff2632/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs#L56)along with other aggregators
  - Add [PrioritizeAndCollectLogEvents](https://github.com/newrelic/newrelic-dotnet-agent/blob/06a3a22acd1f3b4490f263b581e357d4a8ff2632/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs#L476-L490)(ITransaction) to capture and adjust the priority of the logmessage before sending to aggregator
  - Add [call to PrioritizeAndCollectLogEvents ](https://github.com/newrelic/newrelic-dotnet-agent/blob/06a3a22acd1f3b4490f263b581e357d4a8ff2632/src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs#L94)in first Transform immediately after ComputeSampled. We need the final priority. Takes ITransaction, no Immutable since we want to keep the log data around for very log
- Add unit test is possible/logical

Acceptance Criteria

- Transformer still works
- New unit tests pass

jaffinito:
## Description

- Agent/IAgentExperimental now holds a reference to the LogEventAggregator
- Moves RecordLogMessage to Agent/IAgentExperimental to enable capturing logs inside and outside of transactions
- New record log message checks for presence of a transaction and directs the log event to either the Transaction.LogEvents or to the LogEventAggregator directly.
- Transaction now only has a LogEvents ConcurrentList that will batch up transaction  log events until transformation
- Updates wrapper to call RecordLogMessage from Agent and remove the transaction check
- TransactionTransformer now hold a reference to the LogEventAggregator
- TransactionTransformer has PrioritizeAndCollectLogEvents method that captures, adjusts, and sends out the Priority of the log events to the LogEventAggregator.  Uses current sampling algorithm found in engineer brief.
- Adds new unit test for TransactionTransformer andAgentWrapperApi (Agent) to cover adding and storing the message in the right places.
- Also had to update some unit tests to add the new aggregator

Closes #908 

# Author Checklist

# Reviewer Checklist
- [x] Perform code review
- [ ] Review Changelog


src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -408,6 +414,27 @@ public void IncrementLogLinesCount(string logLevel)
             }
         }
 
+        public void RecordLogMessage(DateTime timestamp, string logLevel, string logMessage, string spanId, string traceId)
```
JcolemanNR, line 417: As other log implementations are likely going to use this method as well, I think we should add more validation here. It looks like there are checks for logLevel, and logMessage not being null in log4net only code paths?

JcolemanNR (changes requested):
Looks good, but I think we should add validation to the Agent `RecordLogMessage(...)` method as well.

src/Agent/NewRelic/Agent/Core/Agent.cs
jaffinito: Yeah, I can move it to the Agent class.

jaffinito: updated

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/Experimental/IAgentExperimental.cs
```diff
@@ -17,5 +19,15 @@ public interface IAgentExperimental
         void RecordSupportabilityMetric(string metricName, int count = 1);
 
         void IncrementLogLinesCount(string logLevel);
```
JcolemanNR, line 21: I think this method should be removed, and `RecordLogMessage` should call the `AgentHealthReporter` to increment any counts. I can see it being error prone to have separate calls for these metrics (unless there is a use case I'm not thinking of).

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Logging/Log4netWrapper.cs
```diff
@@ -34,26 +34,16 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
             var xapi = agent.GetExperimentalApi();
             xapi.IncrementLogLinesCount(logLevel);
```
JcolemanNR, line 35: Not sure the wrapper should update the log lines separately from recording the log message. Do we want a count of all logs regardless of if we capture them, or a count of those that we do capture?

JcolemanNR (changes requested):
Looking good, depending on requirements (open questions in review comments), I think we can eliminate the separate call in the experimental API to count log lines by level.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/Experimental/IAgentExperimental.cs
jaffinito: Yeah, now that we attempt to record everything, it makes sense to move it.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Logging/Log4netWrapper.cs
jaffinito: The count should be all log lines the are seen.

JcolemanNR (approved):
I'm liking it! Seems ready for some unit tests :)

PR #945, opened 2022-02-16
Refactors local log decoration to use a single entry

### LINKED ISSUE #943: Refactor llocal og decoration to write the NR-LINKING-METADATA as a single entry
Currently, local log decoration writes out each entry into the properties dictionary, but this causes problems with getting the correctly formatted output in some cases.

We can fix this and simplify setting up the decoration by moving to a single entry that contains the entire output, minus the bits that are null or whitespace.

Given that Serilog (currently in planning) doesn't support hyphens or periods in key names, the recommendation is to use NR_LINKING_METADATA as the key.   This will only affect config and not the actual output

jaffinito:
## Description

Refactors local log decoration to use a single entry which contains the entire metadata string in the proper format and with jsut the data that is available.  This me that if we are missing trace and span IDs, there will not be an empty or null entry in the properties list.  This also enables this to just work for Json layouts/sinks.

Adds a new LoggingHelpers static class with a single method currently.

No unit tests since its a wrapper change and no integration tests as of yet.

Closes #943 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Logging/Log4netWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Logging/Log4netWrapper.cs
@@ -70,12 +70,11 @@ namespace NewRelic.Providers.Wrapper.Logging
                 return;
             }
 
-            // the keys in the metadata match the ones used for decorating
-            var metadata = agent.GetLinkingMetadata();
-            foreach (var entry in metadata)
-            {
-                propertiesDictionary[entry.Key] = entry.Value ?? string.Empty;
-            }
+            // uses the foratted metadata to make a single entry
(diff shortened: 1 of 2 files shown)


nr-ahemsath (approved):
Looks good; I tested this branch locally with a PatternLayout that included `%properties{NR_LINKING_METADATA}` and got the following output:

`2022-02-16 09:32:05,147 [1] INFO  Log4NetPerformanceTesting.Program - This is log message 0 NR-LINKING-METADATA: {trace.id=e0615560e06c9159597b244a95c004ea, span.id=84c3dfc59d1dbcd0, entity.name=Log4NetPerformanceTesting, entity.type=SERVICE, entity.guid=MjczMDcwfEFQTXxBUFBMSUNBVElPTnwyODMwMDQxMQ, hostname=NRC9L46G3.lan}`

PR #948, opened 2022-02-16
Truncate log event message string to 32Kb

### LINKED ISSUE #941: For application logging limit the atttribute sizes to 32kb and truncate
For application log attributes (message, level, etc), limit the size to 32kb and truncate anything after that point.  Realistically, this should only affect the message field. 

This could be done in several different places:
- When the LogEventWireModel is being created
- When the LogEventWireModel is being collected
- During Harvest when the LogEventWireModelCollection is being constructed

There may be other places this could be done, so please use your judgement to pick the best location.

There could be unit tests for this depending on the location selected, but there need to be integration tests.

nr-ahemsath:
## Description

Resolves #941.  The log message part of the LogEventWireModel is truncated to 32K bytes. There's an open question as to whether this means 32 * 1000 (kilobytes) or 32 * 1024 (kibibytes) but I've gone with the more conservative 32 * 1000 at this point.

The story says that integration tests are required; I did a unit test and I think that should be sufficient.    I reused an well-tested internal method for truncating the message string.  I think an integration test would be overkill for this. 

# Author Checklist
- [X] Unit tests, ~Integration tests, and Unbounded tests completed~ See above remarks about testing
- [X] ~Performance testing completed with satisfactory results (if required)~
- [X] ~[Agent Changelog](/src/Agent/CHANGELOG.md) or [Lambda Agent changelog](/src/AwsLambda/CHANGELOG.md) updated~ 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/WireModels/LogEventWireModel.cs
+++ b/src/Agent/NewRelic/Agent/Core/WireModels/LogEventWireModel.cs
@@ -3,11 +3,16 @@
 
 using System;
 using NewRelic.Collections;
+using NewRelic.SystemExtensions;
 
 namespace NewRelic.Agent.Core.WireModels
 {
     public class LogEventWireModel : IHasPriority
     {
+        // Not sure yet if we want 32 kilobytes (32 * 1000) or 32 kibibytes (32 * 1024).
+        // Going with the more conservative option for now.
+        private const uint MaxMessageLengthInBytes = 32 * 1000;
+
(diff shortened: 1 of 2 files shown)


jaffinito (approved):
Awesome! I really like that this lives in the wiremodel itself.  It makes maintaining it real easy.  Not having another integration is great since they are pretty slow overall.

PR #978, opened 2022-02-24
Async Log4net forwarding integration test coverage

### LINKED ISSUE #972: Async/background thread test coverage for log4net
Add integration test coverage of async and background thread log4net usage for logging metrics, forwarding, and decoration.

Should cover:
- Async
- Task.Run
- ThreadPool.QueueUserWorkItem

Review https://www.ben-morris.com/using-asynchronous-log4net-appenders-for-high-performance-logging/ and add coverage for any other recommended usage patterns

JcolemanNR:
## Description

Closes #972 

Adds async use case test coverage for log4net forwarding.

The only undesirable finding is that if logs are delegated to a background thread and not awaited, then they will not show up as part of a transaction that initiated them. I do not think there is much we can do about this.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/Logging/Log4netMetricsAndForwardingTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/Logging/Log4netMetricsAndForwardingTests.cs
@@ -61,7 +61,6 @@ namespace NewRelic.Agent.IntegrationTests.Logging
         private const string AsyncNoAwaitWithDelayInTransactionErrorMessage = "AsyncNoAwaitWithDelayInTransactionErrorLogMessage";
         private const string AsyncNoAwaitWithDelayInTransactionFatalMessage = "AsyncNoAwaitWithDelayInTransactionFatalLogMessage";
 
-
         public Log4NetMetricsAndForwardingTestsBase(TFixture fixture, ITestOutputHelper output, bool metricsEnabled, bool forwardingEnabled) : base(fixture)
         {
             _fixture = fixture;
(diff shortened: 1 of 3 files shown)


tests/Agent/IntegrationTests/IntegrationTests/Logging/Log4netMetricsAndForwardingTests.cs
```diff
@@ -220,6 +283,51 @@ public void LoggingWorksInsideTransaction()
...
+
+        [Fact]
+        public void AsyncNoAwaitLoggingWorksInsideTransaction()
+        {
+            if (_forwardingEnabled)
+            {
+                // NOTE: since the log is not awaited, it shows up outside the transaction
+                var expectedLogLines = new Assertions.ExpectedLogLine[]
+                {
+                new Assertions.ExpectedLogLine { LogLevel = "DEBUG", LogMessage = AsyncNoAwaitInTransactionDebugMessage},
```
nr-ahemsath, line 316: What do you think about making the fact that `HasTraceId` and `HasSpanId` are expected to be false in this case explicit, just to clarify what we mean by "it shows up outside the transaction"?

JcolemanNR: Yes, but I'm going a different direction since this behavior is intermittent based on if the async logging takes place before the calling function returns. I'm planning to refactor the ExpectedLogLine data structure/helpers to use a nullable bool, which will make all the other tests more explicit.

nr-ahemsath (approved):
Looks good to me.  This will be one of the most thoroughly tested features we've launched.  All logging integration tests pass for me locally.  👍

PR #996, opened 2022-03-02
Fixes sampling limits for log_event_data

jaffinito:
## Description

The max samples for log events is supposed to be 10000 per 60 seconds, not 5 seconds.  The config setting is correctly set at 10000, but that needs to be passed up during connect so that the 5 second samples limit (833) is passed back.

- Renamed LogEventsMaximumPerPeriod to LogEventsMaxSamplesStored to match the other settings.
- Adds log_event_data to harvest limits payload using during connect so that it pulls down the correct value

Note: the log_event_data limit has not be added to the connect payload so that raw 10000 value will be used.  If the connect payload is not added before our testing begins, we might want to consider setting it to 833 in the short term

# Author Checklist

# Reviewer Checklist
- [x] Perform code review
- [ ] Review Changelog


JcolemanNR (approved):
Looks good

PR #997, opened 2022-03-02
Avoid unnecessary work in logging wrappers

nr-ahemsath:
## Description

This PR updates the `RecordLogMessage()` experimental agent API to take getter functions that know how to extract the timestamp, log level and log message from an arbitrary log event object, rather than those values themselves.  This allows us to avoid doing unnecessary and expensive reflection work to get values if they aren't needed due to config settings.

I went with this approach rather than just putting the config checks in the wrapper itself for two reasons:
1. Per the current logging instrumentation engineering brief/design doc, we need to create a supportability metric that indicates that our instrumentation matched a given logging framework, even if all logging features are disabled by config.  That happens at the very top of `RecordLogMessage`, and so we can't short-circuit calling this method based on config.
2. It keeps the logic of what to do based on config settings centralized in this API rather than duplicated in all the wrappers.

Some numbers to show that this change has an improvement (taken from running a logging benchmarking app that logs ~1 million log lines as fast as possible, so the total execution time is the measure of performance - smaller is better):

| Agent Version | Config Options | Execution Time (s) |
| --------------- | ----------------- | ----------------- |
| Main Agent | N/A | 47.4 |
| Current feature branch | Overall Logging Feature and Metrics Enabled | 64.27 |
| Current feature branch | All logging features disabled | 61.38 |
| This PR | Overall Logging Feature and Metrics Enabled | 63.55 |
| This PR | All logging features disabled | 52.68 |

# Author Checklist
- [X] Performance testing completed with satisfactory results (if required)

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Agent.cs
+++ b/src/Agent/NewRelic/Agent/Core/Agent.cs
@@ -431,18 +431,18 @@ namespace NewRelic.Agent.Core
                 {
                     return;
                 }
-                var timestamp = getTimestamp(logEvent).ToUniversalTime();
+                var timestamp = getTimestamp(logEvent).ToUnixTimeMilliseconds();
 
                 var transaction = _transactionService.GetCurrentInternalTransaction();
                 if (transaction != null && transaction.IsValid)
                 {
                     // use transaction batching for messages in transactions
-                    transaction.LogEvents.Add(new LogEventWireModel(timestamp.ToUnixTimeMilliseconds(), logMessage, normalizedLevel, spanId, traceId));
+                    transaction.LogEvents.Add(new LogEventWireModel(timestamp, logMessage, normalizedLevel, spanId, traceId));
                     return;
                 }
 
                 // non-transaction messages with proper sanitized priority value
-                _logEventAggregator.Collect(new LogEventWireModel(timestamp.ToUnixTimeMilliseconds(),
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -406,33 +406,46 @@ public void RecordSupportabilityMetric(string metricName, int count)
...
+            if (_configurationService.Configuration.LogEventCollectorEnabled) 
             {
-                return;
-            }
+                var logMessage = getLogMessage(logEvent);
+                if (string.IsNullOrWhiteSpace(logMessage))
+                {
+                    return;
+                }
+                var timestamp = getTimestamp(logEvent).ToUniversalTime();
```
jaffinito, line 434: Couldn't you ToUnixTimeMilliseconds() the time here and just store it as a value type?

jaffinito (commented):
This is a wonderful approach!  Very creative!

src/Agent/NewRelic/Agent/Core/Agent.cs
nr-ahemsath: Good idea!

jaffinito (approved):
Awesome!

JcolemanNR (approved):
Nice! it's cool how we already had functions defined for accessors! :)

Issue #1002, opened 2022-03-03
Serilog logging instrumentation integration test improvements


nr-ahemsath:
Some improvements we can make to the Serilog integration tests:

1. Test multiple versions of Serilog.
2. Unify Serilog and Log4Net integration tests by parameterizing the framework being tested in the base test classes.  Some logic will be required to account for certain differences between the two frameworks, e.g. the names of log levels expected in logging metrics (WARN vs. WARNING).
3. Stretch goal: improve test execution time, which will mainly be about seeing if we can get rid of the built-in 55-second delay in the "MetricsAndForwarding" tests, which is necessary to make the assertions on data usage metrics work.

PR #1004, opened 2022-03-03
Move log data endpoint metric test

nr-ahemsath:
## Description

Part of issue #1002.  The `{Log4Net|Serilog}MetricsAndForwardingTests` included an assertion on the data usage supportability metric for the `log_event_data` endpoint.  This required the test to include a delay of ~60 seconds in order to cause a harvest cycle to happen, triggering the call to the log event data endpoint, which would then generate the expected supportability metric.  However, since this test is subclassed so many times, and because we disable parallel running of tests on the server for reliability, this was causing the logging tests to take a very, very long time to run on the server (at least an hour or two).

I've moved the data usage metric assertions on log event data to the `DataUsageSupportabilityMetricsTests`, which are only subclassed four times at present, so the 60 second delay isn't as big of deal in that context.  I also refactored those tests to use an XUnit `[Theory]` to reduce code duplication.

# Author Checklist
- [X] Unit tests, Integration tests

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


JcolemanNR (approved):
Great idea! This should make developing the tests easier as well

PR #1006, opened 2022-03-04
Swap logs maxsamples config precedence

jaffinito:
## Description

- After some lengthy discussions we decided to have the connect config take precedence over the local values since we want the maxSamplesSize to be spread over the individual faster event harvests.
- Includes setting the default value in the GetValueOrDefault as well to ensure we always get a good value.
- Not test changes since this just swaps precedence.

# Author Checklist

# Reviewer Checklist
- [x] Perform code review
- [ ] Review Changelog


JcolemanNR (approved):
Looks good based on the recent decisions. I'd prefer if we had some test coverage for the overrides, but it sounds like that isn't an established pattern.

PR #1013, opened 2022-03-10
Adds log forwarding for Microsoft.Extensions.Logging in .NET Standard 2.0+ (.NET Core Agent)

jaffinito:
## Description

- Moves LoggingHelpers out of Logging and into Extensions to share it between the new wrappers.
- Changes LogProviders.cs from using const string to use List<string> for the provider names since there are more than one possible value for the provider name for a framework. 
- Replaces LoggerFactoryWrapper with AddProviderRegistrationWrapper.  The method LoggerFactoryWrapper instruments only exists in .NET 5+.  AAddProviderRegistration is a lower level method that is called by all versions.  It is slightly less efficient since it is called for each provider.  it is basically identical to its predecessor otherwise.
- Adds MicrosoftLoggingWrapper to enabled forwarding logs.  MEL does not have the concept of a LogEvent so we pass around the MethodCall instead.  It also doesn't have a timestamp so we create one as early as possible.
 
Tests:
- Added MEL versions of the MaxSamples, HSM, and MetricsAndForwarding tests for .NET Core+ only
- MEL has different names for the various log levels so those have been changed in the MEL versions of the tests
- The new MEL LoggingAdapter also sets up Serilog as a provider to ensure that we are tests the ability to disable the other providers.  This is what led to some changes in this branch compared to the prototype.  If the disable does not work, the counts for various log lines will be off.
- Updates the *MaxSamples tests to use 12 instead of 1 as the max value since the log_event_data faster event harvest is now active.  This means that any values we set are really divided by 12 and returned to us by the server.
- Removed the JSON constants and instead the new work uses the raw .net version preprocessor directives.  The tests were not properly detecting .NET Core 2.1 so it was acting like it was unsupported -  returning NotImplemented.  Using the builtin directives works fine.
- Most of the new MEL test infra is hidden behind a preprocessor directive since it is .NET Core 2.1+ only.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


nr-ahemsath (approved):
Looks good, tests pass locally.  I did have a couple of the flickers you talked about; they have to do with the fire-and-forget async logging scenarios not completing within the amount of time the test sleeps after executing all the relevant commands.  Since this issue affects all of the logging integration tests we can address it separately from this PR.  👍

PR #1019, opened 2022-03-11
Test multiple versions of Serilog back to 2.0.0

nr-ahemsath:
## Description

Final part of #1002.  Test multiple versions of Serilog back to 2.0.0.

We can only test Serilog <2.5.0 in .NET Framework, because of how our tests for Microsoft.Extensions.Logging also depend on Serilog - specifically Serilog.Extensions.Hosting, whose earliest version (2.0.0) depends on Serilog >= 2.5.0.  However, I see no particular reason to suspect that we would not support Serilog 2.0.0 in .NET Core as well.


jaffinito (approved):
Looks good!  The tests themselves already load up different versions so this just works.

nr-ahemsath:
Having some test failures with the MEL local decoration cases after merging that into this branch, will address on Monday.

nr-ahemsath:
Tweaked a couple of versions to get all of the logging integration tests to pass.  Ready for review again.

PR #1030, opened 2022-03-22
Add back entity.name to log data payload and local decoration

nr-ahemsath:
## Description

Add `entity.name` (which is just the first (and usually only) application name the agent knows about) back to the log data payload and the local decoration metadata blob.

I updated the local decoration integration tests to be a bit more specific about the data it is looking for (not just asserting on the regex match, but asserting on the individual values as far as specifically as we can - in the case of the `entity.name` we know it should match the application name so we can be very specific about that one).

# Author Checklist
- [ ] ~[Agent Changelog](/src/Agent/CHANGELOG.md) or [Lambda Agent changelog](/src/AwsLambda/CHANGELOG.md) updated~ N/A 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


jaffinito (approved):
This looks great!  I like how you added the new app naming option for the integration tests.

JcolemanNR (commented):
Looks clean!

PR #1031, opened 2022-03-23
URLEncode entity name in local decoration

nr-ahemsath:
## Description

The `entity.name` value written to the metadata blob in local decoration needs to be URL encoded.  Updated the integration tests for local decoration to account for this.


jaffinito (approved):
Nice catch!

Issue #831, opened 2021-12-02
Performance problems with calls to Assembly::LoadFrom() from instrumented methods


mdomashchenko:
This call gets injected into the beginning of every instrumented method:
https://github.com/newrelic/newrelic-dotnet-agent/blob/813b8210523c2c50942df3da119f4008a86a25dd/src/Agent/NewRelic/Profiler/MethodRewriter/FunctionManipulator.h#L269

In our case `SqlCommand.ExecuteReaderAsync` and `HttpClient.SendAsync` which are executed constantly.

Unfortunately, that quickly hits this `lock` statement in `AssemblyLoadContext`
https://github.com/dotnet/runtime/blob/70d20f1445127d6473283047d0a551ccf85f897f/src/libraries/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyLoadContext.cs#L342

And looks like from time to time it's causing a lock convoy (https://devblogs.microsoft.com/oldnewthing/20061212-00/?p=28753)

I do have a dump file with 77 threads waiting on that `_unloadLock` object and no owner.

Could you, please, cache those types instead of loading them for every execution of every instrumented method?

nr-ahemsath:
This has a similar root cause to issue #533, which is currently marked closed, I believe pending feedback from the customer who was having the problem in that case.

Regarding caching the types, we do in fact do this for .NET Framework applications, but it just moves the problem somewhere else.

Ideally we should find a solution to this problem that works for both the framework and core versions of the agent.

angelatan2:
@mdomashchenko Thank you again for bringing this issue to our attention. We've spiked on the problem space; however, we will need more time to continue to hone in on the solution. An additional spike was created to have the devs continue to make forward progress.  We are closing this ticket as won't-fix until we have the spike resolved.  The spike is planned for this quarter.

srilakshmiprasannaakkineni:
Hello Team,

Is there any further progress on this issue? off late, we have been observing issues with our apps too.

jaffinito:
Thank you for checking in.  We have done additional research and believe we have an approach that would address the issue in .NET Framework. There is still some additional work needed to clean up the prototype and get it to production quality.  As for .NET Core/.NET, the approach we have in mind might also work, but since the two are very different we would need to dig in deeper with an additional spike.

We don't currently have the .NET Core/.NET spike or the .NET Framework prototype work scheduled, though it is on our radar.

Issue #533, opened 2021-04-15
Agent performance bottleneck caused by use of AppDomain.GetData() in high throughput applications


nr-ahemsath:
**Description**

At least one customer has reported seeing a performance bottleneck caused by the agent in a high-throughput application.  The overhead is caused by the agent's use of `AppDomain.GetData()` to cache the results of reflection calls to get `MethodInfo` data.  This is part of the IL-rewriting profiler component of the agent.  The part of the profiler that does this can be seen here: https://github.com/newrelic/newrelic-dotnet-agent/blob/dc85233559d62679252a48e5e55d4140681b2f0f/src/Agent/NewRelic/Profiler/MethodRewriter/FunctionManipulator.h#L328

The method `GetMethodFromAppDomainStorageOrReflectionOrThrow` is one of our helper methods that are themselves bytecode-injected into the `[mscorlib]System.CannotUnloadAppDomainException` class, as can be seen here: https://github.com/newrelic/newrelic-dotnet-agent/blob/8f8ffd1e7612de5b9ff81acea8911a3fcc880175/src/Agent/NewRelic/Profiler/MethodRewriter/HelperFunctionManipulator.h#L133

**Expected Behavior**

Since this feature of the agent is intended to be a performance optimization, it should not be causing a performance bottleneck (due to the locking inside `AppDomain.GetData()`).

**Steps to Reproduce**
There is no specific repro; the customer who reported this issue only sees it intermittently in production and is unable to reproduce in a test environment.  The symptoms of the issue are a major performance bottleneck in an app with the agent attached.  If a memory dump is taken of the app in this state, a "CrashHang" analysis with the DebugDiag tool will show many threads with the following stack at the top:
```
[[HelperMethodFrame] (System.Threading.Monitor.Enter)] System.Threading.Monitor.Enter(System.Object) 
System.AppDomain.GetData(System.String)+c7 
mscorlib.dll!Unknown+3d 
mscorlib.dll!Unknown+37 
```

Note: this behavior would only occur in .NET Framework; the profiler does not do this for .NET Core apps.

Idea: since the profiler method which uses the MethodInfo caching has a boolean `useCache` parameter (currently only used to distinguish between .NET Framework and .NET Core uses of this method), it should be fairly easy to make the use of this cache a configurable parameter that users could set to false if they encounter this kind of thread contention.

Internal note: [original customer ticket](https://newrelic.zendesk.com/agent/tickets/452184)
[...]

PR #1033, opened 2022-03-23
Configuration option to disable use of AppDomain for caching method information

### LINKED ISSUE #1008: Modify environment variables to be more user friendly
I was recently attempting to use environment variables to configure the .NET Agent, and ran in to a scenario where they were not working as I expected/hoped. 

I was trying to use one of the `XYZ_ENABLED` environment variables and had set it's value to `1` thinking that this should enable the option, but alas it did not. Currently the only things that are accepted are `true` to enable, and `false` to disable options exposed via environment variable (case sensitive).

### Feature Description
Modify environment variable logic to allow `0` and `1` to enable/disable features, in addition to `true`/`false`. Also, the `true`/`false` logic should be modified to be case insensitive. 

Example inputs to enable environment variable toggles:
- `1`
- `true`
- `TRUE`
- `True`

Example inputs to disable environment variable toggles:
- `0`
- `false`
- `FALSE`
- `False`

### Additional context
Currently the logic uses the default implementation for  `bool.TryParse(...)` which only accepts `true` or `false`:
https://github.com/newrelic/newrelic-dotnet-agent/blob/6bd398292186ce39992c8a2a584438cdafd5c54c/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs#L1905-L1913

JcolemanNR:
## Description

This PR introduces a secret environment variable in the profiler `NEWRELIC_DISABLE_APPDOMAIN_CACHING` that will disable the use of the AppDomain to cache method information in profiled .NET Framework applications. This effectively makes the .NET agent behave the same for .NET Framework and .NET Core applications

This is intended as an option for customers who are experiencing #533 to try. As a warning, customers who disable AppDomain caching with this new environment variable may encounter #831.

Closes #1008 by accepting 1/0 and case insensitive true/false for Boolean environment variables. 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1969,8 +1977,27 @@ private bool EnvironmentOverrides(bool local, params string[] environmentVariabl
                 .Select(_environment.GetEnvironmentVariable)
                 .FirstOrDefault(value => value != null);
 
-            return bool.TryParse(env, out var parsedValue) ? parsedValue : local;
+            if (env != null)
```
jaffinito, line 1980: Rad!!  This is a great change!

jaffinito (approved):
👍

JcolemanNR:
@sberkovitz: Great idea! :) We actually have a change in the works right now to cache method information in a ConcurrentDictionary directly in the agent, but it isn't ready for release yet. 

This work was started by @tehbio, and I've worked on it a bit as well. We should hopefully be able to roll this out as an improvement to both .NET Core and Framework applications when we're ready to release it:
https://github.com/newrelic/newrelic-dotnet-agent/tree/jcoleman/method-caching-improvement

PR #1040, opened 2022-03-29
Adds test for server sending back 0 for maxSamples

jaffinito:
## Description

Adds tests to ensure that when the server sends back a 0 for maxSamplesStored we stop forwarding log lines.  This is used as part of the system to disable log forwarding account-wide.

It does this in the tests by setting the value locally to 1, which causes the server to sen back 0 (server divides the local value by 12 - 5 seconds harvests)

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


JcolemanNR (approved):
While I think this is a bug that the collector would ever return 0 when a positive value is supplied to it, I think it's valuable to test that we don't send anything when the the collector does return 0...

PR #1049, opened 2022-04-08
Fixes, updates,and adds NoticeError docs - removes 1000 character limit note in docs

### LINKED ISSUE #1048: The documenation for  NoticeError incorrectly describes the message character limits.
The docs indicate that there is a 1000 character limit for the message parameter. This is not correct.
My testing has show the following:

- This 1000 character limit doc was added to the code in v2.0.8 and expanded out into other areas in v8.28.
- The is no 1000 character limit for NoticeError
- NoticeError creates both Error Events and ErrorTraces
- The limit is different for Error Events and Error Traces
  - Error Events: The message is stored as an attribute with a limit of 255 characters per the Attribute spec.  The Errors spec defers to this for the limit.
  - Error Traces: There does not appear to be a limit on message outside of a total payload limit of 1MB.  I was able to send over a 1000 character message with no issues.

The UI displays data from both types in different areas and there is some overlap.  I had Errors Inbox to show one error with a message from an Error Event - limited to 255 chars and a different error shows the Error Trace message at over 1000 characters.   There must be some logic in the background that picks.

I am creating a PR for the docs site to go along with our own changes to address this.

jaffinito:
## Description

The docs indicate that there is a 1000 character limit for the message parameter. This is not correct.
My testing has show the following:

- This 1000 character limit doc was added to the code in v2.0.8 and expanded out into other areas in v8.28.
- The is no 1000 character limit for NoticeError
- NoticeError creates both Error Events and ErrorTraces
- The limit is different for Error Events and Error Traces
  - Error Events: The message is stored as an attribute with a limit of 255 characters per the Attribute spec.  The Errors spec defers to this for the limit.
  - Error Traces: There does not appear to be a limit on message outside of a total payload limit of 1MB.  I was able to send over a 1000 character message with no issues.

The UI displays data from both types in different areas and there is some overlap.  I had Errors Inbox to show one error with a message from an Error Event - limited to 255 chars and a different error shows the Error Trace message at over 1000 characters.   There must be some logic in the background that picks.

I am created a PR for the docs site to go along with our own changes to address this.

I also spread out the documentation blocks to to each NoticeError method so all are not covered.

Closes #1048

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
+++ b/src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
@@ -263,7 +263,8 @@ namespace NewRelic.Agent.Core
         /// Supports web applications only. 
         /// </summary>
         /// <param name="message">The message to be displayed in the traced error.
-        /// Only the first 1000 characters are retained.
+        /// This method creates both Error Events and Error Traces.
+        /// Only the first 255 characters are retained in Error Events while Error Traces will retain the full message.
         /// </param>
         /// <param name="customAttributes">Custom parameters to include in the traced error.
         /// May be null.
@@ -280,6 +281,22 @@ namespace NewRelic.Agent.Core
             TryInvoke(work, apiName, apiMetric);
         }
 
+        /// <summary>
+        /// Notice an error identified by a simple message and report it to the New Relic service.
+        /// If this method is called within a transaction,
+        /// the exception will be reported with the transaction when it finishes.  
+        /// If it is invoked outside of a transaction, a traced error will be created and reported to the New Relic service.
+        /// Only the string/parameter pair for the first call to NoticeError during the course of a transaction is retained.
+        /// Supports web applications only. 
+        /// </summary>
+        /// <param name="message">The message to be displayed in the traced error.
+        /// This method creates both Error Events and Error Traces.
+        /// Only the first 255 characters are retained in Error Events while Error Traces will retain the full message. </param>
+        /// <param name="customAttributes">Custom parameters to include in the traced error.
+        /// May be null.
+        /// Only 10,000 characters of combined key/value data is retained.
+        /// </param>
+        /// <param name="isExpected">Mark error as expected so that it won't affect Apdex score and error rate.</param>
         public static void NoticeError(string message, IDictionary<string, string>? customAttributes, bool isExpected)
         {
             const ApiMethod apiMetric = ApiMethod.NoticeError;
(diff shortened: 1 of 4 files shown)


jaffinito:
Docs PR: https://github.com/newrelic/docs-website/pull/6997

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
```diff
@@ -280,6 +281,22 @@ void work()
             TryInvoke(work, apiName, apiMetric);
         }
 
+        /// <summary>
+        /// Notice an error identified by a simple message and report it to the New Relic service.
+        /// If this method is called within a transaction,
+        /// the exception will be reported with the transaction when it finishes.
```
JcolemanNR, line 287: There is no C# `Exception` in this case, maybe change the verbiage to error, or clarify what this means? Are you referring to an `Exception` in the UI?

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
```diff
@@ -280,6 +281,22 @@ void work()
             TryInvoke(work, apiName, apiMetric);
         }
 
+        /// <summary>
+        /// Notice an error identified by a simple message and report it to the New Relic service.
+        /// If this method is called within a transaction,
+        /// the exception will be reported with the transaction when it finishes.  
+        /// If it is invoked outside of a transaction, a traced error will be created and reported to the New Relic service.
+        /// Only the string/parameter pair for the first call to NoticeError during the course of a transaction is retained.
+        /// Supports web applications only.
```
JcolemanNR, line 290: Woah, what? Why would this only support Web applications? Does this mean that if this is called within any custom non-web instrumentation that the error is not actually Noticed?

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
```diff
@@ -300,7 +317,8 @@ void work()
         /// Supports web applications only.
```
JcolemanNR, line 317: I notice you didn't add this... This sounds like a bug though vs something to document, why would it only work for web applications?

JcolemanNR (approved):
I'm a little worried/confused by the comments about the NoticeError API only supporting web applications, but if this is how it works then this is how it works, and it would be good to document what is in the field.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
jaffinito: NoticeError should work in any type of application.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
jaffinito: I don't think the API call has any detection capability, the message seems odd.

PR #1051, opened 2022-04-11
Fix linux mkdir intermediate paths

### LINKED ISSUE #763: Agent will not run on Linux if profiler log path contains non-existing intermediate paths
**Description**

By default, the native profiler component of the agent logs to `$NEW_RELIC_HOME/Logs` (or `$NEW_RELIC_HOME/logs` on Linux).  This location can be overridden with the environment variable `NEWRELIC_PROFILER_LOG_DIRECTORY`.  On Windows, any non-existing components of the path supplied in this variable will be created (assuming the process being instrumented is running with sufficient permissions to do so).  For example:

`C:\existingDir` <- exists
`C:\existingDir\nonExistingDir` <- doesn't exist
`NEWRELIC_PROFILER_LOG_DIRECTORY=C:\existingDir\nonExistingDir\profilerLogs`

then `C:\existingDir\nonExistingDir` will be created as part of the process of creating the profiler log directory.

On Linux, this doesn't work.  Only the final component of the path can be automatically created.

`NEWRELIC_PROFILER_LOG_DIRECTORY=/home/user/existingDir/profilerLogs` <- works
`NEWRELIC_PROFILER_LOG_DIRECTORY=/home/user/nonExistingDir/profilerLogs` <- doesn't work 

**NOTE**: this problem blocks the profiler from initializing, which means that the agent will not run.

The cause appears to be that on Windows, the system call used to create the profiler log directory is `SHCreateDirectoryExW`, which automatically creates intermediate path components: https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shcreatedirectoryexw

whereas on Linux, the system call used is `mkdir`, which doesn't: https://man7.org/linux/man-pages/man2/mkdir.2.html (under errors, see "ENOENT: A directory component in pathname does not exist")

**Expected Behavior**
The behavior of the agent should be consistent between Windows and Linux.

See this Stack Overflow discussion on how to make a version of `mkdir` that automatically creates intermediate path components: https://stackoverflow.com/questions/2336242/recursive-mkdir-system-call-on-unix

nr-ahemsath:
## Description

Resolves #763.  The `DirectoryCreate` method in `UnixSystemCalls.h` is modifed to be able to handle creating paths recursively.

The reason for creating a sub-method `MakePath(std::string pathToCreate)` is to avoid having to convert back-and-forth between our `xstring_t` string type (which is what `DirectoryCreate` takes as an argument) and `std::string` when recursing through a path...it's bad enough that the `std::string` needs to be converted to two different types of string representations when calling `dirname` and `mkdir` (`dirname` needs a `char*` and `mkdir` needs a `const char*`).

Note:  there are some better methods (that work with `std::string`s) for working with file paths in the `std::filesystem` namespace introduced in C++17, if we could ever get around to migrating the profiler to use it.

The associated integration tests which previously needed to be skipped on Linux because of #763 are now un-skipped.  I verified that these integration tests pass on Ubuntu.  I manually tested the recursive directory creation on both Ubuntu and Alpine.

# Author Checklist
- [X] ~Performance testing completed with satisfactory results (if required)~ N/A

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


jaffinito (approved):
Looks okay to me!

PR #1065, opened 2022-04-21
Adds environment variables for all proxy settings

### LINKED ISSUE #878: Allow for proxy configuration via environment variables
@Julien4218 is looking to implement support for proxy server configuration in the Virtuoso guided install of the .NET Agent. For Windows, this is accomplished via modification of the global `newrelic.config` file. For Linux, the only configurability point currently leveraged is environment variables. As such, it would be nice to allow proxy configuration via environment variable.

### Feature Description
Add environment variable(s) to allow for proxy functionality configuration in the .NET Agent.

There will likely need to be multiple different environment variables due to some of the advanced configuration options (proxy domain, etc).

### Alternatives
Virtuoso could just directly modify the global `newrelic.config`, but this breaks the pattern of using primarily environment variables for Linux usage of the .NET Agent. I believe the team has a desire to expand the environment variable API as well, so adding new env vars for proxy configuration synergizes with that goal.

### Priority
Nice to have. This will allow the .NET agent to be more fully configured in Linux guided installs.

jaffinito:
## Description

* Adds environment variables for all proxy settings
* Updates string version of EnvironmentOverrides to properly handle whitespace
* Adds unit tests for each proxy setting, testing local and env vars
* change log updates!!!!!!

Closes #878

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -2021,16 +2021,16 @@ namespace NewRelic.Agent.Core.Configuration
         {
             var envValue = (environmentVariableNames ?? Enumerable.Empty<string>())
                 .Select(_environment.GetEnvironmentVariable)
-                .Where(value => value != null)
-                .FirstOrDefault();
+                .Where(value => !string.IsNullOrWhiteSpace(value))
+                .FirstOrDefault(); // returns null if no env var found or if enumerable<string> is empty
 
-            // prevent whitespace from leaking through
+            // if we get a null, we use local - should not get whitespace
             if (string.IsNullOrWhiteSpace(envValue))
             {
                 return local;
             }
 
-            return envValue;
+            return envValue.Trim();
         }
 
         private uint? EnvironmentOverrides(uint? local, params string[] environmentVariableNames)
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -1302,10 +1302,38 @@ namespace NewRelic.Agent.Core.Configuration
 
         #region Proxy
 
-        public virtual string ProxyHost { get { return _localConfiguration.service.proxy.host; } }
-        public virtual string ProxyUriPath { get { return _localConfiguration.service.proxy.uriPath; } }
-        public virtual int ProxyPort { get { return _localConfiguration.service.proxy.port; } }
-        public virtual string ProxyUsername { get { return _localConfiguration.service.proxy.user; } }
+        public virtual string ProxyHost
+        {
+            get
+            {
+                return EnvironmentOverrides(_localConfiguration.service.proxy.host, "NEW_RELIC_PROXY_HOST");
+            }
+        }
+
+        public virtual string ProxyUriPath
+        {
+            get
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1984,11 +2019,18 @@ private List<string> EnvironmentOverrides(List<string> local, params string[] en
...
-            return (environmentVariableNames ?? Enumerable.Empty<string>())
+            var envValue = (environmentVariableNames ?? Enumerable.Empty<string>())
                 .Select(_environment.GetEnvironmentVariable)
                 .Where(value => value != null)
-                .FirstOrDefault()
-                ?? local;
+                .FirstOrDefault();
+
+            // prevent whitespace from leaking through
+            if (string.IsNullOrWhiteSpace(envValue))
```
JcolemanNR, line 2028: Thoughts/feelings about trimming the value we read? I can't think of any use cases where we'd want extra whitespace at the start/end of env vars

JcolemanNR (commented):
These changes look good. Please see/respond to my open question about trimming values we read from env vars... I'm just submitting a non-approving review for now

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
jaffinito: Yeah, I was waffling back and forth on using trim vs just detecting.  I am fine using trim, what do you think?

nr-ahemsath: I think trimming makes sense.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1332,15 +1360,16 @@ public virtual string ProxyPassword
                 if (!_proxyPasswordEvaluated)
                 {
                     var hasObscuringKey = !string.IsNullOrWhiteSpace(ObscuringKey);
```
nr-ahemsath, line 1362: It's worth noting, I think, that the obscuring key can also be configured via the env var `NEW_RELIC_CONFIG_OBSCURING_KEY`.  In our config docs, we currently tell users to put the obscuring key in the environment and the obscured values in the config file (otherwise the obfuscation is almost pointless).   We might need to update that statement when we update the config docs with these new env vars.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
JcolemanNR: I agree, that we should trim this. I can only see it helping customers

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
jaffinito: Yeah, I will leave this here so I remember when working on docs

JcolemanNR (approved):
Looks good!

jaffinito:
Updated docs branch (https://github.com/jaffinito/docs-website/tree/dotnet-grass-valley-feature-pack) with new env vars and updated the obscuring portion to better reflect the new settings.

Issue #185, opened 2020-09-08
Agent fails to execute explain plan - MS SQL


angelatan2:
**Description**
The agent fails to execute the explain plan for the following scenarios:

**Scenario 1:** 
The agent can fail to execute explain plan for parameterized stored procedure. This occurs where the parameters are not included in the command text. Because of this, the agent fails to generate correct sql queries to run explain plan. The repro code is as follows. 

```
var connectionString = ApplicationConfiguration.MSSQLConnectionString;

		using (var connection = new SqlConnection(connectionString))
			{
				connection.Open();

				using (var command = new SqlCommand(@"storedprocedurename", connection))
				{
					command.CommandType = CommandType.StoredProcedure;

					command.Parameters.Add(new SqlParameter("@parameter1", 123));
					command.Parameters.Add(new SqlParameter("@parameter2", 456));

					return command.ExecuteNonQuery();
				}
			}
```

**Scenario 2:**
The agent fails to execute explain plan for parameterized sql queries that has parameters of "rowversion" or other special types. The agent incorrectly parses the explain plan queries for these types.

Suggestion: Look into Config setting about this to see if this is part of the issue.

**Expected Behavior**

jaffinito:
Partially improved the second scenario in #477 by not running explain plans when we cannot efficient;y or properly get a serialized version of the parameter value.  The amount of errors generated by incorrect explain plans should go down by a good amount.

PR #1066, opened 2022-04-22
Fix fail to execute explain plan 

vuqtran88:
## Description
This PR mostly resolves the scenario 1 from the issue #185. The reason why Agent fails to execute explain plan for this scenario is, unlike other sql commands, the command text for a parameterized stored procedure doesn't need to have parameters specified, our `bool FixParameterizedSql` function didn't work based on this difference. Since the `bool FixParameterizedSql` function turns parameterized SQLs to non-parameterized and parameterized stored procedures don't need this, in order to fix this, we need to make a special case for stored procedures.

PS: I removed the `binary` type in testing parameterized stored procedures because this type disables the explain plan creation ability in the agent.( https://github.com/newrelic/newrelic-dotnet-agent/blob/0cebc8545acdd8d09d72184cf29a9dd7f3a549f1/src/Agent/NewRelic/Agent/Parsing/SqlParser.cs#L411) We could have seen this issue if we didn't test `binary` type.


JcolemanNR (approved):
Glad this works! I wonder why we need to expand parameters at all, perhaps we can re-evaluate that in the future

PR #1081, opened 2022-05-05
Adds configuration development doc 

michaelgoin:
## Description

Added documentation around how configuration classes are generated in the agent to the `/docs` folder.

 Currently, the only documentation exists in the readme file in `.\build\Tools\xsd2code` which requires knowing that we already use that or searching around once you see the code comments that the classes are generated. I only knew to go track this down due to previous times working on the agent. 

To help out future devs/open-source contributors, I recommend we have something with the docs so submitting as a PR.

This content is very duplicative of the readme but I figured I'd leave the readme with the utility so there is context directly there. Could also just link back from there but I don't think the duplication really hurts here. What is most likely to happen is changing tooling (or removing tooling) which would naturally update this doc and delete the other.

Also noted that this tool requires .NET 3.5 which is not required by any other setup instructions we have thus far.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/docs/config-development.md
+++ b/docs/config-development.md
@@ -15,14 +15,14 @@ Updating the configuration consists of two steps:
 2. Update the class file via `xsd2code`.
 3. Add/restore the copyright to the class file.
 
-## Updating the XSD
+## Update the XSD
 
 Update the relevant XSD with new or modified configuration based on your needs.
 
 * `src\Agent\NewRelic\Agent\Core\Config\Configuration.xsd`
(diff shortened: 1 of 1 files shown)


JcolemanNR (changes requested):
Add a step to add the copyright to the generated class plz

JcolemanNR (approved):
Looks good, thanks for documenting this properly!

PR #1083, opened 2022-05-05
Adds 403 status as indicator of 'alive' URL to markdown link checking.

michaelgoin:
## Description

Adds 403 status as indicator of 'alive' URL to markdown link checking.

GitHub docs pages return a 403 when requests are not from a browser.

I'm guessing this is due to checking for a user agent header or some such to preventing doing a bunch of work for automated scripts. Given 403 indicates the page exists but you are not authorized, it seemed like the right solution here is to add that to the "alive" indicators. It is unlikely we'll add links to truly unauthorized pages we'd need this checking to catch. An alternative approach (I almost went with first) would be to add an ignore rule for GitHub pages in which case we might miss truly bad GitHub links. Lastly one could experiment with manually adding user headers but who knows if GitHub would eventually flag where our requests come from.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/.github/workflows/markdowncheck.config.json
+++ b/.github/workflows/markdowncheck.config.json
@@ -14,5 +14,5 @@
     "retryOn429": true,
     "retryCount": 5,
     "fallbackRetryDelay": "30s",
-    "aliveStatusCodes": [200, 206]
+    "aliveStatusCodes": [200, 206, 403]
   }


vuqtran88 (approved):
👍

Issue #1090, opened 2022-05-17
Default capture of namespace and function attributes on spans based on method data.


michaelgoin:
Acceptance Criteria

- Adds configuration to turn on/off (off by default).
- Adds plumbing of instrumentation points to span attributes.
- Custom instrumentation test coverage.

PR #1098, opened 2022-05-18
Adds default capture of code level metric attributes

michaelgoin:
Plumbs in initial capture of `code.namespace` and `code.function` attributes on spans. Enables mapping of code of custom-instrumented methods.

* Captures instrumented method and type for all spans which apply to instrumented code (not artificial root spans). 
* Plumbs in ability to override to update instrumentation such as MVC Controllers to point to actual customer code.
* Introduces code level metric configuration with enabled defaulting to `false`.

Closes: https://github.com/newrelic/newrelic-dotnet-agent/issues/1090


jaffinito (approved):
This looks good.    We should look at combining the integration tests with a base class to hold the actual test.

PR #1099, opened 2022-05-18
Adds user/customer code level span attributes for ASP.NET Core MVC 

michaelgoin:
Overrides the default capture of `code.namespace` and `code.function` on spans to point to the customer's controller action function that is represented by our instrumentation.

closes: https://github.com/newrelic/newrelic-dotnet-agent/issues/1091


vuqtran88 (approved):
Looks pretty straightforward. 🐑 it!

PR #1100, opened 2022-05-18
Adds user/customer code level span attributes for Web API 2

michaelgoin:
Overrides the default capture of `code.namespace` and `code.function` on spans to point to the customer's controller action function that is represented by our instrumentation.

closes: https://github.com/newrelic/newrelic-dotnet-agent/issues/1096

Went ahead and added Owin-based Web API tests so it is clear that case is supported in the future.


vuqtran88 (approved):
👍

PR #1103, opened 2022-05-19
Adds environment variable for new CLM configuration.

michaelgoin:
Introduces `NEW_RELIC_CODE_LEVEL_METRICS_ENABLED` ENV var which maps to: `codeLevelMetrics.enabled`.

closes: https://github.com/newrelic/newrelic-dotnet-agent/issues/1101

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -2448,7 +2448,19 @@ namespace NewRelic.Agent.Core.Configuration
         public int DatabaseStatementCacheCapcity => _databaseStatementCacheCapcity ?? (_databaseStatementCacheCapcity =
             TryGetAppSettingAsIntWithDefault("SqlStatementCacheCapacity", DefaultSqlStatementCacheCapacity)).Value;
 
-        public bool CodeLevelMetricsEnabled =>  _localConfiguration.codeLevelMetrics.enabled;
+        private bool? _codeLevelMetricsEnabled;
+        public bool CodeLevelMetricsEnabled
+        {
+            get
+            {
+                if (!_codeLevelMetricsEnabled.HasValue)
+                {
+                    _codeLevelMetricsEnabled = EnvironmentOverrides(_localConfiguration.codeLevelMetrics.enabled, "NEW_RELIC_CODE_LEVEL_METRICS_ENABLED");
+                }
+
+                return _codeLevelMetricsEnabled.Value;
+            }
+        }
 
         #endregion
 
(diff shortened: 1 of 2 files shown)


vuqtran88 (approved):

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -2448,7 +2448,19 @@ private void LogDisabledPropertyUse(string disabledPropertyName, string newPrope
...
 
-        public bool CodeLevelMetricsEnabled =>  _localConfiguration.codeLevelMetrics.enabled;
+        private bool? _codeLevelMetricsEnabled;
+        public bool CodeLevelMetricsEnabled
+        {
+            get
+            {
+                if (!_codeLevelMetricsEnabled.HasValue)
+                {
+                    _codeLevelMetricsEnabled = EnvironmentOverrides(_localConfiguration.codeLevelMetrics.enabled, "NEW_RELIC_CODE_LEVEL_METRICS_ENABLED");
```
JcolemanNR, line 2458: I'm not sure about this pattern. While it does reduce the frequency of environment variables being checked, I have a sneaking suspicion it may prevent the configured value from ever being reloaded once the agent starts. 

Will this work when the agent detects a config file change, and reloads?

JcolemanNR (changes requested):
I have a concern that caching the value may prevent values from being reloaded after the agent starts. Can you verify that this won't be an issue, or provide an implementation that will support this use case?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
michaelgoin: I forgot the agent did config file reloading. I believe there are other configuration items that do similar but that doesn't seem ideal if that's a supported use case.

That being said, we also don't want to be processing the ENV string for every single span event. Really, you only want to be processing ENV strings once on configuration load.

michaelgoin: Unsure what sort of testing we have around this... so I'll poke around there but here's the code paths.

When a config change is detected via the `ConfigurationTracker` a new instance of `InternalConfiguration` which inherits from `DefaultConfiguration` is created. So the storing in a member would not be a concern for that specific use case.

This would of course not automatically pick up ENV var changes but seems like a stretch to want to support that. If your agent does attempt to support that then that's a whole bag of worms.

https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Config/ConfigurationTracker.cs#L40

https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs#L364

https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Configuration/ConfigurationService.cs#L100

PR #1108, opened 2022-05-25
Updates the Microsoft.AspNetCore.Mvc.Core reference to a safer version.

vuqtran88:
## Description

Snyk scan reported the agent uses the version of Microsoft.AspNetCore.Mvc.Core that has security issue. Even though the agent references this package version, but actual version of the package that would be loaded is the version specified from the customer's application. The agent targets the lowest version for supportability purposes.

Since the AspNetCore wrapper targets netstandard2.0 so whatever the applications use the AspNetCore wrapper won’t be using the Microsoft.AspNetCore.Mvc.Core version lower than 2.0.0. Because of this, update the package reference is still a safe change to overcome Snyk scan.


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore/AspNetCore.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="Microsoft.AspNetCore.Diagnostics.Abstractions" Version="2.0.0" />
-		<PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="1.1.3" />
+		<PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.0.0" />
 		<PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.0.0" />
```
JcolemanNR, line 10: This package was flagged by SNYK as well

JcolemanNR (changes requested):
The SNYK alert flagged Microsoft.AspNetCore.Hosting@2.0.0 as well

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore/AspNetCore.csproj
JcolemanNR: It looks like we were discussing different Snyk issues :)

JcolemanNR (approved):
If this still passes all integration tests, then I think it will be good to eliminate some Snyk issues!

michaelgoin (approved):

PR #1119, opened 2022-05-27
CLM support for ASP.NET MVC Framework applications

JcolemanNR:
Adds code level metrics support for ASP.NET MVC Framework applications.

Closes: #1092

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/AsyncBeginInvokeActionWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/AsyncBeginInvokeActionWrapper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using NewRelic.Agent.Api;
+using NewRelic.Agent.Api.Experimental;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 using NewRelic.SystemExtensions;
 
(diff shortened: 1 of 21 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/AsyncBeginInvokeActionWrapper.cs
```diff
@@ -29,17 +30,30 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
+                }
 
-                var transactionName = string.Format("{0}/{1}", controllerName, actionName);
+                string transactionName = controllerName + "/" + actionName;
                 transaction.SetWebTransactionName(WebTransactionType.MVC, transactionName, TransactionNamePriority.FrameworkLow);
 
-                var segment = transaction.StartMethodSegment(instrumentedMethodCall.MethodCall, controllerName, actionName);
+                ISegment segment = transaction.StartMethodSegment(instrumentedMethodCall.MethodCall, controllerName, actionName);
+
+                string fullControllerName = MvcRouteNamingHelper.TryGetControllerFullNameFromObject(controllerContext);
```
michaelgoin, line 50: Minor but by duplicating the `TryGet...` pattern for the full name we are executing some of the same code to get the type where we could try to get the type once and then use the short and long names. Perhaps changing existing usages would be painful, though. I do believe `GetType()` has some magic that makes it pretty fast (and only allocate once if I remember right, the bigger concern) and this should not be a hot path.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/AsyncBeginInvokeActionWrapper.cs
```diff
@@ -29,17 +30,30 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
-                var actionName = MvcRouteNamingHelper.TryGetActionNameFromRouteParameters(instrumentedMethodCall.MethodCall, controllerContext.RouteData);
+                // IMPORTANT: Resist the urge to blindly refactor all of this code to use `var`
+                // IMPORTANT: We are being intentional with types over using `var` here due to
+                // IMPORTANT: the effects of handling a `dynamic` object
+                string controllerName = MvcRouteNamingHelper.TryGetControllerNameFromObject(controllerContext);
+                string actionName = MvcRouteNamingHelper.TryGetActionNameFromRouteParameters(instrumentedMethodCall.MethodCall, controllerContext.RouteData);
 
                 var httpContext = controllerContext.HttpContext;
                 if (httpContext == null)
+                {
```
michaelgoin, line 41: +1

tests/Agent/IntegrationTests/IntegrationTests/CodeLevelMetrics/AspNetCoreMvcClmTests.cs
```diff
@@ -13,11 +13,11 @@
 namespace NewRelic.Agent.IntegrationTests.CodeLevelMetrics
 {
     [NetCoreTest]
-    public class AspNetCoreMvcCodeAttributesTest : NewRelicIntegrationTest<RemoteServiceFixtures.AspNetCoreMvcBasicRequestsFixture>
+    public class AspNetCoreMvcClmTests : NewRelicIntegrationTest<RemoteServiceFixtures.AspNetCoreMvcBasicRequestsFixture>
```
michaelgoin, line 16: Not sure I agree with the renaming across the test files/classes, given these are already nested in a `CodeLevelMetrics` folder/namespace (Clm = redundant) and these are testing the `code.*` attributes added for code level metrics... but obviously its up to you and team how you want your code organized.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/AsyncBeginInvokeActionWrapper.cs
```diff
@@ -29,17 +30,30 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
                     throw new NullReferenceException("httpContext");
+                }
 
-                var transactionName = string.Format("{0}/{1}", controllerName, actionName);
+                string transactionName = controllerName + "/" + actionName;
```
michaelgoin, line 45: I thought concat of a null string used to throw but looks like I misremembered, so seems fine.

michaelgoin (approved):
Just some minor comments. Approved.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Mvc3/AsyncBeginInvokeActionWrapper.cs
JcolemanNR: I didn't think of the null case, but verified it does not crash in a test app.

tests/Agent/IntegrationTests/IntegrationTests/CodeLevelMetrics/AspNetCoreMvcClmTests.cs
JcolemanNR: Should be correct now :)

JcolemanNR:
Failing test passes locally, and had never been run in CI before. Merging to feature branch

PR #1132, opened 2022-06-07
Make CatInbound integration test more reliable

JcolemanNR:
The test was calling two web methods and it was a toss up as to which one would take longer and be sampled. This new approach should be more reliable


nr-ahemsath (approved):
Changes make sense.  This issue (needing to make sure we sample the correct transaction) has come up before, and this is the solution we have used. 

Also 👍 for making the assertions individual rather than using NrAssert.Multiple as it should help improve our ability to determine the cause of failures if/when they happen.

PR #1136, opened 2022-06-07
Add logging config to agent settings

nr-ahemsath:
## Description

Add the agent config settings for the application logging feature(s) to the `agent_settings` payload sent to New Relic when the agent connects.  Note that these are the final effective values for these config options, taking into account local config file values, environment variables, HSM/LASP overrides, and server config settings.

Tested manually and the values show up in the agent init page as expected.  One unit test is updated.  We don't have an integration test to cover agent settings, and I think that is okay.

# Author Checklist
- [X] Unit tests, ~Integration tests, and Unbounded tests completed~

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


JcolemanNR (dismissed):
Great work, this will increase the supportability of the Agent! In the future we might be interested in an integration test to verify that the correct values are sent up during normal operation of the agent, but this looks good (and follows the existing paradigm).

JcolemanNR (approved):
Looks good! Now includes changelog functionality

PR #1167, opened 2022-06-23
Change CI to run daily

JcolemanNR:
Per discussion with @angelatan2, we would like to run CI daily. If anyone objects, please add some comments! :)


nr-ahemsath (approved):
Works for me! 👍

PR #1172, opened 2022-06-27
Add MEL logging support for .Net Framework apps.

### LINKED ISSUE #1129: Update Microsoft.Extensions.Logging support to work for Framework
### Feature Description
- MEL support .NET Framework, just not 4.5.
- This work will make the MEL project similar to our AspNetCore instrumentation in that it is only built against netstandard2.0, but is copied over to the framework extensions folder by the [build_home.ps1 script](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/build/build_home.ps1#L72).

No changes are needed to the MEL project itself, just update the following:
- Updated the build_home.ps1 script (see link above) to copy the MEL instrumentation to the framework folders
- Update artifactbuilder with the new files
- **Update the MSI with the new files** (Core agent should already have entries)

vuqtran88:
## Description
This PR resolves #1129 with the following changes:

- [x] Updated the build_home.ps1 script (see link above) to copy the MEL instrumentation to the framework folders
- [x] Update artifactbuilder with the new files
- [x] **Update the MSI with the new files** (Core agent should already have entries)
- [x] Update integration tests to run MEL on Framework for all tests. As much as I wanted to test certain combinations between .Net Framework versions and the MEL versions, the current test setup only gave me limited combinations to pick from, so I had to pick ones that are most important. Given we already had good test coverage for MEL on .NET Core side, I think it is safe to have less tests on the FW side because their functionality are same.


nr-ahemsath (changes requested):

vuqtran88:
Due to serilog dependency conflicts, in the `MultiFunctionApplicationHelpers.csproj` projects, when adding MEL tests for `net462` and `ne471`, only MEL tests against `net48` are added for now.

JcolemanNR (changes requested):
Looks good, but needs changelog

JcolemanNR (approved):
Looks good!

Issue #702, opened 2021-08-28
Remove instrumentation for Castle.Monorail


jaffinito:
[Castle.Monorail](https://github.com/castleproject-deprecated/MonoRail/tree/master) is deprecated and has been for some time.   The [non-deprecated project](https://github.com/castleproject) doesn't even have Monorail.

The repo has not been updated in many years and there is no way outside of sourceforge to even get Castle.Monorail v2 - our supported version.

We don't even have integration tests (or apps) for Monorail.

Given the above, removal should be pretty simple.

- [x] Add deprecation to change log for 9.x
- [ ] ~6 months after 9.x release, remove instrumentation.

jaffinito:
Deprecation done in PR https://github.com/newrelic/newrelic-dotnet-agent/pull/710

PR #1177, opened 2022-06-29
Remove CastleMonoRail2 instrumentation.

vuqtran88:
## Description
Resolves [#702 ]


JcolemanNR (changes requested):
Looks good, but since this a deprecation (similar to the logging instrumentation deprecation), we should do the following:

- [ ] Add the instrumentation XML to the instrumentation ignore list in the profiler.
- [ ] Attempt to remove the DLL/XML in the MSI
- [ ] Attempt to remove the DLL/XML in the Azure Site Extension (I think there is prior art)

vuqtran88:
@JcolemanNR I added the change as requested, except for the Azure Site Extension. With the change in the profiler, the risk is nearly 0 so i don't think it is worth to do the update.

src/Agent/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecations/Removed Features
 * The scriptable installers have been removed. ([#1170](https://github.com/newrelic/newrelic-dotnet-agent/pull/1170))
+* Castle.Monorail v2 instrumentation has been reomoved. ([#1177])(https://github.com/newrelic/newrelic-dotnet-agent/pull/1177)
```
nr-ahemsath, line 17:
```suggestion
* Castle.Monorail instrumentation has been removed. ([#1177])(https://github.com/newrelic/newrelic-dotnet-agent/pull/1177)
```

I think that by specifying "v2" specifically, somebody might think that we still have instrumentation for other versions of Castle.Monorail, when we don't.
```suggestion
* Castle.Monorail v2 instrumentation has been reomoved. ([#1177])(https://github.com/newrelic/newrelic-dotnet-agent/pull/1177)
```

JcolemanNR (approved):
As long as the profiler is rebuilt/checked in before this merges to main then I'm fine with thig going in as is without the profiler binaries

nr-ahemsath (approved):
Looks good to me.  Looks like all review feedback has been handled.  👍

PR #1178, opened 2022-06-30
Fix unbounded integration test user secrets bug

vuqtran88:
The `Shared` project setup used 2 different versions of the `Microsoft.Extensions.Configuration` libraries which caused the following exception when run certain .NET FW related tests:

```
---> System.MissingMethodException: Method not found:
```

This PR fixes integration test failures due to the exception thrown by making the `Shared` project to use the same version of the `Microsoft.Extensions.Configuration` on both .NET Core and .NET FW targets.

Test failure ref: https://github.com/newrelic/newrelic-dotnet-agent/runs/7137605418?check_suite_focus=true


nr-ahemsath (approved):
It definitely makes sense to standardize on one version of these packages going forward.  All integration tests passed in the last CI run for this PR.  Approved.

PR #1182, opened 2022-07-06
Ignore SMSvcHost.exe

### LINKED ISSUE #1120: Add `SMSvcHost.exe` to Agent internal ignore list
### Feature Description
There was a recent customer escalation where they were getting proxy errors for an agent that had been spun up for `SMSvcHost.exe`. After investigating we realized that this was a Microsoft system service that the customer had no interest in profiling. Ideally this windows service would not be profiled by default when environment variables are globally enabled for the agent.

### Additional context
SMSvcHost.exe is a Microsoft system service which manages TCP port sharing functionality. https://answers.microsoft.com/en-us/ie/forum/all/what-is-smsvchostexe/74ca4bbc-99b6-4e86-8a65-080e2c9061f2

### Priority
Nice to have. Will eliminate customer confusion, and support cases related to `SMSvcHost.exe` in the future.

Note: This is a .NET core app.

*Suggested Priority (P1,P2,P3,P4,P5):* P3

vuqtran88:
## Description
Resolves #1120

### DIFF
--- a/src/Agent/NewRelic/Profiler/Configuration/Configuration.h
+++ b/src/Agent/NewRelic/Profiler/Configuration/Configuration.h
@@ -411,6 +411,18 @@ namespace NewRelic { namespace Profiler { namespace Configuration {
             return false;
         }
 
+        bool IsIgnoreProcess(const xstring_t& processName)
+        {
+            //Since instrumenting the SMSvcHost.exe proccess was reported to cause connection failure, force the profiler to ignore this process for safety.
+            if (Strings::EndsWith(processName, _X("SMSvcHost.exe")))
+            {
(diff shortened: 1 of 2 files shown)


vuqtran88:
After investigation, SMSvcHost.exe is a .NET FW app (not .NET Core). The reason why it got instrumented from the first place is a customer specifically specify to instrument it. However, it appears instrumenting this process is problematic, it is still a good idea to ignore it by all cost to save future engineering time from investigation.

JcolemanNR (approved):
I like the way you made a generic ignore process method to capture any future ignore processes easily.

JcolemanNR (approved):
The checked in profilers should be updated before merging to main, but this looks good!

JcolemanNR (approved):
Ship it! 🚢

Issue #1197, opened 2022-07-21
Integration test instability: app_server_wait_for_all_request_done error


nr-ahemsath:
This is one of the common failure modes of our integration tests identified by recent analysis in this internal doc: https://docs.google.com/spreadsheets/d/1TzYeQ1l11TXkQZqfYIo2IOQVFRlJ2aTdsO2Ry9xY6ag/edit#gid=0

More details as they become available.  In general, this is a failure in the mechanism we're using to manage the lifecycle of certain kinds of test applications.

nr-ahemsath:
Closed by linked PR

PR #1200, opened 2022-07-21
Fix three test apps that weren't actually waiting on their wait handles

nr-ahemsath:
Potential fix for #1197 

I found that the test that had been failing semi-regularly in recent CI runs (`CustomInstrumentation.OtherTransaction`) with the "failed to send shutdown signal to $NAME_OF_APP_WAIT_HANDLE" error had a test app that wasn't actually waiting on its event wait handle...so it would be possible to encounter this error if the timing of things was just right.

~Draft PR to see if these changes actually fix the problem in CI~ The tests that had been failing before this didn't fail on the CI run for this PR, so it's ready for review.


JcolemanNR (approved):
Great find! Let's get that CI stability up! :)

PR #1210, opened 2022-08-08
Update all ubuntu 18.04 runners to 22.04

nr-ahemsath:
Github is deprecating its `ubuntu-18.04` runners: https://github.com/actions/runner-images/issues/6002

Upgrade all workflow references of `ubuntu-18.04` to `ubuntu-22.04`.

Note: was not able to upgrade the `distro` parameter in the `run-on-arch` jobs for building and testing the ARM version of the Linux profiler, due to build problems.  Doing this will take more effort as it's unclear what the cause of the problem is, but it seems to be in the general area of cmake/clang/llvm tooling.  This doesn't seem to be a problem since we don't have any indication that the `run-on-arch` action is deprecating the `ubuntu18.04` distro option.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -149,7 +149,7 @@ jobs:
         id: runcmd
         with:
           arch: aarch64
-          distro: ubuntu20.04
+          distro: ubuntu18.04
           githubToken: ${{ github.token }}
           install: |
             apt-get update -q -y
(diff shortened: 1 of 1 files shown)


JcolemanNR (approved):
Awesome, this should keep our CI running past the 18.04 deprecation coming up in a few months! I was surprised that there weren't more issues with this migration, but after looking closer, we are building our linux amd64 profiler on ubuntu:14.04, and like you mentioned our arm64 profiler is using the run-on-arch action targeting ubuntu:18.04.

We should probably refactor our ARM64 build to a dockerfile so it can be built locally in the longrun (and we wouldn't need to use the run-on-arch action potentially), but these changes look good to keep the lights on! :)

PR #1212, opened 2022-08-10
Neilpdx/logging/enhancement

neilpdx:
## Description

Adds DEBUG-level logging for all non-secret optional environment variables.

Currently, when the agent's log level is set to DEBUG, the values of the following environment variables are logged:

NEWRELIC_HOME
NEWRELIC_INSTALL_PATH
CORECLR_NEWRELIC_HOME
COR_PROFILER_PATH

There are an additional number of optional environment variables that could be useful if logged, listed here:

- https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#optional-environment-variables

As users are not always aware that these environment variables are present, and since these variables often take precedence over the config file and control certain configuration settings, logging them could help reduce confusion for users and Support when troubleshooting why agent configuration is different than expected based on the `newrelic.config` file.

For security, the optional environment variables that may contain secrets, such as `NEW_RELIC_LICENSE_KEY`, are not logged.


src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
```diff
@@ -196,6 +196,24 @@ private void LogInitialized()
...
+                Log.DebugFormat("Environment Variable NEW_RELIC_DISABLE_SAMPLERS value: {0}", System.Environment.GetEnvironmentVariable("NEW_RELIC_DISABLE_SAMPLERS"));
+                Log.DebugFormat("Environment Variable NEWRELIC_PROFILER_LOG_DIRECTORY value: {0}", System.Environment.GetEnvironmentVariable("NEWRELIC_PROFILER_LOG_DIRECTORY"));
+                Log.DebugFormat("Environment Variable NEWRELIC_LOG_LEVEL value: {0}", System.Environment.GetEnvironmentVariable("NEWRELIC_LOG_LEVEL"));
+                Log.DebugFormat("Environment Variable NEW_RELIC_APPLICATION_LOGGING_ENABLED value: {0}", System.Environment.GetEnvironmentVariable("NEW_RELIC_APPLICATION_LOGGING_ENABLED"));
+                Log.DebugFormat("Environment Variable NEW_RELIC_APPLICATION_LOGGING_METRICS_ENABLED value: {0}", System.Environment.GetEnvironmentVariable("NEW_RELIC_APPLICATION_LOGGING_METRICS_ENABLED"));
+                Log.DebugFormat("Environment Variable NEW_RELIC_APPLICATION_LOGGING_FORWARDING_ENABLED value: {0}", System.Environment.GetEnvironmentVariable("NEW_RELIC_APPLICATION_LOGGING_FORWARDING_ENABLED"));
+
```
JcolemanNR, line 216: Great idea! There are more environment variables that would be useful to log here as well as they over-ride configured values. These can be identified by searching for `EnvironmentOverrides(` in code.

JcolemanNR (changes requested):
This will add a lot of value, but I think we should make the following changes:
- Add a more exhaustive list of Env vars (search for `EnvironmentOverrides(` in code to see them all).
- Sort the list of env vars (unless you have intentionally ordered them to have high value entries at the top)
- Refactor the code to only log environment variables if they are configured

JcolemanNR (changes requested):
Awesome! Can you convert the lists of env vars to multiline format, and then we'll get this merged? I would have just reformatted it myself and merged it, but it's on a fork.

example:
```
var environmentVariablesSensitive = new List<string> {
                "NEW_RELIC_LICENSE_KEY",
                "NEWRELIC_LICENSEKEY",
                "NEW_RELIC_SECURITY_POLICIES_TOKEN",
                "NEW_RELIC_PROXY_USER",
                "NEW_RELIC_CONFIG_OBSCURING_KEY",
                "NEW_RELIC_PROXY_PASS_OBFUSCATED",
                "NEW_RELIC_PROXY_PASS"
            };
```

JcolemanNR (approved):
This change looks great! Thanks for the contribution

PR #1217, opened 2022-08-15
CI Improvements

JcolemanNR:
## Description

I spent around 10 hours running CI on a powerful VM and dug in to any random failures I encountered locally. Here are the results of my adventures.

Resolves the following issues with CI runs:
- Adds a sleep to a AspNetCore3 test app to a method that is expected to be the traced transaction
- Eliminates test harness retries as the retry obfuscates the initial failure, and causes unrelated failures in some cases. This should allow us to see the actual failures. We may decide to turn retries back on, but for now I believe we should disable retries.
- Upgrades the selenium IE driver. I attempted to migrate to the Chrome driver, but there are issues with incompatibilities with the version of chrome on the GHA CI Runners... Since IE hasn't been updated in forever, it seems to work everywhere (except windows 11).
- Sets a custom (and available) port for the Selenium driver instead of relying on the default port to be available.
- Increase timeouts for selenium webdriver initialization based on timeouts encountered when running all Integration tests in parallel
- Changes a key assertion in the ThreadProfileStressTest that was failing on fast machines. The purpose of the test is to verify that we don't CRASH when threads are finalized before we have a chance to request call stacks. On fast/modern machines, these threads are almost always finalized by the time we try to get call stacks. The test was failing because we expected to be able to capture the callstack atleast once. This is why it worked on slower (CI) machines. This is the comment that I found which makes me believe this is the correct fix: https://github.com/newrelic/newrelic-dotnet-agent/blob/2ba99ea91ac20f79af40259e485949c6b3b1aac3/tests/Agent/IntegrationTests/Applications/ThreadProfileStressTest/Program.cs#L68-L72
- Add log delay check for transaction finalization to AspNetCore web tests
- Add an explicit exclude for `request.headers.cookie` to some tests that failed depending on which framework they were run in due to test client differences (worked in .NET Framework, failed in .NET Core).
- Harden a CAT test by allowing the same metric in two different forms based on a race condition in the test.
- Update HostedWebCore application to use unique application pool names
- Increase WCFClient timeouts based on failures seen during local parallel test runs

# Author Checklist

# Reviewer Checklist
- [x] Perform code review

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
@@ -86,6 +86,7 @@ namespace NewRelic.Agent.IntegrationTestHelpers.RemoteServiceFixtures
             set { RemoteApplication.KeepWorkingDirectory = value; }
         }
 
+        // We think that the test retry loop may be masking real problems and/or actually causing them, so we've disabled it for now
         protected virtual int MaxTries => 1;
 
         public void DisableAsyncLocalCallStack()
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
@@ -86,7 +86,7 @@ namespace NewRelic.Agent.IntegrationTestHelpers.RemoteServiceFixtures
             set { RemoteApplication.KeepWorkingDirectory = value; }
         }
 
-        protected virtual int MaxTries => 3;
+        protected virtual int MaxTries => 1;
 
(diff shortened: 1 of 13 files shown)


tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
```diff
@@ -86,7 +86,7 @@ public bool KeepWorkingDirectory
             set { RemoteApplication.KeepWorkingDirectory = value; }
         }
 
-        protected virtual int MaxTries => 3;
+        protected virtual int MaxTries => 1;
```
nr-ahemsath, line 89: I think a comment would be good here to help un-confuse the next developer to look at this code.

```suggestion
        // We think that the test retry loop may be masking real problems and/or actually causing them, so we've disabled it for now
        protected virtual int MaxTries => 1;
```

nr-ahemsath (approved):
These changes all look great.  I ran the integration tests locally and had zero test failures that weren't caused by known issues with my local config.  Great work! 🥇

PR #1219, opened 2022-08-17
Standalone integrationtests workflow

nr-ahemsath:
## Description

Create a GHA workflow that only runs integration tests (based on the agent and tests built from a previous `all_solutions.yml` workflow run.

It can:
* Run either bounded or unbounded integration tests, or both
* Run all test namespaces, or just a subset (specified as an input)

Here is a run of this workflow from this branch, using a recent set of agent and test artifacts: https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/2876837271

# Author Checklist
- [x] Run workflow and make sure it works (as in all of the desired tests get run and there aren't any failures we wouldn't expect in a regular CI test run)

# Reviewer Checklist
- [ ] Same as above


.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+          repo: ${{ github.repository }}
+      
+      - name: Upload Agent Home Folders Locally
+        uses: actions/upload-artifact@v2
+        with:
+          name: homefolders
+          path: ${{ github.workspace }}/newrelichome*
+          if-no-files-found: error
+
+      - name: Print input vars
```
jaffinito, line 55: I feel like this should go at the start of the job since it doesn't depend on the results of the previous steps..

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        with:
+          name: integrationtests
+          path: |
+            ${{ github.workspace }}/test.runsettings  # Force the artifacts to use repo root as root of package.
+            ${{ github.workspace }}/tests/Agent/IntegrationTests/**/bin/**/*
+            ${{ github.workspace }}/tests/Agent/IntegrationTests/**/Deploy/**/*
+            !${{ github.workspace }}/tests/Agent/IntegrationTests/**/obj/**/*
+          if-no-files-found: error
+
+      - name: Clear integration tests from workspace
```
jaffinito, line 84: I don't think we need this since the runner will be destroyed at the end of the job (two steps after this one).

jaffinito (commented):
This looks really good.  I wonder if we might eventually be able to use this with workflow_call to remove the integration tests code from the main yaml (I don't believe the event has all the features we would need just yet).

.github/workflows/run_integration_tests.yml
nr-ahemsath: Good catch, done.

.github/workflows/run_integration_tests.yml
nr-ahemsath: The reason this step is in there is to avoid including the "bounded" integration test assets in what gets uploaded for the unbounded integration test assets in the subsequent steps.  I can put a comment in the workflow to help clarify that.

jaffinito: Makes sense!

jaffinito (approved):
Awesome work!

jaffinito (approved):

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          workflow: all_solutions.yml
+          run_id: ${{ github.event.inputs.run_id }}
+          name: homefolders
+          path: ${{ github.workspace }}
+          repo: ${{ github.repository }}
+      
+      - name: Upload Agent Home Folders Locally
```
JcolemanNR, line 54: If this step can download the home folders from another workflow/run, why do we need to re-upload them to this workflow/run?

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          workflow: all_solutions.yml
+          run_id: ${{ github.event.inputs.run_id }}
+          name: unboundedintegrationtests
+          path: ${{ github.workspace }}
+          repo: ${{ github.repository }}
+      
+      - name: Upload Unbounded Integration Tests Locally
```
JcolemanNR, line 102: Same question as above. Seems inefficient :(

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+
+      - name: Setup .NET Core 2.2.103
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '2.2.103'
+
+      - name: Setup .NET Core 3.1.100
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '3.1.100'
```
JcolemanNR, line 184: Are you sure we need to install this? I think the server 2019 runner has 3.1.x installed already: https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md#net-core-sdk

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        shell: pwsh #this doesn't work with normal powershell due to UTF-8 BOM handling
+
+      - name: Download Agent Home Folders
+        uses: actions/download-artifact@v2
+        with:
+          name: homefolders
+          path: src/Agent
+
+      - name: Download Integration Test Artifacts
+        uses: actions/download-artifact/@v2
```
JcolemanNR, line 211: could we just download from the other workflow run here?

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+            C:\IntegrationTestWorkingDirectory\**\*.log
+            C:\IntegrationTestWorkingDirectory\**\*.config
+          if-no-files-found: error
+
+      - name: Archive Test Artifacts
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2.2.4
+        with:
+          name: integration-test-artifacts
+          path: C:\IntegrationTestWorkingDirectory\TestResults\**\*TestResults.xml
```
JcolemanNR, line 264: Do we really need to archive the test results? I don't think anyone ever uses them. (thinking of optimizations)

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Setup .NET 6.0
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '6.0.x'
+          include-prerelease: true
```
JcolemanNR, line 299: include-prerelease? probably left over from when we were adding .NET 6 support

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        shell: powershell
+
+      - name: Install dependencies
+        run: |
+          Write-Host "Installing HostableWebCore Feature"
+          Enable-WindowsOptionalFeature -Online -FeatureName IIS-HostableWebCore
+          Write-Host "Installing Msmq Features"
+          Enable-WindowsOptionalFeature -Online -FeatureName MSMQ-Server -All
+          Enable-WindowsOptionalFeature -Online -FeatureName MSMQ-HTTP -All
+          Enable-WindowsOptionalFeature -Online -FeatureName MSMQ-Triggers -All
```
JcolemanNR, line 349: I think we only need MSMQ for a subset of tests?

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        uses: southpolesteve/cosmos-emulator-github-action@v1
+
+      - name: Run Unbounded Integration Tests
+        run: |
+          if ($Env:enhanced_logging -eq $True) {
+            Write-Host "List ports in use"
+            netstat -no  
+          }
+
+          ${{ env.xunit_console }} ${{ env.unbounded_tests_dll }} -namespace NewRelic.Agent.UnboundedIntegrationTests.${{ matrix.namespace }} -parallel none -xml C:\IntegrationTestWorkingDirectory\TestResults\${{ matrix.namespace }}_testResults.xml
```
JcolemanNR, line 376: pass the parallel option in to this command if you implement that idea

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        shell: powershell
+
+      - name: Archive IntegrationTestWorkingDirectory on Failure
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2.2.4
+        with:
+          name: integration-test-artifacts
+          path: |
+            C:\IntegrationTestWorkingDirectory\**\*.log
+            C:\IntegrationTestWorkingDirectory\**\*.config
```
JcolemanNR, line 394: AppSettings.json

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+          if-no-files-found: error
+
+      - name: Archive Test Artifacts
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2.2.4
+        with:
+          name: integration-test-artifacts
+          path: C:\IntegrationTestWorkingDirectory\TestResults\**\*TestResults.xml
```
JcolemanNR, line 402: Remove if un-needed

JcolemanNR (changes requested):
I left lots of comments, it probably doesn't make sense to implement everything I asked for, but please review and incorporate what makes sense to you. The goal for this workflow is to be as lightweight and fast as possible, so I have lots of comments/suggestions to help it meet that goal.

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+    runs-on: ubuntu-latest
+    outputs:
+      integration-tests-matrix: ${{ steps.configure_integration_tests_matrix.outputs.matrix }}
+      unbounded-tests-matrix: ${{ steps.configure_unbounded_tests_matrix.outputs.matrix }}
+    steps:
+      - id: configure_integration_tests_matrix
+        run: |
+          if [ "${{ github.event.inputs.integration-test-namespaces }}" == "ALL" ] ; then
+            # Use the full list of namespaces
+            namespaces="[ 'AgentFeatures', 'AgentLogs', 'AgentMetrics', 'Api', 'AspNetCore', 'BasicInstrumentation', 'CatInbound', 'CatOutbound', 'CSP', 'CustomAttributes', 'CustomInstrumentation', 'DataTransmission', 'DistributedTracing', \
```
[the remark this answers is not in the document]
nr-ahemsath: We could do something like this:
`Get-ChildItem -Path C:\workspace\newrelic-dotnet-agent\tests\Agent\IntegrationTests\IntegrationTests -Directory -Exclude bin,obj,RemoteServiceFixtures` to get a list of the namespace directories.  However, that doesn't work for the sub-namespaces that we've exploded out for more parallelism in Logging and WCF.

We could store a file on disk and read that in, but we'd have to remember to update it whenever we added more namespaces (just like we currently have to do for the workflow files).  I am aware of the fact that this new standalone workflow creates an additional maintenance burden due to duplication.  I'd like to explore the possibility of having `all_solutions.yml` call this workflow to run the integration tests as a way to eliminate that redundancy.

All of the above, however, is beyond the scope of what I could accomplish in two days of innovation time.

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+            namespaces="[ ${{ github.event.inputs.integration-test-namespaces }} ]"
+          fi
+          echo "::set-output name=matrix::$namespaces"
+        shell: bash
+
+      - id: configure_unbounded_tests_matrix
+        run: |
+          if [ "${{ github.event.inputs.unbounded-test-namespaces }}" == "ALL" ] ; then
+            # Use the full list of namespaces
+            namespaces="[ 'CosmosDB', 'Couchbase', 'MongoDB', 'Msmq', 'MsSql', 'MySql', 'NServiceBus', 'NServiceBus5', 'Oracle', 'Postgres', 'RabbitMq', 'Redis' ]"
```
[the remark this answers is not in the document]
nr-ahemsath: Same answer 😸

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+
+      - name: Download Integration Test Artifacts
+        uses: actions/download-artifact/@v2
+        with:
+          name: integrationtests
+          # Should not need a path because the integration test artifacts are archived with the full directory structure
+
+      - name: Install dependencies
+        run: |
+          Enable-WindowsOptionalFeature -Online -FeatureName IIS-HostableWebCore
```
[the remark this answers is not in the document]
nr-ahemsath: The tag used the XUnit traits mechanism.  It would be nice if tests that required use of HWC were isolated to a few namespaces and then we could conditionally install HWC based on that.  We currently have a lot of tests that use a .NET framework web app (which generally means using HWC) that aren't really testing web instrumentation and could be refactored to use a console app instead, and this could make it easier to isolate the namespaces that require HWC.

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+        shell: powershell
+
+      - name: Archive IntegrationTestWorkingDirectory on Failure
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2.2.4
+        with:
+          name: integration-test-artifacts
+          path: |
+            C:\IntegrationTestWorkingDirectory\**\*.log
+            C:\IntegrationTestWorkingDirectory\**\*.config
```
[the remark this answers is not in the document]
nr-ahemsath: Good idea, done.

.github/workflows/run_integration_tests.yml
nr-ahemsath: The tool I was working on to help analyze test failures across runs depends on being able to download the test results.  Also, this step takes approximately zero seconds according to GHA.

.github/workflows/run_integration_tests.yml
nr-ahemsath: Done

.github/workflows/run_integration_tests.yml
nr-ahemsath: Done

.github/workflows/run_integration_tests.yml
```diff
@@ -2,13 +2,406 @@ name: Run Integration Tests
...
+            netstat -no  
+          }
+
+          Write-Host "Run tests"
+          $parallelXunitOption = "none" 
+
+          # Test parallelization is disabled until we can solve concurrent dotnet publish issues with ConsoleMF usage
+          # if ("${{ matrix.namespace }}" -like "Logging.*" ) {
+          #   Write-Host "Logging tests detected, running xunit tests in parallel."
+          #   $parallelXunitOption = "all"
```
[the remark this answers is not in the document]
nr-ahemsath: Good idea, done.

.github/workflows/run_integration_tests.yml
nr-ahemsath: Yes, should just be Msmq, NServiceBus and NServiceBus5

.github/workflows/run_integration_tests.yml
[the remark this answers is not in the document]
JcolemanNR: Makes sense. Yeah, I like the long term idea of removing the integration test steps from all_solutions, and having them live in one place.

.github/workflows/run_integration_tests.yml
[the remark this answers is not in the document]
JcolemanNR: Yeah, maybe we could figure something out in the future, but I can see how it isn't easy to determine... In the future we might want to try removing this step and seeing what fails to get the list of what needs it 🤷‍♂️

JcolemanNR (approved):
Looking good! I think this might be the future of running Integration tests :)

PR #1228, opened 2022-08-25
Implements sending exception attributes for log events

jaffinito:
## Description

Updates our RecordLogMessage api to accept a Func to get the exception data attached to the message, if present.  All existing frameworks have been updated to capture this information.  

* Captures message, stack, and class of the exception
* Stack is capture using our StackTraces.ScrubAndTruncate system so it will report similar results compared to error traces. This also enabled it to dig into the inner exceptions and pull together a complete stack.
* Wiremodel updated to add new values.  Stack is stored as a string, not string[], here since that is what the spec is asking for.
* Unit tests (AgentWrapperApiTests) have been added.
* Integration tests and our LoggingAdapter have been updated to take an exception for "Error" level logs instead of a string. 
* Integration test Assertions have been updated to check for for exception data in log lines, but only for "Error" level.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


JcolemanNR (changes requested):
Awesome, this will add a lot of value for customers calling log methods and supplying exceptions!

In addition to my inline comments requesting changes, I think we should add unit and integration test coverage for log events with a null/empty message and an exception per the spec.

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -431,26 +432,44 @@ public void RecordLogMessage(string frameworkName, object logEvent, Func<object,
...
                 }
+
                 var timestamp = getTimestamp(logEvent).ToUnixTimeMilliseconds();
 
+                var logException = getLogException(logEvent);
+
+                LogEventWireModel logEventWireModel = null;
+                if (logException != null)
+                {
+                    // logException.StackTrace.Split(new[] { '\n' }, 300)
```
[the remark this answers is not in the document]
jaffinito: I thought I had gotten all of those comments...

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -431,26 +432,44 @@ public void RecordLogMessage(string frameworkName, object logEvent, Func<object,
...
+                    logEventWireModel = new LogEventWireModel(timestamp, logMessage, normalizedLevel,
+                        StackTraces.ScrubAndTruncate(logException, 300), logException.Message, logException.GetType().ToString(),
```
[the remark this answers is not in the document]
jaffinito: Good catch!

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/LogInstrumentation/SerilogLoggingWebAdapter.cs
```diff
@@ -40,9 +40,12 @@ public void Warn(string message)
...
         }
 
-        public void Error(string message)
+        public void Error(Exception exception)
         {
-            var result = _client.GetStringAsync(_uriBase + "test?logLevel=ERROR&message=" + message).Result;
+            // In this case we are not passing the exact Exception to the test app, just the message.
+            // The test app will create an Exception for us.
+            // As long as it has the same message and class with a stacktrace of any kind, its good for test.
+            var result = _client.GetStringAsync(_uriBase + "test?logLevel=ERROR&message=" + exception.Message).Result;
```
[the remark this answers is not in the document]
jaffinito: I was actually thinking of removing all but info and error(exception) .  Having more than one level is not testing anything more in instrumentation, instead it is testing the logging framework itself.  if we can capture info level, we will capture any of the others.  The same goes for capturing exceptions, since all the levels take an exception in some form - using error just made the most sense thematically.  The logging tests are also way to large at this point, so adding more would be less than ideal.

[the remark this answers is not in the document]
JcolemanNR: I don't see how we can test a log message with an exception, but no message with the current model (in this webapp, and the ConsoleMF api). In this case I was thinking we could just add another query parameter to the test app for exceptions. 

Regarding testing different log levels, I believe there is value in this, especially when we add some tests that verify the configured log level is respected. While I agree that testing all log levels may be exhaustive, there is little overhead (except possibly in maintaining the tests and building them initially).

I'm not saying we need a whole separate test harness (which would add more time), I was just saying it should be clear by reading the tests that we are testing an exception, and log events with/without messages.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/LogInstrumentation/Controllers/TestController.cs
```diff
@@ -37,7 +37,8 @@ public string Get(string logLevel, string message)
                     _logger.LogWarning(message);
                     break;
                 case "ERROR":
-                    _logger.LogError(message);
+                    var exception = ExceptionBuilder.BuildException(message);
+                    _logger.LogError(exception, exception.Message);
```
JcolemanNR, line 41: The current implementation does not allow us to test an exception without a message

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -427,30 +429,43 @@ public void RecordLogMessage(string frameworkName, object logEvent, Func<object,
             if (_configurationService.Configuration.LogEventCollectorEnabled)
             {
                 var logMessage = getLogMessage(logEvent);
```
JcolemanNR, line 431: We only want to forward a log event without a message if it has a non-null exception.

JcolemanNR (changes requested):
I believe the logic for when to forward log events still needs some changes to be conformant to the specifications.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/LogInstrumentation/Controllers/TestController.cs
jaffinito: This should be resolved in the new commit.

src/Agent/NewRelic/Agent/Core/WireModels/LogEventWireModel.cs
```diff
@@ -70,5 +86,30 @@ public LogEventWireModel(long unixTimestampMS, string message, string level, str
...
         }
+
+        public LogEventWireModel(long unixTimestampMS, string message, string level, ICollection<string> errorStack, string errorMessage, string errorClass, string spanId, string traceId)
+        {
+            TimeStamp = unixTimestampMS;
+            Message = message.TruncateUnicodeStringByBytes(MaxMessageLengthInBytes);
+            Level = level;
+            SpanId = spanId;
+            TraceId = traceId;
+            ErrorStack = string.Join(" \n", errorStack);
```
chynesNR, line 97: Are there other places where we concatenate multi-line text? If so, maybe it would make sense to have a helper function to do the formatting?

chynesNR, line 97: Since this will be going out as JSON, does the newline need to be escaped?

jaffinito: Other stack traces are sent up as an array, not a single string - the results from the ScrubAndTruncate call.  Unfortunately, we also need to keep the stack lines to no more than 300 so we use the existing ScrubAndTruncate  to get the lines truncated and in the same format as out existing stacks and then join it back to a single string when we build the model.

Given its a such a small bit of code, I don;t think a helper function makes much sense here.

jaffinito: I have not seen any errors in testing, the data comes in properly.

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -427,30 +429,49 @@ public void RecordLogMessage(string frameworkName, object logEvent, Func<object,
...
-                if (string.IsNullOrWhiteSpace(logMessage))
+
+                var timestamp = getTimestamp(logEvent).ToUnixTimeMilliseconds();
```
nr-ahemsath, line 433: Would it make sense to move the getting and conversion of the timestamp below the short-circuit to avoid unnecessary work (however trivial)?

JcolemanNR (dismissed):
These changes look good to me!

tests/Agent/IntegrationTests/IntegrationTests/Logging/MetricsAndForwardingTests.cs
```diff
@@ -372,18 +398,23 @@ private void LoggingWorksOutsideTransaction()
...
+
+                    // 2 expected NOMESSAGE log lines since there is no way to tell the inside transaction line from the outside transaction line
+                    // One line needs to have HasTraceId and HasSpanId set to true
+                    new Assertions.ExpectedLogLine { Level = LogUtils.GetLevelName(_loggingFramework, "NOMESSAGE"), LogMessage = null, HasSpanId = true, HasTraceId = true, HasException = true, ErrorStack = "\t   at NotNewRelic.ExceptionBuilder.BuildException(String message) in C:\\dev\\github\\newrelic-dotnet-agent\\tests\\Agent\\UnitTests\\Core.UnitTest\\Wrapper\\AgentWrapperApi\\AgentWrapperApiTests.cs:line 1597", ErrorMessage = _loggingFramework == LoggingFramework.SerilogWeb ? SerilogWebErrorNoMessage : InTransactionErrorNoMessage, ErrorClass = "Exception" },
```
nr-ahemsath, line 406: 1. I'm having a hard time understanding how these tests are working with a hardcoded error stack that includes a path that doesn't match where my tests live (or where they live when run in CI).  They do pass for me locally, so there must be something I don't get.
2. Since it seems like this stack string is repeated several times verbatim, I suggest pulling it out into a constant for readability and maintainability.

vuqtran88 (dismissed):

src/Agent/NewRelic/Agent/Core/Agent.cs
jaffinito: That is a great catch! I'll make the adjustment.

tests/Agent/IntegrationTests/IntegrationTests/Logging/MetricsAndForwardingTests.cs
jaffinito: Your right, the stack string could be a constant.
Let's chat about the exception bit!

nr-ahemsath (approved):
Changes look good to me.  Looks good in the UI too.  👍

PR #1237, opened 2022-09-01
Resolve issue where more logs were forwarded than desired from MEL

### LINKED ISSUE #1230: Log forwarding does not respect configured level for Microsoft Extensions Logging
**Description**
When automatic log forwarding is enabled, the agent is supposed to only capture and forward logs at the lowest configured level in an application. Based on recent testing we have identified that this automatic detection of the lowest configured log level is not functioning properly when instrumenting Microsoft Extensions Logging.

Consider the following `appsettings.json` from a .NET 6 ASP.NET Core MVC web application which configures the default log level to `Information`:
```
{
  "Logging": {
    "LogLevel": {
      "Default": "Information"     
    }
  }
}
```

When the application is run, only `Information` and above log events are written to the console. When the .NET Agent attaches, and the automatic log forwarding feature sends data, it is capturing and forwarding all levels of logs including `Trace` and `Debug`.

The following code results in log information incorrectly being forwarded to New Relic, while nothing is written to the Console by Microsoft Extensions Logging:

```
public IActionResult Index()
{
    _logger.LogTrace("This should not be sent via log forwarding...");
    return View();
}
```

**Expected Behavior**
The lowest configured log level for a log library is used to forward log events, or there is another mechanism in place to manually filter logs by level in the .NET agent.

*Suggested Priority (P1,P2,P3,P4,P5):* P2

JcolemanNR:
## Description

Our instrumentation point for Microsoft Extensions Logging in agent log forwarding is invoked before any of the built in log level filtering has taken place. Due to this, logs with a lower level than what was configured in MEL were being forwarded to New Relic. 

These changes resolve that issue by querying MEL to see if the level of observed log events are enabled or disabled.

Resolves #1230 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/.github/workflows/run_integration_tests.yml
+++ b/.github/workflows/run_integration_tests.yml
@@ -48,7 +48,7 @@ jobs:
             'Errors', 'HttpClientInstrumentation.NetCore', 'HttpClientInstrumentation.NetFramework', 'Owin', 'ReJit.NetCore', 'Logging.HsmAndCsp', 'Logging.LocalDecoration', 'Logging.MaxSamplesStored', 'Logging.ZeroMaxSamplesStored', \
             'Logging.MetricsAndForwarding.log4net', 'Logging.MetricsAndForwarding.MicrosoftLogging', 'Logging.MetricsAndForwarding.Serilog', 'Logging.MetricsAndForwarding.NLog', 'ReJit.NetFramework', 'RemoteServiceFixtures', 'RestSharp', \
             'WCF.Client.IIS.ASPDisabled', 'WCF.Client.IIS.ASPEnabled', 'WCF.Client.Self', 'WCF.Service.IIS.ASPDisabled', 'WCF.Service.IIS.ASPEnabled', 'WCF.Service.Self', 'RequestHeadersCapture.WCF', 'RequestHeadersCapture.Owin', \
-            'RequestHeadersCapture.AspNetCore', 'RequestHeadersCapture.Asp35', 'RequestHeadersCapture.EnvironmentVariables', 'RequestHandling', 'CodeLevelMetrics' ]"
+            'RequestHeadersCapture.AspNetCore', 'RequestHeadersCapture.Asp35', 'RequestHeadersCapture.EnvironmentVariables', 'RequestHandling', 'CodeLevelMetrics', 'Logging.LogLevelDetection' ]"
           else
             # Just use the supplied list of namespaces
             namespaces="[ ${{ github.event.inputs.integration-test-namespaces }} ]"
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
@@ -8,7 +8,7 @@ using NewRelic.Agent.Api.Experimental;
 using NewRelic.Agent.Extensions.Logging;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 using System.Collections.Generic;
-using Microsoft.Extensions.Caching.Memory;
+using System.Collections.Concurrent;
 
 namespace NewRelic.Providers.Wrapper.MicrosoftExtensionsLogging
 {
@@ -17,7 +17,6 @@ namespace NewRelic.Providers.Wrapper.MicrosoftExtensionsLogging
         public bool IsTransactionRequired => false;
 
         private const string WrapperName = "MicrosoftLogging";
-        private static readonly IMemoryCache MemoryCache = new MemoryCache(new MemoryCacheOptions());
 
         public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
         {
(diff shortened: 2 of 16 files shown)


tests/Agent/IntegrationTests/IntegrationTests/Logging/Common.cs
```diff
@@ -11,4 +11,72 @@ public enum LoggingFramework
...
+                        case "DEBUG":
+                            return "DEBUG";
+                        case "INFO":
+                            return "INFORMATION";
+                        case "WARN":
+                            return "WARNING";
+                        case "ERROR":
+                            return "ERROR";
+                        case "FATAL":
+                            return "CRITICAL";
```
JcolemanNR, line 54: NOTE: We used to map "FATAL" to "TRACE", this has been corrected

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
```diff
@@ -33,16 +33,24 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
 
         private void RecordLogMessage(MethodCall methodCall, IAgent agent)
         {
-            // MSE Logging doesn't have a timestamp for us to pull so we fudge it here.
-            Func<object, DateTime> getTimestampFunc = mc => DateTime.UtcNow;
+            // We need to manually check if each log message is enabled since our MEL instrumentation takes place before
+            // logs have been filtered to enabled levels
+            var melLoggerInstance = (MEL.ILogger)methodCall.InvocationTarget;
+            var logLevelIsEnabled = melLoggerInstance.IsEnabled((MEL.LogLevel)methodCall.MethodArguments[0]);
```
JcolemanNR, line 39: We could potentially cache these responses.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
```diff
@@ -33,16 +33,24 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
 
         private void RecordLogMessage(MethodCall methodCall, IAgent agent)
         {
-            // MSE Logging doesn't have a timestamp for us to pull so we fudge it here.
-            Func<object, DateTime> getTimestampFunc = mc => DateTime.UtcNow;
+            // We need to manually check if each log message is enabled since our MEL instrumentation takes place before
+            // logs have been filtered to enabled levels
+            var melLoggerInstance = (MEL.ILogger)methodCall.InvocationTarget;
```
chynesNR, line 38: Is it worth putting this in a try/catch? If we fail to cast for some reason (unlikely, but say they add a new log level), we probably want to handle that?

JcolemanNR: Great question! The common pattern is to not hide exceptions like this as they would occur every time this instrumentation is invoked with potentially large overhead (exceptions aren't lightweight). The instrumentation wrapper framework has top level exception handlers that will disable instrumentation if an exception threshold is reached.

chynesNR: Oh that's good failsafe to have, and you're right, exceptions aren't cheap. We could use `as` and `Enum.TryParse()` here, though if you just want it to fail outright so it can be addressed, that's valid, too.

JcolemanNR: If this does fail, I think the logged exception for failed type conversions would be more useful in logs (though the same could be achieved with `as`). I do see we are casting to ILogger twice though, so will only do that once.

tests/Agent/IntegrationTests/IntegrationTests/Logging/Common.cs
```diff
@@ -11,4 +11,72 @@ public enum LoggingFramework
         SerilogWeb,
         NLog
     }
+
+    public class LogUtils
```
jaffinito, line 15: Really like seeing this pulled out into Common.

tests/Agent/IntegrationTests/IntegrationTestHelpers/Assertions.cs
```diff
@@ -568,7 +568,7 @@ public static void LogLinesExist(IEnumerable<ExpectedLogLine> expectedLogLines,
             var succeeded = true;
             var builder = new StringBuilder();
 
-            if (!actualLogLines.Any() && actualLogLines.Any())
+            if (!actualLogLines.Any() && expectedLogLines.Any())
```
jaffinito, line 571: Looks like this was a bug we might have missed, nice catch.

jaffinito (approved):

.github/workflows/run_integration_tests.yml
```diff
@@ -48,7 +48,7 @@ jobs:
             'Errors', 'HttpClientInstrumentation.NetCore', 'HttpClientInstrumentation.NetFramework', 'Owin', 'ReJit.NetCore', 'Logging.HsmAndCsp', 'Logging.LocalDecoration', 'Logging.MaxSamplesStored', 'Logging.ZeroMaxSamplesStored', \
             'Logging.MetricsAndForwarding.log4net', 'Logging.MetricsAndForwarding.MicrosoftLogging', 'Logging.MetricsAndForwarding.Serilog', 'Logging.MetricsAndForwarding.NLog', 'ReJit.NetFramework', 'RemoteServiceFixtures', 'RestSharp', \
             'WCF.Client.IIS.ASPDisabled', 'WCF.Client.IIS.ASPEnabled', 'WCF.Client.Self', 'WCF.Service.IIS.ASPDisabled', 'WCF.Service.IIS.ASPEnabled', 'WCF.Service.Self', 'RequestHeadersCapture.WCF', 'RequestHeadersCapture.Owin', \
-            'RequestHeadersCapture.AspNetCore', 'RequestHeadersCapture.Asp35', 'RequestHeadersCapture.EnvironmentVariables', 'RequestHandling', 'CodeLevelMetrics' ]"
+            'RequestHeadersCapture.AspNetCore', 'RequestHeadersCapture.Asp35', 'RequestHeadersCapture.EnvironmentVariables', 'RequestHandling', 'CodeLevelMetrics', 'Logging.LogLevelDetection' ]"
```
nr-ahemsath, line 51: Thanks for remembering to need to duplicate work.  Hopefully we will find the time soon to eliminate this source of toil that I created. 🐑

nr-ahemsath (approved):
Changes look good to me!

JcolemanNR:
Performance test results:

With Changes (additional log enabled check):
Logged 873129 messages in 2 minutes. 9-10% CPU
Logged 1015070 messages in 2 minutes. 8-10% CPU
Logged 995464 messages in 2 minutes. 8-11% CPU

Without changes:
Logged 872923 messages in 2 minutes. 8-11% CPU
Logged 1007570 messages in 2 minutes. 8-10% CPU

PR #1239, opened 2022-09-08
Setting from NR backend should override setting from MAX_EVENT_SAMPLES_STORED and MAX_TRANSACTION_SAMPLES_STORED env var

vuqtran88:
## Description

At the current state, when MAX_EVENT_SAMPLES_STORED env var is set, the agent uses value from the env var for the preconnect request. The server replies with a smaller value (whatver the value sent up get divided by 12). However, the agent won't use the value from the server because our current logic allows the env to override any config values. This causes configuration mismatched as well as incorrect `...\EvenHarvest\CustomEventData\HarvestLimit` supportability metric value being reported.

The same story happens to the MAX_TRANSACTION_SAMPLES_STORED env var usage.

The right approach is to let the server config to dictate the final value that the agent will use.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -1528,10 +1528,8 @@ namespace NewRelic.Agent.Core.Configuration
         {
             get
             {
-                return (int)EnvironmentOverrides(
-                    ServerOverrides(_serverConfiguration.EventHarvestConfig?.CustomEventHarvestLimit(),
-                        _localConfiguration.customEvents.maximumSamplesStored),
-                    "MAX_EVENT_SAMPLES_STORED");
+                var maxValue = _localConfiguration.customEvents.maximumSamplesStored;
+                return ServerOverrides(_serverConfiguration.EventHarvestConfig?.CustomEventHarvestLimit(), (int)EnvironmentOverrides(maxValue, "MAX_EVENT_SAMPLES_STORED"));
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1528,10 +1528,8 @@ public virtual int CustomEventsMaximumSamplesStored
         {
             get
             {
-                return (int)EnvironmentOverrides(
-                    ServerOverrides(_serverConfiguration.EventHarvestConfig?.CustomEventHarvestLimit(),
-                        _localConfiguration.customEvents.maximumSamplesStored),
-                    "MAX_EVENT_SAMPLES_STORED");
+                var maxValue = _localConfiguration.customEvents.maximumSamplesStored;
```
nr-ahemsath, line 1531:
```suggestion
               // Faster Event Harvest configuration rules apply here, which is why ServerOverrides takes precedence over EnvironmentOverrides
                var maxValue = _localConfiguration.customEvents.maximumSamplesStored;
```

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1600,9 +1598,8 @@ public virtual int TransactionEventsMaximumSamplesStored
             get
             {
                 var maxValue = _localConfiguration.transactionEvents.maximumSamplesStored;
```
nr-ahemsath, line 1600:
```suggestion
                // Faster Event Harvest configuration rules apply here, which is why ServerOverrides takes precedence over EnvironmentOverrides
                var maxValue = _localConfiguration.transactionEvents.maximumSamplesStored;
```

nr-ahemsath (changes requested):
I'd like to add some clarifying comments to help the next person to look at this understand the precedence order of ServerOverrides vs. EnvironmentOverrides and why it is different from other cases.

nr-ahemsath (dismissed):
Looks good!

nr-ahemsath (approved):
Looks even better than before! :shipit:

nr-ahemsath (changes requested):
@vuqtran88 Josh reminded me that we need a changelog entry for this

nr-ahemsath (approved):
Thanks for updating the changelog.  💯

PR #1249, opened 2022-09-22
Include original message in NLog output when local decoration is enabled

nr-ahemsath:
## Description

Fixes a bug in NLog instrumentation where the original message was not being included in output if the local decoration option was enabled.

Integration tests are updated to verify the inclusion of the original message in the output for local decoration scenarios.

This PR is in draft until I verify end-to-end functionality.

# Author Checklist

# Reviewer Checklist
- [x] Perform code review
- [x] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
@@ -79,7 +79,7 @@ namespace NewRelic.Providers.Wrapper.NLogLogging
 
             // this cannot be made a static since it is unique to each logEvent
             var messageSetter = VisibilityBypasser.Instance.GeneratePropertySetter<string>(logEvent, "Message");
-            messageSetter(messageGetter + " " + formattedMetadata);
+            messageSetter(originalMessage + " " + formattedMetadata);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
```diff
@@ -79,7 +79,7 @@ private void DecorateLogMessage(object logEvent, Type logEventType, IAgent agent
 
             // this cannot be made a static since it is unique to each logEvent
             var messageSetter = VisibilityBypasser.Instance.GeneratePropertySetter<string>(logEvent, "Message");
-            messageSetter(messageGetter + " " + formattedMetadata);
```
JcolemanNR, line 82: Good catch not just calling the getter again, as we already have!

tests/Agent/IntegrationTests/IntegrationTests/Logging/LocalDecorationTests.cs
```diff
@@ -76,6 +81,10 @@ public void LogIsDecorated()
                 Assert.NotNull(traceId);
                 Assert.NotNull(spanId);
                 Assert.Equal(_primaryApplicationName, entityName);
+
+                // Make sure the original message is there too
+                var commandResults = Regex.Split(testOutput, System.Environment.NewLine).Where(l => !l.Contains("EXECUTING"));
+                Assert.Contains(_testMessage, string.Join(System.Environment.NewLine, commandResults));
```
JcolemanNR, line 87: I think this check should be moved out of any if/else logic since we always want to make sure the original log message exists

JcolemanNR (approved):
Looks good! I think we should make a minor logic change to the test, but I don't feel strongly enough about it to 'Request Changes'

tests/Agent/IntegrationTests/IntegrationTests/Logging/LocalDecorationTests.cs
nr-ahemsath: Oh, that's a good catch.  Will do.

JcolemanNR (approved):
Looks good!

PR #1251, opened 2022-09-27
Stop any harvests while reconnecting

vuqtran88:
## Description
In an edge case where a user sets to disable event collection for specific event types from the New Relic APM UI and the agent is issued a reconnect, the agent still sends up the disabled events during the agent reconnecting period.

During the agent reconnecting window, all of the configuration settings get reset to the default values while waiting for the configuration setting backend negotiation to complete. This causes some events still being collected and sent up during this time period regardless some events collection were explicitly disabled.

This PR proposes a fix where the agent stops any scheduled `Harvest` during the agent's reconnecting process.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Aggregators/AbstractAggregator.cs
+++ b/src/Agent/NewRelic/Agent/Core/Aggregators/AbstractAggregator.cs
@@ -73,9 +73,5 @@ namespace NewRelic.Agent.Core.Aggregators
             _scheduler.StopExecuting(Harvest);
         }
 
-        protected void Stop()
-        {
-            _scheduler.StopExecuting(Harvest, TimeSpan.FromSeconds(2));
-        }
     }
 }
--- a/src/Agent/NewRelic/Agent/Core/Aggregators/AbstractAggregator.cs
+++ b/src/Agent/NewRelic/Agent/Core/Aggregators/AbstractAggregator.cs
@@ -23,8 +23,14 @@ namespace NewRelic.Agent.Core.Aggregators
             _scheduler = scheduler;
             _processStatic = processStatic;
 
+            _subscriptions.Add<StopHarvestEvent>(OnStopHarvestEvent);
             _subscriptions.Add<AgentConnectedEvent>(OnAgentConnected);
             _subscriptions.Add<PreCleanShutdownEvent>(OnPreCleanShutdown);
+        }
+
+        private void OnStopHarvestEvent(StopHarvestEvent obj)
+        {
+            _scheduler.StopExecuting(Harvest, TimeSpan.FromSeconds(2));
         }
 
         public abstract void Collect(T wireModel);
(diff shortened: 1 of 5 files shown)


src/Agent/NewRelic/Agent/Core/Aggregators/LogEventAggregator.cs
```diff
@@ -41,7 +41,7 @@ public LogEventAggregator(IDataTransportService dataTransportService, IScheduler
         }
 
         protected override TimeSpan HarvestCycle => _configuration.LogEventsHarvestCycle;
-        protected override bool IsEnabled => _configuration.LogEventCollectorEnabled;
+        protected override bool IsEnabled => _configuration.LogEventCollectorEnabled && _configuration.LogEventsMaxSamplesStored > 0;
```
JcolemanNR, line 44: I thind DefaultConfiguration.LogEventCollectorEnabled already includes this in it's logic

src/Agent/NewRelic/Agent/Core/Aggregators/AbstractAggregator.cs
```diff
@@ -66,5 +72,10 @@ public override void Dispose()
             base.Dispose();
             _scheduler.StopExecuting(Harvest);
         }
+
+        protected void Stop()
+        {
+            _scheduler.StopExecuting(Harvest, TimeSpan.FromSeconds(2));
+        }
```
JcolemanNR, line 79: Can this be removed?

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -135,6 +135,8 @@ private void Disconnect()
 
         private void Reconnect()
         {
+            EventBus<StopHarvestEvent>.Publish(new StopHarvestEvent());
```
chynesNR, line 138: Would it make more sense to have this in Disconnect? That way it would also stop harvesting during shutdown. Or maybe that already happens?

vuqtran88: > Would it make more sense to have this in Disconnect? That way it would also stop harvesting during shutdown. Or maybe that already happens?

After putting some thought, I think I still like have this as where it is. The reasons are:
1. I would like this to happen as soon as the agent receives a reconnect event. Currently the Disconnect call is wrapped inside a lock which potentially delays the call.
2. The PreCleanShutdownEvent event already calls `_scheduler.StopExecuting(Harvest, TimeSpan.FromSeconds(2));` and it happens before the CleanShutdownEvent, so it would not make any different to make such change.

JcolemanNR (approved):
This looks good to me and should prevent any rogue harvest cycles during agent reconnect!

chynesNR (approved):

PR #1260, opened 2022-10-05
Update RestSharp Testing

vuqtran88:
- Updates RestSharp testing with the latest working version (`106.6.10`).
- Due to a limitation of the Muti Version Scanner (MVS), I can't update it to validate some RestSharp packages. MVS currently fails to separate the validation between different TargetFrameworks. Therefore, I updated the Scanner with commented code which can be uncommented later after the limitation is lifted.


chynesNR (approved):
This is fine, though I don't think it's adding much value at the moment.

JcolemanNR (approved):
Looks good. It would be good to convert these to ConsoleMF and get better test coverage as this is only currently tested in .NET Framework.

PR #1263, opened 2022-10-06
Add utility script for generating ConsoleMF test classes

nr-ahemsath:
## Description

I figure people might be tired of creating all the test subclasses for ConsoleMF tests, so I made a script to do it for us.  I've placed it in `tests/Agent/IntegrationTests` and you can run it like (using RabbitMqTests as an example):

`./generateTestSubclasses.ps1 -BaseClassName RabbitMqTestsBase`

and you'll get output you can copy/paste into the test file.

Of course, this would be much cooler if we made it a Visual Studio extension so it could just happen in VS without having to open up a Powershell console, run a command with the correct input parameter, copy/paste text, etc.  But I figured I could get this done in a few minutes, I don't know how hard writing a VS extension is.  Going for "good enough" first.


chynesNR (approved):

JcolemanNR (approved):
Cool utility!

PR #1273, opened 2022-10-12
Update MSSQL tests to use the ConsoleMultiFunctionApplicationFW/ConsoleMultiFunctionApplicationCore utility application.

vuqtran88:
## Description:

The goal of this PR is to swap the test application used in MSSQL tests with the ConsoleMultiFunctionApplicationFW/ConsoleMultiFunctionApplicationCore application. This will enable us to be able to test an array of different versions of different MSSQL client libraries.

This PR also removes the old test `MsSqlAsyncLegacyAspPipelineTests` as the test appears be no longer relevant as we stopped support old .NET Framework. 

### Disclaimer:
This PR does not intend to add more test coverage .ie no additional tests to test different versions of a specific libraries. These tests will be added in followup PRs.


tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/MsSql/MicrosoftDataSqlClientExerciser.cs
```diff
@@ -0,0 +1,293 @@
...
+                }
+            }
+
+            return string.Join(",", teamMembers);
+        }
+
+        [LibraryMethod]
+        [Transaction]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        public string MsSql_WithParameterizedQuery(bool paramsWithAtSign)
```
jaffinito, line 131: Minor nitpick: I don't think we should use underscores in the names of the library methods

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/MsSql/MicrosoftDataSqlClientExerciser.cs
```diff
@@ -0,0 +1,293 @@
...
+                }
+            }
+
+            return string.Join(",", teamMembers);
+        }
+
+        [LibraryMethod]
+        [Transaction]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        public async Task<string> MsSqlAsync_WithParameterizedQuery(string tableName, bool paramsWithAtSign)
```
jaffinito, line 162: Same minor nitpick

nr-ahemsath:
@vuqtran88 So, I noticed this:

```
$ diff MicrosoftDataSqlClientExerciser.cs SystemDataExerciser.cs
3a4,5
> #if NETFRAMEWORK
>
8c10
< using Microsoft.Data.SqlClient;
---
> using System.Data.SqlClient;
10d11
<
18c19
<     public class MicrosoftDataSqlClientExerciser
---
>     public class SystemDataExerciser
20d20
<
153a154
>
215d215
<
293a294,295
>
> #endif
```

In other words, the two exerciser libraries are identical except for their names and which namespace (client library) they are using.

Is there any way to avoid the code duplication?  Can C# code conditionally have different `using`s at runtime?

tests/Agent/IntegrationTests/UnboundedApplications/BasicMvcApplication/Controllers/MsSqlController.cs
```diff
@@ -10,7 +10,6 @@
 using System.Data.SqlClient;
```
chynesNR, line 10: I don't think we need this file at all any more, do we?

vuqtran88:
> @vuqtran88 So, I noticed this:
> 
> ```
> $ diff MicrosoftDataSqlClientExerciser.cs SystemDataExerciser.cs
> 3a4,5
> > #if NETFRAMEWORK
> >
> 8c10
> < using Microsoft.Data.SqlClient;
> ---
> > using System.Data.SqlClient;
> 10d11
> <
> 18c19
> <     public class MicrosoftDataSqlClientExerciser
> ---
> >     public class SystemDataExerciser
> 20d20
> <
> 153a154
> >
> 215d215
> <
> 293a294,295
> >
> > #endif
> ```
> 
> In other words, the two exerciser libraries are identical except for their names and which namespace (client library) they are using.
> 
> Is there any way to avoid the code duplication? Unfortunately we can have different `using`s at runtime, but maybe there is another solution.

Good point. I'm sure we can do something to remove the duplication.

tests/Agent/IntegrationTests/UnboundedApplications/BasicMvcApplication/Controllers/MsSqlController.cs
vuqtran88: We can remove this controller after the OLEDDB and ODBC tests are migrated. I don't think it is hard to migrate. I just haven't migrated them yet :-)

vuqtran88:
> @vuqtran88 So, I noticed this:
> 
> ```
> $ diff MicrosoftDataSqlClientExerciser.cs SystemDataExerciser.cs
> 3a4,5
> > #if NETFRAMEWORK
> >
> 8c10
> < using Microsoft.Data.SqlClient;
> ---
> > using System.Data.SqlClient;
> 10d11
> <
> 18c19
> <     public class MicrosoftDataSqlClientExerciser
> ---
> >     public class SystemDataExerciser
> 20d20
> <
> 153a154
> >
> 215d215
> <
> 293a294,295
> >
> > #endif
> ```
> 
> In other words, the two exerciser libraries are identical except for their names and which namespace (client library) they are using.
> 
> Is there any way to avoid the code duplication? Unfortunately we can have different `using`s at runtime, but maybe there is another solution.
@nr-ahemsath 
I spent some time to address this but I'm sure I can find a good way. The challenge is we use `System.Data` for only FW and `Microsoft.Data.SqlClient` for both Core and FW. They both share usage in FW and I don't know how to make class to choose which one to use in FW.

nr-ahemsath:
@vuqtran88 Yes, I agree.

tests/Agent/IntegrationTests/UnboundedApplications/BasicMvcApplication/Controllers/MsSqlController.cs
vuqtran88: I also realized that I was unsuccessfully migrating the `EnterpriseLibraryMsSqlTests` tests due to some dependencies  errors so we might not yet be able to remove this file.

chynesNR (approved):

PR #1278, opened 2022-10-17
MVS updates for mongocsharpdriver/mongodb.driver

nr-ahemsath:
## Description

This PR:

1. Reverts a recent change to ignore assemblies starting with `net3`.  This was causing us to lose visibility into the `mongocsharpdriver` package because its assemblies are `net35`.  (Note: I don't know if ignoring `net3` was put in to fix a specific bug or if it just seemed like a good idea at the time; if it was to fix a specific bug we'll need to discuss further.)
2. Adds a new, optional instrumentation set config property called "DownloadLatest" (`download-latest: true/false` in YAML).  This controls whether to download and scan the latest NuGet package for the package name.  The default value is true and preserves the original behavior.  I set it to "false" for `mongocsharpdriver` because we don't care what MongoDB is doing for this package in 2.x+.
3. Updates the latest versions scanned for MongoDB.Driver

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/tests/Agent/MultiverseTesting/ConsoleScanner/InstrumentationSet.cs
+++ b/tests/Agent/MultiverseTesting/ConsoleScanner/InstrumentationSet.cs
@@ -15,5 +15,7 @@ namespace NewRelic.Agent.ConsoleScanner
         public List<NugetSet> NugetPackages { get; set; }
 
         public List<string> LocalAssemblies { get; set; }
+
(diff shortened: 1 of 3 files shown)


chynesNR:
Hmm, that's an interesting point about mongocsharpdriver. 1.10.0 _only_ has a net35 target, which we no longer support. Should we update that dependency?

PR #1280, opened 2022-10-20
System.Data.SqlClient (.NET Core/5+) tests

nr-ahemsath:
## Description

Add integration tests for `System.Data.SqlClient` in .NET Core/5+.
Testing driver package versions 4.4.0 through 4.8.4.  4.4.0 is the oldest version where our explain plan feature works.

I also renamed other tests that already existed to help normalize the test class names according to the following pattern:
TestFileName_ExerciserName_{Options}_TargetFrameworkMoniker

I added some comments to explain why we don't have ODBC or OleDb tests for System.Data.SqlClient (our instrumentation currently doesn't support .NET Core).


chynesNR (approved):
Looks good! Thanks for the ticket on the missing instrumentation.

PR #1284, opened 2022-10-24
Update Custom Event sample size default value.

vuqtran88:
## Description

Update Custom Event sample size default value from 10,000 to 30,000.

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New Features
 * Custom Event Limit Increase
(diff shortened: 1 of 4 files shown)


src/Agent/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### New Features
+* Custom Event Limit Increase
+  * This version increases the default limit of custom events from X events per minute to Y events per minute. In the scenario that custom events were being limited, this change will allow more custom events to be sent to New Relic. There is also a new configurable maximum limit of 100,000 events per minute. To change the limits, see the documentation for [max_samples_stored](https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#custom_events). To learn more about the change and how to determine if custom events are being dropped, see our Explorers Hub [post](https://discuss.newrelic.com/t/send-more-custom-events-with-the-latest-apm-agents/190497). [#1284](https://github.com/newrelic/newrelic-dotnet-agent/pull/1284)
```
chynesNR: We should probably fill in `X` and `Y` here. I believe they were left as variables because some Agents use different values.

chynesNR (dismissed):

src/Agent/CHANGELOG.md
vuqtran88: I also found this right after and made a change.

chynesNR (approved):

PR #1291, opened 2022-10-27
Add app logging custom attributes settings

vuqtran88:
## Description

Add app logging custom attributes settings.


src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1615,6 +1615,38 @@
...
+                                                    <xs:annotation>
+                                                        <xs:documentation>
+                                                            Controls whether or not context attributes filtering is used. Defaults to false.
+                                                        </xs:documentation>
+                                                    </xs:annotation>
+                                                </xs:attribute>
+                                                <xs:attribute name="include" type="xs:string" default="">
+                                                    <xs:annotation>
+                                                        <xs:documentation>
+                                                            A comma-separated and case-insensitive string that contains the context attributes that should be added to the log record. Empty means include all
```
nr-ahemsath, line 1636:
```suggestion
                                                            A comma-separated and case-insensitive string that contains the context attributes that should be added to the log record. Empty means include all.
```

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1615,6 +1615,38 @@
...
+                                                        </xs:documentation>
+                                                    </xs:annotation>
+                                                </xs:attribute>
+                                                <xs:attribute name="exclude" type="xs:string" default="">
+                                                    <xs:annotation>
+                                                        <xs:documentation>
+                                                            A comma-separated and case-insensitive string that contains the context attributes that should NOT be added to the log record. Denylist wins over Allowlist.
```
nr-ahemsath, line 1643:
```suggestion
                                                            A comma-separated and case-insensitive string that contains the context attributes that should NOT be added to the log record. "exclude" wins over "include".
```

nr-ahemsath (changes requested):
Found a few typos to fix.

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1615,6 +1615,38 @@
...
+                                                    <xs:annotation>
+                                                        <xs:documentation>
+                                                            Controls whether or not context attributes filtering is used. Defaults to false.
+                                                        </xs:documentation>
+                                                    </xs:annotation>
+                                                </xs:attribute>
+                                                <xs:attribute name="include" type="xs:string" default="">
+                                                    <xs:annotation>
+                                                        <xs:documentation>
+                                                            A comma-separated and case-insensitive string that contains the context attributes that should be added to the log record. Empty means include all.
```
nr-ahemsath, line 1636:
```suggestion
                                                            A comma-separated list of case-sensitive strings that define the context attributes that should be added to the log record. Empty means include all.
```

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1615,6 +1615,38 @@
...
+                                                        </xs:documentation>
+                                                    </xs:annotation>
+                                                </xs:attribute>
+                                                <xs:attribute name="exclude" type="xs:string" default="">
+                                                    <xs:annotation>
+                                                        <xs:documentation>
+                                                            A comma-separated and case-insensitive string that contains the context attributes that should NOT be added to the log record. "exclude" wins over "include".
```
nr-ahemsath, line 1643:
```suggestion
                                                            A comma-separated list of case-sensitive strings that define the context attributes that should NOT be added to the log record. "exclude" wins over "include".
```

nr-ahemsath (approved):
Approved. 👍

chynesNR (approved):

shegeley:
If you can please show an example (full or just small conceptual piece of code) with on how to add custom attributes to logs? I can't deduct anything from changes on this PR. 

In [Ruby](https://github.com/newrelic/newrelic-ruby-agent/pull/2084) it's really nice: 
```ruby
NewRelic::Agent.add_custom_log_attributes
```
Is there something like this?

jaffinito:
The .NET agent doesn't have an API for adding custom attributes to log events directly.  Instead, if you add attributes/properties to the log event in your logging framework, the agent can be configured to capture those automatically.  You can find details on enabling this feature on [this page](https://docs.newrelic.com/docs/logs/logs-context/net-configure-logs-context-all/#1-agent), about a quarter of the way down (just after the first yellow important box),  It also has some information on how to include or exclude specific properties if you don't want all of them sent up.

[PR where this was added](https://github.com/newrelic/newrelic-dotnet-agent/pull/1317).

shegeley:
Thanks for the answer 

I am not an expert with .NET, the platform and it's setup ways, still can't understand much from the original doc. As far as I've got only "static" attributes (that are encoded in XML) can be added with this agent. 

I believe an example added to this conversation or the agent repo itself would be highly appreciated by anyone who wants to add custom log attributes to the .NET add hussle-free.

jaffinito:
Since each logging framework we support (log4net, NLog, Serilog, Microsoft.Extensions.Logging) often has many different ways to add properties it makes it difficult to add examples that are up to date and extensive enough.  We recommend looking at the individual logging library's documentation for how to add properties to the messages.

PR #1292, opened 2022-10-27
Add arm64 profiler to nuget package

JcolemanNR:
## Description

Adds the linux-arm64 profiler and grpc libraries to the Agent Nuget package. This will allow customers to deploy to ARM using only the nuget package.

The AgentComponents class likely needs to be refactored to be specific to Linux/Windows targets in the future, but I worked around that for now.

### DIFF
--- a/build/ArtifactBuilder/CoreAgentComponents.cs
+++ b/build/ArtifactBuilder/CoreAgentComponents.cs
@@ -8,7 +8,18 @@ namespace ArtifactBuilder
         public CoreAgentComponents(string configuration, string platform, string repoRootDirectory, string homeRootPath)
             : base(configuration, platform, repoRootDirectory, homeRootPath) { }
 
-        protected override string SourceHomeBuilderPath => $@"{HomeRootPath}\newrelichome_{Platform}_coreclr";
+        protected override string SourceHomeBuilderPath
+        {
+            get
+            {
+                // arm64 is only supported on linux, so this is a special case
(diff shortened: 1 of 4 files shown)


build/ArtifactBuilder/CoreAgentComponents.cs
```diff
@@ -111,16 +130,26 @@ protected override void CreateAgentComponents()
...
+            else if (Platform == "arm64") 
+            {
+                LinuxProfiler = $@"{HomeRootPath}\newrelichome_arm64_coreclr_linux\libNewRelicProfiler.so";
+                GRPCExtensionsLibLinux = new[]
+                {
+                    $@"{HomeRootPath}\newrelichome_arm64_coreclr_linux\libgrpc_csharp_ext.arm64.so"
+                };
+            }
+
+            AgentInfoJson = Platform == "arm64" ? null : $@"{SourcePath}\..\src\Agent\Miscellaneous\{Platform}\agentinfo.json";
```
chynesNR, line 152: Does having a `null` here cause a problem? It looks like it gets put in a HashSet, which I know is allowed, but not always expected.

JcolemanNR: This is kind of uncharted territory, since the HashSet isn't actually used in any code for the arm64 platform. I'll change the impl to not add any potential nulls though so this doesn't break something in the future

chynesNR (approved):

PR #1293, opened 2022-10-27
Fix lambda test assertions

JcolemanNR:
## Description

Updates lambda test assertions to match new exception type introduced by https://github.com/newrelic/newrelic-dotnet-agent/pull/1287


nr-ahemsath (approved):
Approved 👍

PR #1299, opened 2022-10-31
Serialize ContextData from LogEventWireModel

vuqtran88:
## Description

Serialize ContextData from LogEventWireModel.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
+++ b/src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
@@ -24,6 +24,7 @@ namespace NewRelic.Agent.Core.JsonConverters
         private const string ErrorStack = "error.stack";
         private const string ErrorMessage = "error.message";
         private const string ErrorClass = "error.class";
+        private const string Context = "context";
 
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
```diff
@@ -100,6 +101,20 @@ private static void WriteJsonImpl(JsonWriter jsonWriter, LogEventWireModelCollec
                     jsonWriter.WriteValue(logEvent.TraceId);
                 }
 
+                if (logEvent.ContextData?.Count > 0)
+                {
+                    jsonWriter.WritePropertyName(Attributes);
```
chynesNR, line 106: I assume it's ok that we're reusing the property name "attributes" here? (It's also in the metadata above.)

vuqtran88: I believe so. In fact, they should match. The outer "attributes" are global attributes, the inner ones are specific to log events.

chynesNR (approved):

nr-ahemsath (approved):
Looks good to me!

PR #1297, opened 2022-10-28
Update RecordLogMessage() API to take getter function for context data

nr-ahemsath:
## Description

Implements https://issues.newrelic.com/browse/NEWRELIC-4758.  Includes unit tests to test context data filtration based on include and exclude lists in config (caveat: wildcard filtering support not implemented yet, will be implemented in a follow-on PR).

# Author Checklist
- [X] Unit tests, Integration tests, and Unbounded tests completed

# Reviewer Checklist
- [ ] Perform code review
- [ ] Pull request was adequately tested (new/existing tests, performance tests)
- [ ] Review Changelog

PR #1300, opened 2022-11-02
Logging custom attributes: attribute include/exclude filtering per the spec

nr-ahemsath:

This is a follow-on PR to #1297, which had a broken (per the agent spec) implementation of log context filtering rules.

See the comments in FilterLogContextData and its helper methods for details on how I've coded this to work.

I did look into reusing the code we already have for filtering other attributes in the agent.  However, that code is very entangled with the specific business of determining what data payloads each attribute is going to be attached to, and it would have been more effort, I think, to refactor that code to be more generic and usable by the log context data filtering feature.

I think that the unit tests cover all the necessary cases but I'd be happy to explore expanding them.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Agent.cs
+++ b/src/Agent/NewRelic/Agent/Core/Agent.cs
@@ -2,10 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using NewRelic.Agent.Api;
-using NewRelic.Agent.Api.Experimental;
 using NewRelic.Agent.Configuration;
 using NewRelic.Agent.Core.AgentHealth;
 using NewRelic.Agent.Core.Aggregators;
+using NewRelic.Agent.Core.Attributes;
 using NewRelic.Agent.Core.BrowserMonitoring;
 using NewRelic.Agent.Core.DistributedTracing;
 using NewRelic.Agent.Core.Logging;
@@ -23,14 +23,12 @@ using NewRelic.Agent.Core.Wrapper.AgentWrapperApi.Synthetics;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 using NewRelic.Core;
 using NewRelic.Core.Logging;
-using NewRelic.SystemExtensions.Collections.Generic;
 using NewRelic.SystemInterfaces;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Text.RegularExpressions;
 using System.Threading;
 
 namespace NewRelic.Agent.Core
@@ -62,6 +60,7 @@ namespace NewRelic.Agent.Core
         private readonly ICATSupportabilityMetricCounters _catMetricCounters;
         private readonly Api.ITraceMetadataFactory _traceMetadataFactory;
         private readonly ILogEventAggregator _logEventAggregator;
+        private readonly ILogContextDataFilter _logContextDataFilter;
         private Extensions.Logging.ILogger _logger;
 
         public Agent(ITransactionService transactionService, ITransactionTransformer transactionTransformer,
@@ -71,7 +70,7 @@ namespace NewRelic.Agent.Core
             IBrowserMonitoringPrereqChecker browserMonitoringPrereqChecker, IBrowserMonitoringScriptMaker browserMonitoringScriptMaker,
             IConfigurationService configurationService, IAgentHealthReporter agentHealthReporter, IAgentTimerService agentTimerService,
             IMetricNameService metricNameService, Api.ITraceMetadataFactory traceMetadataFactory, ICATSupportabilityMetricCounters catMetricCounters,
-            ILogEventAggregator logEventAggregator)
+            ILogEventAggregator logEventAggregator, ILogContextDataFilter logContextDataFilter)
         {
             _transactionService = transactionService;
             _transactionTransformer = transactionTransformer;
@@ -91,6 +90,7 @@ namespace NewRelic.Agent.Core
             _traceMetadataFactory = traceMetadataFactory;
             _catMetricCounters = catMetricCounters;
             _logEventAggregator = logEventAggregator;
+            _logContextDataFilter = logContextDataFilter;
 
             Instance = this;
         }
@@ -98,10 +98,6 @@ namespace NewRelic.Agent.Core
         public IConfiguration Configuration => _configurationService.Configuration;
 
         public Extensions.Logging.ILogger Logger => _logger ?? (_logger = new Logger());
-
-        private List<LogContextDataFilterRule> _includeRuleList;
-        private List<LogContextDataFilterRule> _excludeRuleList;
-        private List<LogContextDataFilterRule> _orderedCludeRuleList;
 
         #region Transaction management
 
@@ -451,11 +447,11 @@ namespace NewRelic.Agent.Core
                 {
                     logEventWireModel = new LogEventWireModel(timestamp, logMessage, normalizedLevel,
                         StackTraces.ScrubAndTruncate(logException, LogExceptionStackLimit), logException.Message, logException.GetType().ToString(),
-                        spanId, traceId, FilterLogContextData(logContextData));
+                        spanId, traceId, _logContextDataFilter.FilterLogContextData(logContextData));
                 }
                 else
                 {
-                    logEventWireModel = new LogEventWireModel(timestamp, logMessage, normalizedLevel, spanId, traceId, FilterLogContextData(logContextData));
+                    logEventWireModel = new LogEventWireModel(timestamp, logMessage, normalizedLevel, spanId, traceId, _logContextDataFilter.FilterLogContextData(logContextData));
                 }
 
                 var transaction = _transactionService.GetCurrentInternalTransaction();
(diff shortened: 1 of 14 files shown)


src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -562,19 +567,82 @@ internal void Detach(bool removeAsync, bool removePrimary)
...
-            else
+
+            // Now apply the rules in order
+            // TODO: maybe optimize by just working with keys as lists of strings until it comes time to return the actual data
+            foreach (var rule in allRulesInOrder)
             {
-                filteredContextData = unfilteredContextData.Where(property => _configurationService.Configuration.ContextDataInclude.Contains(property.Key)).ToDictionary();
+                if (rule.Include)
+                {
+                    filteredContextData = filteredContextData.Concat(unfilteredContextData.Where(x => Regex.IsMatch(x.Key, rule.Text))).ToDictionary();
```
chynesNR, line 594: I'm not sure about using `Regex.IsMatch` for a few reasons:
* These are simple string comparisons (either `Equals` or `StartsWith`) and regex will be slower
* It will break on special characters
* It allows users to actually use a regex, which isn't the intent

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -562,19 +567,82 @@ internal void Detach(bool removeAsync, bool removePrimary)
 
             var filteredContextData = new Dictionary<string, object>();
 
-            // first, handle the "includes"
-            if (_configurationService.Configuration.ContextDataInclude.Count() == 0) // empty "include" list, so include everything
+            // Pseudocode
```
chynesNR, line 570: There's enough specialized code here that you might want to split it out into a new class/file.

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -562,19 +567,82 @@ internal void Detach(bool removeAsync, bool removePrimary)
...
+
+        private List<LogContextDataFilterRule> GetOrderedCludeRuleList()
+        {
+            var includeRuleList = _includeRuleList ?? (_includeRuleList = GetCludeRulesFromConfig(true));
+            var excludeRuleList = _excludeRuleList ?? (_excludeRuleList = GetCludeRulesFromConfig(false));
+
+            // This should work because OrderBy uses a stable sort.  By sorting each list individually,
+            // concactating them, and then sorting again, we get a list of rules ordered by length (specificity),
+            // with includes before excludes for each level of specificity
+            return includeRuleList.OrderBy(rule => rule.Specificity).Concat(
```
chynesNR, line 623: This feels a bit clunky to me. While we probably won't have many include/exclude items, this requires three sorts and relies on (perhaps) non-intuitive behavior of the `OrderBy` function. Maybe there's a simpler way to do this?

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -562,19 +567,82 @@ internal void Detach(bool removeAsync, bool removePrimary)
...
+            var allRulesInOrder = _orderedCludeRuleList ?? (_orderedCludeRuleList = GetOrderedCludeRuleList()); 
+
+            if (_includeRuleList.Count() == 0) // empty "include" list, so include everything
             {
                 filteredContextData = unfilteredContextData;
             }
-            else
+
+            // Now apply the rules in order
+            // TODO: maybe optimize by just working with keys as lists of strings until it comes time to return the actual data
```
chynesNR, line 589: Yeah I'm a little concerned about the performance of doing a Where/Concat/ToDictionary for every rule, if there are lot of rules and/or a lot of logging.

vuqtran88: I feel like if we just do string comparisons on the keys, we might be able to optimize it by caching previous matches somewhere so that the agent doesn't have to do Where/Concat/ToDictionary again.

nr-ahemsath:
Converting to draft while I work on implementing PR feedback changes.

src/Agent/NewRelic/Agent/Core/Agent.cs
nr-ahemsath: Thanks for pointing this out; I've fixed it.

src/Agent/NewRelic/Agent/Core/Agent.cs
nr-ahemsath: Done.

src/Agent/NewRelic/Agent/Core/Agent.cs
nr-ahemsath: My latest push introduces a cache for include/exclude results, and reverses the logic - instead of going through every rule and applying it to the context data dictionary, it goes through every key/value pair in the context dictionary and looks for a cache result or matching rule.  I think this should be more efficient.

nr-ahemsath:
Ready for review again.

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
```diff
@@ -0,0 +1,136 @@
...
+            // Algorithm:
+            // Create list of rules, sorted by most specific to least specific, with excludes first and then includes within each level of specificity
+            // Specificity is just the length of the string, minus any wildcard character at the end
+            // For each context data key/value pair:
+            //   First, check the "clusion cache" to see if we already know whether or not to include or exclude this key, and if so handle it appropriately
+            //   Otherwise, go through the list of rules and see if we find a matching rule to exclude or include, cache the result, and handle it appropriately
+
+            foreach (var kvp in unfilteredContextData)
+            {
+                if (!_clusionCache.ContainsKey(kvp.Key))
```
JcolemanNR, line 55: _clusionCache should be reset during configuration change events as well

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
```diff
@@ -0,0 +1,136 @@
...
+            // 1. Exclude wins over include
+            // 2. Can have a '*' wildcard at the end of an include/exclude definition
+            // 3. "more specific" rules win over "less specific".  Example: include=AB, exclude=A*, AB would be included but not ABC or AAA
+
+            // Algorithm:
+            // Create list of rules, sorted by most specific to least specific, with excludes first and then includes within each level of specificity
+            // Specificity is just the length of the string, minus any wildcard character at the end
+            // For each context data key/value pair:
+            //   First, check the "clusion cache" to see if we already know whether or not to include or exclude this key, and if so handle it appropriately
+            //   Otherwise, go through the list of rules and see if we find a matching rule to exclude or include, cache the result, and handle it appropriately
```
JcolemanNR, line 51: general concern (which will be partially alleviated when config changes reset all the cached values): If a customer continually generates unique key names in log events then every value ever generated will be cached for the lifecycle of the application. This could use a significant amount of memory....

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
```diff
@@ -0,0 +1,136 @@
...
+using NewRelic.Agent.Configuration;
+
+namespace NewRelic.Agent.Core.Attributes
+{
+    public interface ILogContextDataFilter
+    {
+        Dictionary<string, object> FilterLogContextData(Dictionary<string, object> contextData);
+    }
+
+    public class LogContextDataFilter : ILogContextDataFilter
```
JcolemanNR, line 16: This filter needs its own test harness to validate all conditions and use cases. This will (hopefully) prevent regressions if anyone attempts to refactor this in the future.

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            var filteredContextData = new Dictionary<string, object>();
+
+            // Pseudocode
+            // Include/exclude rules:
+            // 1. Exclude wins over include
+            // 2. Can have a '*' wildcard at the end of an include/exclude definition
+            // 3. "more specific" rules win over "less specific".  Example: include=AB, exclude=A*, AB would be included but not ABC or AAA
+
```
JcolemanNR, line 45: Not serious comment: We could make this even more efficient if we used a TRIE! (don't refactor to use a TRIE)

tests/Agent/UnitTests/Core.UnitTest/Wrapper/AgentWrapperApi/AgentWrapperApiTests.cs
```diff
@@ -1761,14 +1763,16 @@ public void RecordLogMessage_WithTransaction_WithException_ThatHasHadLogsHarvest
...
+        [TestCase(true, "key1", "", "key1", TestName = "Explicit include, empty exclude")]
+        [TestCase(true, "key1,key2", "key2", "key1", TestName = "Explicit include and exclude")]
+        [TestCase(true, "", "key1", "key2", TestName = "Empty include, explicit exclude")]
+        [TestCase(true, "", "key1,key2", "", TestName = "Exclude all")]
+        [TestCase(true, "key*", "", "key1,key2", TestName = "Wildcard include, empty exclude")]
+        [TestCase(true, "key1", "key*", "key1", TestName = "More-specific explicit include overrides widlcard exclude")]
+        [TestCase(true, "key3", "", "", TestName = "Explicit include of non-existent key")]
+        [TestCase(false, "", "", "", TestName = "Context data disabled with empty include and exclude")]
+        [TestCase(false, "key1,key2", "key2", "", TestName = "Context data disabled overrides explicit include and exclude")]
         public void RecordLogMessage_ContextDataConfiguration(bool contextDataEnabled, string includeList, string excludeList, string expectedAttributeNames)
```
JcolemanNR, line 1776: This test coverage is good, but I would make a whole new test harness that focuses on the capabilities of the filter

JcolemanNR (changes requested):

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                rules.Add(new LogContextDataFilterRule(configRule, isIncludeRules));
+            }
+            return rules;
+        }
+
+        private List<LogContextDataFilterRule> GetOrderedCludeRuleList()
+        {
+            var includeRuleList = _includeRuleList ?? (_includeRuleList = GetCludeRulesFromConfig(true));
+            var excludeRuleList = _excludeRuleList ?? (_excludeRuleList = GetCludeRulesFromConfig(false));
```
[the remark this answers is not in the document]
nr-ahemsath: Good catch, done

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
```diff
@@ -0,0 +1,136 @@
...
+                    filteredContextData[kvp.Key] = kvp.Value;
+                }
+            }
+
+            return filteredContextData;
+        }
+
+        private bool GetClusionResult(string key)
+        {
+            var allRulesInOrder = _orderedCludeRuleList ?? (_orderedCludeRuleList = GetOrderedCludeRuleList());
```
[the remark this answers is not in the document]
nr-ahemsath: Good catch, done

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
nr-ahemsath: I added a max cache size, logic and hardcoded size of 1000 cribbed from AttributeFilter.cs.  I also updated the cache to be a ConcurrentDictionary since that is what AttributeFilter.cs uses.

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
nr-ahemsath: Done.

tests/Agent/UnitTests/Core.UnitTest/Wrapper/AgentWrapperApi/AgentWrapperApiTests.cs
nr-ahemsath: Done.

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
nr-ahemsath: Done

src/Agent/NewRelic/Agent/Core/Attributes/LogContextDataFilter.cs
```diff
@@ -0,0 +1,156 @@
...
+            // Create list of rules, sorted by most specific to least specific, with excludes first and then includes within each level of specificity
+            // Specificity is just the length of the string, minus any wildcard character at the end
+            // For each context data key/value pair:
+            //   First, check the "clusion cache" to see if we already know whether or not to include or exclude this key, and if so handle it appropriately
+            //   Otherwise, go through the list of rules and see if we find a matching rule to exclude or include, cache the result, and handle it appropriately
+
+            foreach (var kvp in unfilteredContextData)
+            {
+                bool clusionResult;
+                if (!_clusionCache.ContainsKey(kvp.Key))
```
chynesNR, line 60: Just FYI, you could save a lookup by using `_clusionCache.TryGetValue()`

nr-ahemsath: Good idea, done.

PR #1301, opened 2022-11-03
Updates attribute limit error log messages with transaction guid

jaffinito:
## Description

When troubleshooting attribute limit errors in busy applications, it is difficult to know which transaction the errors are coming from.  This PR updates those log messages to include the transaction guid at finest like most other log messages,

- Attribute limit error messages include the transaction guid
- Updates AttributeValueCollection.cs with a new constructor that takes and stores the transaction guid.  It leaves the other constructors alone.
- Updates AttributeValueCollection.ValidateCollectionLimits to use a helper method to decide when the new finest logs format should be used
- Updates TransactionMetadata to have an explicit constructor that takes in and stores the transaction guid while also instantiating the AttributeValueCollection using its new constructor.  
- Updates Core/Transactions/Transaction.cs to use the new TransactionMetadata constructor to pass in the guid.
- Updates all the unit tests to use the new TransactionMetadata constructor

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New Features
 * The [NewRelic.Agent NuGet package](https://www.nuget.org/packages/NewRelic.Agent) now includes the Linux Arm64 profiler. This can be found in the `newrelic/linux-arm64` directory. Configure your `CORECLR_PROFILER_PATH` environment variable to use this version of the profiler when deploying to linux ARM64 targets.
-* When finest logs is enabled, the transaction guid will be applied to attribute limit log messages, if present.
+* When finest logs are enabled, the transaction guid will be applied to attribute limit log messages, if present.
 
 ### Fixes
 *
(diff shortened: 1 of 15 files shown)


chynesNR (approved):

src/Agent/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New Features
 * The [NewRelic.Agent NuGet package](https://www.nuget.org/packages/NewRelic.Agent) now includes the Linux Arm64 profiler. This can be found in the `newrelic/linux-arm64` directory. Configure your `CORECLR_PROFILER_PATH` environment variable to use this version of the profiler when deploying to linux ARM64 targets.
+* When finest logs is enabled, the transaction guid will be applied to attribute limit log messages, if present.
```
JcolemanNR, line 11:
```suggestion
* When finest logs are enabled, the transaction guid will be applied to attribute limit log messages, if present.
```

src/Agent/NewRelic/Agent/Core/Attributes/AttributeValueCollection.cs
```diff
@@ -275,6 +288,18 @@ public void MakeImmutable()
 
             IsImmutable = true;
         }
+
+        private void LogTransactionIfFinest(string message)
```
JcolemanNR, line 292: I would suggest just logging the transaction guid always. Also, when would there not be a transaction guid?

JcolemanNR (changes requested):
Definitely change the wording in the changelog, consider feedback about always logging guid

chynesNR (approved):

src/Agent/NewRelic/Agent/Core/Attributes/AttributeValueCollection.cs
jaffinito: There are several different constructors for this class and only one of them (the new one) takes in the guid which is used by TransactionMetadata.   The AttributeValueCollection class is used by many different parts of the agent that don't have access to the guid so this change is limited to just the transaction bit to keep it from spider-webbing.

PR #1303, opened 2022-11-10
Dotty improvements

nr-ahemsath:
## Description

Improvements to our tool that checks for new releases of our core instrumented libraries:

1. Add console logging so you can see what the tool is doing.
2. Changed the name of the env var for the Slack webhook from "webhook" to "DOTTY_WEBHOOK".
3. Added env var "DOTTY_DAYS_TO_SEARCH" to allow setting the number of days of NuGet history to search for package updates.  Defaults to the previously hardcoded value of 1.
4. Added env var "DOTTY_TEST_MODE" to allow running the tool without sending the Slack notification.
5. Modified the action workflow to allow setting these env vars when run manually.
6. Added monitoring the tool with the .NET agent:
   1. Added a newrelic.config with the necessary sendDataOnExit config that gets copied to the output dir.
   2. Modified the action workflow to install the (latest) .NET agent via nuget
   3. Added the necessary profiling env vars

### DIFF
--- a/.github/workflows/nuget_slack_notifications.yml
+++ b/.github/workflows/nuget_slack_notifications.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Check for updates to core technology packages
         run: dotnet run --project ${{ env.scan-tool-path }} ${{ env.nugets }}
         env:
-            webhook: ${{ secrets.SLACK_NUGET_NOTIFICATIONS_WEBHOOK }}
+            DOTTY_WEBHOOK: ${{ secrets.SLACK_NUGET_NOTIFICATIONS_WEBHOOK }}
             nugets:
                  "system.data.sqlclient
                  microsoft.data.sqlclient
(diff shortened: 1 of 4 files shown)


chynesNR:
Can we get a Slack notification if Dotty encounters an error? Having it in the log is good but it might take us a while to notice.

nr-ahemsath:
> Can we get a Slack notification if Dotty encounters an error? Having it in the log is good but it might take us a while to notice.

That's a good idea.  Just pushed a change.  Looking at the code, I think there's only a couple of places where an error can occur:
1. Getting and processing data from NuGet
2. Sending the Slack notification itself

If it's a problem with sending the Slack notification, we can't send a Slack notification, so I only addressed cause 1.  Since the agent will be instrumenting the process, we should also get errors in New Relic, and can explore getting notifications on those as well.

chynesNR (approved):
Looks good!

PR #1305, opened 2022-11-16
Adds context data to forwarded log events with log4net

jaffinito:
## Description

TODO: needs attribute filtering tests - will be in a follow up PR

Agent
- Updates Agent.cs and the experimental Agent API to handles check for and process context data (attributes)
- Adds configuration options to enabled context data (off by default) and set include/exclude lists - all with env vars and server-side overrides
- Updates LogEventWireModel and LogEventWireModelCollectionJsonConverter to handle the new attribute data
- Updates Log4netWrapper with a new method, GetContextData to capture attributes from the GetProperties method.  This uses GetProperties method instead of the Properties property since it will capture the other contexts along with the event specific data.  Important:  GetProperties returns a new merged IDictionary, not the real Properties value so we need to have DecorateLogMessage use Properties.
- Updates the other logging wrappers to support the new API with a placeholder.

Tests
- Adds many new unit tests
- Adds context data integration tests for log4net
- Updates logging adapters and LoggingTester to support context data
- Updates Assertions.cs with new assertions to test if log lines exist or not.
- Updates Assertions.cs to allow not checking Attribute countssince log4net automatically adds some Properties by default.  Used by the ContextData tests.
- LogData model now has attributes
- NewRelicConfigModifier added for new config settings
- Adds missing net481 csproj entry for log4net

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Log4NetLogging/Log4netWrapper.cs
```diff
@@ -18,7 +18,8 @@ public class Log4netWrapper : IWrapper
         private static Func<object, string> _getRenderedMessage;
         private static Func<object, DateTime> _getTimestamp;
         private static Func<object, Exception> _getLogException;
-        private static Func<object, IDictionary> _getProperties;
+        private static Func<object, IDictionary> _getGetProperties; // calls GetProperties method
+        private static Func<object, IDictionary> _getProperties; // getter for Properties property
```
chynesNR, line 22: It doesn't look like this is being used?

jaffinito: We are using both _getGetProperties and _getProperties.
_getGetProperties  is in GetContextData
_getProperties is in DecorateLogMessage

chynesNR: Ah! My mistake.

jaffinito:
There are gong to be failures in some of the logging tests due to the new Info method that takes context data throwing NotImplementedException.  This is expected.

nr-ahemsath (approved):
Looks good! 👍

PR #1306, opened 2022-11-17
NLog instrumentation adds context data to log event data

nr-ahemsath:
## Description

1. Update the NLog instrumentation wrapper to access a log event's `IDictionary<object,object> Properties` and add those values to the log event data being recorded in the agent.
2. Add subclasses of `ContextDataTestsBase` (integration tests) for NLog
3. Update `LogEventWireModelCollectionJsonConverter` to handle non-primitive value objects.   Even though we're not testing this in integration tests, I did ad-hoc testing of this use case, and this change is necessary to avoid errors being logged in the agent and log data being lost if the customer's use case looks like this:

```
class Person
{
    public int Id { get; set; }
    public string Name { get; set; }
}

var message = $"Log message with string key and object value";
LogEventInfo logEvent = new LogEventInfo(LogLevel.Info, null, message);
logEvent.Properties["person"] = new Person { Id = 1, Name = "Alice" };
logger.Log(logEvent);
```

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
+++ b/src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
@@ -109,7 +109,16 @@ namespace NewRelic.Agent.Core.JsonConverters
                     foreach (var item in logEvent.ContextData)
                     {
                         jsonWriter.WritePropertyName(Context + "." + item.Key);
-                        jsonWriter.WriteValue(item.Value);
+                        string contextValueJson;
+                        try
+                        {
+                            contextValueJson = JsonConvert.SerializeObject(item.Value);
+                        }
+                        catch
+                        {
+                            contextValueJson = item.Value.GetType().ToString();
+                        }
+                        jsonWriter.WriteRawValue(contextValueJson);
                     }
 
                     jsonWriter.WriteEndObject();
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
@@ -104,7 +104,8 @@ namespace NewRelic.Providers.Wrapper.NLogLogging
                 else
                 {
                     // This probably isn't a good idea, should just have an "ignore" fallback
-                    contextData[property.Key.ToString()] = property.Value;
+                    var insertKey = property.Key.ToString();
+                    contextData[insertKey] = property.Value;
                 }
             }
 
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
@@ -93,20 +93,7 @@ namespace NewRelic.Providers.Wrapper.NLogLogging
             var properties = getPropertiesDictionary(logEvent);
             foreach (var property in properties)
             {
-                if (property.Key is string keyName)
-                {
-                    contextData[keyName] = property.Value;
-                }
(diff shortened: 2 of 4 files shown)


src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
```diff
@@ -109,7 +109,16 @@ private static void WriteJsonImpl(JsonWriter jsonWriter, LogEventWireModelCollec
...
                         jsonWriter.WritePropertyName(Context + "." + item.Key);
-                        jsonWriter.WriteValue(item.Value);
+                        string contextValueJson;
+                        try
+                        {
+                            contextValueJson = JsonConvert.SerializeObject(item.Value);
+                        }
+                        catch
+                        {
+                            contextValueJson = item.Value.GetType().ToString();
```
chynesNR, line 119: Did you mean to use `GetType()` here?

nr-ahemsath: Yes, my thinking is that if we can get valid JSON from JsonConvert.SerializeObject for whatever the value object is, as a fallback just send the string representation of the object's type.  I suppose that's what `Object.ToString()` does anyway if there's no override of ToString() for that object, so maybe we could skip the GetType() altogether.

chynesNR: Yeah, that's kind of what I was thinking. It's possible they've implemented `ToString()` which would be more useful than the type. But if we can't serialize it, it'll most likely end up just being the type.

JcolemanNR (approved):
The parts I didn't write look good :)

src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
```diff
@@ -109,7 +109,25 @@ private static void WriteJsonImpl(JsonWriter jsonWriter, LogEventWireModelCollec
...
+                        {
+                            // If JsonConvert can't serialize it, maybe it has a ToString()
+                            try
+                            {
+                                contextValueJson = item.Value.ToString();
+                            }
+                            catch
+                            {
+                                // If that didn't work, just use the type name
+                                contextValueJson = item.Value.GetType().ToString();
```
jaffinito, line 127: I don't really like just ToStringing something and dropping the value in even if its just as a fallback.  In other areas of the agent we use switches with specific actions for different types and for complex types that action is usually "drop".

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1637,7 +1637,7 @@
                                                         </xs:documentation>
                                                     </xs:annotation>
                                                 </xs:attribute>
-                                                <xs:attribute name="exclude" type="xs:string" default="">
+                                                <xs:attribute name="exclude" type="xs:string" default="SpanId,TraceId,ParentId">
```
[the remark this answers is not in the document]
JcolemanNR: This is hidden in the schema only. We currently don't write any of the log forwarding defaults to config files (I believe).

src/Agent/NewRelic/Agent/Core/JsonConverters/LogEventWireModelCollectionJsonConverter.cs
nr-ahemsath: Discussed as a team and we agreed to keep the current behavior.

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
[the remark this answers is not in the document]
jaffinito: 👍

jaffinito (approved):

PR #1315, opened 2022-12-05
Replaced Trace calls with Debug Calls

ajmcateer:
## Changes Trace calls to debug calls

While tracking down some thread starvation issues with our app, I noticed this 

```
Ben.Diagnostics.BlockingMonitor: Warning: Blocking method has been invoked and blocked, this can lead to threadpool starvation.
   at System.Threading.Monitor.ReliableEnter(Object obj, Boolean& lockTaken)
   at System.Threading.Monitor.Enter(Object obj, Boolean& lockTaken)
   at System.Diagnostics.TraceInternal.WriteLine(String message)
   at System.Diagnostics.Trace.WriteLine(String message)
   at NewRelic.Api.Agent.NewRelic.IncrementCounter(String name)
```

Trace is enabled by default in release mode and unfortunately it contains a blocking call.

[MSDN documentation](https://learn.microsoft.com/en-us/visualstudio/debugger/project-settings-for-csharp-debug-configurations-dotnetcore?view=vs-2022)

[Link to TraceInternal.WriteLine](https://github.com/dotnet/runtime/blob/7c9b2522ab52e6452e3d39bca9f91111d73bed53/src/libraries/System.Diagnostics.TraceSource/src/System/Diagnostics/TraceInternal.cs#L530)

While its possible to side step this issue, by clearing the Listeners list and setting Global Lock to false. I figured switching these calls to Debug which is not enabled by default on C# release builds wouldn't hurt as well.

In case anyone stumbles onto this issue and wants the code to sidestep the issue. This will disable all trace logging in your release or debug build, so make sure you aren't relying on it in some way.

```
Trace.Listeners.Clear();
Trace.UseGlobalLock = false; 
```

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic.Api.Agent/NewRelic.cs
+++ b/src/Agent/NewRelic.Api.Agent/NewRelic.cs
@@ -42,7 +42,7 @@ namespace NewRelic.Api.Agent
         {
             try
             {
-                System.Diagnostics.Trace.WriteLine($"NewRelic.InitializePublicAgent({nameof(publicAgent)})");
+                System.Diagnostics.Debug.WriteLine($"NewRelic.InitializePublicAgent({nameof(publicAgent)})");
             }
             catch
             {
@@ -86,7 +86,7 @@ namespace NewRelic.Api.Agent
         {
             try
             {
-                System.Diagnostics.Trace.WriteLine($"NewRelic.RecordMetric({nameof(name)},{nameof(value)})");
+                System.Diagnostics.Debug.WriteLine($"NewRelic.RecordMetric({nameof(name)},{nameof(value)})");
             }
             catch
             {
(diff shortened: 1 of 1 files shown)


chynesNR:
Hi @ajmcateer - thanks for your submission! I think there may be some confusion about what's happening in your test. The Agent API function that calls `System.Diagnostics.Trace.WriteLine` is a shim. At runtime, our Profiler (the native component) swaps that function out for the real API function in the .NET Agent. That function, which does the work, does not call into `System.Diagnostics`: https://github.com/newrelic/newrelic-dotnet-agent/blob/1525135b38f93938ddb567394a1031c47e1ff15e/src/Agent/NewRelic/Agent/Core/Api/AgentApiImplementation.cs#L159
The fact that you are seeing it on your call stack suggests that the profiler is not loaded and did not perform the swap. Since you were doing performance testing, is it possible that you had another profiler loaded? .NET only allows one profiler to be loaded at a time, which would mean the .NET Agent was not running during your test.

chynesNR:
Hi again @ajmcateer - I'm going to close this, because I don't believe anything needs to be changed, but if you're still having issues, please open a ticket so we can assist you. Thanks!

Issue #1202, opened 2022-07-27
Agent log showing error: An error occurred while attempting......


vuqtran88:
**Description**
Customer reported a strange issue with their app when recycle app pool. They do this on daily bases to clear built up memory and also to keep the performance of the app up. Once the server has been restarted IIS is unstable.

Customer also reported that the only solution for this had been to replace the newrelic.config file in every website – no changes, just put the exact same file back. Once this is done the Application Pools stabilise and CPU usage drops. They also saw memory issues with the app pools dropping memory until this config is replaced.

Agent is writing thousands of log entries with the below error. 
```
2022-07-25 16:31:16,464 NewRelic  ERROR: [pid: 106828, tid: 20] Unable to connect to the New Relic service at collector.newrelic.com:443 (Proxy: newrelic-proxy.advhc.net:3128) : System.Web.HttpException (0x80004005): An error occurred while attempting to impersonate.  Execution of this request cannot continue.
   at System.Web.ImpersonationContext.ImpersonateToken(HandleRef token)
   at System.Web.ApplicationImpersonationContext..ctor()
   at System.Web.Util.FileUtil.PhysicalPathStatus(String physicalPath, Boolean directoryExistsOnError, Boolean fileExistsOnError, Boolean& exists, Boolean& isDirectory)
   at System.Web.CachedPathData.GetConfigPathData(String configPath)
   at System.Web.CachedPathData.GetConfigPathData(String configPath)
   at System.Web.CachedPathData.GetApplicationPathData()
   at System.Web.Configuration.HttpConfigurationSystem.GetApplicationSection(String sectionName)
   at System.Web.Configuration.HttpConfigurationSystem.GetSection(String sectionName)
   at System.Web.Configuration.HttpConfigurationSystem.System.Configuration.Internal.IInternalConfigSystem.GetSection(String configKey)
   at System.Configuration.ConfigurationManager.GetSection(String sectionName)
   at System.Configuration.ConfigurationManager.get_AppSettings()
   at System.Net.ConnectStream.GetResponseDrainTimeout()
   at System.Net.ConnectStream.DrainSocket()
   at System.Net.ConnectStream.CloseInternal(Boolean internalCall, Boolean aborting)
   at System.Net.ConnectStream.System.Net.ICloseEx.CloseEx(CloseExState closeState)
   at System.Net.ConnectStream.Dispose(Boolean disposing)
   at System.IO.Stream.Close()
[...]

Issue #897, opened 2022-01-27
The .Net Agent alongside usage of HttpClient is causing Unknown Socket Error in the .Net runtime


sliderhouserules:
**Description**

We have a .Net 5/6 application that is experiencing an inordinate number of `SocketExceptions`. We use a Web API base template and implement background processing that interacts heavily with AWS resources. We are using New Relic functionality extensively throughout.

Removing any and all calls to the New Relic Agent, and all New Relic instrumentation, has alleviated the issue almost completely.

We have an issue filed directly with Microsoft that details the issue [here](https://github.com/dotnet/runtime/issues/64305).

**Expected Behavior**

The NR .Net Agent should play nice with the usage of `HttpClient` and the `IHttpClientFactory` and not contribute to, or cause `SocketExceptions` deep in the .Net runtime.

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/diagnostics-cli-nrdiag/diagnostics-cli-nrdiag/) results**

We have gone through our code intensively to make sure we are properly using the `IHttpClientFactory` which uses pooling under the hood to correctly use `HttpClient`/`HttpRequestMessage`. This is able to be plugged into all of the AWS clients so they are using it properly as well. We have removed all (other) 3rd party libraries that we can determine do not properly use these patterns. The errors persist.

Since we use New Relic extensively, we refrained from removing it. Until I found [this](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs#L157). You are apparently using `WebRequest`, which has been deprecated and is recommended to not be used, as it is known to cause socket exhaustion and `SocketExceptions`.

Once we completely removed all calls to the agent, and any references to NR namespaces and projects, _the problems have ceased_. Several hours of load testing has confirmed this.

**Steps to Reproduce**

I don't have a sample application to share with you, I can only suggest that:
- a basic Web API project (`dotnet create webapi`) be created
- some background processing implemented in `IHostedService`s that:
interacts with external web resources (using GETs, PUTs, and POSTs to match our use case)
properly uses either singleton `HttpClient` or the `IHttpClientFactory`
- test both with and without usage of the NR agent integrated in there somewhere (preferably liberally)

[...]

PR #1325, opened 2022-12-15
Use HTTPClient for agent communications with New Relic services

nr-ahemsath:
## Description

See issues #897 and #1202 for background, but this is more of a technical debt item than anything else.  The agent currently uses `System.Net.WebRequest` to communicate with New Relic back-end services.  This library is deprecated by MSFT in favor of `System.Net.Http.HttpClient`.  This PR replaces our use of WebRequest with HTTPClient.

As part of getting our integration tests to pass with this change, the `MockNewRelic` service which is used by some tests (that need to simulate agent commands being sent from the collector to the agent) is updated to use the default .NET development SSL certificate rather than our own (expired) self-signed certificate, and our CI workflow is updated to trust the development certificate.

Other integration testing documentation is updated as well since we missed updating it for .NET 7, and to add the step of needing to trust the development SSL certificate with the `dotnet dev-certs https --trust` command when running integration tests locally.

This PR is a draft, as we still need to do performance testing.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
```diff
@@ -46,76 +51,108 @@ private enum Direction
...
                 {
-                    // Log that the payload is being dropped. 
+                    // Log that the payload is being dropped
                     Log.ErrorFormat("Request({0}): Dropped large payload: size: {1}, max_payload_size_bytes={2}", requestGuid, requestPayload.Data.Length, _configuration.CollectorMaxPayloadSizeInBytes);
                     _agentHealthReporter.ReportSupportabilityPayloadsDroppeDueToMaxPayloadSizeLimit(method);
                     return "{}"; // mimics what is sent by NoOpCollectorWire
                 }
 
-                var response = SendRequest(request, requestPayload.Data);
+                var httpClient = _httpClientFactory.CreateClient(connectionInfo.Proxy);
```
tippmar-nr, line 92: Since `HttpClient` implements `IDisposable`, this block (down to line 144 or so) should probably be wrapped in a `using()` to ensure the client gets disposed when it goes out of scope.

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
```diff
@@ -46,76 +51,108 @@ private enum Direction
...
+                if (_configuration.PutForDataSend)
+                {
+                    request.Method = HttpMethod.Put;
+                }
+                else
+                {
+                    request.Method = HttpMethod.Post;
+                }
+
+                var response = httpClient.SendAsync(request).GetAwaiter().GetResult();
```
tippmar-nr, line 129: This is the line I mentioned in the meeting this morning - in certain circumstances - can lead to a deadlock and/or thread pool starvation. Unfortunately, at this point in my tenure at NR, I don't have enough context to know whether this bit of code ever *might* be subject to the potential cases where a deadlock would occur. So this may be a non-issue.

Some relevant articles: 
https://stackoverflow.com/questions/39007006/is-getawaiter-getresult-safe-for-general-use
https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
https://www.nikouusitalo.com/blog/why-is-getawaiter-getresult-bad-in-c/

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
```diff
@@ -157,98 +194,60 @@ private CollectorRequestPayload GetRequestPayload(byte[] bytes)
...
-            using (var outputStream = request.GetRequestStream())
+            try
             {
-                if (outputStream == null)
-                    throw new NullReferenceException("outputStream");
-
-                outputStream.Write(payload, 0, (int)request.ContentLength);
-            }
-        }
+                var responseStream = response.Content.ReadAsStreamAsync().GetAwaiter().GetResult();
```
tippmar-nr, line 201: Same issue here with `.GetAwaiter().GetResult()`

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
```diff
@@ -46,76 +51,108 @@ private enum Direction
...
                 throw;
             }
-        }
-
-        private static string SendRequest(WebRequest request, byte[] requestPayloadData)
-        {
-            SendRequestPayload(request, requestPayloadData);
-            var response = GetResponse(request);
-            return response;
+            catch (Exception ex)
```
tippmar-nr, line 152: This `catch` block is unnecessary - no reason to catch and rethrow, as it's the same as not catching at all.

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
chynesNR: Further, is it necessary to create a new client object for every Send?

src/Agent/NewRelic/Agent/Core/DataTransport/IHttpClientFactory.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+
+namespace NewRelic.Agent.Core.DataTransport
+{
+    public interface IHttpClientFactory
```
tippmar-nr, line 10: New guy ignorance time: Why are we using a home-grown `IHttpClientFactory` rather than the built-in `Microsoft.Extensions.Http.IHttpClientFactory`?

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
chynesNR, line 152: In the old code it looks like other kinds of exceptions were ignored. Is there someone above that will catch and/or log this?

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
tippmar-nr, line 129: `response` should be inside a `using()` block to ensure it is correctly disposed.

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
tippmar-nr: (I deleted my original comment - once I dug through the implementation of `HttpClientFactory`, I saw that we are re-using a single `httpClient` instance for the duration of the application, which is what MS recommends. 

Maybe it wouldn't hurt to rename the method from `CreateClient()` to `GetClient()` ? (I'm also a little bit bothered by the creation of a `IHttpClientFactory` interface that might (did, in my case) cause confusion with the Microsoft implementation...)

chynesNR: Whoops, I made the same mistake, sorry.

src/Agent/NewRelic/Agent/Core/DataTransport/IHttpClientFactory.cs
vuqtran88: Mostly I tried to avoid referencing external assemblies as much as possible so that we didn't have to ILRepack.

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
vuqtran88: I was aware of the deadlock that might happen, but I couldn't find a better way to force async call to sync call.

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
vuqtran88: agreed

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
nr-ahemsath: @vuqtran88 Did you look into doing async-await and what would have to happen to make the `SendData` method async?

chynesNR: Thanks for bringing this up, @tippmar-nr, as I don't think it's something we consider enough. I don't believe this functionality can deadlock -- this happens on a timer, not a UI thread or in a request context, and there should only ever be one Send happening at a time. And I agree with Vu that doing it property as an async call would touch too much code. I do think it's worth putting in a tech debt Story to revisit this later, though.

chynesNR:
We should probably add this to the changelog. It's technically an implementation detail but also a behavior change.

chynesNR (approved):
Looks good to me!

PR #1340, opened 2023-01-06
Fixed XML External Entity (XXE) Injection in log4net()

wica-sufatmawati:
## Description
Apache log4net before 2.0.10 does not disable XML external entities when parsing log4net configuration files. This could allow for XXE-based attacks in applications that accept arbitrary configuration files from users.

XXE Injection is a type of attack against an application that parses XML input. XML is a markup language that defines a set of rules for encoding documents in a format that is both human-readable and machine-readable. By default, many XML processors allow specification of an external entity, a URI that is dereferenced and evaluated during XML processing. When an XML document is being parsed, the parser can make a request and include the content at the specified URI inside of the XML document. Attacks can include disclosing local files, which may contain sensitive data such as passwords or private user data, using file: schemes or relative paths in the system identifier.

below is a sample XML document, containing an XML element- username.
```
<xml> <?xml version="1.0" encoding="ISO-8859-1"?> <username>John</username> </xml>
```
An external XML entity - xxe, is defined using a system identifier and present within a DOCTYPE header. These entities can access local or remote content. For example the below code contains an external XML entity that would fetch the content of /etc/passwd and display it to the user rendered by username.
```
<xml> <?xml version="1.0" encoding="ISO-8859-1"?> <!DOCTYPE foo [ <!ENTITY xxe SYSTEM "file:///etc/passwd" >]> <username>&xxe;</username> </xml>
```
Other XXE Injection attacks can access local resources that may not stop returning data, possibly impacting application availability and leading to Denial of Service.

CVE-2018-1285
[CWE-611](https://cwe.mitre.org/data/definitions/611.html)
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

# Author Checklist

# Reviewer Checklist
- [x] Perform code review
- [x] Review Changelog

### DIFF
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
@@ -16,22 +16,22 @@
     <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="1.0.205" />
     
     <!-- log4net .NET framework references -->
-    <PackageReference Include="log4net" Version="1.2.10" Condition="'$(TargetFramework)' == 'net462'" />
-    <PackageReference Include="log4net" Version="2.0.5" Condition="'$(TargetFramework)' == 'net471'" />
+    <PackageReference Include="log4net" Version="2.0.10" Condition="'$(TargetFramework)' == 'net462'" />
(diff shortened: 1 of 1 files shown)


JcolemanNR:
Thanks for submission @wica-sufatmawati! This update targets our test applications which are not shipped with the .NET Agent, so there is no real world security vulnerability as these applications are not hosted for more than a few minutes in sandboxed (local/ci) environments. In this case, we are intentionally testing old versions of log4net so we can verify publicly communicated version compatibility with our agent. 

We appreciate the submission, but will be closing this PR as the use of an outdated version of log4net is intentional here.

PR #1352, opened 2023-01-19
Add support for RestSharp versions between 106.6.10 and 107.x

nr-ahemsath:
## Description

This PR uses the instrumentation assembly version filtering to add support for RestSharp clients with versions greater than 106.6.10 (the previous highest tested/supported version) and less than 107.x.  

The two main changes required were:

- Instrument the `ExecuteAsync(RestSharp.IRestRequest,System.Threading.CancellationToken)` method in newer 106.x versions instead of `ExecuteTaskAsync(RestSharp.IRestRequest,System.Threading.CancellationToken)`, using the same wrapper.
- Instrument the `ConfigureWebRequest(System.String,System.Uri)` method as a point to add our tracing data headers in newer 106.x versions, since the `AppendHeaders` method we previously instrumented was changed to a local method with no arguments.  A new wrapper is used for this, since ConfigureWebRequest returns an HttpWebRequest rather than accepting one as an argument as AppendHeaders did.

Note: because we are filtering on assembly version, not package version, and because some of the RestSharp versions between 106.6.10 and 107.x have the AssemblyVersion set to "106.0.0.0", we will only be able to claim support for some of the newer versions, full list TBD (and will definitely be documented on our public docs).  I'll also put that info in the changelog when I do that before this PR is merged.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### New Features
+* RestSharp client instrumentation support has been extended to include the following versions: 106.11.x, 106.12.0, 106.13.0, and 106.15.0. [#1352](https://github.com/newrelic/newrelic-dotnet-agent/pull/1352)
+
 ## [10.4.0]
 
 ### New Features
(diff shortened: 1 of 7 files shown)


chynesNR (approved):
Looks good!

PR #1357, opened 2023-01-24
Log effective TLS configuration on startup

JcolemanNR:

## Description

As we are close to ending support for TLS 1.0 & 1.1 there will likely be several customer issues with how TLS is configured. This PR logs a useful piece of TLS configuration at startup to aid in troubleshooting. This is logged at the info level since this only occurs during startup, and we want to reduce churn with GTS & customers.

Thanks to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1860 for the idea!

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


JcolemanNR:
Converting to draft so we can add more information to agent logs:
- Query known TLS related registry entries. Log key existence and values.
- Review HttpClient and log any useful pieces of information (if any)

chynesNR (dismissed):

tippmar-nr (dismissed):
Looks like a good addition to me!

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
```diff
@@ -184,6 +184,12 @@ private void Initialize()
 
             StartServices();
             LogInitialized();
+            LogTlsConfiguration();
+        }
+
+        private void LogTlsConfiguration()
+        {
+            Log.Info($"TLS Configuration (System.Net.ServicePointManager.SecurityProtocol): {System.Net.ServicePointManager.SecurityProtocol}");
```
nrcventura, line 192: Should we call out that this is the system configuration?
Should this be at the `Debug` level so that it is only being emitted when we suspect there is a Tls issue?

jaffinito (dismissed):
This is great!  So simple and it will make support's lives a bit easier.  Nice!

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
JcolemanNR: I don't think so, as we don't know if people have configured anything... Also, it may not be the system configuration as peoples applications may have changed the security protocol. Also, we should log this at info so it's easy for people and support to see without round tripping with a customer

PR #1361, opened 2023-01-25
Fix a race condition when using SetApplicationName

jaffinito:
## Description

Now, there is a bool that can be set if a runtime update has occurred, meaning the names have changed.  If this occurs during a Connect, a restart is scheduled.  I used ScheduleRestart so that we will follow the backoff pattern and hopefully not spam connect attempts of there is an issue.  This also lets the connect process complete before we start another.  I didn't think we needed a flag for "connecting" since all the work is done in the Connect method.

The code isn't really unit testable.  I tried to see if I could capture some part of the reconnect happening, but the methods just aren't available.

Adds a new integration test that ensures they process is working.  I only added for the oldest (net462, net6) and latest supported frameworks since framework version shouldn't affect anything.  

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
@@ -8,6 +8,7 @@ using NewRelic.Agent.Core.Utilities;
 using NewRelic.Core.Logging;
 using System;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Net.Sockets;
 
@@ -40,7 +41,7 @@ namespace NewRelic.Agent.Core.DataTransport
         {
             _connectionHandler = connectionHandler;
             _scheduler = scheduler;
-
+            
             _subscriptions.Add<StartAgentEvent>(OnStartAgent);
             _subscriptions.Add<RestartAgentEvent>(OnRestartAgent);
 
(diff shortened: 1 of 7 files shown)


JcolemanNR (commented):
I reviewed the execution context and I don't think we need to lock around the boolean. It needs changelog, and ideally test coverage, but I'm not sure how to reliably test this. This code does not look like it is under unit test (or built to be unit testable), and timing this to occur in an integration test sounds challenging to do reliably.

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -84,12 +86,19 @@ private void Connect()
         {
             try
             {
+                _runtimeConfigurationUpdated = false;
```
JcolemanNR, line 89: Move this in to the lock starting on line 90

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -84,12 +86,19 @@ private void Connect()
...
                 }
 
-                _connectionAttempt = 0;
+                // If the runtime configuration has changed, the app names have updated, so we schedule a restart
+                // This uses the existing ScheduleRestart logic so the current Connect can finish and we follow the backoff pattern and don't spam reconnect attempts.
+                if (_runtimeConfigurationUpdated)
+                {
+                    Log.Warn("The runtime configuration was updated during connect");
+                    ScheduleRestart();
+                }
```
JcolemanNR, line 101: Move this in to the lock starting on line 90

JcolemanNR (commented):
To avoid the possibility of extra unexpected (or missed) reconnects, we should move the reset/check of `_runtimeConfigurationUpdated` in to the lock

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -35,19 +36,20 @@ public class ConnectionManager : ConfigurationBasedService, IConnectionManager
         private int _connectionAttempt = 0;
         private bool _started;
         private readonly object _syncObject = new object();
+        private bool _runtimeConfigurationUpdated = false;
```
JcolemanNR, line 39: After reviewing the expected usage patterns of this field, I believe read/write access to it should be wrapped in a lock. There is a low possibility that a read and write may occur at the exact same time. The logic in the rest of the PR is solid, but we want to avoid any 1:1mil crashes due to read/write conflicts.

jaffinito: I tried a few combinations of locking around the bool, but the method in the PR was the one that worked.

chynesNR (approved):
Very nice!

Issue #1366, opened 2023-01-31
Index was out of range in StackExchange.Redis v2


michal-bajer1:
[NOTE]: # ( ^^ After upgrading to 10.6.0 new exception started appearing from StackExchange.Redis v2^^ )

**Description**
We are getting sporadically thrown exception during NR agent's attempts to profile Redis connections after upgrading to 10.6.0.

**Expected Behavior**
There is no reason for exception in this spot.

**[Troubleshooting]**
```
ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
  Module "System.ThrowHelper", line 0, col 0, in ThrowArgumentOutOfRange_IndexException
    Void ThrowArgumentOutOfRange_IndexException()
  Module "System.Collections.Generic.List`1", line 9, col 0, in get_Item
    T get_Item(Int32)
  Module "NewRelic.Collections.ConcurrentList`1", line 12, col 0, in get_Item
    T get_Item(Int32)
  Module "NewRelic.Agent.Core.Transactions.Transaction", line 0, col 0, in get_CurrentSegment
    NewRelic.Agent.Api.ISegment get_CurrentSegment()
  Module "NewRelic.Providers.Wrapper.StackExchangeRedis2Plus.SessionCache", line 38, col 0, in GetProfilingSession { <lambda> }
    StackExchange.Redis.Profiling.ProfilingSession <GetProfilingSession>b__7_0()
  File "/_/src/StackExchange.Redis/ConnectionMultiplexer.cs", line 1878, col 17, in PrepareToPushMessageToBridge
    Boolean PrepareToPushMessageToBridge[T](StackExchange.Redis.Message, StackExchange.Redis.ResultProcessor`1[T], StackExchange.Redis.IResultBox`1[T], StackExchange.Redis.ServerEndPoint ByRef)
  File "/_/src/StackExchange.Redis/ConnectionMultiplexer.cs", line 1905, col 16, in TryPushMessageToBridgeSync
    StackExchange.Redis.WriteResult TryPushMessageToBridgeSync[T](StackExchange.Redis.Message, StackExchange.Redis.ResultProcessor`1[T], StackExchange.Redis.IResultBox`1[T], StackExchange.Redis.ServerEndPoint ByRef)
  File "/_/src/StackExchange.Redis/ConnectionMultiplexer.cs", line 1988, col 17, in ExecuteSyncImpl
[...]

PR #1390, opened 2023-02-13
Partial fix for Segment out-of-range error

chynesNR:
Partial fix for issue with Transactions created in an async context. See [#1366](https://github.com/newrelic/newrelic-dotnet-agent/issues/1366)

I believe there is still an general issue with Transactions created in an async context; this is just one of the places where it can happen.

Because this is an async timing issue, I wasn't able to come up with a consistent unit or integration test to capture it, unfortunately.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
+++ b/src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
@@ -59,9 +59,15 @@ namespace NewRelic.Agent.Core.Transactions
                     return Segment.NoOpSegment;
                 }
 
-                if (Segments[currentSegmentIndex.Value] != null)
+                int idx = currentSegmentIndex.Value;
+                if (idx >= Segments.Count)
                 {
-                    return Segments[currentSegmentIndex.Value];
+                    Log.Warn($"Transaction {Guid} is out of sync with the current segment [looking for {idx} out of {Segments.Count}]");
+                    return Segment.NoOpSegment;
+                }
+                if (Segments[idx] != null)
+                {
+                    return Segments[idx];
                 }
 
                 return Segment.NoOpSegment;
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -59,9 +59,15 @@ public ISegment CurrentSegment
                     return Segment.NoOpSegment;
                 }
 
-                if (Segments[currentSegmentIndex.Value] != null)
+                int idx = currentSegmentIndex.Value;
+                if (idx >= Segments.Count)
                 {
-                    return Segments[currentSegmentIndex.Value];
+                    Log.Warn($"Transaction {Guid} is out of sync with the current segment [looking for {idx} out of {Segments.Count}]");
```
jaffinito, line 65: Do we want this to be a warn or a debug level message?  Warns will show up in info level logs.

nrcventura (approved):

tippmar-nr (approved):

jaffinito (approved):

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
nrcventura: I think that this should be rare enough that it should be ok at the warn level. It will also make it easier for us to be aware that it is happening. I think something similar is done for transactions that are captured by finalier/destructor.

chynesNR: It was intentional, unless we have guidelines for what constitutes a Warning? Although I fixed the exception, if this happens it means these Segments are being associated with the wrong Transaction, which is pretty bad.

PR #1393, opened 2023-02-21
Update NLog to improve local log decoration coverage

jaffinito:
## Description

Changes how the NLog instrumentation updates the log message for local log decoration.

* Instead of examining and using LogEventInfo.Message, the new instrumentation uses LogEventInfo.FormattedMessage and its backing field, (v4.5+ _formattedMessage)|(Pre v4.5 formattedMessage).  This property has additional logic that Message does not and appears to be the source for the actual log message we see.  Calling FormattedMessage causes NLog to build the message and store it in the backing field.  
* Using the visibility bypasser, we get the setter for the backing field and use it to set the field to itself with our decoration token.  The Action is stored staticly, same as with other bypasser setups.
* Since we can only get the major assembly version from NLog assembly we cannot use minVersion or maxVersion.  The cleanest solution was to check for _formattedMessage first since it is the going forward name and if that fails, use formattedMessage.  This is done using reflection, but is will only happen once per run so the impact is minimal.

Caveats:
This does not 100% cover all the log lines.  It does cover the vast majority of them.  

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
@@ -8,17 +8,13 @@ using NewRelic.Agent.Api.Experimental;
 using NewRelic.Agent.Extensions.Logging;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 using NewRelic.Reflection;
+using NLog;
 
 namespace NewRelic.Providers.Wrapper.NLogLogging
 {
     public class NLogWrapper : IWrapper
     {
-        private static Func<object, object> _getLevel;
-        private static Func<object, string> _getRenderedMessage;
-        private static Func<object, DateTime> _getTimestamp;
-        private static Func<object, string> _messageGetter;
-        private static Func<object, Exception> _getLogException;
-        private static Func<object, IDictionary<object, object>> _getPropertiesDictionary;
+        private static Action<object, string> _setFormattedMessage;
 
         public bool IsTransactionRequired => false;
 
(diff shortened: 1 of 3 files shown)


tippmar-nr (approved):

nrcventura (approved):
It sounds like there are integration tests that exercise this code, but we need an item on our backlog to add the necessary assertions to validate this behavior.

jaffinito:
I added a story to our backlog.

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/NLogLogging/NLogWrapper.cs
```diff
@@ -68,28 +70,38 @@ private void DecorateLogMessage(object logEvent, Type logEventType, IAgent agent
...
             {
                 return;
             }
 
-            // this cannot be made a static since it is unique to each logEvent
-            var messageSetter = VisibilityBypasser.Instance.GeneratePropertySetter<string>(logEvent, "Message");
-            messageSetter(originalMessage + " " + formattedMetadata);
+            // NLog version strings are not setup to allow using min/max version in instrumentation - they only report major version.
+            // This will default to the 4.5+ field name and if that is not found, use the pre4.5 field name.
+            var formattedMessageName = FormattedMessage45Plus;
```
chynesNR, line 82: I think this is a good solution. Should we have either a fallback or a try/catch in case they change the field name again?

jaffinito: I don't think it will be needed.  It looks like they started following Microsoft's format, like we did, and that was the reason for the change.

PR #1410, opened 2023-02-27
Automate generation of index.html during deploy

tippmar-nr:

## Description

Updates the `deploy_agent.yml` workflow to generate the `index.html` file for the current and previous release folders.

Note that testing of the workflow update isn't possible until we do the next release. I tested my changes for this PR in a (now deleted) temporary workflow file.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


.github/workflows/deploy_agent.yml
```diff
@@ -121,6 +125,19 @@ jobs:
           }
         shell: pwsh
 
+  index-download-site:
+    name: Rebuild indexes on the download site
+    if: ${{ github.event.inputs.indexdownloadsite == 'true' }}
```
nrcventura, line 130: This job is implicitly related to the deploy-download-site job. We don't check for that setting here, but I thing not checking that setting also makes sense if we want the flexibility to run the indexer without redeploying to the download site. I'm mostly calling this out to verify if that is the intention here.

nrcventura (approved):

.github/workflows/deploy_agent.yml
tippmar-nr: Yeah, my intention was to allow re-indexing without actually deploying. But then I realized that the `get-external-artifacts` job isn't conditional and would do a lot of throw-away work if all you wanted to do was update the indexes.

I could certainly update so that the index rebuilding is triggered off of the same parameter as the deploy download site job uses.

jaffinito (approved):

Issue #1399, opened 2023-02-24
Error Fingerprinting: Setup attribute system


jaffinito:
### Describe the story, please be clear on scope of the story.
Setup the attribute definitions and enums for the error fingerprinting.

ApiSupportabilityMetricCounters.cs

1. Update ApiMethod enum to add “ErrorGroupCallback” method.

AttributeDefinitionService.cs

1. Add definition for ErrorGroup with a private backing field. Name of attribute is "error_group".  Type should be AttributeDefinition<string, string>.  Destinations should be ErrorEvent and.ErrorTrace. AttributeClassification should be AgentAttributes.
2. Add ErrorGroupFingerprint  to IAttributeDefinitions

### Acceptance Criteria

- Agent should build
- new definition is available from interface.

### Additional context
**Feature Branch: feature/errorsinbox-fingerprinting-usertracking**

Experimental branch: https://github.com/newrelic/newrelic-dotnet-agent/tree/experimental-errorsinbox-improvements

### Estimates
S

angelatan2:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/APM-Agent-Overview-614b054ff9a50b000f9039d7/issues/newrelic/newrelic-dotnet-agent/1399?planning-poker) with Zenhub @jaffinito

angelatan2:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/APM-Agent-Overview-614b054ff9a50b000f9039d7/issues/newrelic/newrelic-dotnet-agent/1399?planning-poker) with Zenhub @JcolemanNR

angelatan2:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Dotnet-Agent-61560f15ddd6a90010ada14e/issues/newrelic/newrelic-dotnet-agent/1399?planning-poker) with Zenhub @tippmar-nr

jaffinito:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Dotnet-Agent-61560f15ddd6a90010ada14e/issues/newrelic/newrelic-dotnet-agent/1399?planning-poker) with Zenhub @nrcventura

jaffinito:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Dotnet-Agent-61560f15ddd6a90010ada14e/issues/newrelic/newrelic-dotnet-agent/1399?planning-poker) with Zenhub @chynesNR

PR #1413, opened 2023-02-28
Rename error group fingerprint to error group

nrcventura:

## Description

Updates the error group fingerprint attribute name, definition, and enum to match the names specified in #1399. 

# Author Checklist
- ~~[ ] Unit tests, Integration tests, and Unbounded tests completed~~ Just a rename.
- ~~[ ] Performance testing completed with satisfactory results (if required)~~ Just a rename
- ~~[ ] [Agent Changelog](/src/Agent/CHANGELOG.md) or [Lambda Agent changelog](/src/AwsLambda/CHANGELOG.md) updated~~ Just a rename

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
+++ b/src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
@@ -632,7 +632,7 @@ namespace NewRelic.Agent.Core
             {
                 InternalApi.ErrorFingerprintingCallback(callback);
             }
-            TryInvoke(work, apiName, ApiMethod.ErrorGroupCallback);
+            TryInvoke(work, apiName, ApiMethod.SetErrorGroupCallback);
         }
     }
 }
--- a/src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
+++ b/src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
@@ -632,7 +632,7 @@ namespace NewRelic.Agent.Core
             {
                 InternalApi.ErrorFingerprintingCallback(callback);
             }
-            TryInvoke(work, apiName, ApiMethod.ErrorFingerprintingCallback);
+            TryInvoke(work, apiName, ApiMethod.ErrorGroupCallback);
         }
     }
 }
--- a/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core.Metric/ApiSupportabilityMetricCounters.cs
+++ b/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core.Metric/ApiSupportabilityMetricCounters.cs
@@ -38,7 +38,7 @@ namespace NewRelic.Agent.Core.Metric
         InsertDistributedTraceHeaders = 24,
         AcceptDistributedTraceHeaders = 25,
         SpanSetName = 26,
-        ErrorGroupCallback = 27,
+        SetErrorGroupCallback = 27,
         SetUserId = 28
     }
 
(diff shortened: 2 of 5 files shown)


tippmar-nr (approved):

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
```diff
@@ -632,7 +632,7 @@ void work()
             {
                 InternalApi.ErrorFingerprintingCallback(callback);
             }
-            TryInvoke(work, apiName, ApiMethod.ErrorFingerprintingCallback);
+            TryInvoke(work, apiName, ApiMethod.ErrorGroupCallback);
```
chynesNR, line 635: Could we have `Set` in the name, so it's clear what it's doing? This makes it seem like we're invoking the callback here.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
```diff
@@ -632,7 +632,7 @@ void work()
             {
                 InternalApi.ErrorFingerprintingCallback(callback);
             }
-            TryInvoke(work, apiName, ApiMethod.ErrorFingerprintingCallback);
+            TryInvoke(work, apiName, ApiMethod.SetErrorGroupCallback);
```
jaffinito, line 635: In the most recent spec update, the callback name is still ErrorGroupCallback (error_group_callback), so I don't think we should call this "SetErrorGroupCallback".

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core.Metric/ApiSupportabilityMetricCounters.cs
```diff
@@ -38,7 +38,7 @@ public enum ApiMethod
         InsertDistributedTraceHeaders = 24,
         AcceptDistributedTraceHeaders = 25,
         SpanSetName = 26,
-        ErrorFingerprintingCallback = 27,
+        SetErrorGroupCallback = 27,
```
jaffinito, line 41: Same a previous comment.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
jaffinito: The spec currently doesn't have the set verb.

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
nrcventura: Is the intention of the API to be used as `Agent.ErrorGroupCallback(MyType.MyCallback)` or `Agent.ErrorGroupCallback = MyType.MyCallback`? Given the name, I think a property syntax makes the most sense, in which case the `Set` is implied. The method form is kind of confusing to me.

chynesNR: Ah, I took that to mean the name of the parameter, not the API call. It does look like the spec typically calls out the function name. I guess I should have raised this then. :)

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
jaffinito: We discussed this in pairing and decided to go with Set...

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
jaffinito: We discussed this in pairing and decided to go with Set...

src/Agent/NewRelic/Agent/Core/Api/AgentApi.cs
nrcventura: I plan on updating the API itself (and not just the enum) with a separate PR.

tippmar-nr (approved):

Issue #1401, opened 2023-02-24
Error Fingerprinting: Update error types with new callback


jaffinito:
### Describe the story, please be clear on scope of the story.
Update the ErrorData to store the fingerprint, ErrorService to invoke the callback and set the fingerprint, and the makers to set the attribute.

ErrorData.cs

1. Add a string Property called “ErrorGroup” that only has a getter.
2. Update constructor to take an errorGroup and assign it.
3. There will be several tests that need fixed to use the updated constructor

ErrorService.cs

1. Update FromExceptionInternal to invoke the Agent.Instance.ErrorGroupCallback callback and pass it into the ErrorData constructor.

ErrorEventMaker.cs

Note: if the string is null or empty, do not add the attrtibute!

1. Add a “SetErrorGroup” method that takes “ErrorData errorData, IAttributeValueCollection attribValues” and sets the _attribDefs.ErrorGroup to the errorData.ErrorGroup value.
2. SetErrorGroupshould be called immediately before “return new ErrorEventWireModel” in both GetErrorEvent methods.

ErrorTraceMaker.cs

Note: if the string is null or empty, do not add the attrtibute!

1. Update constructor to take in IAttributeDefinitionService so that we can access the definitions. Similar to other uses of this class, recommend using a field that calls “_attribDefs => _attribDefSvc.AttributeDefs” to access the definitions.
2. Update GetErrorTraceAttributes signature to take in an ErrorData and use that to set the _attribDefs.ErrorGroup to the errorData.ErrorGroup value.

### Acceptance Criteria

- Agent should build
- Smoke test: Load API into app and make sure methods are available and work

### Additional context
**Feature Branch: feature/errorsinbox-fingerprinting-usertracking**

Experimental branch: https://github.com/newrelic/newrelic-dotnet-agent/tree/experimental-errorsinbox-improvements

### Estimates
S

angelatan2:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Dotnet-Agent-61560f15ddd6a90010ada14e/issues/newrelic/newrelic-dotnet-agent/1401?planning-poker) with Zenhub @jaffinito @JcolemanNR @tippmar-nr

jaffinito:
[...]

PR #1418, opened 2023-03-03
Use configured errorgroupcallback

nrcventura:

## Description

Updates the error group functionality mentioned in #1401. 
- Renames groupFingerprint to errorGroup
- Update error group attribute to not get generated if the error group is null, empty string, or just whitespace
- Gets the error group callback from configuration instead of the agent class

# Author Checklist
- ~~[ ] [Agent Changelog](/src/Agent/CHANGELOG.md) or [Lambda Agent changelog](/src/AwsLambda/CHANGELOG.md) updated~~ to be updated in a separate PR 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/tests/Agent/UnitTests/Core.UnitTest/Errors/ErrorServiceTests.cs
+++ b/tests/Agent/UnitTests/Core.UnitTest/Errors/ErrorServiceTests.cs
@@ -317,6 +317,79 @@ namespace NewRelic.Agent.Core.Errors
 
         }
 
+        [TestCase("test group")]
+        [TestCase(null)]
+        [TestCase("")]
+        [TestCase("     ")]
+        public void FromException_ShouldHaveErrorGroup(string callbackReturnValue)
+        {
+            SetupConfiguration(null, null, null, false, null, null, null, true);
+            SetupErrorGroupCallback(ex => callbackReturnValue);
+
+            var testException = new Exception("Test exception message");
+            var errorData = _errorService.FromException(testException);
+
+            Assert.AreEqual(callbackReturnValue, errorData.ErrorGroup);
+        }
+
+        [Test]
+        public void FromException_ShouldNotHaveErrorGroup()
+        {
+            SetupConfiguration(null, null, null, false, null, null, null, true);
+            SetupErrorGroupCallback(null);
+
+            var testException = new Exception("Test exception message");
+            var errorData = _errorService.FromException(testException);
+
+            Assert.IsNull(errorData.ErrorGroup);
+        }
+
+        [Test]
+        public void FromException_ErrorGroupShouldUseOuterException()
+        {
+            SetupConfiguration(null, null, null, false, null, null, null, true);
+            SetupErrorGroupCallback(TestErrorGroupCallback);
+
+            var innerException = new Exception("Inner exception message");
+            var outerException = new Exception("Outer exception message", innerException);
+
+            var errorData = _errorService.FromException(outerException);
+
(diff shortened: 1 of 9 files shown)


tests/Agent/UnitTests/Core.UnitTest/Errors/ErrorServiceTests.cs
```diff
@@ -317,6 +317,79 @@ public void FromException_MarkErrorDataAsExpected_SameErrorClass_In_ExpectedClas
 
         }
 
+        [TestCase("test group")]
+        [TestCase(null)]
```
tippmar-nr, line 321: Shouldn't `errorData.ErrorGroup` come back `null` for all three of these test cases?

tippmar-nr (approved):

tests/Agent/UnitTests/Core.UnitTest/Errors/ErrorServiceTests.cs
nrcventura: I'm keeping the abstraction very thin at this point and enforcing the data at the attribute level. So The AttributeDefinition has a conversion function to enforce that the invalid data is coerced to null in a single location. I considered putting it in the ErrorService, but it just seemed too easy to bypass the ErrorService and create an ErrorData instance directly. In fact a lot of the existing tests bypassed the ErrorService, which led to inconsistent behavior between tests.

chynesNR:
Based on `AttributeDefinition.TrySetValue()`, it looks like we'll log whenever an attribute is set to null, even though it won't get sent because it's filtered out. If you don't have a callback set, I think we'll get a lot of log spam. Does that seem like the case?

chynesNR:
Oh also, it doesn't have to be done in this PR, but the most recent version of the spec says we should be using the name `error.group.name` rather than `error_group` for the attribute.

nrcventura:
> Based on `AttributeDefinition.TrySetValue()`, it looks like we'll log whenever an attribute is set to null, even though it won't get sent because it's filtered out. If you don't have a callback set, I think we'll get a lot of log spam. Does that seem like the case?

Based on the code in the HandleNullValue method and because the attribute is not a user attribute (it's an agent attribute) nothing should get logged. https://github.com/newrelic/newrelic-dotnet-agent/blob/90b8018fbb433aa554c24d133ad7df63147b648c/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinition.cs#L354-L363.

nrcventura:
> Oh also, it doesn't have to be done in this PR, but the most recent version of the spec says we should be using the name `error.group.name` rather than `error_group` for the attribute.

Thanks for the update. Yes, I'd like to handle the rename in a separate PR to make reviewing the changes easier.

chynesNR:
> > Based on `AttributeDefinition.TrySetValue()`, it looks like we'll log whenever an attribute is set to null, even though it won't get sent because it's filtered out. If you don't have a callback set, I think we'll get a lot of log spam. Does that seem like the case?
> 
> Based on the code in the HandleNullValue method and because the attribute is not a user attribute (it's an agent attribute) nothing should get logged.
> 
> https://github.com/newrelic/newrelic-dotnet-agent/blob/90b8018fbb433aa554c24d133ad7df63147b648c/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinition.cs#L354-L363

Whoops, my bad. I misread `!=` as `==`

chynesNR (approved):

PR #1420, opened 2023-03-03
User Tracking Unit and Integration Tests

tippmar-nr:

## Description

Adds unit and integration tests for User Tracking.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/tests/Agent/UnitTests/Core.UnitTest/Api/TransactionBridgeApiTests.cs
+++ b/tests/Agent/UnitTests/Core.UnitTest/Api/TransactionBridgeApiTests.cs
@@ -1,0 +1,53 @@
+using System;
+using NewRelic.Agent.Api;
+using NewRelic.Agent.Configuration;
+using NewRelic.Agent.Core.Configuration;
+using NewRelic.Agent.Core.Metric;
+using NUnit.Framework;
+using Telerik.JustMock;
+
+namespace NewRelic.Agent.Core.Api
+{
+    [TestFixture]
+    public class TransactionBridgeApiTests
+    {
+        private IApiSupportabilityMetricCounters _apiSupportabilityMetricCounters;
+        private ITransaction _transaction;
+        private TransactionBridgeApi _transactionBridgeApi;
+
+        [SetUp]
+        public void Setup()
+        {
+            _apiSupportabilityMetricCounters = Mock.Create<IApiSupportabilityMetricCounters>();
+            _transaction = Mock.Create<ITransaction>();
+
+            _transactionBridgeApi = new TransactionBridgeApi(_transaction, _apiSupportabilityMetricCounters, Mock.Create<IConfigurationService>());
+        }
+
+        [Test]
+        public void SetUserId_AddsSupportabilityMetric()
+        {
+            _transactionBridgeApi.SetUserId("CustomUserId");
+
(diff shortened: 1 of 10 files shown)


tests/Agent/UnitTests/Core.UnitTest/Wrapper/AgentWrapperApi/Builders/TransactionTests.cs
```diff
@@ -296,5 +296,54 @@ public void AddUserAttribute_LastInWins()
...
+            var hasUserIdAttribute = userAttributes.TryGetValue(_attribDefs.EndUserId.Name, out var userIdValue);
+
+            Assert.AreEqual(endUserAttributeShouldExist, hasUserIdAttribute);
+            if (endUserAttributeShouldExist)
+            {
+                Assert.AreEqual(expectedUserId, userIdValue);
+            }
+        }
+
+        [Test]
```
chynesNR, line 332: Might be overkill, but should we test the case where we set the user ID to null after it previously had a value?

chynesNR (dismissed):

tests/Agent/UnitTests/Core.UnitTest/Api/TransactionBridgeApiTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            Mock.Assert(() => _transaction.SetUserId(Arg.Matches<string>(userId => userId == expectedCustomUserId)));
+        }
+
+        [Test]
+        public void SetUserId_HandlesException()
+        {
+            Mock.Arrange(() => _apiSupportabilityMetricCounters.Record(Arg.IsAny<ApiMethod>())).Throws<Exception>();
+
+            var expectedCustomUserId = "CustomUserId";
+            _transactionBridgeApi.SetUserId(expectedCustomUserId);
```
nrcventura, line 53: Should we assert that no user id was set after this api call is made? Or is it enough that this method bubble up the exception thrown in the api call?

tippmar-nr: Seems reasonable.

nrcventura (approved):

tests/Agent/UnitTests/Core.UnitTest/Wrapper/AgentWrapperApi/Builders/TransactionTests.cs
nrcventura: Having a test like that might serve as additional documentation of that behavior.

tippmar-nr: So I added a test for this and find that the previously-set user id is still present in the attributes after being set to null (since the code short-circuits when the value is null). 

I wonder if we should actually ensure the attribute gets removed when setting to null or whitespace?

nrcventura: I thought that was behavior that we chose to remove in a previous PR. Either way, it's a good idea to confirm the intention.

tippmar-nr: After thinking about it for a minute, I don't see any valid use case where the user would want to set a user id and then clear it later during the same transaction. So for now, we'll leave things alone.

chynesNR (approved):

PR #1422, opened 2023-03-06
Update GitHub workflows

tippmar-nr:

## Description

* Updates various actions referenced in our workflows to use the newer Node v16 compliant versions. 
* Replaces deprecated usages of `::set-output` with [proper usages of GitHub environment files](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions?tool=powershell).

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -40,12 +40,12 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
       - name: Add msbuild to PATH
-        uses: microsoft/setup-msbuild@v1
+        uses: microsoft/setup-msbuild@v1.3.1
 
       - name: Clean out _profilerBuild directory
         run: |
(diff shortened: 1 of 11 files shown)


nrcventura (approved):

chynesNR:
It looks fine, but I think we either need to find a way to do some additional testing, or set aside extra time for the next release to make sure there aren't any surprises.

tippmar-nr:
@chynesNR So, with only a couple of exceptions, the changes in the workflows that I've been unable to test are simple version upgrades that are also mirrored in the workflows that I could test. I would agree that there is some risk with not testing these changes, but I think it's very low. Several of the workflows don't appear to be used at all, a couple are only seldomly used and the others seem to be related to the release process. 

I'll be happy to help with (or run) the next release and deal with any issues that might come up, though I'm confident there won't be any. 

Thoughts?

nr-ahemsath:
Since the changes in the workflow files we haven't been able to test are so similar to the changes in the workflows we *have* been able to test, most importantly the all_solutions_build (main CI), I believe the risk in merging this is acceptable.  Approved.

nr-ahemsath (approved):
Changes look good, verified that the green CI for this PR used the new workflow. 👍

PR #1424, opened 2023-03-09
fix: don't throw when unable to get parent process path

tippmar-nr:

## Description

Modify `GetParentProcessPath()` to log a message and return empty string instead of throwing an exception when unable to obtain a handle to the parent process or unable to get the parent process image filename. 

This fix only affects the Windows x86/x64 builds, so there are no updates to the Linux shared libraries.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog


nrcventura (approved):

src/Agent/NewRelic/Profiler/Profiler/SystemCalls.h
```diff
@@ -109,16 +109,18 @@ namespace NewRelic { namespace Profiler
 
                 if (!result)
                 {
+                    // log the error but don't throw
                     auto error = ::GetLastError();
-                    LogError(L"Unable to get the parent process path (1).  Error number: ", std::hex, error, std::resetiosflags(std::ios_base::basefield));
-                    throw ProfilerException();
+                    LogTrace(L"Unable to get the parent process path (1).  Error number: ", std::hex, error, std::resetiosflags(std::ios_base::basefield));
```
nr-ahemsath, line 114: I'm wondering if "warning" or "info" might be the appropriate level so that if any customers are impacted by this in the field, they'll see the message at the default agent logging level.

chynesNR: I don't think that's necessary. We don't even know if the parent path is relevant until later in the decision process, and we already log the process and parent name as part of the w3wp check. If the parent name comes up blank, that should be a good enough clue.

chynesNR (approved):

nr-ahemsath (approved):
Looks good.  I verified that this fixes the problem on my system.  👍

PR #1468, opened 2023-03-20
chore: cleaned up all_solutions workflow to include previously missing integration test namespaces

tippmar-nr:

## Description

There were a few integration test namespaces missing from the all_solutions workflow. Added those and sorted the entire list to make it easier to track in the future. 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -469,17 +469,13 @@ jobs:
     runs-on: windows-2022
     strategy:
       matrix:
-        namespace: [ AgentFeatures, AgentLogs, AgentMetrics, Api, AspNetCore, BasicInstrumentation, CatInbound, CatOutbound, 
-          CSP, CustomAttributes, CustomInstrumentation, DataTransmission, DistributedTracing, Errors, 
-          HttpClientInstrumentation.NetCore, HttpClientInstrumentation.NetFramework, Owin, ReJit.NetCore, 
-          Logging.HsmAndCsp, Logging.LocalDecoration, Logging.MaxSamplesStored, Logging.ZeroMaxSamplesStored,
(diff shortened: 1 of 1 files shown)


chynesNR (approved):
Good catch!

nrcventura (dismissed):

.github/workflows/run_integration_tests.yml
```diff
@@ -93,10 +95,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup .NET 6.0
-        uses: actions/setup-dotnet@v1
+      - name: Add .NET 7 SDK
```
chynesNR, line 98: We don't have to do this now, but the 2022 runner already includes .NET 7: https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#net-core-tools

chynesNR (dismissed):

.github/workflows/run_integration_tests.yml
tippmar-nr: @chynesNR I went ahead and made that change, both in this file and in `all_solutions.yml` - there were several references that I removed, so maybe the build will be a few seconds faster now... ;-)

chynesNR (approved):

JcolemanNR (approved):

nr-ahemsath:
> We define this list in 2 separate jobs. Would it make sense to create a single sorted file that we could read and use in both locations? It looks like there are a few ways to dynamically define these matrix jobs.

Alternatively, could we use reflection into the integration tests DLL to find all of the namespaces dynamically?

tippmar-nr:
We can investigate ways to optimize the list of namespaces at some future point. The current changes get us in sync between the two workflows and in sync with the integration tests solution, which was the goal of this PR.

PR #1470, opened 2023-03-20
feat: support for Supportability/Logging/Forwarding/Dropped metric

tippmar-nr:

## Description

Adds support for the `Supportability/Logging/Forwarding/Dropped` metric by reporting the number of log messages that get dropped due to capacity constraints. Also included in the metric are logs that are dropped because they have an empty message and no exception.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review
- [ ] Review Changelog

### DIFF
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### New Features
-* Adds the `Logging/Forwarding/Dropped` metric to track the number of log messages that were dropped due to capacity constraints. [#1470](https://github.com/newrelic/newrelic-dotnet-agent/pull/1470)
+* Adds the `Supportability/Logging/Forwarding/Dropped` metric to track the number of log messages that were dropped due to capacity constraints. [#1470](https://github.com/newrelic/newrelic-dotnet-agent/pull/1470)
 
 ### Fixes
 * Reduce redundant collector request data payload logging in the agent log at DEBUG level. [#1449](https://github.com/newrelic/newrelic-dotnet-agent/pull/1449)
--- a/src/Agent/CHANGELOG.md
+++ b/src/Agent/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### New Features
+* Adds the `Logging/Forwarding/Dropped` metric to track the number of log messages that were dropped due to capacity constraints. [#1470](https://github.com/newrelic/newrelic-dotnet-agent/pull/1470)
 
 ### Fixes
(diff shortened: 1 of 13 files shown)


JcolemanNR (approved):
Looks good! I don't see any other aggregators creating metrics in the `Discard` collector response case, but we could hypothetically track some some more dropped logs here:
https://github.com/newrelic/newrelic-dotnet-agent/blob/23ed7f99e3f0d7ea8583c8e80a36ca6ea7fc35e0/src/Agent/NewRelic/Agent/Core/Aggregators/LogEventAggregator.cs#L149

src/Agent/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### New Features
+* Adds the `Logging/Forwarding/Dropped` metric to track the number of log messages that were dropped due to capacity constraints. [#1470](https://github.com/newrelic/newrelic-dotnet-agent/pull/1470)
```
nr-ahemsath, line 10: This is a supportability metric so the full name would be `Supportability/Logging/Forwarding/Dropped`.  I don't recall us typically announcing supportability metrics in the CHANGELOG or release notes.

nr-ahemsath: Never mind, should have searched our docs instead of relying on my memory!  We should still correct the name of the metric.
```suggestion
* Adds the `Supportability/Logging/Forwarding/Dropped` metric to track the number of log messages that were dropped due to capacity constraints. [#1470](https://github.com/newrelic/newrelic-dotnet-agent/pull/1470)
```

nrcventura:
I'm wondering if it makes sense to use the same supportability metric for both dropping scenarios, because I think they both require different actions. In the case of dropping logs because of an empty message or missing exception, the solution requires a feature request to add that support. In the case of dropping logs because the reservoir is full, it likely requires increasing the size of the reservoir or adjusting the collected log level target.

src/NewRelic.Core/NewRelic.Collections/ConcurrentPriorityQueue.cs
```diff
@@ -122,13 +126,13 @@ private void RemoveItemsFromBottomOfSortedSet()
             if (Size == 0)
             {
                 Reset();
```
chynesNR, line 128: If Size is 0, won't `Reset()` clear `_sortedSet`, meaning you won't increment the dropped counter?

tippmar-nr: Good catch! I reworked the logic in `Reset()` to capture the size of the collection before clearing it.

tippmar-nr:
@nrcventura 
> I'm wondering if it makes sense to use the same supportability metric for both dropping scenarios

Could we add our own supportability metric for the empty messages? 

Though I wonder if it's really a concern out in the wild that's worth worrying about? Have we had instances in the past where a customer complained that messages weren't showing up, only to find out they were sending empty messages?

chynesNR (approved):

chynesNR (approved):

nrcventura (approved):

PR #1479, opened 2023-03-23
Adding workflow to test release automation

chynesNR:
I'm in the process of configuring release-please, but I need the workflow to exist on `main` in order to be able to run it in my feature branch. This shouldn't do anything to `main` in its current state.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,21 @@
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  pull-requests: write
+
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        with:
+          release-type: go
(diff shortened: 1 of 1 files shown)


nrcventura (approved):

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,21 @@
...
+  contents: write
+  pull-requests: write
+
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
```
nr-ahemsath, line 16: I think one of the things recently suggested by security was pinning actions to very specific versions.  Might as well get in the habit now.  Just a suggestion, I'm not going to hold up approval for it.

nr-ahemsath (approved):
Approved.

.github/workflows/release-please.yml
chynesNR: Yeah that's a good idea, I'll think about that

PR #1504, opened 2023-04-03
Fix multiple StackExchange.Redis v2+ profiling issues

jaffinito:
## Description
Allow StackExchange.Redis v2+ profiling to start outside of a transaction.
The new v2+ instrumentation has "IsTransactionRequired" set incorrectly to "true" which prevented some customers from getting redis metrics.

Fix a memory leak when using StackExchange.Redis v2+.
A threading issue was created/discovered with the Segment.SpanId getter.  If spanid is null, it creates a new spanid, but if the getter is called rapidly from different threads, more than one ID can be created.  Prior to adding the new SessionProfiling instrumnetation for SE.Redis v2+ this never happened as that parameter was access in very specific ways.  The new instrumentation accessed it repeatedly and very rapidly depending on how many redis operations were occurring.

Instead of locking around the getter and potentially affecting other parts of the agent (locking and lock contention have been issues in the past), I took the approach of documenting the potential for trouble and making other changes that are narrower in scope.

This updates the SessionCache to no longer use SpanId as the cache key, instead using the actual segment.  This removes the call to SpanId and lessens the need to lock around the getter.  The cache key is only used as a key for a very short amount of time so should not pose an issue.

The second issue this uncovered was that the logic for getting and setting the cache item in the ConcurrentDictionary was not behaving as expected.  The TryGetValue method does not lock and so can allow several calls across threads to get into the code that makes new ProfilingSessions.  This was the source of the leak when combined with the false/bad SpanId issue above.  This resulted in orphaned cache entries that increased memory.    Moving to segment doesn't resolve this completely.

The way the logic worked could return a bad session that was not stored anywhere resulting in missed metrics and possibly  more leaking memory.  To resolve this, we create a session by default (called sessionCandidate) and then attempt to add it using TryAdd, which does need to "lock" to prevent duplicate keys.  If this returns true we can know that this is the first time the segment has made it here and can return the new session.  In the case of false, we know that a session already exists and can call TryGetValue knowing that it will have a sessions for us and return that one.  The sessionCandidate that we didn't use will be GC'ed after we leave the method since it was never passed back to SE.Redis and drop its hold on the transaction.

Testing:
I setup an app that creates ~8000 redis calls (Add, expire, get, delete) per transaction with an RPM of ~400.  With the old code this rapidly caused memory to grow to massive levels as the ConcurrentDictionary filled and transactions were kept alive.  In the new setup, memory grows very little since the ConcurrentDictionary rarely ever had more than 10 items in it (each segment cleans out its own stuff).

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -19,7 +19,7 @@ public class SessionCache : IStackExchangeRedisCache
     {
         private readonly EventWaitHandle _stopHandle = new EventWaitHandle(false, EventResetMode.ManualReset);
 
-        private readonly ConcurrentDictionary<string, ProfilingSession> _sessionCache = new ConcurrentDictionary<string, ProfilingSession>();
+        private readonly ConcurrentDictionary<ISegment, ProfilingSession> _sessionCache = new ConcurrentDictionary<ISegment, ProfilingSession>();
```
chynesNR, line 22: Do we ever copy/clone a Segment, where using a reference as the key might be a problem?

jaffinito: I don't believe we clone, but we do "combine segments", and that uses the existing segment data with some timing changes.  That looks to be called only after the segment has ended and is being processed in the TransactionTransformer.

tombatron:
Hey @jaffinito thanks for the effort here. 

In addition to the known memory leak, do you think this could also lead to sporadic excessive CPU utilization and reduced throughput to Redis?

jaffinito:
The initial changes (PR #1351) were done to reduce the agent's impact on Redis network utilization (see the PR for details on why/how).  This PR should only eliminate the memory leak, however, it might reduce memory pressure a small amount since it has slightly less allocations (better logic and use of features).

I don't think the problems with the previous session cache would have had much impact on overall redis throughput, though I would imagine that the profiling SE.Redis is running does have an impact.  The new work in this PR is more efficient so it might lead to a boost.  I never saw any CPU issues though I wasn't really on the lookout for them.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -34,28 +34,26 @@ public SessionCache(IAgent agent, int invocationTargetHashCode)
...
             }
 
             // Get the transaction from the session
-            var transaction = sessionData.UserToken as ITransaction;
-
-            // We want to make sure to finish the session even if the transaction is done so that it is not orphaned.
-            var commands = sessionData.FinishProfiling();
-            if (transaction.IsFinished)
+            var weakTransaction = sessionData.UserToken as WeakReference<ITransaction>;
+            if (!weakTransaction.TryGetTarget(out var transaction) || transaction.IsFinished)
```
nrcventura, line 50: WeakTransaction can potentially be null, so might want to do something like the following.
```suggestion
            if (!weakTransaction?.TryGetTarget(out var transaction) || transaction.IsFinished)
```

chynesNR (approved):
Looks good!

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -63,7 +63,7 @@ public class Agent : IAgent // any changes to api, update the interface in exten
         private readonly ILogEventAggregator _logEventAggregator;
         private readonly ILogContextDataFilter _logContextDataFilter;
         private Extensions.Logging.ILogger _logger;
-        private IStackExchangeRedisCache _stackExchangeRedisCache;
+        private volatile IStackExchangeRedisCache _stackExchangeRedisCache;
```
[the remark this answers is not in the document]
jaffinito: This was just for safely.  The fixes work with or without the keyword.

tippmar-nr (dismissed):
Looks good.

nr-ahemsath (approved):
LGTM 👍

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -47,7 +47,7 @@ public void Harvest(ISegment hostSegment)
 
             // Get the transaction from the session
             var weakTransaction = sessionData.UserToken as WeakReference<ITransaction>;
-            if (!weakTransaction?.TryGetTarget(out var transaction) || transaction.IsFinished)
+            if (!(weakTransaction?.TryGetTarget(out var transaction) ?? false) || transaction.IsFinished)
```
chynesNR, line 50: Thanks, I noticed this and forgot to comment on it

chynesNR (approved):

tippmar-nr (approved):

nrcventura (approved):

PR #1505, opened 2023-04-03
Added elastic search and kibana to unbounded services

tippmar-nr:
Adds elastic search and kibana to the UnboundedServices `docker-compose.yml`. All security and SSL/TLS configurations are disabled in elastic search to work around the auto-generated certificate file issue. Shouldn't be a problem for our environment. 

Tested that a test app can send data to Elastic Search via http://localhost:9200 and that I can view the data using Kibana (http://localhost:5601).

### DIFF
--- a/tests/Agent/IntegrationTests/UnboundedServices/docker-compose.yml
+++ b/tests/Agent/IntegrationTests/UnboundedServices/docker-compose.yml
@@ -131,4 +131,26 @@ services:
             # the connection string used by those tests
             - MYSQL_ROOT_PASSWORD=${MYSQL_ROOT_PASSWORD:-MysqlPassword}
         container_name: MySqlServer
-        
+
+    elastic:
+        image: elastic/elasticsearch:8.6.2
+        ports:
+            - "9200:9200"
+            - "9300:9300"
+        environment:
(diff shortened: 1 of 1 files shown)


tests/Agent/IntegrationTests/UnboundedServices/docker-compose.yml
```diff
@@ -131,4 +131,26 @@ services:
...
+        environment:
+            - discovery.type=single-node
+            - xpack.security.enabled=false
+            - xpack.security.enrollment.enabled=false
+            - xpack.security.http.ssl.enabled=false
+            - xpack.security.transport.ssl.enabled=false
+        container_name: ElasticServer
+
+    kibana:
+        image: kibana:8.6.2
```
jaffinito, line 149: Formatting is off for the kibana sentry.  I think it will still work though.

tippmar-nr: Yeah, it still works but I'll fix it anyway.

nr-ahemsath (approved):
Works with my 7.x NEST test app.  Thanks for figuring out how to bypass the SSL requirements.  👍

PR #1512, opened 2023-04-06
Support Elastic.Clients.Elasticsearch (8.x) client

### LINKED ISSUE #1492: Instrument Elastic.Clients.Elasticsearch
This is a stretch goal for the April 2023 milestone.  Elastic.Clients.Elasticsearch is the new 8.x client.  It was in beta when this work was initially planned last year.

### Acceptance Criteria
Instrumentation and tests for Elastic.Clients.Elasticsearch on par with 7.x NEST and Elasticsearch.Net.

### Estimates
L

nr-ahemsath:
Resolves #1492 by adding an instrumentation point that works for the 8.x `Elastic.Clients.Elasticsearch`.

The instrumentation wrapper is updated to handle differences between 8.x and 7.x, mainly assembly names, type names and property names.  It also:

* Gets the operation name by a simple string replacement on the request parameters type name (note: when using the low-level `Elasticsearch.Net` client (7.x) instead of  NEST, this is null, so we fall back to a generic operation name.)
* Introduces caching of the getter methods in the AfterWrappedMethodDelegate


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
```diff
@@ -95,5 +75,37 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
                         segment.End(exception);
                     });
         }
+
+        private string GetOperationFromRequestParams(object requestParams)
+        {
+            if (requestParams == null)
```
chynesNR, line 81: Nice, I just ran into this

chynesNR (approved):

nrcventura (approved):
Looks good to me. There were just a couple of local variables that began with an uppercase letter instead of lowercase that could be potentially updated with a follow up PR.

Issue #1484, opened 2023-03-28
ConnectionManager.Connect does not retry 'HttpRequestException's


rwtema:
**Description**
Our services run on an OpenServiceMesh that use an envoy sidecar to route and intercept traffic. This container has a startup time before it is read, and the initial New Relic connection call attempt for collector.newrelic.com seems to occur before the container is ready.

**Expected Behavior**
We expect that the New Relic Agent should retry the connection after 15 seconds.

However, the New Relic Agent simply immediately shuts down.

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/diagnostics-cli-nrdiag/diagnostics-cli-nrdiag/) results**
```
﻿2023-03-28 11:12:16,670 NewRelic   INFO: [pid: 1, tid: 1] The following assembly will be loaded on-demand: *************/newrelic/extensions/NewRelic.Providers.Wrapper.AspNetCore.dll
2023-03-28 11:12:16,773 NewRelic   INFO: [pid: 1, tid: 1] The following assembly will be loaded on-demand: *************/newrelic/extensions/NewRelic.Providers.Wrapper.Owin.dll
2023-03-28 11:12:16,774 NewRelic   INFO: [pid: 1, tid: 1] The following assembly will be loaded on-demand: *************/newrelic/extensions/NewRelic.Providers.Wrapper.Asp35.dll
2023-03-28 11:12:16,774 NewRelic   INFO: [pid: 1, tid: 1] The following assembly will be loaded on-demand: *************/newrelic/extensions/NewRelic.Providers.Wrapper.Sql.dll
2023-03-28 11:12:16,774 NewRelic   INFO: [pid: 1, tid: 1] The following assembly will be loaded on-demand: *************/newrelic/extensions/NewRelic.Providers.Wrapper.SerilogLogging.dll
2023-03-28 11:12:16,822 NewRelic   INFO: [pid: 1, tid: 1] Loading extensions of type NewRelic.Agent.Extensions.Providers.IContextStorageFactory from folder: *************/newrelic/extensions
2023-03-28 11:12:16,832 NewRelic   INFO: [pid: 1, tid: 1] Loading extensions of type NewRelic.Agent.Extensions.Providers.IContextStorageFactory from folder: *************/newrelic/extensions
[...]

PR #1514, opened 2023-04-07
fix: Retry connection on HttpRequestException error (#1484)

tippmar-nr:

## Description

Adds a missing handler for `HttpRequestException` during connect, so that we schedule a restart instead of immediately shutting down the agent.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


chynesNR:
This is probably fine, but I wonder how many scenarios where a `HttpRequestException` is thrown are actually recoverable. It's not great for us to be retrying forever if we're never going to connect. Interested in what others think.

tippmar-nr:
@chynesNR Take a look at the associated bug report, the customer has a use case where there is a delay until the network is "ready" - so this fixes the problem for them. There's also a current support ticket referencing the same issue.

jaffinito (approved):
Makes sense to me.

chynesNR (approved):

PR #1532, opened 2023-04-10
Elasticsearch integration tests framework

chynesNR:
This supports unbounded integration tests for NEST, Elasticsearch.net, and Elastic.Clients.Elasticsearch. At the moment the tests are pretty basic, and only the sync tests for NEST are passing.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
@@ -69,15 +69,8 @@ namespace NewRelic.Agent.Core
         {
             get
             {
-                try
-                {
-                    return Singleton.ExistingInstance;
-                }
-                catch (NullReferenceException)
-                {
-                    // The singleton pointer may be null if we instrument a method that's invoked in the agent constructor.
-                    return DisabledAgentManager;
-                }
+                // The singleton pointer may be null if we instrument a method that's invoked in the agent constructor.
+                return Singleton?.ExistingInstance ?? DisabledAgentManager;
             }
         }
 
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchExerciser.cs
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchExerciser.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
+using System.Runtime.CompilerServices;
 using NewRelic.Agent.IntegrationTests.Shared.ReflectionHelpers;
 using NewRelic.Api.Agent;
 
@@ -21,6 +22,11 @@ namespace MultiFunctionApplicationHelpers.NetStandardLibraries.Elasticsearch
         private const string ASYNC_MODE = "async";
 
         [LibraryMethod]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        /// <summary>
+        /// Sets the library to use for further actions
+        /// </summary>
+        /// <param name="client">ElasticsearchNet, NEST, or ElasticClients</param>
         public void SetClient(string clientType)
         {
             if (Enum.TryParse(clientType, out ClientType client))
@@ -48,6 +54,11 @@ namespace MultiFunctionApplicationHelpers.NetStandardLibraries.Elasticsearch
 
         [LibraryMethod]
         [Transaction]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        /// <summary>
+        /// Performs an Elasticsearch Search operation
+        /// </summary>
+        /// <param name="syncMode">sync or async</param>
         public void Search(string syncMode)
         {
             if (syncMode == ASYNC_MODE)
@@ -62,6 +73,11 @@ namespace MultiFunctionApplicationHelpers.NetStandardLibraries.Elasticsearch
 
         [LibraryMethod]
         [Transaction]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        /// <summary>
+        /// Performs an Elasticsearch Index operation
+        /// </summary>
+        /// <param name="syncMode">sync or async</param>
         public void Index(string syncMode)
         {
             if (syncMode == ASYNC_MODE)
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchExerciser.cs
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchExerciser.cs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
-using Nest;
 using NewRelic.Agent.IntegrationTests.Shared.ReflectionHelpers;
 using NewRelic.Api.Agent;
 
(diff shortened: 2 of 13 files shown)


src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
```diff
@@ -69,15 +69,8 @@ public static IAgentManager Instance
...
-                {
-                    return Singleton.ExistingInstance;
-                }
-                catch (NullReferenceException)
-                {
-                    // The singleton pointer may be null if we instrument a method that's invoked in the agent constructor.
-                    return DisabledAgentManager;
-                }
+                // The singleton pointer may be null if we instrument a method that's invoked in the agent constructor.
+                return Singleton?.ExistingInstance ?? DisabledAgentManager;
```
chynesNR, line 73: Not related to this task but I got the ok during pairing last week.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchSyncTests.cs
```diff
@@ -0,0 +1,279 @@
...
+            NrAssert.Multiple
+            (
+                () => Assertions.MetricsExist(expectedMetrics, metrics),
+                () => Assert.Equal(1, operationDatastoreSpans.Count())
+            );
+        }
+
+    }
+
+    #region NEST
```
chynesNR, line 54: This is currently too many tests (3 libraries x 6 .NET frameworks x 2 (sync/async). We can pare down once we decide on how many library versions we want to check.

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
nrcventura: It might be the case that the previous code relied on the try catch statement instead of the conditional to avoid needing to evaluate the conditional every time this code is executed. It might have been a performance optimization because the problem only occurs during the shorter period where the agent constructor is still executing.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchAsyncTests.cs
```diff
@@ -0,0 +1,280 @@
...
+using Xunit.Abstractions;
+
+namespace NewRelic.Agent.UnboundedIntegrationTests.Elasticsearch
+{
+    public abstract class ElasticsearchAsyncTestsBase<TFixture> : ElasticsearchTestsBase<TFixture>
+        where TFixture : ConsoleDynamicMethodFixture
+    {
+        const string SYNC_MODE = "async";
+
+        protected ElasticsearchAsyncTestsBase(TFixture fixture, ITestOutputHelper output, string clientType) : base(fixture, output, clientType, SYNC_MODE)
```
nrcventura, line 22: If we have separate classes for testing async and sync, we will spin up a separate process for each one. Is it possible to test the sync calls and async calls during the same run of the application? If we can do that, it might help with this test suite running faster.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchAsyncTests.cs
```diff
@@ -0,0 +1,280 @@
...
+            : base(fixture, output, ClientType.NEST.ToString())
+        {
+        }
+    }
+
+    [NetFrameworkTest]
+    public class ElasticsearchNestAsyncTestsFW462 : ElasticsearchAsyncTestsBase<ConsoleDynamicMethodFixtureFW462>
+    {
+        public ElasticsearchNestAsyncTestsFW462(ConsoleDynamicMethodFixtureFW462 fixture, ITestOutputHelper output)
+            // FW462 is testing MongoDB.Driver version 2.3, which needs to connect to the 3.2 server
```
nrcventura, line 87: This comment might have been accidentally included.

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
chynesNR: I suppose that could be, though I'd be surprised if it was a measurable difference. I can wrap it in a `DEBUG` if you think that would be better.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchAsyncTests.cs
chynesNR: True. I was trying to maximize code reuse. I could still do that with multiple tests with a little refactoring. Surprisingly, the whole thing completes in 30 seconds on my machine.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchAsyncTests.cs
chynesNR: Yeah, lot of copy-paste :)

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchAsyncTests.cs
nrcventura: Yeah, I think the impact is more noticeable in our CI because things are not run in parallel there.

nrcventura: I'm also fine with merging the tests as is, and then with a separate PR we could see how much of an impact combining the sync/async tests have on the CI testing time.

chynesNR: I should be able to get it done today. If not we can split it out.

nrcventura: If you do get it done it will be interesting to know just how much of an impact it has on the timing of our integration tests (even if it's just this one sample point).

chynesNR: Now that I've thought about it, I'm not sure how to accomplish this with the ConsoleMF pattern. We need a different set of library calls for sync and async, but we set all the commands to run at process creation time. We can brainstorm in pairing.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchExerciser.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        [LibraryMethod]
+        [Transaction]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        /// <summary>
+        /// Performs an Elasticsearch Search operation
+        /// </summary>
+        /// <param name="syncMode">sync or async</param>
+        public void Search(string syncMode)
```
JcolemanNR, line 62: Hmm.. What about separate sync/async methods at this API level so we can easily differentiate data?

chynesNR: We could, but my understanding is they're both async calls under the hood, so the data should look the same.

nrcventura: It might be the same data, but if they're in different transactions, that's how you can differentiate them.

chynesNR: Fair enough.

nrcventura (dismissed):

JcolemanNR (dismissed):
LGTM

JcolemanNR (approved):

nrcventura (approved):

PR #1535, opened 2023-04-12
Instrument async Elasticsearch methods

nr-ahemsath:
## Description

Adds instrumentation points for `RequestAsync` in Elasticsearch 7.x and 8.x clients.  URI capture from async response is working, but still not being used.  All of the new elasticsearch integration tests are passing except the ones for the low-level `Elasticsearch.net` client (7.x) because the operation name isn't being captured as expected.


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
```diff
@@ -27,10 +30,22 @@ public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
 
         public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
         {
-            //Elasticsearch.Net.HttpMethod,System.String,Elasticsearch.Net.PostData,Elasticsearch.Net.IRequestParameters
+            var indexOfRequestParams = 3; // unless it's Elasticsearch.Net/NEST and async
+
+            if (instrumentedMethodCall.IsAsync)
```
chynesNR, line 35: I think we might want to do some bounds checking, just in case they make some drastic change in a future release, but that can be done later.

chynesNR (approved):
Looks good!

PR #1538, opened 2023-04-12
Set URI after datastore segment creation

nr-ahemsath:
## Description

* Adds the necessary interface and methods to support being able to update the ConnectionInfo (which contains the URI) of a `DatastoreSegment` after initial creation.
* Updates integration tests for Elasticsearch to check for the `peer.address` attribute in the datastore segment span, which is what the URI is used to populate
* Comment cleanup in the wrapper

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
@@ -53,7 +53,7 @@ namespace NewRelic.Providers.Wrapper.Elasticsearch
             var operation = GetOperationFromRequestParams(requestParams);
 
             var transactionExperimental = transaction.GetExperimentalApi();
-            var datastoreSegmentData = transactionExperimental.CreateDatastoreSegmentData(new ParsedSqlStatement(DatastoreVendor.Elasticsearch, model, operation), new ConnectionInfo(string.Empty, string.Empty, databaseName), string.Empty, null);
+            var datastoreSegmentData = transactionExperimental.CreateDatastoreSegmentData(new ParsedSqlStatement(DatastoreVendor.Elasticsearch, model, operation), new ConnectionInfo(string.Empty, string.Empty, string.Empty), string.Empty, null);
             var segment = transactionExperimental.StartSegment(instrumentedMethodCall.MethodCall);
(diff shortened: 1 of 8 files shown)


tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchAsyncTests.cs
```diff
@@ -24,6 +25,7 @@ protected enum ClientType
         }
 
         protected readonly ConsoleDynamicMethodFixture _fixture;
+        protected readonly string _host = ElasticSearchConfiguration.ElasticServer.Remove(0, "http://".Length);
```
chynesNR, line 28: Probably need to account for `https`, too?

chynesNR (approved):
Works for me!

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
```diff
@@ -91,8 +78,10 @@ void InvokeTryProcessResponse(Task responseTask)
                         onSuccess: response =>
                         {
                             var uri = GetUriFromResponse(response);
-                            // TODO: need to figure out how to plumb things so that we can set the uri on the segment after it has already been created.
-                            // See comments above about extending the experimental API
+
+                            var data = segment.GetExperimentalApi().SegmentData as IDatastoreSegmentData;
```
nrcventura, line 82: You only need to call `GetExperimentalApi` once if you save the reference to it.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
```diff
@@ -115,8 +104,9 @@ private static void TryProcessResponse(IAgent agent, object response, ITransacti
 
                 var uri = GetUriFromResponse(response);
 
-                // TODO: need to figure out how to plumb things so that we can set the uri on the segment after it has already been created.
-                // See comments above about extending the experimental API
+                var data = segment.GetExperimentalApi().SegmentData as IDatastoreSegmentData;
```
nrcventura, line 107: You could also just call `GetExperimentalApi` once if you save it to a local variable.

nrcventura (approved):

PR #1540, opened 2023-04-13
ci: apply harden-runner egress policy config

tippmar-nr:
Applies the recommended configuration for the `harden-runner` action to most of our workflows. Note that some workflows haven't been executed yet, so there is no recommended configuration available at this time.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -85,10 +85,10 @@ jobs:
 
     steps:
       # intentionally disabled for this job
-      # - name: Harden Runner
-      #   uses: step-security/harden-runner@03bee3930647ebbf994244c21ddbc0d4933aab4f # v2.3.0
-      #   with:
-      #     egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+      - name: Harden Runner
+        uses: step-security/harden-runner@03bee3930647ebbf994244c21ddbc0d4933aab4f # v2.3.0
+        with:
+          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
 
       - name: Checkout
         uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3 # v3.5.0
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -92,6 +92,7 @@ jobs:
       #<other env vars here>
 
     steps:
+      # intentionally disabled for this job
       # - name: Harden Runner
       #   uses: step-security/harden-runner@03bee3930647ebbf994244c21ddbc0d4933aab4f # v2.3.0
       #   with:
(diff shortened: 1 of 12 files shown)


.github/workflows/all_solutions.yml
```diff
@@ -84,11 +84,11 @@ jobs:
       #<other env vars here>
 
     steps:
-      # intentionally disabled for this job
-      - name: Harden Runner
-        uses: step-security/harden-runner@03bee3930647ebbf994244c21ddbc0d4933aab4f # v2.3.0
-        with:
-          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+      # intentionally disabled for this job, when enabled it causes a failure in the Build Linux Profiler step
```
chynesNR, line 87: Are we planning on revisiting this later?

chynesNR (dismissed):

nrcventura (approved):

.github/workflows/nuget_slack_notifications.yml
```diff
@@ -33,7 +33,15 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@03bee3930647ebbf994244c21ddbc0d4933aab4f # v2.3.0
         with:
-          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+          disable-sudo: true
+          egress-policy: block
+          allowed-endpoints: >
+            api.nuget.org:443
+            collector-1.staging-collectors.newrelic.com:443
```
nr-ahemsath, line 40: This could be a problem if the initial call to the collector (`staging-collector.newrelic.com`) returns something other than `collector-1....` for the actual collector instance the agent should talk to.  Can this action accept wildcards in allowed-endpoints?

FYI, the only reason this job needs to talk to the collector is because I got the bright idea back in November to instrument the Dotty process with the agent to detect errors.  If it becomes a lot of toil to maintain we can just undo that change.

tippmar-nr: Good question. I'll see what I can do.

nrcventura: This sounds like something we'll need to solve for the integration tests anyways.

tippmar-nr: Nope, doesn't look like they support wildcards. I'm starting to wonder if using this action is really going to be worth the effort - I definitely don't want to constantly have to come back and tweak the configuration...

varunsh-coder: Hi @tippmar-nr, I am the one who had emailed you :smile:. For most jobs, the outbound connections don't change often. If you have a job where the outbound calls might be different in each run, you can still run it in `audit` mode, so if there is something odd and you need to investigate, you at least have the insights data (outbound calls/ file overwrites) for analysis. 

We have an [issue related to supporting wildcards](https://github.com/step-security/harden-runner/issues/236) and I can let you know once that is done. We also support storing the policy in a [policy store](https://docs.stepsecurity.io/harden-runner/how-tos/block-egress-traffic#2-add-the-policy-using-the-policy-store) so you can change it without the need to change the workflow file. 

Let me know if you have any other feedback, and I can create an issue to track it. Thanks!

tippmar-nr: Thanks @varunsh-coder - I'm updating to leave the action in `audit` mode. Will make this change in a few other places where we had otherwise disabled the harden-runner action.

.github/workflows/all_solutions.yml
```diff
@@ -588,8 +601,8 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@03bee3930647ebbf994244c21ddbc0d4933aab4f # v2.3.0
         with:
-          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
-
+          egress-policy: audit
```
nr-ahemsath, line 604: Is this intentionally being left on audit?

nr-ahemsath (commented):
Just one comment, otherwise looks good.

.github/workflows/all_solutions.yml
tippmar-nr: It is; this job is currently disabled, so until we run it a few times, we can't use any other policy.

nr-ahemsath (approved):
👍

nrcventura (approved):

PR #1541, opened 2023-04-13
Update set_community_label.yml

angelatan2:
Constraint writes permission to issue updates.  Currently, the community label is not working.

### DIFF
--- a/.github/workflows/set_community_label.yml
+++ b/.github/workflows/set_community_label.yml
@@ -5,7 +5,7 @@ on:
     types: [opened]
 
 permissions:
-  issues: write
+  contents: read
 
 jobs:
   set-community-label:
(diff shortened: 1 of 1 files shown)


.github/workflows/set_community_label.yml
```diff
@@ -5,7 +5,7 @@ on:
     types: [opened]
 
 permissions:
-  contents: read
+  issues: write
```
tippmar-nr, line 8: We should probably add a `permissions:` block to the `Set Community Label If Needed` step rather than at the top-level for this action. I would suggest reverting this change and adding
```
permissions:
    issues: write
``` 
at line 35

tippmar-nr (changes requested):

tippmar-nr (approved):

chynesNR (approved):

nrcventura (approved):

angelatan2:
Please note that I did not rerun the failed integration test since this change does not impact the agent functionality.

PR #1563, opened 2023-04-19
Improve Elasticsearch integration tests

nr-ahemsath:
## Description

Improves Elasticsearch integration tests by:

- Adding tests for the following operations: `IndexMany`, `IndexManyAsync`, `MultiSearch`, `MultiSearchAsync`
- Creating a single `ValidateOperation` method that takes the operation name as an argument, and performs assertions accordingly.  This replaces multiple nearly identical validation methods
- Combining the sync and async tests into a single test class
- Increasing the range of client versions being tested
- Reducing the overall number of test app executions by only testing the oldest and newest framework and core versions

The exerciser code still isn't validating that its operations are actually succeeding.  I started down this path for a bit and ran into a number of issues, so I want to get what I've done PR'ed before continuing with that work.

Note: for the `Elastic.Clients.Elasticsearch` (8.x) case, the code for MultiSearch(Async) isn't actually working.  I was unable to find documentation on how to use this API in 8.x, nor could I find any examples online, and translating working 7.x examples to 8.x did not work.  The empty invocation of the API is still enough to trigger the instrumentation wrapper, and the test assertions succeed accordingly.  

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


chynesNR (approved):
Looks good!

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchTests.cs
```diff
@@ -0,0 +1,324 @@
...
+            _fixture.AddCommand($"ElasticsearchExerciser IndexAsync");
+
+            _fixture.AddCommand($"ElasticsearchExerciser SearchAsync");
+
+            _fixture.AddCommand($"ElasticsearchExerciser IndexManyAsync");
+
+            _fixture.AddCommand($"ElasticsearchExerciser MultiSearchAsync");
+
+            // Don't like having to do this but it makes the async tests pass reliably
+            _fixture.AddCommand("RootCommands DelaySeconds 5");
```
nrcventura, line 68: If you execute the async commands before the sync commands, does that also work consistently?

nr-ahemsath: That's a great idea, and I tried it.  Unfortunately it looks like the five second delay is necessary after the async commands for the tests to pass reliably even if we do the async ones first.

nrcventura (approved):

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchExerciser.cs
```diff
@@ -60,6 +60,16 @@ public void SetClient(string clientType)
...
 
+        [LibraryMethod]
+        [Transaction]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        public void MultiSearch() => _client.MultiSearch();
+
+        [LibraryMethod]
+        [Transaction]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        public void MultiSearchAsync() => _client.MultiSearchAsync();
```
tippmar-nr, line 71: I know we use this pattern a lot where we call an async method without awaiting it, but I don't understand why. Is it because the MFA console app doesn't know how to do async?

nrcventura: I think that this pattern is part of the problem with at least some of the flaky tests.

tippmar-nr: Looks like the pattern could be modified to use proper async and the MFA console app would handle it -- `DynamicMethodExecutor` checks to see if the return is of type `Task` and `.Wait()`s it if so. Looks like quite a few of the other "Exerciser" classes do this already - MySqlExerciser, RestSharpExerciser, etc.

PR #1568, opened 2023-04-21
Elasticsearch error reporting

chynesNR:
Elasticsearch's error handling is a bit unusual. As far as I can tell, if the library layer can detect an error before it sends the request to the server, it doesn't throw an exception or set any kind of error object. Though some of the libraries look like they fake an HTTP status code? In any case, this approach should give us enough error info to be useful.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
@@ -23,6 +23,8 @@ namespace NewRelic.Providers.Wrapper.Elasticsearch
         private const string WrapperName = "RequestWrapper";
 
         private static Func<object, object> _apiCallDetailsGetter;
+        private static Func<object, bool> _successGetter;
+        private static Func<object, object> _exceptionGetter;
         private static Func<object, Uri> _uriGetter;
         private static ConcurrentDictionary<Type, Func<object, object>> _getRequestResponseFromGeneric = new ConcurrentDictionary<Type, Func<object, object>>();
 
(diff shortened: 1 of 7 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/RequestWrapper.cs
```diff
@@ -100,17 +104,42 @@ void InvokeTryProcessResponse(Task responseTask)
...
 
-        private static void TryProcessResponse(IAgent agent, object response, ITransaction transaction, ISegment segment)
+        private static void ReportError(object apiCallDetails)
+        {
+            var exceptionGetter = _exceptionGetter ??= GetExceptionGetterFromApiCallDetails(apiCallDetails);
+            var ex = exceptionGetter(apiCallDetails);
+
+            if ((ex != null) && (ex is Exception exception))
+            {
+                InternalApi.NoticeError(exception);
```
nrcventura, line 114: By calling the InternalApi directly here, the error can potentially get missed or reported to the wrong transaction when we are in an async code path. For the async use case it's often safer to hold a reference to the transaction, and call NoticeError directly on the transaction.

chynesNR: That's a good point, but the `NoticeError` in `ITransaction` only takes an `Exception`. I'll have to think on a clean way to do it.

nrcventura: Yeah, we might want to expand that API so that it can accept non-exception errors.

chynesNR: Done.

nr-ahemsath (approved):
Looks good.  Tests passed locally for me reliably (three consecutive all-green passes). 👍

PR #1578, opened 2023-04-26
test: Add msi content validation.

nrcventura:

## Description

Adds logic to the ArtifactBuilder so that it can extract the contents of the msi and validate that it contains all of the items that we expect. The validation logic was tested locally.

# Author Checklist
- ~~[ ] Unit tests, Integration tests, and Unbounded tests completed~~

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/build/ArtifactBuilder/Artifacts/MsiInstaller.cs
+++ b/build/ArtifactBuilder/Artifacts/MsiInstaller.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Xml;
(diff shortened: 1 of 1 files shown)


build/ArtifactBuilder/Artifacts/MsiInstaller.cs
```diff
@@ -289,14 +287,163 @@ private static IReadOnlyCollection<string> CreateReadOnlyCollection(params strin
...
+
+            var installedFilesRoot = unpackedLocation + $@"\New Relic\.NET Agent\";
+
+            var expectedComponents = GetExpectedComponents(installedFilesRoot);
+
+            var unpackedComponents = GetUnpackedComponents(installedFilesRoot);
+
+            var missingExpectedComponents = new SortedSet<string>(expectedComponents, StringComparer.OrdinalIgnoreCase);
+            missingExpectedComponents.ExceptWith(unpackedComponents);
+            foreach (var missingComponent in missingExpectedComponents)
```
chynesNR, line 321: Maybe it would be better to collect all the errors and then report? So we don't have to fix-build-fix-etc

nrcventura: I thought about that, but for simplicity I kept it following the same behavior as the other similar errors. When we add instrumentation and forget to add it to the MSI we'll see an error that lets us know we missed that adding that instrumentation. Technically there would be 4 possible errors generated, but the first one is good enough for letting us know what the problem is. If we're working on a bigger msi change, we probably need to run the artifact builder locally and treat it as the test project for the msi.

chynesNR (approved):

nr-ahemsath (approved):
Approved.  Ran it locally from this branch and it worked.

PR #1584, opened 2023-04-27
test: complete Windows zip archive content validation from POC

nr-ahemsath:
## Description

This PR completes the POC work for doing Windows ZIP archive content validation.  I liked @nrcventura's approach used for the MSI better, so I ended up doing that instead.  This created some code duplication so I refactored the common code into a helper class.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/build/ArtifactBuilder/Artifacts/ZipArchive.cs
+++ b/build/ArtifactBuilder/Artifacts/ZipArchive.cs
@@ -74,21 +74,8 @@ namespace ArtifactBuilder.Artifacts
 
             var unpackedComponents = ValidationHelpers.GetUnpackedComponents(installedFilesRoot);
 
-            var missingExpectedComponents = new SortedSet<string>(expectedComponents, StringComparer.OrdinalIgnoreCase);
-            missingExpectedComponents.ExceptWith(unpackedComponents);
-            foreach (var missingComponent in missingExpectedComponents)
-            {
-                throw new PackagingException($"The unpacked ZIP archive was missing the expected component {missingComponent}");
-            }
+            ValidationHelpers.ValidateComponents(expectedComponents, unpackedComponents, "ZIP archive");
 
-            var unexpectedUnpackedComponents = new SortedSet<string>(unpackedComponents, StringComparer.OrdinalIgnoreCase);
-            unexpectedUnpackedComponents.ExceptWith(expectedComponents);
-            foreach (var unexpectedComponent in unexpectedUnpackedComponents)
-            {
-                throw new PackagingException($"The unpacked ZIP archive contained an unexpected component {unexpectedComponent}");
-            }
-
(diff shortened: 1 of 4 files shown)


build/ArtifactBuilder/Artifacts/ZipArchive.cs
```diff
@@ -56,73 +55,73 @@ protected override void InternalBuild()
...
-            // 2. Why do the RootInstallDirectoryComponents differ between framework and core?
-            // 3. How to account for "agentInfo.json" in the home dir (both framework and core)?
+            ValidationHelpers.AddFilesToCollectionWithNewPath(expectedComponents, installedFilesRoot, _frameworkAgentComponents.RootInstallDirectoryComponents);
 
-            // cleanup
-            FileHelpers.DeleteDirectories(unpackDir);
-        }
+            var netframeworkFolder = Path.Join(installedFilesRoot, FrameworkSubDirectoryName);
+            ValidationHelpers.AddFilesToCollectionWithNewPath(expectedComponents, netframeworkFolder, _frameworkAgentComponents.AgentHomeDirComponents);
+            expectedComponents.Add(Path.Join(netframeworkFolder, AgentInfo.AgentInfoFilename));
```
nrcventura, line 105: You might not need to special case the agentinfo filename. The file is already available as a property on the frameworkcomponents class. Alternatively, the agentinfo json file is part of the ConfigurationComponents property too.

build/ArtifactBuilder/Artifacts/ZipArchive.cs
```diff
@@ -56,73 +55,73 @@ protected override void InternalBuild()
...
+                throw new PackagingException($"The unpacked ZIP archive was missing the expected component {missingComponent}");
+            }
 
-            // core agent
+            var unexpectedUnpackedComponents = new SortedSet<string>(unpackedComponents, StringComparer.OrdinalIgnoreCase);
+            unexpectedUnpackedComponents.ExceptWith(expectedComponents);
+            foreach (var unexpectedComponent in unexpectedUnpackedComponents)
+            {
+                throw new PackagingException($"The unpacked ZIP archive contained an unexpected component {unexpectedComponent}");
+            }
```
nrcventura, line 89: We might be able to extract this code into the helper too, if we can parameterize the part of the exception messages.

build/ArtifactBuilder/Artifacts/ZipArchive.cs
nr-ahemsath: Thanks, that works better.

build/ArtifactBuilder/Artifacts/ZipArchive.cs
nr-ahemsath: Good idea.

PR #1597, opened 2023-05-03
test: eliminate use of arbitrary delays in ConsoleMF integration tests, part 1

nr-ahemsath:
## Description

In order to help reduce the overall execution time of our CI, as well as be more deterministic, this PR eliminates the use of `RootCommands DelaySeconds` from a bunch of ConsoleMF-based integration tests.  The primary way this has been accomplished is to have the test harness wait for certain agent log lines to appear instead of just sleeping for an arbitrary amount of time.

A few other things:
* The NServiceBus exerciser has been updated to do a more proper async-await pattern that returns a `Task` to the ConsoleMF app that waits on it, rather than doing its own waiting.
* In order to eliminate the need for a 15-second sleep in the 8T tests, which was to allow the SpanStreamingService to connect to the trace observer before exercising the app to generate spans, a new pattern of usage for ConsoleMF is introduced where test application commands can be sent *after* waiting for the appearance of a given agent log line.

The only remaining uses of DelaySeconds are found in the application logging instrumentation tests, where they are being used to work around some un-awaited async calls.  Those tests are complex enough that I plan to tackle solving them in a separate PR.

All of these affected tests are passing for me reliably locally after these changes; we'll see how they go in CI.

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingFlakyTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingFlakyTests.cs
@@ -31,16 +31,6 @@ namespace NewRelic.Agent.IntegrationTests.InfiniteTracing
 
             _fixture.AddCommand("InfiniteTracingTester StartAgent");
 
-            // Give the agent time to warm up... If we send a span too soon, it will be sent via DT (span_event_data) instead of 8T (gRPC)
-            _fixture.AddCommand("RootCommands DelaySeconds 15");
-
-            _fixture.AddCommand("InfiniteTracingTester Make8TSpan");
-            _fixture.AddCommand("InfiniteTracingTester Make8TSpan");
-            _fixture.AddCommand("InfiniteTracingTester Make8TSpan");
-            _fixture.AddCommand("InfiniteTracingTester Make8TSpan");
-            _fixture.AddCommand("InfiniteTracingTester Make8TSpan");
-            _fixture.AddCommand("InfiniteTracingTester Make8TSpan");
-
             _fixture.AddActions
             (
                 setupConfiguration: () =>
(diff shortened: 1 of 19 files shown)


tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/NServiceBus/NServiceBusDriver.cs
```diff
@@ -112,35 +104,33 @@ public void StopNServiceBus()
         [LibraryMethod]
         [Transaction]
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
-        public void PublishEventInTransaction()
+        public async void PublishEventInTransaction()
```
tippmar-nr, line 107: `async void` is [generally considered bad](https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void). 

I think these can all safely be changed to `async Task` instead.

tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingFlakyTests.cs
```diff
@@ -56,6 +46,16 @@ public InfiniteTracingFlakyTestsBase(TFixture fixture, ITestOutputHelper output)
                 ,
                 exerciseApplication: () =>
                 {
+                    // Wait for 8T to connect
+                    _fixture.AgentLog.WaitForLogLine(AgentLogBase.SpanStreamingServiceConnectedLogLineRegex, TimeSpan.FromSeconds(15));
```
tippmar-nr, line 50: Nice!

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/NServiceBus/NServiceBusDriver.cs
nr-ahemsath: Good catch, done.

tippmar-nr (approved):
Looks good. You definitely came up with some more clever solutions than I would have!

nrcventura (approved):

nrcventura (approved):

chynesNR (approved):
Looks good!

tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingTests.cs
```diff
@@ -43,6 +43,7 @@ public InfiniteTracingTestsBase(TFixture fixture, ITestOutputHelper output) : ba
                     // Now send the command to make the 8T Span
                     _fixture.SendCommand("InfiniteTracingTester Make8TSpan");
                     // Now wait to see that the 8T spans were sent successfully
+                    _fixture.AgentLog.WaitForLogLinesCapturedIntCount(AgentLogBase.SpanStreamingSuccessfullySentLogLineRegex, TimeSpan.FromMinutes(1), ExpectedSentCount);
```
chynesNR, line 46: Not for right now, but we should consider adding the ability to wait for multiple conditions, so we don't have to read and regex the whole log more than once.

chynesNR (approved):

PR #1610, opened 2023-05-08
test: Reduce Integration Test Execution Time 

jaffinito:
## Description

Results:  Most jobs saw a reduction in time spent.  GHA does time a bit oddly, but overall it went from 815 total minutes to 737 minutes.  These are GHA usage amounts so it is total time spent running.  Overall, most updated test are taking less than 30 seconds to run.

### Agent changes
Adds harvest cycle overrides for the following using appSettings in the newrelic.config:
* Metrics: `OverrideMetricsHarvestCycle`
* Transaction traces: `OverrideTransactionTracesHarvestCycle`
* Error traces: `OverrideErrorTracesHarvestCycle`
* Span events: `OverrideSpanEventsHarvestCycle`

Adds a similar override for GetAgentCommands allowing the polling interval to be defined in appSettings in the newrelic.config
* `OverrideGetAgentCommandsCycle`

When GetAgentCommands has been overridden the thread profiling duration lower bound is also removed (was hardcoded to 2 minutes).  This allows the tests to request profiling sessions that are much shorter.

All of the above come with private backing fields to limit reading the config repeatedly.  These are not meant to be setting in a running agent and only work if set prior to startup.  There are also log entries when the overrides occurs.  If a value of 0 is set in the override, it will ignore it and use the default values.

Aggregators
The MetricAggregator, TransactionTraceAggregator, and ErrorTraceAggregator were updated to check the new config entries directly.    The SpanEventAggregator was not changed since it already read from a "SpanEventsHarvestCycle" config entry.  That item in config was updated instead.

Commands
The CommandService was updated to use the new `GetsAgentCommandsCycle` config item when starting up or the original default if it was not set.  The required having the CommandService  take a dependency on IConfigurationService.

ThreadProfiling
Since ThreadProfilingService already had access to the configuration, I added a new property called `IgnoreMinMinimumSamplingDuration` that is set only if the GetsAgentCommandsCycle override has been set.    The value is then passed down via the args to the profiling start and stop commands which use it to set the "MinimumSamplingDurationSeconds".  If true, the duration is set to 0 otherwise it is 120.

### Test fixture changes
* NewRelicConfigModifier has new settings for the various overrides.
* The AgentListnerController used for thread profile testing was changed to use 30 seconds as the duration of of the profiling session.

HostedWebCore
While digging into some tests that did not seem to improve when the overrides where in place I found that much of the time was spent in HostedWebCore.  Digging into the issue, I found that all the time was after the shutdown had been requested.  Even with the "immediate shutdown" flag set hostable web core would take nearly 90 seconds to shutdown in some instances.   My reading on this indicates that it is trying to close and release the ports during this time, the immediate flag should force close an connections.

To get around this issue, I updated HostedWebCore to not trigger a shutdown of hostable web core.  Instead, the app is just shutdown.  This resolved the seemingly random slowdowns in the tests, but lead to some additional issues.  Since hostable web core  is not allowed to gracefully shutdown, it is possible that some ports could be orphaned.  Windows will clean these up eventually, but there can be conflicts rarely.  To get around this, I changed the RandomPortGenerator to track and not reuse ports for a given test run.  This reduced my instances of port conflicts considerably.  In our CI we should experience none of these issues since the servers are always fresh and only used for one run.

### Test changes
Many many tests where changed.

Some highlights:
* The thread profiling tests also had their durations all reduced considerably resulting in tests that take ~1 minute to complete vs. 3.5 minutes previously.
* Most tests that used "WaitForLogLine" have been change to use the new shorted harvest cycles - only where appropriate.  
* Most cycles are set to 10 seconds, but the value can and should be tuned to match the data being captured so some will be 15 or even higher where it made sense.  
* Most of the WaitForLogLine timeouts have been reduced from 2 or 3 minutes to 1 minute.
* Some tests had polling or other intervals and those have been reduced as well.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


tests/Agent/IntegrationTests/IntegrationTestHelpers/RandomPortGenerator.cs
```diff
@@ -32,21 +32,27 @@ public static int NextPort()
         {
             lock (_usedPortLock)
             {
-                for (var countAttempts = 0; countAttempts < maxAttempts; countAttempts++)
+                var countAttempts = 0;
```
chynesNR, line 35: Since the ports don't actually need to be random, they just need to be unique, I think we could come up with a way to block off ports that would minimize conflicts. Something based on the timestamp and an incremental counter, for example. That can be a future optimization, though. I don't have it fully fleshed out in my head.

chynesNR (approved):
Very thorough. Nicely done!

tests/Agent/IntegrationTests/IntegrationTestHelpers/RandomPortGenerator.cs
jaffinito: I definitely took the easy route here and just bolted on ignoring the previous ports.

nrcventura (changes requested):
I think that we should add some unit tests to cover the production code changes.

jaffinito:
I agree.

chynesNR (approved):

tests/Agent/IntegrationTests/Applications/ThreadProfileStressTest/Program.cs
```diff
@@ -73,7 +73,7 @@ static void DoTheThing(string port)
 
             //Only hammer the system for 90s instead of the full thread profile duration of 2 minutes so that the agent's
```
nrcventura, line 74: The comment no longer matches the code.

src/Agent/NewRelic/Agent/Core/Aggregators/ErrorTraceAggregator.cs
```diff
@@ -41,6 +42,19 @@ public override void Dispose()
...
+        protected override TimeSpan HarvestCycle
+        {
+            get
+            {
+                if (_configuration.ErrorTracesHarvestCycle.HasValue)
+                {
+                    return _configuration.ErrorTracesHarvestCycle.Value;
+                }
+
+                return DefaultHarvestCycle;
```
nrcventura, line 54: Should we move the default value into DefaultConfiguration so that the logic for the harvest cycle is in just one location?

src/Agent/NewRelic/Agent/Core/Aggregators/MetricAggregator.cs
```diff
@@ -44,6 +45,19 @@ public MetricAggregator(IDataTransportService dataTransportService, IMetricBuild
...
+        protected override TimeSpan HarvestCycle
+        {
+            get
+            {
+                if (_configuration.MetricsHarvestCycle.HasValue)
+                {
+                    return _configuration.MetricsHarvestCycle.Value;
+                }
+
+                return DefaultHarvestCycle;
```
nrcventura, line 57: Should we move the default value into DefaultConfiguration so that the logic for the harvest cycle is in just one location?

src/Agent/NewRelic/Agent/Core/Aggregators/TransactionTraceAggregator.cs
```diff
@@ -33,6 +33,19 @@ public TransactionTraceAggregator(IDataTransportService dataTransportService, IS
...
+        protected override TimeSpan HarvestCycle
+        {
+            get
+            {
+                if (_configuration.TransactionTracesHarvestCycle.HasValue)
+                {
+                    return _configuration.TransactionTracesHarvestCycle.Value;
+                }
+
+                return DefaultHarvestCycle;
```
nrcventura, line 45: Should we move the default value into DefaultConfiguration so that the logic for the harvest cycle is in just one location?

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core.Commands/CommandService.cs
```diff
@@ -20,12 +21,15 @@ public class CommandService : DisposableService
...
+        private readonly IConfigurationService _configurationService;
+
+        public CommandService(IDataTransportService dataTransportService, IScheduler scheduler, IConfigurationService configurationService)
         {
             _dataTransportService = dataTransportService;
+            _configurationService = configurationService;
             _scheduler = scheduler;
 
-            _scheduler.ExecuteEvery(GetAndExecuteAgentCommands, TimeSpan.FromMinutes(1));
+            _scheduler.ExecuteEvery(GetAndExecuteAgentCommands, _configurationService.Configuration.GetAgentCommandsCycle ?? TimeSpan.FromMinutes(1));
```
nrcventura, line 32: Should we move the default value into DefaultConfiguration so that the logic for the harvest cycle is in just one location?

src/Agent/NewRelic/Agent/Core/Aggregators/ErrorTraceAggregator.cs
jaffinito: I considered that (and did that in my first attempt), but I didn't want to lock the different cycles to one setting (1 minute).  I thought it would be more flexible this way.  If we want to centralize things, I can totally do that.

nrcventura: What I'm thinking of is that `ErrorTracesHarvestCycle` should always have a value. If an override is present, it will return the override value. If no override is present, then it will return what should be the default value. I'm not describing having a single setting that controls all of the harvests.

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Core/ThreadProfiling/ThreadProfilingService.cs
```diff
@@ -533,5 +533,13 @@ private FidTypeMethodName[] GetFunctionInfo(UIntPtr[] functionIDs)
             }
             return new FidTypeMethodName[0];
         }
+
+        public bool IgnoreMinMinimumSamplingDuration
+        {
+            get
+            {
+                return _configuration.GetAgentCommandsCycle != TimeSpan.FromMinutes(1);
```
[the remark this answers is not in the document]
jaffinito: I think that could work, let me try something.

chynesNR (approved):

nr-ahemsath (approved):
Great work! 👍

PR #1616, opened 2023-05-09
test: Remove netcoreapp3.1 and net5.0 TFMs from all integration tests

tippmar-nr:
(Apologies in advance for the size of this PR - it got a little bigger than I was expecting it to...)

### TL;DR: 
We now target only .NET 6 and .NET 7 in our integration test suite. References to .NET Core 3.1 and .NET 5.0, both of which are no longer supported by Microsoft, have been removed. Additionally, usages of `WebClient` and `WebRequest` were globally refactored in the integration tests to use `HttpClient` instead.

### Details

* Updated `MultiFunctionApplicationHelpers.csproj` to remove references to `netcoreapp3.1` and `net5.0` build targets and package references. Tests targeting .NET 6 will test the oldest supported package version (typically what the .NET Core 3.1 target was previously testing), while tests targeting .NET 7 will test the newest supported package version. Intermediate versions are no longer tested on .NET 6/7. 
* Refactored projects under the Applications folder that previously targeted `netcoreapp3.1` to target `net7.0` instead. Deleted / renamed / refactored projects that had specific .NET versions in the name (like `AspNetCore3Features`) to have version-agnostic names. All of these projects now target .NET 7 instead.
* Refactored usages of `WebClient` and `WebRequest` throughout all tests and fixtures to use `HttpClient` instead. This turned into a bigger effort than I was expecting, but ultimately will be worth the time spent.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


chynesNR (approved):
Looks good to me!
Going forward, I do think it's important to break up a PR once you see that the size is going to make it difficult to review. 
For this one I would have recommended something like:
* Convert Core 3.1 to .NET  7 where necessary
* Remove Core 3.1 and .NET 5 integration tests
* Switch to using HttpClient

tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatEnabledChainedRequestsHttpClient.cs
nrcventura, line 1: It's not clear if this file was "renamed" or if this test was removed.

tests/Agent/IntegrationTests/IntegrationTests/DistributedTracing/W3CInstrumentationTests/HttpWebRequestW3CTests.cs
nrcventura, line 1: Was this test intentionally removed? It might be a meaningful test on .net framework.

tests/Agent/IntegrationTests/IntegrationTests/DistributedTracing/W3CInstrumentationTests/HttpWebRequestW3CTestsNetCore.cs
nrcventura, line 1: I'm assuming this test was removed because HttpWebRequest is deprecated on newer versions of .net?

tests/Agent/IntegrationTests/IntegrationTests/RemoteServiceFixtures/ServiceStackApplicationFixture.cs
```diff
@@ -15,8 +16,8 @@ public ServiceStackApplicationFixture() : base(new RemoteWebApplication("Service
 
         public void GetHello()
         {
-            var address = $"http://{DestinationServerName}:{Port}/hello/Worked";
-            DownloadJsonAndAssertEqual(address, new HelloResponse("Worked"));
+            var address = $"http://{DestinationServerName}:{Port}/hello/Worked";//"?format=json";
```
nrcventura, line 19: Is the comment leftover code from some older work in progress stuff?

tests/Agent/IntegrationTests/IntegrationTests/RemoteServiceFixtures/TracingChainFixture.cs
```diff
@@ -106,25 +106,6 @@ public RemoteApplication SetupReceiverApplication(bool isDistributedTracing, boo
             return receiverApplication;
         }
 
-        public HttpResponseHeaders ExecuteTraceRequestChainHttpWebRequest(IEnumerable<KeyValuePair<string, string>> headers)
```
nrcventura, line 109: I'm assuming this was removed with some of the W3C chaining tests, but I'm not sure if we want this removed? Or if this should be migrated to a .net framework app?

tests/Agent/IntegrationTests/Shared/WCF/WcfService.cs
```diff
@@ -230,15 +230,16 @@ private async Task<string> DoWork(string value, bool ignoreTransaction, bool thr
             }
 
             ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
-            using (var wc = new WebClient())
+
+            using (var client = new HttpClient())
             {
-                var s = wc.DownloadString(new Uri("https://www.google.com/"));
+                var s = client.GetStringAsync(new Uri("https://www.google.com/")).Result;
```
nrcventura, line 236: You could probably keep this endpoint as async if you just await this call.

tests/Agent/IntegrationTests/Applications/BasicMvcApplication/Controllers/CustomInstrumentationAsyncController.cs
```diff
@@ -57,7 +57,7 @@ public string GetBackgroundThreadWithError()
 
             // Intentionally run this method on a background thread so it won't be caught up as part of the current transaction.
             // Don't wait for it to finish or else it won't be able to generate a transaction trace (because it will be eclipsed by the request)
-            Task.Run(async () => await CustomMethodBackgroundThreadWithError()).Wait();
+            Task.Run(CustomMethodBackgroundThreadWithError);
```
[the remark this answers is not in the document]
tippmar-nr: Actually, it may not be. It looks like the test that exercises this endpoint is failing. Still investigating.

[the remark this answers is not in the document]
tippmar-nr: So the resolution for this is to put the code back like it was and modify how I'm making the call via HttpClient -- the original test fixture had an exception handler because WebClient would throw if there was an error making the remote call. My updates to use HttpClient were using a different method to make the call that didn't throw, thereby either failing in the response validation or requiring that I take the .Wait() off. 

I'm not convinced this test actually does what we want it to, but I've been able to put this controller back the way it was and get the test to pass like it did before. So I guess that's good enough.

tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatEnabledChainedRequestsHttpClient.cs
tippmar-nr: It was removed, being redundant to the existing `CatDisabledChainedRequests` tests. Previously, that test called an endpoint in `BasicMvcWebApplicationTestFixture` that used `WebRequest` and this file (the one I deleted) called a similar endpoing that used `HttpClient` - after refactoring to use `HttpClient` instead, the two endpoints were identical. So I deleted the endpoint that originally used `HttpClient` as well as this file. 

Long explanation, but we're all good here.

tests/Agent/IntegrationTests/IntegrationTests/DistributedTracing/W3CInstrumentationTests/HttpWebRequestW3CTests.cs
tippmar-nr: This file was deleted because `WebClient` usage was removed from the related fixture. The tests previously performed here (via `WebClient`) are now performed in `HttpClientW3CTests`. So we still have .NET Framework and .NET Core coverage.

tests/Agent/IntegrationTests/IntegrationTests/DistributedTracing/W3CInstrumentationTests/HttpWebRequestW3CTestsNetCore.cs
tippmar-nr: Yes, `HttpClientW3CTestsNetCore` is performing the same tests, just using `HttpClient` instead of `WebClient`.

tests/Agent/IntegrationTests/IntegrationTests/RemoteServiceFixtures/ServiceStackApplicationFixture.cs
tippmar-nr: Yep, thanks for catching that.

tests/Agent/IntegrationTests/IntegrationTests/RemoteServiceFixtures/TracingChainFixture.cs
tippmar-nr: It was removed because `ExecuteTraceRequestChainHttpClient()` does the exact same thing and the other `XxxWebRequest` tests were removed as well.

tests/Agent/IntegrationTests/Shared/WCF/WcfService.cs
tippmar-nr: Good catch. I'll update that.

tests/Agent/IntegrationTests/Applications/BasicMvcApplication/BasicMvcApplication.csproj
```diff
@@ -30,6 +30,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <LangVersion>default</LangVersion>
```
jaffinito, line 33: Is this required?  My reading of https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version is that it will default to this setting automatically if it is not present.

tests/Agent/IntegrationTests/IntegrationTests/BasicInstrumentation/MsCorLibTests.cs
```diff
@@ -37,7 +36,7 @@ public MsCorLibTests(RemoteServiceFixtures.BasicMvcApplicationTestFixture fixtur
                     {
                         _fixture.Get();
                     }
-                    catch (WebException ex)
+                    catch (AggregateException ex)
```
jaffinito, line 39: Do we want to pull out the individual exceptions in this?

tests/Agent/IntegrationTests/IntegrationTests/Logging/LocalDecorationTests.cs
```diff
@@ -146,27 +146,9 @@ public Log4netJsonLayoutDecorationEnabledTestsNetCoreLatestTests(ConsoleDynamicM
     }
 
     [NetCoreTest]
-    public class Log4netJsonLayoutDecorationEnabledTestsNetCore60Tests : LocalDecorationTestsBase<ConsoleDynamicMethodFixtureCore60>
+    public class Log4netJsonLayoutDecorationEnabledTestsNetCoreOldestTests : LocalDecorationTestsBase<ConsoleDynamicMethodFixtureCoreOldest>
```
jaffinito, line 149: Nice to see this test get shrunk down a bit.

jaffinito (commented):
All the tests pass locally for me, but I am still working through all the changes.

tests/Agent/IntegrationTests/Applications/BasicMvcApplication/BasicMvcApplication.csproj
tippmar-nr: It was a necessary change so that I could use the "modern" `using` declaration syntax over in https://github.com/newrelic/newrelic-dotnet-agent/pull/1616/files#diff-30592c3c36f251b029ff1e2ba1b7fe563b9225c6d41c106e55b32b6ec76ab3acR66 -- I have no idea why it doesn't default to `default` by default... ;-)

tests/Agent/IntegrationTests/IntegrationTests/BasicInstrumentation/MsCorLibTests.cs
tippmar-nr: I don't think it really matters here -- the top-level exception that's getting thrown from `HttpClient` is `AggregateException` - which is as close to "equivalent" as I can get to `WebException` from `WebClient`.

I think this is a case of optimistic coding for a case that never happens -- at least, I've never seen this test fail with the exception that gets thrown on line 41.

tests/Agent/IntegrationTests/IntegrationTests/RemoteServiceFixtures/TracingChainFixture.cs
nrcventura: I think what's going on is that we had 2 different tests for distributed tracing where we chained multiple calls together. One test was meant to ensure that our HttpWebRequest instrumentation handled chaining correctly, and the other was meant to ensure that our HttpClient instrumentation handled chaining correctly. With the migration to HttpClient from HttpWebRequest, I think that we may have inadvertently removed our distributed tracing chaining test that was meant to validate our HttpWebRequest instrumentation.

tests/Agent/IntegrationTests/IntegrationTests/DistributedTracing/W3CInstrumentationTests/HttpWebRequestW3CTests.cs
nrcventura: Do we still have tests that confirm that our HttWebRequest instrumentation is compatible with the W3C spec?

tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatEnabledChainedRequestsHttpClient.cs
nrcventura: Since you are saying that this test was replaced by a test that previously used HttpWebRequest, do we still have tests that confirm this logic for our HttpWebRequest instrumentation?

tippmar-nr: At this point, we don't have any tests anywhere that exercise `HttpWebRequest` - if you think it's important, I can roll back the changes that removed it. But, since it's deprecated in favor of `HttpClient`, I'm not sure I see any reason to continue testing it.

tippmar-nr: Ok... 

After looking through all the places in `main` that reference `WebRequest`, I understand now where I went astray. I'm working on putting back the tests and supporting fixture / app code that were specific to testing `WebRequest` instrumentation, which will involve about a dozen file updates. 

Once I get all the tests passing locally, I'll commit to the PR branch.

tests/Agent/IntegrationTests/IntegrationTests/DistributedTracing/W3CInstrumentationTests/HttpWebRequestW3CTests.cs
tippmar-nr: This should be resolved with my latest push. WebRequest tests have been restored.

tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatEnabledChainedRequestsHttpClient.cs
tippmar-nr: This should be resolved with my latest push. WebRequest tests have been restored.

jaffinito (approved):
LGTM

nrcventura (approved):
LGTM. 
I think that we should get a story added to our backlog to move the WebRequest tests over to a .net framework application since that API is marked as obsolete.

PR #1630, opened 2023-05-16
ci: Fixed issues with building debug locally

tippmar-nr:
I accidentally hard-coded the build configuration in the paths for my  changes to `build_functions.ps1` - this PR addresses that issue. 

Also, I forgot to remove the profiler build from `build.ps1` after we moved it to a NuGet package. I intentionally left the profiler build section commented out in case we ever need to come back to it.

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -9,7 +9,7 @@ Param(
     [ValidateSet("All","x64","x86")][string]$Architecture = "All",
     [string]$HomePath = "$env:NR_DEV_HOMEROOT",
     [string]$gpgKeyPath = "",
-    [switch]$SkipProfilerBuild = $false,
+    # [switch]$SkipProfilerBuild = $false,
     [switch]$KeepNewRelicConfig = $false,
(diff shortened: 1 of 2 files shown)


build/build.ps1
```diff
@@ -9,7 +9,7 @@ Param(
     [ValidateSet("All","x64","x86")][string]$Architecture = "All",
     [string]$HomePath = "$env:NR_DEV_HOMEROOT",
     [string]$gpgKeyPath = "",
-    [switch]$SkipProfilerBuild = $false,
+    # [switch]$SkipProfilerBuild = $false,
```
chynesNR, line 12: Can we delete this if it's no longer needed?

tippmar-nr: I left this and the related block below commented out in case we need to come back someday.

nrcventura: I think deleting it is fine. We can use source control to bring it back.

chynesNR (approved):

nrcventura (approved):

PR #1633, opened 2023-05-18
test: Clean up async usages in integration tests.

tippmar-nr:

## Description

* Added `Microsoft.VisualStudio.Threading.Analyzers` package to all integration test projects. This is a static code analysis tool that finds and warns on improper async usages. Updated `Directory.Build.props` file to suppress a few warnings from the analysis tool that we don't care about. 
* Updated a few async issues pointed out by the analysis tool.
~* Updated a few places in async methods where we weren't using an equivalent async overload (i.e., `OpenAsync()` instead of `Open()`)~
~* Updated a few places where we were using obsolete methods (primarily MongoDb).~
* Suppressed `Obsolete` warnings in a couple of places that I didn't want to mess with due to complexity of using the recommended  replacement.
* Deleted all files related to IBM DB2, since the code in those files was completely commented out. 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


tests/Agent/IntegrationTests/Applications/WebApiAsyncApplication/Controllers/AsyncFireAndForgetController.cs
```diff
@@ -63,14 +61,12 @@ public async Task<string> Async_FireAndForget()
...
-            return "Worked";
+            return Task.FromResult("Worked");
         }
 
         [HttpGet]
         [Route("AsyncFireAndForget/Async_Sync")]
-#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
-        public async Task<string> Async_Sync()
-#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+        public Task<string> Async_Sync()
```
nr-ahemsath, line 69: Are we sure that this is still testing the scenario it was designed to?  The name of the method being "Async_Sync()" leads me to think that this needs to be an async method to meet the goals of the test.

nrcventura: That's a good question. Our instrumentation looks for the property `IsAsync` to be set to true, and adjusts its behavior. The `IsAsync` property is only true when we're in an `async` method I believe. I don't think that it's enough for a method to just return a `Task` to trigger that behavior.

nr-ahemsath (approved):
Looks good to me. 👍

nrcventura (approved):

PR #1646, opened 2023-05-19
chore: Log Runtime Version at Startup, wait longer for agent log in tests

tippmar-nr:
Two small changes in this PR: 
* Log `System.Environment.Version` at Agent startup, potentially useful for support and debugging.
* Extend the timeout in integration tests when waiting for the Agent logfile from 30 seconds to 1 minute, may help with some tests that take a while to spin up.


src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
```diff
@@ -281,6 +281,7 @@ private void LogInitialized()
                     }
                 }
 
+                Log.Debug($".NET Runtime Version (from System.Environment.Version): {System.Environment.Version}");
```
nrcventura, line 284: Not a requirement, but just a question... Should we only do this for the NETSTANDARD build of the agent?

nrcventura: We already have some code in DotnetVersion.cs that looks up .net versions, we might just able to add the logging code there, or you can look up the data already stored in AgentInstallConfiguration (although you might need to store the raw return value of that environment property in AgentInstallConfiguration).

tippmar-nr: That code only gets you the generic version for Framework, not the specific build version. It could get the major.minor.patch version for .NET Core, but so can either of the two methods I've proposed. 

I'm testing out some changes where I added `System.Runtime.InteropService.RuntimeInformation` to the ILRepack list, seems to work just fine so far.

nrcventura: If that works, should we refactor the code in DotnetVersion to leverage that too (with a separate PR later)?

tippmar-nr: I think we could do that. The ILRepack works fine, check my latest commit.

src/Agent/NewRelic/Agent/Core/Core.csproj
```diff
@@ -124,7 +125,7 @@
     </ItemGroup>
 
     <PropertyGroup>
-      <ILRepackIncludeCount Condition="'$(TargetFramework)' == 'net462'">14</ILRepackIncludeCount>
+      <ILRepackIncludeCount Condition="'$(TargetFramework)' == 'net462'">15</ILRepackIncludeCount>
       <ILRepackIncludeCount Condition="'$(TargetFramework)' == 'netstandard2.0'">11</ILRepackIncludeCount>
```
nrcventura, line 129: It looks like we're only repacking that dependency on .net framework.

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
```diff
@@ -281,6 +282,7 @@ private void LogInitialized()
                     }
                 }
 
+                Log.Debug($".NET Runtime Version: {RuntimeInformation.FrameworkDescription}");
```
nrcventura, line 285: Do we need to wrap this in a try catch block? It also looks like that API was only made available begging in `4.7.1`. You might need to test this on a system that just has 4.6.2 installed (and not newer framework versions).

tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogFile.cs
```diff
@@ -49,7 +49,7 @@ public AgentLogFile(string logDirectoryPath, string fileName = "", TimeSpan? tim
             } while (timeTaken.Elapsed < timeout);
 
             if (throwIfNotFound)
-                throw new Exception("No agent log file found.");
+                throw new Exception($"Waited {timeout.TotalSeconds:N0}s but didn't find an agent log matching {Path.Combine(logDirectoryPath, searchPattern)}.");
```
jaffinito, line 52: I like this! Seems so obvious that it should have been more descriptive.

src/Agent/NewRelic/Agent/Core/Core.csproj
tippmar-nr: It only needs to be packed for the net462 target - I also had to add a package reference for that target. (I thought I had it in the previous commit, but I didn't...)

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
tippmar-nr: I think we're good with the package ref and ILRepack.

Testing on a machine with 4.6.2 only on it is just a bit obnoxious... ;-)

src/Agent/NewRelic/Agent/Core/Core.csproj
nrcventura: Are you sure about that for .net core? It's hard to know which dependencies are shipped with the different "frameworks" available in .net core. You might try testing it with a simple Console application that doesn't add any additional libraries except for maybe the NewRelic.Api nuget and then uses the transaction attribute to create a transaction. Simple console application seem to pull in the least amount of dependencies (from the framework) for .net core.

tippmar-nr: I've been testing in a console app, targeting every tfm I can - net462, net471, coreapp31, net50, net60, net70 . Only need the package for net462.

src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
nrcventura: I made that comment before the nuget package ref was added. I think that with the nuget package it should be fine.

nrcventura (approved):

jaffinito (approved):

PR #1657, opened 2023-05-22
feat: Implement simple in-memory Serilog sink.

jaffinito:
## Description

- Adds a simple in-memory Serilog sink using a ConcurrerntQueue for storage and to handle threading.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Logging/InMemorySink.cs
+++ b/src/Agent/NewRelic/Agent/Core/Logging/InMemorySink.cs
@@ -1,0 +1,43 @@
+﻿// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using Serilog.Configuration;
+using Serilog;
+using Serilog.Core;
+using Serilog.Events;
+using System.Collections.Concurrent;
+
+namespace NewRelic.Agent.Core.Logging
+{
+    public class InMemorySink : ILogEventSink, IDisposable
+    {
+        private static readonly InMemorySink _instance = new InMemorySink();
+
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/Logging/InMemorySink.cs
```diff
@@ -0,0 +1,43 @@
...
+using Serilog.Events;
+using System.Collections.Concurrent;
+
+namespace NewRelic.Agent.Core.Logging
+{
+    public class InMemorySink : ILogEventSink, IDisposable
+    {
+        private static readonly InMemorySink _instance = new InMemorySink();
+
+        public readonly ConcurrentQueue<LogEvent> LogEvents;
```
tippmar-nr, line 17: This might need to be a different container. I discovered yesterday that we'll also need to use this sink in some unit tests, where the tests may ask for all of the log events multiple times while looking for certain log lines.  

Or we might make the queue a private field and expose it as an `IEnumerable<LogEvent>` instead, and offer a separate `public void Clear()` method to drain the queue if needed.

tippmar-nr: Actually... after looking at the unit tests a little bit closer, I think we'll need to refactor the global singleton pattern and just create a private field in `LoggerBootstrapper` for the `InMemorySink` that we can then reference in `EchoInMemoryLogsToConfiguredLogger()`. 

The unit tests are expecting a separate instance of the in-memory sink each time the `NewRelic.Agent.TestUtilities.Logging` class is instantiated.

jaffinito: I'll refactor and update.

PR #1662, opened 2023-05-24
Add Serilog.Async sink, wrap Console and File sinks with Async sink.

tippmar-nr:
Also enables Event Log output for .NET Framework apps only.


jaffinito (approved):
This looks good!

PR #1663, opened 2023-05-25
log4net to Serilog: Unit tests plus code cleanup for serilog implementation

tippmar-nr:
Adds unit tests for the `Logger` and `InMemorySink` class, as well as tests for `LogLevelExtensions`.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Logging/Logger.cs
+++ b/src/Agent/NewRelic/Agent/Core/Logging/Logger.cs
@@ -53,8 +53,6 @@ namespace NewRelic.Agent.Core.Logging
                 case Level.Error:
                     _logger.Error(messageString);
                     break;
-                default:
-                    break;
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/Agent/NewRelic/Agent/Core/Logging/Logger.cs
```diff
@@ -230,11 +216,7 @@ public void Finest(Exception exception)
         /// </summary>
         public void FinestFormat(string format, params object[] args)
         {
-            if (IsFinestEnabled)
```
chynesNR, line 233: I suspect these checks were deliberate, in order to avoid the overhead of formatting the string if we aren't going to log the line. I don't know how much of a performance hit it is, but I could see it being noticeable at Finest.

nrcventura: Yes, they were added a while ago as part of a performance improvement.

tippmar-nr: We have many other places where the strings are formatted directly in `Log.Debug()` or `Log.Finest()`.

`Log.FinestFormat()` is called in about 10 places and `Log.Finest()` with a string format is called in about 12 places.

`Log.DebugFormat()` is called in about 40 places, and `Log.Debug()` with a direct format string is called a similar number of times. 

So if performance is *really* a concern and if checking whether the log is enabled makes a noticeable difference, we ought to go back through the code and  clean up all the stray usages of `Log.Finest()` and `Log.Debug()`.

PR #1672, opened 2023-05-26
log4net to Serilog: Log format cleanup / rework

tippmar-nr:
Fixed some issues with log formatting, particularly in the Audit log output. Added another Serilog package to help with defining the log format similarly to the way we did with log4net.

Modified the various `XxxFormat()` methods to pass the format string and args through to Serilog, which ensures proper formatting of the string especially when the format string contains embedded JSON.

Also re-worked the guard checks in `Logger.cs` and `Log.cs` so that the check is made at the lowest level.


chynesNR (approved):

tippmar-nr:
Waiting on a green CI run before I merge. Not positive I've got all the edge cases handled yet.

jaffinito (approved):
Nice work here!

PR #1679, opened 2023-05-31
fix: Stop double injecting headers with HttpClient on .NET Framework

nrcventura:

## Description

When HttpClient is used in a .NET Framework application it executes both our HttpClient and HttpWebRequest, because on .NET Framework HttpClient relies on code in HttpWebRequest. HttpClient instrumentation always attempts to create the external segment and inject the appropriate outbound headers (and process the response). HttpWebRequest instrumentation uses 2 separate instrumentation points: 1. to create the external segment (and process the response), and 2. to inject the outbound headers. Most of the time when an application uses HttpClient on .net framework, the HttpWebRequest instrumentation does not find a valid transaction (because of the async behavior in HttpClient) and the HttpWebRequest instrumentation does not attempt to inject any outbound headers. Sometimes, HttpWebRequest code to serialize headers is executed on a thread that has access to the transaction, in this situation the HttpWebRequest instrumentation will attempt to inject outbound headers even though the HttpClient instrumentation already injected the outbound headers. This attempt to inject the outbound headers, will use the wrong segment/span as the parent span in that outbound context, and can lead to a situation where the distributed trace may show some incorrect parenting in the trace.

This problem seems to be the root cause for the WCF server tests sometimes reporting an unexpected metric value.

The fix has 2 parts:

1. Makes the external segment created by HttpClient instrumentation a leaf segment to prevent other instrumentation from executing when that segment is the current segment.
2. Update the HttpWebRequest instrumentation to only inject headers when the current segment is an external segment, which should only happen if the HttpWebRequest instrumentation created the current segment.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
@@ -61,7 +61,9 @@ namespace NewRelic.Providers.Wrapper.HttpClient
 
             var externalSegmentData = transactionExperimental.CreateExternalSegmentData(uri, method);
             var segment = transactionExperimental.StartSegment(instrumentedMethodCall.MethodCall);
-            segment.GetExperimentalApi().SetSegmentData(externalSegmentData);
+            segment.GetExperimentalApi()
+                .SetSegmentData(externalSegmentData)
+                .MakeLeaf();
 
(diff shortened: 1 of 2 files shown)


tippmar-nr (approved):
Excellent!

nr-ahemsath (approved):
Looks good!  Thanks for the detailed PR explanation and the code comments.

PR #1685, opened 2023-06-01
feat: Add detailed assembly reporting to enable Vulnerability Management support.

jaffinito:
## Description

- Adds support for Vulnerability Management via a new service, UpdateLoadedModulesService and supported types.  Sends new assembly data on a one minute cycle.  Only new assemblies are reported so it is expected that nothing will be sent from longer running apps.
- Adds unit tests for new code.
- Adds configuration to the UpdateLoadedModulesService so that we can change the cycle time for testing.
- Adds an integration test for Vulnerability Management to AgentFeatures group for Framework and Core.  Only one test was needed since the unit test coverage is high.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/JsonConverters/LoadedModuleWireModelCollectionJsonConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System;
+using NewRelic.Agent.Core.WireModels;
+using Newtonsoft.Json;
+
+namespace NewRelic.Agent.Core.JsonConverters
+{
+    public class LoadedModuleWireModelCollectionJsonConverter : JsonConverter<LoadedModuleWireModelCollection>
+    {
+        private const string JarsName = "Jars";
```
chynesNR, line 12: I'm not sure where the best place is for this, but we might want to document why we're sending a list of "jars" in .NET. :)

jaffinito: Yeah, that is a quirk of the system.  It was meant for Java.  Once the VM team gets some time with the data we send, they will loop back around and let us know what is required and we can trim down what we send.  At the same time we can push for the team to update the endpoint to have an "Assemblies" or similar label for us,

I'll add a note.

src/Agent/NewRelic/Agent/Core/Utilities/UpdatedLoadedModulesService.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            _configurationService = configurationService;
+            _dataTransportService = dataTransportService;
+            _scheduler = scheduler;
+            _scheduler.ExecuteEvery(GetLoadedModules, _configuration.UpdateLoadedModulesCycle);
+        }
+
+        private void GetLoadedModules()
+        {
+            var assemblies = AppDomain.CurrentDomain.GetAssemblies()
```
chynesNR, line 32: I'm still a little concerned about performance if a large number of assemblies are loaded, but I think it's ok just to keep an eye on it for now.

chynesNR (approved):

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Core/Utilities/UpdatedLoadedModulesService.cs
jaffinito: If I pulled in the AgentHealthReporter I could make a few supportability metrics.  Add them to Angler so we could track them.

src/Agent/NewRelic/Agent/Core/JsonConverters/LoadedModuleWireModelCollectionJsonConverter.cs
jaffinito: Updated.

src/Agent/NewRelic/Agent/Core/Utilities/UpdatedLoadedModulesService.cs
nrcventura: I think that we already have timing code in the agent that we could leverage for this. We used them in some infinite tracing code.

nrcventura: The AgentTimerService is what I'm thinking of.

PR #1696, opened 2023-06-06
ci: Dotty will now create issues in our repo for new package versions it finds

nr-ahemsath:
Part of June 2023 Innovation Days.  Our nightly workflow that checks for new versions of core tech libraries will create issues in our repo so we don't lose track of the need to update the versions we test.


tippmar-nr (approved):
Nice work!

chynesNR (approved):
Very cool!

nrcventura (approved):

PR #1697, opened 2023-06-06
ci: let the ArtifactBuilder's content validation logic handle semantic versioning of NuGet packages

nr-ahemsath:
The Unpack() method in the content validation logic for NuGet packages needs to handle the possibility that the version string might only have three components (major.minor.patch) if the fourth component of the agent DLL string is zero.


jaffinito (approved):

chynesNR (approved):
Thanks for the fix!

PR #1698, opened 2023-06-07
ci: ArtifactBuilder parse nuget pack output to find package file name

tippmar-nr:
Modifies ArtifactBuilder to parse the output from the `nuget pack` command to find the actual `.nupkg` filename that was created, eliminating the need to manually build the filename in the `Unpack()` step.

Validated by running `.\build\package.ps1` locally and verifying that all NuGet packages were successfully created and validated.

### DIFF
--- a/build/ArtifactBuilder/NuGetHelpers.cs
+++ b/build/ArtifactBuilder/NuGetHelpers.cs
@@ -8,10 +8,18 @@ namespace ArtifactBuilder
     {
         private static readonly string _nugetPath = Path.Combine(FileHelpers.GetRepoRootDirectory(), @"Build\Tools\nuget.exe");
 
-        public static void Pack(string nuspecFilePath, string outputDirectory)
+        public static string Pack(string nuspecFilePath, string outputDirectory)
         {
             var parameters = $@"Pack -NoPackageAnalysis ""{nuspecFilePath}"" -OutputDirectory ""{outputDirectory}""";
-            var process = Process.Start(_nugetPath, parameters);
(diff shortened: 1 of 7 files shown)


build/ArtifactBuilder/NuGetHelpers.cs
```diff
@@ -22,6 +30,23 @@ public static void Pack(string nuspecFilePath, string outputDirectory)
...
+
+            var stdOut = process.StandardOutput;
+            var output = stdOut.ReadToEnd();
+            Console.WriteLine(output);
+
+            // output is expected to look like:
+            // Attempting to build package from 'NewRelic.Azure.WebSites.x64.nuspec'.
+            // Successfully created package 'C:\Source\Repos\newrelic-dotnet-agent\Build\BuildArtifacts\NugetAzureWebSites-x64\NewRelic.Azure.WebSites.x64.10.11.0.nupkg'.
+
+            // so split on single quotes and take the 4th element in the resulting array
```
chynesNR, line 42: I like this approach, I think it's a lot cleaner. I can think of a couple things to make the parsing less error-prone (in case the output format changes):
* Look for `.nupkg` rather than counting quotes
* Do a `File.Exists()` check to make sure you got the full path right (`GetFileName` sort of does the same thing, but I believe it gives you a result even if the file isn't there)

tippmar-nr: Modified to use a regex for pattern matching and added the `File.Exists()` check.

PR #1708, opened 2023-06-08
feat: add support for MySql.Data version 8.0.33+

nr-ahemsath:
## Description

MySql.Data version 8.0.33 reorganized its methods internally so that all of the methods we instrumented previously (ExecuteReader, ExecuteNonQuery, Read, NextResult) became pass-throughs to async versions that perform the same work.

This PR updates the instrumentation file for the Sql wrapper to add the new async instrumentation points for MySql.Data 8.0.33 and higher.

### DIFF
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
@@ -60,11 +60,11 @@
     <PackageReference Include="MySql.Data" Version="6.10.7" Condition="'$(TargetFramework)' == 'net462'" />
     <PackageReference Include="MySql.Data" Version="8.0.15" Condition="'$(TargetFramework)' == 'net471'" />
     <PackageReference Include="MySql.Data" Version="8.0.30" Condition="'$(TargetFramework)' == 'net48'" />
-    <PackageReference Include="MySql.Data" Version="8.0.32.1" Condition="'$(TargetFramework)' == 'net481'" />
+    <PackageReference Include="MySql.Data" Version="8.0.33" Condition="'$(TargetFramework)' == 'net481'" />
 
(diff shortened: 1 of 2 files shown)


tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
```diff
@@ -60,11 +60,11 @@
     <PackageReference Include="MySql.Data" Version="6.10.7" Condition="'$(TargetFramework)' == 'net462'" />
     <PackageReference Include="MySql.Data" Version="8.0.15" Condition="'$(TargetFramework)' == 'net471'" />
     <PackageReference Include="MySql.Data" Version="8.0.30" Condition="'$(TargetFramework)' == 'net48'" />
-    <PackageReference Include="MySql.Data" Version="8.0.32.1" Condition="'$(TargetFramework)' == 'net481'" />
+    <PackageReference Include="MySql.Data" Version="8.0.33" Condition="'$(TargetFramework)' == 'net481'" />
 
     <!-- MySql.Data .NET/Core references -->
     <PackageReference Include="MySql.Data" Version="6.10.7" Condition="'$(TargetFramework)' == 'net6.0'" />
-    <PackageReference Include="MySql.Data" Version="8.0.32.1" Condition="'$(TargetFramework)' == 'net7.0'" />
+    <PackageReference Include="MySql.Data" Version="8.0.33" Condition="'$(TargetFramework)' == 'net7.0'" />
```
chynesNR, line 67: We should probably leave a comment in here that 8.0.33 is a breaking change for us. Otherwise looks good!

jaffinito (approved):

chynesNR (approved):

nrcventura (changes requested):
@nr-ahemsath and I talked about this and noticed that the async methods should be instrumented using the async tracer factory instead of the sync tracer factory so that the correct datastore segment timing will be captured.

nrcventura (approved):

Issue #1513, opened 2023-04-06
Injection of browser monitoring script into AXD files in version 10.2 and above breaks pages


jaffinito:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

**Description**
In 10.2, we updated our injection logic to avoid a regex problem (see #1247) allowing the agent to use the newer versions of the browser agent.  Prior to v10.2, the agent is pinned to a specific version of the browser agent and this problem does not happen.  Using the RequestPathExclusionList to block AXD files eliminates the issue.

Under normal circumstances, the agent does not inject into items that are not of the contect-type "text/html".  AXD files can be many different types (all at once) depending on what they are returning. For example, "text/css" or "text/javascript".  They also appear to be plain CS files in reality based on examples.

During troubleshooting it, it was also discovered that the AXD can make use of single-quotes in places which our injected browser agent JS also uses. Example:  `xmlRequestFrame.document.write('<html>...</html>'`.  This does not seem consist since most of the other quotes are of the double variety.

Workaround:
Until this issue is resolved, but workaround us to use the [RequestPathExclusionList](https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#browser-path-exclusion) to block AXD files.

Potential Resolutions:
There are two approaches that we could take to resolve this.

1. Using logic similar to the RequestPathExclusionList, block all AXD files from injection.  This is the simplest approach, but also very heavy handed.
2. Update the inject logic to detect the usage of single-quotes and reformat the injected JS to escape its own usage of single-quotes.  This is the most difficult and risky, but also has the best outcome for customers.

3-ish option:  Combine both approaches and create a config setting to determine which is in use.

**Expected Behavior**
Mainly the page doesn't break.  

**Steps to Reproduce**
Did not repro the issue.

**Additional context**
From a ticket.

**For Maintainers Only or Hero Triaging this bug**
*Suggested Priority (P1,P2,P3,P4,P5):* P4
*Suggested T-Shirt size (S, M, L, XL, Unknown):* S or L depending on approach

PR #1710, opened 2023-06-09
fix: Escape single quotes in RUM script when html content is contained in single quotes. (#1513)

tippmar-nr:

## Description

In certain cases (such as a `WebResource.axd` file that is serving javascript that writes an html document to a frame), the html content may be enclosed in single quotes. Even though the content is javascript, the content type may show up as `text/html`, which triggers our injection logic to fire.

An example (from a GTSE) might look like this:
```
xmlRequestFrame = document.frames[callbackFrameID];
if (xmlRequestFrame && xmlRequestFrame.document) {
  window.clearInterval(interval);
  xmlRequestFrame.document.write("");
  xmlRequestFrame.document.close();
  xmlRequestFrame.document.write('<html><body><form method="post"><input type="hidden" name="__CALLBACKLOADSCRIPT" value="t"></form></body></html>');
  xmlRequestFrame.document.close();
}
```

When we inject the RUM script into that quoted html content, we need to escape any single quotes that occur in the script or the resulting javascript will be invalid. 

This fix looks at the content to see if the `<html>` tag is preceded by a single quote and that the subsequent `</html>` tag is followed by a single quote. If so, the RUM script is passed through a function that escapes all single quotes in it (i.e.,  `'` gets replaced with `\'`).

Unit tests were added for this condition but I was unable to create an integration test that exercises this scenario in a "real world" condition.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/BrowserMonitoring/BrowserMonitoringWriter.cs
+++ b/src/Agent/NewRelic/Agent/Core/BrowserMonitoring/BrowserMonitoringWriter.cs
@@ -9,7 +9,7 @@ namespace NewRelic.Agent.Core.BrowserMonitoring
 {
     public class BrowserMonitoringWriter
     {
-        private readonly Func<string> _getJsScript;
+        private Func<string> _getJsScript;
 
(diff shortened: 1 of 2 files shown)


tippmar-nr:
Nevermind... ;-)

PR #1723, opened 2023-06-14
fix: Cache the AgentEnabled setting value.

tippmar-nr:

## Description

Caches the value of the `AgentEnabled` configuration setting when it is first read, rather than reading the value from config every time it is accessed.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -197,7 +197,8 @@ namespace NewRelic.Agent.Core.Configuration
                 {
                     lock (_lockObj)
                     {
-                        _agentEnabledAppSettingParsed = bool.TryParse(_configurationManagerStatic.GetAppSetting("NewRelic.AgentEnabled"), out _appSettingAgentEnabled);
+                        _agentEnabledAppSettingParsed ??= bool.TryParse(_configurationManagerStatic.GetAppSetting("NewRelic.AgentEnabled"),
+                            out _appSettingAgentEnabled);
                     }
                 }
 
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -183,11 +183,10 @@ namespace NewRelic.Agent.Core.Configuration
 
         public object AgentRunId { get { return _serverConfiguration.AgentRunId; } }
 
-        private bool _agentEnabled;
-
-        private static string _agentEnabledAppSettingString;
         private static bool? _agentEnabledAppSettingRead;
         private static bool _agentEnabledAppSettingParsed;
+        private static bool _appSettingAgentEnabled;
+        private static readonly object _lockObj = new object();
 
 
         public virtual bool AgentEnabled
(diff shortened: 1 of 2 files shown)


chynesNR (commented):
My only concern is if you change the value while the Agent is running, does it pick it up? I'm not sure if this getter is in that code path, though. Or, even if you can change the `enabled` status on the fly currently.

tippmar-nr:
@chynesNR I just pushed a pretty significant update. I've been consulting with @nrcventura and think the latest update is good.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -183,17 +183,31 @@ private int TryGetAppSettingAsIntWithDefault(string key, int defaultValue)
 
         public object AgentRunId { get { return _serverConfiguration.AgentRunId; } }
 
+        private bool _agentEnabled;
+
+        private static string _agentEnabledAppSettingString;
+        private static bool? _agentEnabledAppSettingRead;
+        private static bool _agentEnabledAppSettingParsed;
```
nrcventura, line 190: We might need to make these `volatile` since we are not using any operations with memory barriers.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -183,17 +183,31 @@ private int TryGetAppSettingAsIntWithDefault(string key, int defaultValue)
 
         public object AgentRunId { get { return _serverConfiguration.AgentRunId; } }
 
+        private bool _agentEnabled;
```
nrcventura, line 186: We might not need this field anymore because we are not using its value to short-circuit any logic.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -183,17 +183,31 @@ private int TryGetAppSettingAsIntWithDefault(string key, int defaultValue)
...
+        private static bool _agentEnabledAppSettingParsed;
+
+
         public virtual bool AgentEnabled
         {
             get
             {
-                var agentEnabledAsString = _configurationManagerStatic.GetAppSetting("NewRelic.AgentEnabled");
+                // read from app setting one time only and cache the result in a static
+                if (!_agentEnabledAppSettingRead.HasValue)
```
chynesNR, line 198: I think this can probably be simplified a bit? I don't think `_agentEnabledAppSettingRead` needs to be nullable, and `_agentEnabledAppSettingString` can be a local var.

tippmar-nr: I cleaned things up a bit.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -183,17 +183,26 @@ private int TryGetAppSettingAsIntWithDefault(string key, int defaultValue)
 
         public object AgentRunId { get { return _serverConfiguration.AgentRunId; } }
 
+        private static bool? _agentEnabledAppSettingParsed;
```
nrcventura, line 186: Does this field need to be `volatile` since it is only read outside of the lock?

tippmar-nr: `bool?` can't be `volatile`. No idea why...

nrcventura: That surprised me. Should we be checking the value again inside the lock just to be safe (double-checked lock pattern)?

tippmar-nr: Yeah, I'm adding that now. Good catch!

nrcventura (approved):

jaffinito (approved):
Nice!

PR #1727, opened 2023-06-16
fix: Include config file path in the "Agent is disabled " message on all platforms.

tippmar-nr:

## Description

At startup, an exception is thrown when the agent is disabled via config. That message includes the path to the config file where the agent is disabled. For .NET 6+ applications, the path is empty. 

This PR modifies the logic in `ConfigurationLoader.GetConfigSetting()` to include looking at the `appsettings.json` file for .NET 6+ applications and to include the path to the file in the message. As a final fallback, the location of the `newrelic.config` file is used in the message.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
+++ b/src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
@@ -111,6 +111,14 @@ namespace NewRelic.Agent.Core.Config
 				value = new ValueWithProvenance<string>(ConfigurationManager.AppSettings[key],
 					"ConfigurationManager app setting");
 			}
+#else
+            if (value?.Value == null)
+            {
+                var configMgrStatic = new ConfigurationManagerStatic();
+                var configValue = configMgrStatic.GetAppSetting(key);
+                if (configValue != null)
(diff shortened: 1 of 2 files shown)


nrcventura:
Can you explain a little more about how this fixes the problem? The change in this PR looks like it is enabling yet another location where configuration can be defined, which I did not think was the intention of this PR. I might just be missing the necessary context though.

tippmar-nr:
@nrcventura The changes in this PR only extend existing "startup" code that looks for the source of the `agentEnabled` setting so that it checks the `appSettings.json` file when running on .NET 6+ -- previously, that code didn't do anything on .NET 6+.

nrcventura (approved):
Thank you @tippmar-nr for walking through this code with me. We confirmed that the code is already looking at that configuration location in another place, and these code changes are meant to mimic that lookup so that we can provide a more helpful log message. We also discussed that it would be great if we could refactor the code in this file to make it easier to understand and maintain. It is also one of the reasons that #1731 would be helpful so that we could more safely refactor this code.

nr-ahemsath (approved):
Looks good. 👍

PR #1751, opened 2023-06-28
Fix/se redis memory growth

jaffinito:
## Description

Adds a SimpleSchedulingService that makes our Scheduler available from IAgent.  This makes it possible to use the Scheduler in wrappers and extensions.

Adds some new properties to Segment to make it possible to block finished segments and datastore segments from getting added to the cache.

Several StackExchange.Redis 2+ improvements and fixes.
- Moves the storage of the Transaction out of the ProfilingSession (Redis class) and into the ConcurrentDictionary, still as a WeakReference.  This helps ensure that the transaction is not being held onto by the session.
- Implements locks to guard against threading issues with segments in GetProfilingSession and Cleanup.
- Adds a Cleanup method to SessionCache that runs on a 1 minute schedule.  This looks for finished segments and transactions as well as garbage collected transactions and removes them from the cache. 
- Adds new supportability metric to track number of items removed from cache on CleanUp called "Supportability/Dotnet/RedisSessionCacheCleanup/Count"

Adds unit tests for various states of segments and transactions as well as tests that make sure Harvest doesn't throw.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/ISegment.cs
```diff
@@ -68,5 +68,12 @@ public interface ISegment : ISpan
         /// Removes this segment from the top of the agent's internal call stack. Should only be used for asynchronous methods. Calling EndSegment is sufficient for synchronous methods.
         /// </summary>
         void RemoveSegmentFromCallStack();
+
+        string GetCategory();
+
+        /// <summary>
+        /// Will be true if a relative end time has been set on the segment.  In most situations, this is only set when a segment is ended.
+        /// </summary>
+        bool IsDone { get; }
```
nrcventura, line 77: Should we add these to ISegmentExperimental instead?

tests/Agent/UnitTests/CompositeTests/StackExchangeRedisSessionCacheTests.cs
nrcventura, line 1: We should probably address the warnings in this file.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -45,15 +49,14 @@ public void Harvest(ISegment hostSegment)
                 return;
```
nrcventura, line 49: I think that we need to have a lock around this entire conditional. Without the lock, it's possible that the Cleanup method and the method to get a profiling session will be checking the state of things while this code decides to remove harvest the segment.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -71,6 +75,28 @@ public void Harvest(ISegment hostSegment)
...
+        private void CleanUp()
+        {
+            var cleanedSessions = 0;
+            foreach (var pair in _sessionCache)
+            {
+                lock (pair.Key)
+                {
+                    if (pair.Key.IsDone
+                        || !(pair.Value.transaction?.TryGetTarget(out var txn) ?? false)
+                        || txn.IsFinished)
```
nrcventura, line 87: We might want to move the transaction check outside of the lock. I don't think that we want to remove the segments if they are not done, even if the transaction is done. If we cleanup segments that are not done here, it may be the case that the transaction finalization process has not forcibly ended the child segments yet, which would prevent us from capturing the redis segments since this code does not try to capture the redis segments.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -111,25 +137,28 @@ public Func<ProfilingSession> GetProfilingSession()
...
                 // Don't want to save data to a session to a NoOp - no way to clean it up easily or reliably.
+                // Don't want to save to a Datastore segment - could be another Redis segment or something else.
                 var segment = transaction.CurrentSegment;
-                if (!segment.IsValid)
+                ProfilingSession session = null;
+
+                lock (segment)
                 {
-                    return null;
+                    if (segment.IsValid && !segment.IsDone && segment.GetCategory() != "Datastore")
```
nrcventura, line 146: segment.IsValid and segment.GetCategory does not change over time, we might want to pull these checks outside of the lock, and only lock around the conditional that checks that the segment is not done.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -71,6 +75,28 @@ public void Harvest(ISegment hostSegment)
             }
         }
 
+        private void CleanUp()
+        {
+            var cleanedSessions = 0;
+            foreach (var pair in _sessionCache)
```
chynesNR, line 81: To follow up on Chris' above comment, I think we risk this enumerable changing due to `Harvest` as we loop through it.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/ISegment.cs
jaffinito: It works both ways, but experimental does give us options so I am going to see about moving it over.

nrcventura (approved):

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
chynesNR: My mistake, this enumerable is threadsafe.

chynesNR (approved):

chynesNR (approved):

nrcventura (approved):

PR #1752, opened 2023-06-29
ci: Add MSI certificate validation to ArtifactBuilder

tippmar-nr:
* Adds code to ArtifactBuilder to validate the code signing certificate
* Modifies `all_solutions.yml` to run rpm / deb package builders, artifact builder and multiverse scanner on a release or on a manual run when the "Build for release" option is selected.
* Modifies `all_solutions.yml` to add a build step for NewRelic.NuGetHelper and remove previous archive / unarchive steps. 
* Adds a new `BuildTools.sln` containing ArtifactBuilder and NuGetHelper.


build/ArtifactBuilder/SecurityHelpers.cs
```diff
@@ -0,0 +1,199 @@
...
+        //public static bool CheckFile(string filename)
+        //{
+        //    // check digital signature
+        //    var ret = WinTrust.VerifyEmbeddedSignature(filename);
+        //    if (!ret) return false;
+
+        //    // do some other checks - for example verify the subject
+        //    var cert = new X509Certificate2(filename);
+        //    return cert.Verify() && cert.Subject == "foo";
+        //}
```
nrcventura, line 197: Should this be removed too?

tippmar-nr: I left in as a reminder, in case we decide that we need to do more with the cert than just a simple verification.

build/ArtifactBuilder/SecurityHelpers.cs
```diff
@@ -0,0 +1,199 @@
...
+            try
+            {
+                // specify the WinVerifyTrust function/action that we want
+                var action = new Guid(WINTRUST_ACTION_GENERIC_VERIFY_V2);
+
+                // instantiate our WinTrustFileInfo and WinTrustData data structures
+                winTrustFileInfo = new WinTrustFileInfo(fileName);
+                winTrustData = new WinTrustData(winTrustFileInfo);
+
+                var result = WinVerifyTrust(INVALID_HANDLE_VALUE, action, winTrustData);
```
chynesNR, line 177: Can we log the error code here if it fails?

nrcventura (approved):

build/ArtifactBuilder/SecurityHelpers.cs
tippmar-nr: Yeah, I saw some code that takes the result and gets the corresponding system message. I'll add that back in.
Error exceptions will have a message that looks like 
```
Code signing certificate is not valid.  Error result: 800B0109 - A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider.
```

chynesNR: Oh great! I would have been fine with an error code :)

chynesNR (approved):

jaffinito (approved):

PR #1759, opened 2023-07-10
ci: Update AwsLamba unit tests to .NET 7.0

tippmar-nr:
Fixes unit test workflow failure related to .NET Core 3.1 no longer being available on the test runners. (A better solution would be to either remove or exclude the AwsLambda unit test projects since we don't need them, but this will work for now.)

### DIFF
--- a/.github/workflows/run_unit_tests.yml
+++ b/.github/workflows/run_unit_tests.yml
@@ -44,7 +44,7 @@ jobs:
         run: dotnet restore
 
       - name: Build & Run Unit Tests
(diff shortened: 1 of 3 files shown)


nrcventura (approved):
This seems like a good enough compromise for now. It's good to keep these tests around to help verify that changes to shared code (between the agent and the lambda stuff) don't affect the lambda stuff.

jaffinito (approved):

chynesNR (approved):

PR #1763, opened 2023-07-10
chore: Merge log level filtering poc branch to new feature branch

tippmar-nr:
Merged work in the log level filtering poc branch to a new feature branch. Only one minor merge conflict. But might need to run `xsd2code` again just for good measure.


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1990,6 +1990,16 @@ public virtual bool LogDecoratorEnabled
             }
         }
 
+        public virtual IEnumerable<string> LogLevelDenylist
```
tippmar-nr, line 1993: @jaffinito Is this method expected to pick up dynamic changes to the environment variable setting at runtime? If not, we should probably initialize this property one time and `.ToList()` the results -- otherwise every time we check the list in `Agent.cs`, we'll be re-running this enumeration.

jaffinito: I think that having it be dynamic makes more sense than not with this setting.  I imagine that customers would expect this to behave similarly to other config settings and allow updates with an app restart.  What do you think?

tippmar-nr: I'm just concerned about performance - for every log line that we consider for forwarding, we will have to read the environment variable (doing a linear search for it from a list containing every available environment variable) and parse it, then go through the entire enumeration looking to see whether the log line should be excluded.

tippmar-nr: I'm working on a refactor, going to try to use `HashSet<string>` if possible rather than `IEnumerable<string>` - just have to figure out whether JSON deserialization will work correctly or not.

jaffinito (approved):
PoC seems to be workable as a base.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
jaffinito: Sounds great!

PR #1767, opened 2023-07-11
Feature work: log level filtering metrics

tippmar-nr:
Adds metrics to track log lines discarded due to LogLevelDenyList configuration. Includes unit and integration tests. 

I'm open to suggestions for different metric naming if folks don't like what I chose... ;-)


jaffinito (approved):

nrcventura:
Should the "dropped" metrics be incremented too?

tippmar-nr:
> Should the "dropped" metrics be incremented too?

I don't think so, as that metric is for tracking logs that were dropped due to capacity constraints.

Issue #1766, opened 2023-07-11
Add supportability metric for log level filtering


tippmar-nr:
### Describe the story, please be clear on scope of the story.
Add a supportability metric to track the number of log lines discarded due to the log level filtering configuration

Add new unit and/or integration tests to verify the metric is correctly populated.

### Acceptance Criteria
Unit and Integration tests pass and supportability metric is sent to the collector.

### Estimates
Please provide initial t-shirt size
S

Issue #1762, opened 2023-07-10
Update logging configuration docs for log level filtering


jaffinito:
### Describe the story, please be clear on scope of the story.
Update logging configuration docs for log level filtering

### Acceptance Criteria
Docs cover setting up the deny list and its limitations.

Issue #1761, opened 2023-07-10
Update Agent.cs to filter logs using level


jaffinito:
### Describe the story, please be clear on scope of the story.

This is one possible approach:
https://github.com/newrelic/newrelic-dotnet-agent/blob/1a4fed5ae6618e8d0fef0adb8e40ff6416f6b3b2/src/Agent/NewRelic/Agent/Core/Agent.cs#L423-L427

Basically just add a .Contains using StringComparer.OrdinalIgnoreCase.

We want to avoid regex and wildcard to keep the logic simple and fast.

Add unit and integration tests.  This could be updating existing tests to cover the new work or adding a new test completely.

### Acceptance Criteria

- Existing and new unit tests pass
- Existing integration tests pass
- New integration tests pass

Issue #1760, opened 2023-07-10
Update Configuration to support log level filtering


jaffinito:
### Describe the story, please be clear on scope of the story.

- Configuration.xsd
  - Add a new attribute called “logLevelDenylist” to applicationLogging.forwarding.
  - Type: xs:string
  - Marked as optional
  - Documentation string something like “A comma-separated, case-insensitive, list of log levels that should be ignored and not sent up to New Relic.”
  - Add the license header back to Configuration.cs
- IConfiguration.cs
  - Add IEnumerable<string> LogLevelDenylist
- DefaultConfiguration.cs
  - Add a new IEnumerable<string> called “LogLevelDenylist” next to other “application logging” settings
  - Has an EnvironmentOverrides called “NEW_RELIC_APPLICATION_LOGGING_FORWARDING_LOG_LEVEL_DENYLIST”
  - Will need to string.Split the value with “.Split(new[] { StringSeparators.CommaChar, ' ' }, StringSplitOptions.RemoveEmptyEntries);”
- ReportedConfiguration.cs
  - Add IEnumerable<string> LogLevelDenylist
  - Add a JsonProperty of “application_logging.log_level_denylist”
- tests/Agent/UnitTests/Core.UnitTest/DataTransport/ExhaustiveTestConfiguration.cs
  - Add IEnumerable<string> LogLevelDenylist

Old PoC: https://github.com/newrelic/newrelic-dotnet-agent/compare/proto-log-level-deny-list

###Acceptance Criteria

- Have run xsd2code
- Confirm that new entry appears in Configuration.cs
- EnvironmentOverrides is setup properly
- Existing unit test pass
- New unit tests pass

PR #1770, opened 2023-07-12
summary: Log Level Filtering

tippmar-nr:
feat: Add support for filtering log events based on a list of log levels so that they are not forwarded to New Relic. Configurable via `newrelic.config` or environment variable (`NEW_RELIC_APPLICATION_LOGGING_FORWARDING_LOG_LEVEL_DENYLIST`). The list of log levels is comma-delimited and case-insensitive and must match the log level names used by the selected logging framework. Also adds new logging metrics to count the total number of filtered log events (`Logging/denied`), and the number of filtered log events by log level (`Logging/denied/<logLevel>`). Refer to our [log forwarding configuration](https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#applog-forwarding) documents for more details.  (#1760) (#1761) (#1762) (#1766)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1990,6 +1990,20 @@ public virtual bool LogDecoratorEnabled
             }
         }
 
+        private HashSet<string> _logLevelDenyList;
+        public virtual HashSet<string> LogLevelDenyList
+        {
+            get
```
chynesNR, line 1996: Could we log the deny list if we found one? I'm guessing we'll run into some "why doesn't this work" questions where the answer is a typo.

chynesNR:
Looking good!

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
tippmar-nr: @chynesNR I suppose we could, just not sure where we would do that at. Ideas?

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -420,6 +420,15 @@ public void RecordLogMessage(string frameworkName, object logEvent, Func<object,
                 normalizedLevel = string.IsNullOrWhiteSpace(level) ? "UNKNOWN" : level.ToUpper();
             }
 
+            // LogLevelDenyList is already uppercase
```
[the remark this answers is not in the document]
tippmar-nr: Not sure I understand your comment here. What would you do differently?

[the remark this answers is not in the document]
chynesNR: If there's nothing in the deny list, we don't need to check if `normalizedLevel` is blank, or do the `Contains()` call. I'm sure it won't make any practical difference here, but I think it's a good thing to keep in mind.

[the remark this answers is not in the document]
tippmar-nr: Gotcha. 

I would expect `HashSet.Contains()` to be pretty smart when there's no data, so that checking `HashSet.Count > 0` might not be any more efficient. 

I also pushed a change to move where we are checking the deny list so that the `string.Empty` isn't necessary at all.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1990,6 +1990,20 @@ public virtual bool LogDecoratorEnabled
             }
         }
 
+        private HashSet<string> _logLevelDenyList;
+        public virtual HashSet<string> LogLevelDenyList
```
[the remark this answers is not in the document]
tippmar-nr: Added some new unit tests.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
chynesNR: Here is probably ok, since it only happens once. It looks like we log some similar things in DefaultConfiguration

chynesNR: Rather, this is the place we could enforce it only happening once

tippmar-nr: I didn't see the other logging that's already happening in DefaultConfiguration; I'll add it here also.

chynesNR (approved):

jaffinito (approved):

src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -1655,6 +1655,15 @@
                                             </xs:documentation>
                                         </xs:annotation>
                                     </xs:attribute>
+
+                                    <xs:attribute name="logLevelDenyList" type="xs:string" use="optional">
+                                        <xs:annotation>
+                                            <xs:documentation>
+                                                A comma-separated, case-insensitive, list of log levels that should be ignored and not sent up to New Relic.
```
nrcventura, line 1662: Should this call out that the level name used by their logging framework should be used?

nrcventura (approved):

PR #1779, opened 2023-07-14
fix: Remove invalid trailing comma added to W3C tracestate header.

### LINKED ISSUE #1775: Trailing comma added to tracestate header makes it invalid
When `otherVendorTracestates` is empty here, you end up with an invalid tracestate header

https://github.com/newrelic/newrelic-dotnet-agent/blob/6c98ba2820a3f511ffcb2399692088dc510097f9/src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs#L173

jaffinito:
## Description

- Refactors DistributedTracePayloadHandler.BuildTracestate to detect empty vendor states and not use string.Join to create the tracestate which could result in trailing commas.
- Adds unit tests to confirm that we don't add trailing commas.

Resolves #1775 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs
+++ b/src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs
@@ -12,6 +12,7 @@ using NewRelic.Core.DistributedTracing;
 using NewRelic.Core.Logging;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace NewRelic.Agent.Core.DistributedTracing
 {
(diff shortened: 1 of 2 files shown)


chynesNR (approved):

src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs
```diff
@@ -162,12 +162,15 @@ private string BuildTracestate(IInternalTransaction transaction, DateTime timest
             var newRelicTracestate = $"{accountKey}@nr={version}-{parentType}-{parentAccountId}-{appId}-{spanId}-{transactionId}-{sampled}-{priority}-{timestampInMillis}";
             var otherVendorTracestates = string.Empty;
 
-            if (transaction.TracingState != null)
+            if (transaction.TracingState?.VendorStateEntries != null)
```
tippmar-nr, line 165: You might also do a `&& transaction.TracingState?.VendorStateEntries.Any()` and refactor the `if` block on line 171.

jaffinito: I'll take a look!

jaffinito: I updated the if on line 165 with the additional check.
I left the if on line 171 as it was since an empty item in VendorStateEntries could still trigger a trailing comma.  That should never happen, but better to be safe and check.

tippmar-nr (approved):
Looks good!

PR #1783, opened 2023-07-19
test: add an Elasticsearch 7.x server to test infrastructure and use it to test older clients

nr-ahemsath:
This PR allows us to test older versions (all the way back to 7.0.0) of the 7.x clients (NEST and Elasticsearch.Net) to verify our instrumentation's compatibility with the full range of versions we want to support.  We needed to add a 7.x Elasticsearch server to our test infrastructure to enable this.

### DIFF
--- a/tests/Agent/IntegrationTests/Shared/ElasticSearchConfiguration.cs
+++ b/tests/Agent/IntegrationTests/Shared/ElasticSearchConfiguration.cs
@@ -33,7 +33,8 @@ namespace NewRelic.Agent.IntegrationTests.Shared
             }
         }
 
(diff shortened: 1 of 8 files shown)


tippmar-nr (approved):
Looks solid!

chynesNR:
It looks like some of the Elasticsearch tests are failing. Maybe something isn't hooked up quite right?

nr-ahemsath:
> It looks like some of the Elasticsearch tests are failing. Maybe something isn't hooked up quite right?

I forgot to open up the new port on the server.  It should pass on the re-run.

chynesNR (approved):

PR #1795, opened 2023-07-25
feat: Add support for Sitecore.Logging (#1790)

### LINKED ISSUE #1790: Sitecore support
This is the parent ticket to capture work to be done for adding Sitecore support.

chynesNR:
Adds logging support (including context data) for Sitecore.Logging, which uses a very old fork of log4net. Resolves #1790.

Note that the Sitecore.Logging library is only available as a standalone package via myget, but this is only needed for the integration tests.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Log4NetLogging/Log4netWrapper.cs
```diff
@@ -57,7 +64,27 @@ private void RecordLogMessage(object logEvent, Type logEventType, IAgent agent)
...
             var getTimestampFunc = _getTimestamp ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<DateTime>(logEventType, "TimeStamp");
 
-            var getLogExceptionFunc = _getLogException ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<Exception>(logEventType, "ExceptionObject");
+            Func<object, Exception> getLogExceptionFunc;
+
+            try
+            {
+                getLogExceptionFunc = _getLogException ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<Exception>(logEventType, "ExceptionObject");
+            }
+            catch
```
nrcventura, line 73: Would it be better if we used reflection to see if that Property (or field in the other block) exists instead of relying on this sequence of try-catch blocks?

chynesNR: That would probably be a little cleaner, but would mean some refactoring in the VisibilityBypasser. Maybe something for the backlog? This will only throw once if you're running Sitecore, and then the function will be cached, so I don't think there's much overhead (unless I'm missing something).

nrcventura: Overall it would be a 1-time cost so the difference probably would not be noticeable. In the past we didn't perform that check in the visibility bypasser. There is at least one other spot in instrumentation where we needed to do something similar, and took the approach of using reflection to check for the existence of something before attempting to setup the visibility bypasser. I don't have a strong reason to prefer one approach over the other. However, it might make sense like you're suggesting to have something in the backlog so that it's easier to use a consistent pattern (possibly in the VisibilityBypasser or something similar).

tests/Agent/IntegrationTests/nuget.config
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!-- For testing purposes, we want to pull in just the Sitecore logging package -->
+    <add key="Sitecore-MyGet.org" value="https://sitecore.myget.org/F/sc-packages/api/v3/index.json"/>
+  </packageSources>
+  <packageSourceMapping>
+    <!-- key value for <packageSource> should match key values from <packageSources> element -->
+    <packageSource key="Sitecore-MyGet.org">
```
tippmar-nr, line 9: Mentioned in a Slack thread, but looks like Sitecore is moving away from MyGet. https://support.sitecore.com/kb?id=kb_article_view&sysparm_article=KB1002999

jaffinito (approved):

tests/Agent/IntegrationTests/nuget.config
chynesNR: Thanks for the heads up! I added a ticket to fix it later: https://github.com/newrelic/newrelic-dotnet-agent/issues/1796

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Log4NetLogging/Log4netWrapper.cs
chynesNR: Yeah I think a common pattern would be a good idea. I've created a ticket to come back to it: https://github.com/newrelic/newrelic-dotnet-agent/issues/1797

tests/Agent/IntegrationTests/nuget.config
chynesNR: My mistake, I didn't realize the new feed was already available. Fixed!

tippmar-nr (approved):
Looks good!

nrcventura (approved):

PR #1801, opened 2023-07-28
test: Run smoke tests from all solutions build

tippmar-nr:
Invokes the linux smoke test workflow from the all solutions build.

### DIFF
--- a/.github/workflows/run_linux_smoke_tests.yml
+++ b/.github/workflows/run_linux_smoke_tests.yml
@@ -1,6 +1,12 @@
 name: .NET Agent Linux Smoke Tests
 
 on:
+  workflow_call:
+    inputs:
+      external_call:
+        type: boolean
+        default: true
+        required: false
   workflow_dispatch:
     inputs:
       run_id:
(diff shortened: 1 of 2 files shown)


.github/workflows/run_linux_smoke_tests.yml
```diff
@@ -48,7 +54,15 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Download Agent Home Folders
+      - name: Download Agent Home Folders (Call)
+        if: ${{ inputs.external_call }}
+        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
```
nrcventura, line 59: Should the 2 download home folder steps use the same action?

tippmar-nr: No, the first one `(Call)` is used when the workflow is invoked from all_solutions and is downloading an artifact from the "current" run. The other `(Dispatch)` is when running the workflow manually and passing the ID of some previous all_solutions run.

nrcventura: I meant they are using different action libraries. One comes from `actions/download-artifact` and the other comes from `dawidd6/action-download-artifact`.

tippmar-nr: Right, they're two different actions that do different things - one downloads from the current build, one downloads from any specified build. See https://github.com/dawidd6/action-download-artifact#download-workflow-artifact-github-action

nrcventura (approved):

PR #1811, opened 2023-08-01
Add S3 Validation tool.

jaffinito:
## Description

Add S3 Validation tool.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/build/S3Validator/Program.cs
+++ b/build/S3Validator/Program.cs
@@ -18,11 +18,8 @@ namespace S3Validator
                 .WithNotParsed(HandleParseError);
 
             var version = result.Value.Version;
-
             var configuration = LoadConfiguration(result.Value.ConfigurationPath);
-
             var isValid = Validate(version, configuration);
-
             if (!isValid)
             {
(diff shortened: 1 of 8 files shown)


chynesNR (approved):
Very cool!

build/S3Validator/Program.cs
```diff
@@ -0,0 +1,133 @@
...
+
+                results.Add($"{status,-12}{task.Result.RequestMessage?.RequestUri}");
+            }
+
+            if (!isValid)
+            {
+                ExitWithError(ExitCode.Error, "Validation failed. Results:" + Environment.NewLine + string.Join(Environment.NewLine, results));
+            }
+
+            return isValid;
```
nrcventura, line 81: Do we need to return isValid here? It seems like the pattern being used is to log errors and exit immediately, so the method would always return true.

jaffinito: Updated the logic and removed the need for a return.

nrcventura (approved):

PR #1825, opened 2023-08-04
test: Improve integration test reliability

chynesNR:
Explanation of changes added as inline comments.

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
@@ -262,7 +262,6 @@ namespace NewRelic.Agent.IntegrationTestHelpers.RemoteServiceFixtures
                         }
                         catch (Exception ex)
                         {
-                            retryTest = true;
                             TestLogger?.WriteLine("Exception occurred in try number " + (numberOfTries + 1) + " : " + ex.ToString());
                             retryMessage = "Exception thrown.";
                         }
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/Timing.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/Timing.cs
@@ -10,7 +10,7 @@ namespace NewRelic.Agent.IntegrationTestHelpers
     {
         public static readonly TimeSpan TimeToColdStart = TimeSpan.FromMinutes(5);
         public static readonly TimeSpan TimeToConnect = TimeSpan.FromSeconds(30);
-        public static readonly TimeSpan TimeToWaitForLog = TimeSpan.FromMinutes(1);
+        public static readonly TimeSpan TimeToWaitForLog = TimeSpan.FromMinutes(3);
         public static readonly TimeSpan TimeToApm = TimeSpan.FromMinutes(7);
         public static readonly TimeSpan TimeBetweenHarvests = TimeSpan.FromMinutes(1);
         public static readonly TimeSpan ApmCheckInterval = TimeSpan.FromSeconds(5);
--- a/tests/Agent/IntegrationTests/IntegrationTests/AgentFeatures/ThreadProfileTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/AgentFeatures/ThreadProfileTests.cs
@@ -78,8 +78,7 @@ namespace NewRelic.Agent.IntegrationTests.AgentFeatures
             NrAssert.Multiple(
                 () => Assert.Contains(@"""OTHER"":[[[""Native"",""Function Call"",0]", _threadProfileString),
                 () => Assert.Contains(@"[""HostedWebCore.Program"",""Main"",0]", _threadProfileString),
-                () => Assert.Contains(@"System.Threading", _threadProfileString),
-                () => Assert.Contains(@"System.Web", _threadProfileString)
+                () => Assert.Contains(@"System.Threading", _threadProfileString)
             );
         }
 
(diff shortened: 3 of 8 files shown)


tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
```diff
@@ -262,7 +262,6 @@ public virtual void Initialize()
                         }
                         catch (Exception ex)
                         {
```
chynesNR, line 264: This was causing some failing tests to take 45 minutes due to the retries

tests/Agent/IntegrationTests/IntegrationTestHelpers/Timing.cs
```diff
@@ -10,7 +10,7 @@ public static class Timing
     {
         public static readonly TimeSpan TimeToColdStart = TimeSpan.FromMinutes(5);
         public static readonly TimeSpan TimeToConnect = TimeSpan.FromSeconds(30);
-        public static readonly TimeSpan TimeToWaitForLog = TimeSpan.FromMinutes(1);
+        public static readonly TimeSpan TimeToWaitForLog = TimeSpan.FromMinutes(3);
```
chynesNR, line 13: I found some test runs where it took over 2 minutes for the Agent to start

tests/Agent/IntegrationTests/IntegrationTests/AgentFeatures/ThreadProfileTests.cs
```diff
@@ -78,8 +78,7 @@ public void Test()
             NrAssert.Multiple(
                 () => Assert.Contains(@"""OTHER"":[[[""Native"",""Function Call"",0]", _threadProfileString),
                 () => Assert.Contains(@"[""HostedWebCore.Program"",""Main"",0]", _threadProfileString),
-                () => Assert.Contains(@"System.Threading", _threadProfileString),
-                () => Assert.Contains(@"System.Web", _threadProfileString)
+                () => Assert.Contains(@"System.Threading", _threadProfileString)
```
chynesNR, line 81: There are some runs where System.Web is not on the stack

tests/Agent/IntegrationTests/IntegrationTests/InfiniteTracing/InfiniteTracingTests.cs
```diff
@@ -44,7 +44,6 @@ public InfiniteTracingTestsBase(TFixture fixture, ITestOutputHelper output) : ba
                     _fixture.SendCommand("InfiniteTracingTester Make8TSpan");
                     // Now wait to see that the 8T spans were sent successfully
                     _fixture.AgentLog.WaitForLogLinesCapturedIntCount(AgentLogBase.SpanStreamingSuccessfullySentLogLineRegex, TimeSpan.FromMinutes(1), ExpectedSentCount);
-                    _fixture.AgentLog.WaitForLogLinesCapturedIntCount(AgentLogBase.SpanStreamingSuccessfullyProcessedByServerResponseLogLineRegex, TimeSpan.FromMinutes(1), ExpectedSentCount);
                 }
```
chynesNR, line 47: The server does not always return a correct "processed" count

tests/Agent/IntegrationTests/IntegrationTests/Owin/OwinWebApiAsyncTests.cs
```diff
@@ -81,15 +81,15 @@ public void Test()
 
             var generalMetrics = new List<Assertions.ExpectedMetric>
             {
-                new Assertions.ExpectedMetric { metricName = @"Supportability/AnalyticsEvents/TotalEventsSeen", callCount = ExpectedTransactionCount },
+                new Assertions.ExpectedMetric { metricName = @"Supportability/AnalyticsEvents/TotalEventsSeen", CallCountAllHarvests = ExpectedTransactionCount },
```
chynesNR, line 84: These are sometimes split across two harvest cycles, so we need to take the total count

tests/Agent/IntegrationTests/IntegrationTests/RemoteServiceFixtures/AspNetCoreDistTraceRequestChainFixture.cs
```diff
@@ -106,9 +106,11 @@ private void WriteApplicationAgentLogToTestLogger(string applicationName, Remote
         public override void ShutdownRemoteApplication()
         {
             FirstCallApplication.Shutdown();
+            FirstCallApplication.WaitForExit();
```
chynesNR, line 109: We were trying to get data from the log immediately after signaling shutdown, but since the process is set to send data on exit, sometimes we would miss it

tests/Agent/IntegrationTests/UnboundedIntegrationTests/CosmosDB/CosmosDBTests.cs
```diff
@@ -198,7 +198,6 @@ public void BulkCreatingItemsTests()
                 new Assertions.ExpectedMetric { metricName = $"Datastore/operation/CosmosDB/ReadDatabase", metricScope = expectedTransactionName, callCount = 1 },
                 new Assertions.ExpectedMetric { metricName = $"Datastore/operation/CosmosDB/DeleteDatabase", metricScope = expectedTransactionName, callCount = 1 },
                 new Assertions.ExpectedMetric { metricName = $"Datastore/operation/CosmosDB/CreateCollection", metricScope = expectedTransactionName, callCount = 1 },
-                new Assertions.ExpectedMetric { metricName = $"Datastore/statement/CosmosDB/{_testContainerName}/BatchDocument", metricScope = expectedTransactionName, callCount = 1 }
```
chynesNR, line 201: Sometimes this call isn't triggered, but I'm not sure why. We can add a tech debt ticket to investigate if anyone is concerned.

nrcventura: Do we think this is possibly related to an instrumentation bug?

chynesNR: I don't think so. The test seems to be performing operations in two concurrent tasks and hoping they get batched. There is actual batching functionality in the library that we aren't using for whatever reason. I'll put in a task to revisit. Maybe we can make the test more reliable.

chynesNR: https://github.com/newrelic/newrelic-dotnet-agent/issues/1831

tippmar-nr (approved):
Good detective work!

nrcventura (approved):

PR #1830, opened 2023-08-07
Feature work: NuGet Package Deprecator

tippmar-nr:
A tool that queries NR1 for a list of all agent releases with EOL dates, then cross-references that list with non-deprecated NuGet packages to report a list of NuGet package versions that should be deprecated. Creates a new GH issue with the list of versions to be deprecated. Invoked from the post_deploy_agent workflow.


build/NugetVersionDeprecator/Program.cs
```diff
@@ -0,0 +1,1016 @@
...
+    }
+
+    static void ExitWithError(ExitCode exitCode, string message)
+    {
+        Console.WriteLine(message);
+        Environment.Exit((int)exitCode);
+    }
+
+    // for testing only
+    private const string NerdGraphSample = @"
```
jaffinito, line 209: I think this can be removed or moved to a text file/other file to make the class cleaner.

chynesNR:
If we make a second release without resolving the ticket this creates, we'll get a duplicate ticket, right? (Not saying we need to address that, just verifying.)

tippmar-nr:
@chynesNR Yeah, we'll get a second ticket. Don't know any easy way to keep track of which releases were already included in a different ticket...

jaffinito (approved):

nrcventura (approved):

PR #1852, opened 2023-08-16
feat: Log a warning when an unsupported .NET version is detected.

### LINKED ISSUE #1850: Log a warning at startup if application is running an unsupported .NET version
We've run into some issues with customers experiencing unexpected behavior because they are running the latest version of the .NET Agent with an unsupported version of .NET or .NET Framework. We should log this as a warning so it's more obvious during support/troubleshooting. The latest agent currently supports .NET Framework 4.6.2 or higher, and .NET 6 or higher.

### Estimates
S/M

nrcventura:

## Description

Resolves #1850.

Adds a warning to the managed agent logs when an unsupported version of .net is detected. 
- For .net and .net core it checks the version of the runtime used by the application. 
- For .net framework it checks the installed .net framework version (not the targeted version) that the application is running on.

Additional profiler warnings are not generated.

- For .net/.net core the profiler will only attach if the application is running on netcoreapp3.1 or greater. We cannot specifically check for net6 in the profiler, but we could check for net5. However the differences between those versions are unlikely to trigger a scenario where we could publish a warning in the profiler log and the managed log would not also contain a warning.
- For .net framework the profiler will not attach unless .net framework >= 4.6.1 is installed. We cannot easily check for .net framework 4.6.2. No additional warning is necessary in this case.

Other notes:

- There is not a simple way for the agent to automatically detect if a .net framework application is targeting an unsupported version. This would involve knowing which assembly to check for the TargetFrameworkAttribute.
- If a supported .net framework version is installed and the application is targeting an unsupported framework, the agent can still dynamically load dependencies from the installed framework version. However, the compatibility quirks mode that is enabled by the unsupported target version may lead to unexpected behavior differences.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/NewRelic.Agent.Core/AgentManager.cs
@@ -268,7 +268,7 @@ namespace NewRelic.Agent.Core
 
                 foreach (var ev in environmentVariables)
                 {
-                    if (!String.IsNullOrEmpty(System.Environment.GetEnvironmentVariable(ev)))
+                    if (!string.IsNullOrEmpty(System.Environment.GetEnvironmentVariable(ev)))
                     {
                         Log.DebugFormat("Environment Variable {0} value: {1}", ev, System.Environment.GetEnvironmentVariable(ev));
                     }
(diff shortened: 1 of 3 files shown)


chynesNR (approved):
Thanks for the thorough analysis!

tippmar-nr (approved):

PR #1853, opened 2023-08-16
feat: Use HttpWebRequest instead of HttpClient on .NET Framework

### LINKED ISSUE #1844: HttpWebRequest should be used to send data to New Relic on .NET Framework

Older versions of the agent used to use HttpWebRequest to send data to New Relic for both .net framework and .net core applications. On .net/.net core HttpWebRequest is obsolete, so the agent was updated to use HttpClient instead. However, HttpClient in some .net framework apps that do not use the version of HttpClient that is built into .net framework, have experienced problems with the dotnet agent's usage of HttpClient. As a team, we decided that the agent should go back to sending data to New Relic using HttpWebRequest on .net framework.

### Acceptance Criteria

1. .Net Framework applications will send data to New Relic using HttpWebRequest, just like older agent versions used to.
2. .NET/.Net core applications will send data to New Relic using HttpClient, just the like the current agent version does.

### Dependencies

N/A

Alternatives were considered and documented in an internal doc. This approach was determined to be the most helpful with the fewest risks.

### Estimates

t-shirt size: M

tippmar-nr:
* Major refactor of `HttpCollectorWire` to introduce an abstraction layer that allows usage of either `System.Net.HttpClient` (on .NET 6+) or `System.Net.HttpWebRequest` (on .NET Framework 4.6.2+) for sending data to the NR1 collector. 
* Adds a number of new interfaces and related implementations to make the abstraction layer more testable
* Adds unit tests for the new code.

Resolves #1844


src/Agent/NewRelic/Agent/Core/DataTransport/Client/HttpClientBase.cs
```diff
@@ -0,0 +1,103 @@
...
+            catch (Exception)
+            {
+                Log.ErrorFormat("Unable to resolve host name \"{0}\"", host);
+            }
+
+            TestConnection();
+        }
+
+#if NETFRAMEWORK
+        // use WebClient for .NETFramework builds
```
chynesNR, line 49: Nitpick: Maybe we could combine these and have the `#if` inside the call? It's really just the Get that's different

chynesNR (approved):
Looks good! Very clean

jaffinito (approved):
Great work!

src/Agent/NewRelic/Agent/Core/DataTransport/Client/NRWebRequestClient.cs
```diff
@@ -0,0 +1,98 @@
...
+                throw;
+            }
+        }
+
+        public override void Dispose()
+        {
+        }
+
+        // for unit testing only
+        public void SetHttpWebRequestFunc(Func<Uri, HttpWebRequest> fooFunc )
```
nrcventura, line 92: We should probably rename `fooFunc` to something else.

nrcventura: Alternatively you might be able to avoid needing this method if you just passed the create method as a .ctor parameter. That method can be registered with the IOC container, and the test can define its own method.

tests/Agent/UnitTests/Core.UnitTest/DataTransport/HttpCollectorWireWebRequestTests.cs
```diff
@@ -0,0 +1,377 @@
...
+
+        public byte[] ByteArrayContent { get; set; }
+
+        public long ContentLength => ByteArrayContent?.Length ?? StringContent?.Length ?? 0;
+
+        public string ContentEncoding { get; set; }
+    }
+
+    [TestFixture]
+    public class HttpCollectorWireWebRequestTests
```
nrcventura, line 40: I'm assuming these tests are very similar to the HttpClient tests. Should we try to ensure that we are running the same tests for both? Right now it is easier for the tests to diverge.

tippmar-nr: After looking at the tests, I can do a significant refactor that will eliminate the need for two sets of nearly parallel tests.

src/Agent/NewRelic/Agent/Core/DataTransport/Client/NRWebRequestClient.cs
tippmar-nr: Gonna leave the method in place for now.

nrcventura (dismissed):

src/Agent/NewRelic/Agent/Core/DataTransport/Client/NRHttpClient.cs
```diff
@@ -22,7 +23,10 @@ public class NRHttpClient : HttpClientBase
         public NRHttpClient(IWebProxy proxy, IConfiguration configuration) : base(proxy)
         {
             _configuration = configuration;
-            _httpClientWrapper = new HttpClientWrapper(new HttpClient(new HttpClientHandler { Proxy = proxy }, true));
+
+            // set the default timeout to "infinite", but specify the configured collector timeout as the actual timeout for SendAsync() calls
```
nrcventura, line 27: Do we need to set this timeout to infinite? It seems like it would be dangerous to do if some bug prevented the request level timeout from being applied.

tippmar-nr: The default timeout has to be greater than any timeout we'd want to specify. But since we always specify a timeout for the cancellation token passed to the `.SendAsync()` call (in HttpClientWrapper), it's not an issue.

nrcventura (approved):

PR #1855, opened 2023-08-18
ci: Run all solutions build only on weekdays

tippmar-nr:
Updates the schedule for `all_solutions.yml` so that it only runs on weekdays rather than every day. 

image


chynesNR (approved):
Nice!

jaffinito (approved):

Issue #1859, opened 2023-08-22
De-prioritize Microsoft.Extensions.Logging instrumentation


chynesNR:
Based on our investigation of https://github.com/newrelic/newrelic-dotnet-agent/issues/1785, it looks like the best approach is:
* For logging libraries we already instrument that are registered with Microsoft.Extensions.Logging (i.e. Serilog, log4net, and NLog), use that library's instrumentation
* For all other loggers registered with MEL, use our MEL instrumentation

This allows us much cleaner access to the scope/context data, and covers the case where the logger is invoked directly, outside of MEL.

### Acceptance Criteria
* Passing integration tests for Serilog.Extensions.Logging and NLog.Extensions.Logging instrumentation
* Existing MEL integration tests still pass

### Estimates
M

PR #1860, opened 2023-08-23
feat: Add support for Serilog.Extensions.Logging and NLog.Extensions.Logging

chynesNR:
Previously, if we detected Microsoft.Extensions.Logging and another logger, we would only instrument MEL, in order to avoid double collection of logs. This turned out to be a flawed approach, because MEL does not store any context data. It relies on the underlying logger to handle it. So now, if we detect that scenario, we disable MEL instrumentation. Only if we don't find another logger that we know how to instrument, will we instrument MEL.

In order for us to collect context data from an ILogger via MEL, the logger must implement `IExternalScopeProvider` (specifically the `ForEachScope` function). Scope data can be any object, but we will only collect it if it's a key-value pair (or collection of key-value pairs). This is how NLog and Serilog implement their scope provider, so this seems like a reasonable restriction.

I also found that NLog treats context data and scope data differently. We weren't recording NLog's scope data at all before; now we are.

Addresses #1859

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Logging/LogProviders.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Logging/LogProviders.cs
@@ -6,21 +6,18 @@ using System.Collections.Generic;
 
 namespace NewRelic.Agent.Extensions.Logging
 {
-    public enum LogProvider
-    {
-        Log4Net,
-        Serilog,
-        NLog
-    }
-
     public static class LogProviders
     {
-        public static readonly bool[] RegisteredLogProvider = new bool[Enum.GetNames(typeof(LogProvider)).Length];
+        public static bool KnownLogProvider = false;
 
-        public static readonly List<string> Log4NetProviderNames = new List<string> { "Microsoft.Extensions.Logging.Log4NetProvider", "log4net.Extensions.Logging.Log4NetProvider" };
-
-        public static readonly List<string> SerilogProviderNames = new List<string> { "Microsoft.Extensions.Logging.SerilogLoggerProvider", "Serilog.Extensions.Logging.SerilogLoggerProvider" };
-
-        public static readonly List<string> NLogProviderNames = new List<string> { "Microsoft.Extensions.Logging.NLogLoggerProvider", "NLog.Extensions.Logging.NLogLoggerProvider" };
+        public static readonly List<string> KnownLogProviders = new List<string>
+        {
+            "Microsoft.Extensions.Logging.Log4NetProvider",
+            "log4net.Extensions.Logging.Log4NetProvider",
+            "Microsoft.Extensions.Logging.SerilogLoggerProvider",
+            "Serilog.Extensions.Logging.SerilogLoggerProvider",
+            "Microsoft.Extensions.Logging.NLogLoggerProvider",
+            "NLog.Extensions.Logging.NLogLoggerProvider"
+        };
     }
 }
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Logging/LogProviders.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Logging/LogProviders.cs
@@ -8,9 +8,11 @@ namespace NewRelic.Agent.Extensions.Logging
 {
     public static class LogProviders
     {
-        public static bool KnownLogProvider = false;
+        // This will only be set once, when a Microsoft.Extensions.Logging registration method is called.
+        // It can safely be read concurrently.
+        public static bool KnownMELProviderEnabled = false;
 
-        public static readonly List<string> KnownLogProviders = new List<string>
+        public static readonly List<string> KnownMELProviders = new List<string>
         {
             "Microsoft.Extensions.Logging.Log4NetProvider",
             "log4net.Extensions.Logging.Log4NetProvider",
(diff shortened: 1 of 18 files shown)


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Logging/LogProviders.cs
```diff
@@ -6,21 +6,18 @@
...
-    }
-
     public static class LogProviders
     {
-        public static readonly bool[] RegisteredLogProvider = new bool[Enum.GetNames(typeof(LogProvider)).Length];
-
-        public static readonly List<string> Log4NetProviderNames = new List<string> { "Microsoft.Extensions.Logging.Log4NetProvider", "log4net.Extensions.Logging.Log4NetProvider" };
-
-        public static readonly List<string> SerilogProviderNames = new List<string> { "Microsoft.Extensions.Logging.SerilogLoggerProvider", "Serilog.Extensions.Logging.SerilogLoggerProvider" };
+        public static bool KnownLogProvider = false;
```
nrcventura, line 11: Should we add a comment here explaining that this static field is currently only used by MEL instrumentation, and it is not accessed concurrently (there are only concurrent reads)? I'm bringing this up because the field is not protected by any interlocked operations or locks, and if we introduce a new usage for it, we need to reassess whether or not there is concurrent usage of this field.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
```diff
@@ -71,27 +75,24 @@ private void RecordLogMessage(MethodCall methodCall, MEL.ILogger logger, IAgent
 
             try
             {
-                // We are trying to access this property:
-                // logger.Loggers[0].Logger.ScopeProvider
+                // MEL keeps an array of scope handlers. We can ask one of them for the current scope data.
 
-                // Get the array of Loggers (logger.Loggers[])
```
chynesNR, line 77: The property we were reading is only available in Microsoft.Extensions.Logging.ConsoleLogger, which won't necessarily be loaded

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
```diff
@@ -28,7 +28,11 @@ public class MicrosoftLoggingWrapper : IWrapper
 
         public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
         {
-            return new CanWrapResponse(WrapperName.Equals(methodInfo.RequestedWrapperName));
+            if (!LogProviders.KnownLogProvider)
```
nrcventura, line 31: This might read clearer if `KnownLogProvider` was changed to something like `KnownLogProviderEnabled` or `KnownMELProviderEnabled`. Renaming this field would also make it easier to see the difference between that field and the list of known providers `KnownLogProviders`.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
```diff
@@ -71,27 +75,24 @@ private void RecordLogMessage(MethodCall methodCall, MEL.ILogger logger, IAgent
...
-                var getLoggersArrayFunc = _getLoggersArray ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<dynamic>(logger.GetType(), "Loggers");
+                // Get the array of ScopeLoggers (logger.ScopeLoggers[])
+                var getLoggersArrayFunc = _getLoggersArray ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<dynamic>(logger.GetType(), "ScopeLoggers");
                 var loggers = getLoggersArrayFunc(logger);
 
-                // Get the first logger in the array (logger.Loggers[0])
+                // Get the first ScopeLogger in the array (logger.ScopeLoggers[0])
+                // If there is more than one scope logger, they've all received the same data, so the first
+                // one should be fine
                 object firstLogger = loggers.GetValue(0);
```
nrcventura, line 87: Is it possible that this array has a length of 0?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
chynesNR: Sure, that makes sense.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
chynesNR: It's a bit confusing, but as far as I can tell, no. The ILogger still seems to be stored as a scope logger, even if it isn't one (There's an integration test that does this). Worst case, this will throw and be caught, resulting in us disabling trying to read scope data going forward. If there are no scope loggers, that would be correct since we wouldn't have any way to get the data.

nr-ahemsath (approved):
Looks good!  I appreciate all of the comments to help understand what is going on. 👍

nrcventura (approved):

PR #1862, opened 2023-08-24
test: Extend harvest from 15 to 45 seconds for MySql tests.

tippmar-nr:
The overnight run failed again, even after adding the "warmup" code. So we'll just extend the MySql harvest interval to 45 seconds which should make things more stable. Maybe. I hope.


chynesNR (approved):

nrcventura (approved):
As a follow up, should we remove the warmup code that didn't help?

PR #1873, opened 2023-08-29
fix: Fix NRHttpClientFactory so that it creates only one client.

jaffinito:
## Description

Adds a log line to NRHttpClient that reports the `HttpClientHandler.SslProtocols`.  This value can differ from the TLS settings capture in previous logging since `System.Net.ServicePointManager.SecurityProtocol` changes don't pass through to the HttpClientHandler.  This new log line should appear in the same general area as the current TLS message.  it does not replace that message.  it also only exists for HttpClient and not HttpWebRequest.

While adding this log line, I noticed that it was appearing multiple times when it should only occur once per the lifetime of the IoC container.  There was a bug in the NRHttpClientFactory that resulted in it creating a new client every time CreateClient was called instead of returning the cached client.

Refactored NRHttpClientFactory to return the cached client unless the a proxy configuration is added or removed.  Added some unit tests around NRHttpClientFactory to confirm that the client is only replaced when the proxy is added or removed.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/Client/NRHttpClient.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/Client/NRHttpClient.cs
@@ -9,6 +9,7 @@ using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using NewRelic.Agent.Configuration;
 using NewRelic.Agent.Core.DataTransport.Client.Interfaces;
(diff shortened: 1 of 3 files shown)


chynesNR (approved):
Good catch!

nrcventura (approved):

tippmar-nr (approved):

PR #1876, opened 2023-08-30
ci: Workflow Updates

tippmar-nr:
* Adds a new reusable workflow to check whether files outside the `.github/**` folder were modified.
* Modifies several workflows to skip running if the reusable workflow indicates nothing of merit was modified.
* Modifies `markdowncheck.yml` so that it only runs when `**.md` files are modified in a push to any branch and schedules it to run nightly to help catch any dead links.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -47,8 +47,9 @@ jobs:
     name: Build FullAgent and MSIInstaller
     runs-on: windows-2022
     needs: check-modified-files
+    # don't run this job if triggered by Dependabot, will cause all other jobs to be skipped as well
     # run this job if non-workflow files were modified, or if triggered by a release or a manual execution
-    if: needs.check-modified-files.outputs.non-workflow-files-changed == 'true' || github.event.release || github.event_name == 'workflow_dispatch'
+    if: github.actor != 'dependabot[bot]' && (needs.check-modified-files.outputs.non-workflow-files-changed == 'true' || github.event.release || github.event_name == 'workflow_dispatch')
 
     env:
       fullagent_solution_path: ${{ github.workspace }}\FullAgent.sln
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -6,8 +6,6 @@ on:
     branches:
       - main
       - "feature/**"
-    paths-ignore:
-      - ".github/**" # skip changes to the .github branch (workflows, etc.)
   release:
     types: [published]
   workflow_dispatch:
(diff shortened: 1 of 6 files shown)


.github/workflows/markdowncheck.yml
```diff
@@ -1,6 +1,13 @@
 name: Check Markdown links
 
-on: push
+on: 
+  push:
+    paths:
+      - '**.md' # only run when there are changes to markdown files in a push to any branch
```
nr-ahemsath, line 6: We rarely modify the markdown files, and sometimes the markdown linter lets us know that links which used to be correct have gone away.  Something to consider.

tippmar-nr: Yeah, that's why I added the "sanity check" a couple of lines farther down - we'll run markdown check on every PR, but not necessarily every push.

.github/workflows/all_solutions.yml
```diff
@@ -33,13 +33,43 @@ env:
...
 
 jobs:
+  check-modified-files:
+    name: Check if source files were modified, skip remaining jobs if not
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    # don't run this job if triggered by Dependabot, will cause all other jobs to be skipped as well
+    # don't run if triggered by a release or manual execution, as we're going to do the full build in those cases anyway
+    if: ${{ github.actor != 'dependabot[bot]' }} || !(github.event.release || github.event_name == 'workflow_dispatch')
```
jaffinito, line 43: I don't think we need the braces and they also don't cover the whole statement.

```suggestion
    if: github.actor != 'dependabot[bot]' || !(github.event.release || github.event_name == 'workflow_dispatch')
```

.github/workflows/markdowncheck.yml
tippmar-nr: Actually, I think a better way would be to change the workflow to run anytime a .md file is modified and then as a scheduled nightly check.

.github/workflows/repolinter.yml
```diff
@@ -6,8 +6,11 @@ name: Repolinter Action
 # Currently there is no elegant way to specify the default
 # branch in the event filtering, so branches are instead
 # filtered in the "Test Default Branch" step.
-on: [push, workflow_dispatch]
-
+on: 
+  push: 
+    paths-ignore:
+      - ".github/**" # skip changes to the .github branch (workflows, etc.)
```
nrcventura, line 12: I'm not sure if this is something that we are always required to run.

tippmar-nr: Shouldn't be an issue - the [rules set that repolinter](https://raw.githubusercontent.com/newrelic/.github/main/repolinter-rulesets/community-plus.yml) is using only checks for existence and required content in license files, etc.

nrcventura (approved):

PR #1882, opened 2023-08-31
chore: Add code samples.

tippmar-nr:
Adds a set of sample apps to demonstrate using the .NET Agent. More will be added as time allows
* Console - Simple console app with a background service that does some work. Demonstrates attribute instrumentation.
* Docker (Alpine, Centos and Debian/Ubuntu) - shows the required incantations in `Dockerfile` to build and deploy an app with the .NET Agent for various Linux distros.

Also tweaks our markdown check configuration to ignore `localhost` and to fix a broken cron schedule from a change I made last week.


jaffinito (approved):
Very nice!

samples/docker/alpine/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN dotnet restore
+RUN dotnet publish -c Release -o /app/publish
+
+FROM base AS final
+
+# download and install the New Relic .NET Agent using the tarball
+# find the latest version at https://download.newrelic.com/dot_net_agent/latest_release and update the filename as needed
+# the agent is installed at /usr/local/new-relic-dotnet-agent
+RUN apk update && apk add --no-cache wget tar \
+    && wget https://download.newrelic.com/dot_net_agent/latest_release/newrelic-dotnet-agent_10.15.0_amd64.tar.gz \
```
chynesNR, line 19: I'm trying to think of a way we could keep this somewhat up to date... Part of the release process? Tech debt ticket?

tippmar-nr: Personally, I'd like to see some kind of symlink (or equivalent) on the download site that is an unchanging URL that always points to the latest version... ;-)

nr-ahemsath: As noted in team pairing, since the URL is for the `latest_release` folder but has a specific version named, this will break the next time we do a release.  I'll work on a more automagical way of doing things.

tippmar-nr: Resolved with a little `wget` magic.

samples/console/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+                    services.AddHostedService<Worker>();
+                })
+            .Build();
+
+        var logger = host.Services.GetService(typeof(ILogger<Program>)) as ILogger<Program>;
+
+        // run the hosted service until the application is terminated
+        logger!.LogInformation("Hello world!");
+        await host.RunAsync();
+        logger!.LogInformation("Goodby world!");
```
nr-ahemsath, line 24:
```suggestion
        logger!.LogInformation("Goodbye world!");
```

nr-ahemsath (approved):
Looks good; all the samples run as documented and send data to New Relic (if configured correctly with a valid license key). 💯

PR #1884, opened 2023-09-01
feat: Add more detail to "heartbeat" message in log

chynesNR:
Right now there is very little information logged by the Agent at Info level, by design. One of the things that does get logged is a simple `Metric harvest starting.`, which can be used to show that the Agent is still alive and doing work. The goal with this change is to record something a little more useful. Instead, every (5) minutes, the Agent will log how many events it has seen in that window.

Before:
```
2023-06-12 07:44:14,961 NewRelic   INFO: [pid: 5832, tid: 10] Metric harvest starting.
2023-06-12 07:45:15,501 NewRelic   INFO: [pid: 5832, tid: 17] Metric harvest starting.
2023-06-12 07:46:15,730 NewRelic   INFO: [pid: 5832, tid: 29] Metric harvest starting.
2023-06-12 07:47:15,947 NewRelic   INFO: [pid: 5832, tid: 11] Metric harvest starting.
2023-06-12 07:48:16,182 NewRelic   INFO: [pid: 5832, tid: 11] Metric harvest starting.
2023-06-12 07:49:16,763 NewRelic   INFO: [pid: 5832, tid: 25] Metric harvest starting.
```

After:
```
2023-08-31 23:23:11,003 NewRelic   INFO: [pid: 50188, tid: 47] In the last 5 minutes: 112 Transaction events, 1 Log event, 44 Span events
2023-08-31 23:28:11,003 NewRelic   INFO: [pid: 25376, tid: 40] In the last 5 minutes: No events
2023-08-31 23:33:11,021 NewRelic   INFO: [pid: 50188, tid: 29] In the last 5 minutes: 35 Transaction events, 15 Span events
```
This has a few troubleshooting benefits, even if logging is only set to Info level:
* Can tell quickly that the Agent is running and collecting data
* Gives you a rough idea of how busy the Agent is
* Can see what kinds of data collection are enabled

Some other approaches I considered but decided against:
* Logging our Supportability metrics (too noisy)
* Metadata about the agent (is it connected, how long has it been running, etc)

But I am open to other suggestions for things to report. When I started I hadn't realized we already had a periodic logging mechanism in place, so it's very easy to add new data.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
@@ -22,7 +22,7 @@ namespace NewRelic.Agent.Core.AgentHealth
 {
     public class AgentHealthReporter : ConfigurationBasedService, IAgentHealthReporter
     {
-        private static readonly TimeSpan _timeBetweenExecutions = TimeSpan.FromMinutes(5);
+        private static readonly TimeSpan _timeBetweenExecutions = TimeSpan.FromMinutes(2);
 
         private readonly IMetricBuilder _metricBuilder;
         private readonly IScheduler _scheduler;
--- a/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
@@ -22,7 +22,7 @@ namespace NewRelic.Agent.Core.AgentHealth
 {
     public class AgentHealthReporter : ConfigurationBasedService, IAgentHealthReporter
     {
-        private static readonly TimeSpan _timeBetweenExecutions = TimeSpan.FromMinutes(1);
+        private static readonly TimeSpan _timeBetweenExecutions = TimeSpan.FromMinutes(5);
 
         private readonly IMetricBuilder _metricBuilder;
         private readonly IScheduler _scheduler;
@@ -42,7 +42,7 @@ namespace NewRelic.Agent.Core.AgentHealth
         {
             _metricBuilder = metricBuilder;
             _scheduler = scheduler;
-            _scheduler.ExecuteEvery(LogRecurringLogs, _timeBetweenExecutions);
+            _scheduler.ExecuteEvery(LogPeriodicReport, _timeBetweenExecutions);
             var agentHealthEvents = Enum.GetValues(typeof(AgentHealthEvent)) as AgentHealthEvent[];
             foreach (var agentHealthEvent in agentHealthEvents)
             {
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Agent/Core/Aggregators/MetricAggregator.cs
```diff
@@ -63,7 +63,7 @@ public override void Collect(IAllMetricStatsCollection metric)
         }
         protected override void Harvest()
         {
-            Log.Info("Metric harvest starting.");
+            Log.Debug("Metric harvest starting.");
```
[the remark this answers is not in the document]
chynesNR: Good point! Fixed.

nr-ahemsath:
Regarding "Nothing is reported if no events were sent" - I would think that for this to be a "heartbeat" message it should report something every interval regardless if there is no instrumentation data being sent to NR.  Otherwise it's unclear whether the absence of a heartbeat message means "nothing was sent to NR this interval" or "something else is broken".

nrcventura:
Is 5 minutes an appropriate cadence? 

At startup we can see that the managed agent is doing something, so we probably don't need it faster to know if the agent has started. Maybe it's important to know when the agent has its first harvest? 5 minutes may be a relatively long time to wait for that. Even in that scenario, 5 minutes might be good enough.

If you're looking at an already running agent and wanting to know if the agent is still running, 5 minutes might be a relatively long time to wait, but it's probably ok too.

chynesNR:
> Is 5 minutes an appropriate cadence?
> 
> At startup we can see that the managed agent is doing something, so we probably don't need it faster to know if the agent has started. Maybe it's important to know when the agent has its first harvest? 5 minutes may be a relatively long time to wait for that. Even in that scenario, 5 minutes might be good enough.
> 
> If you're looking at an already running agent and wanting to know if the agent is still running, 5 minutes might be a relatively long time to wait, but it's probably ok too.

I think if we want an indicator for the first harvest we might want to special-case that rather than rely on this timer. Maybe a 3 minute interval would be better? I'm trying to strike a balance between useful and not clogging up the log too much. Some options on the time windows:
5 minutes => 288 lines/day
3 minutes => 480 lines/day
2 minutes => 720 lines/day
1 minute => 1440 lines/day

The old line is every minute, which I think is too much, but that may be personal preference. :)

chynesNR:
> > Is 5 minutes an appropriate cadence?
> > At startup we can see that the managed agent is doing something, so we probably don't need it faster to know if the agent has started. Maybe it's important to know when the agent has its first harvest? 5 minutes may be a relatively long time to wait for that. Even in that scenario, 5 minutes might be good enough.
> > If you're looking at an already running agent and wanting to know if the agent is still running, 5 minutes might be a relatively long time to wait, but it's probably ok too.
> 
> I think if we want an indicator for the first harvest we might want to special-case that rather than rely on this timer. Maybe a 3 minute interval would be better? I'm trying to strike a balance between useful and not clogging up the log too much. Some options on the time windows: 5 minutes => 288 lines/day 3 minutes => 480 lines/day 2 minutes => 720 lines/day 1 minute => 1440 lines/day
> 
> The old line is every minute, which I think is too much, but that may be personal preference. :)

Let's try every 2 minutes to start.

nrcventura (commented):
There's enough logic involved in producing these log messages that it would be great if we could get some unit test coverage for it.

src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -58,25 +58,28 @@ public AgentHealthReporter(IMetricBuilder metricBuilder, IScheduler scheduler)
...
+            
+            List<string> events = new List<string>();
             foreach (var counter in _agentHealthEventCounters)
             {
                 if (counter.Value != null && counter.Value.Value > 0)
                 {
                     var agentHealthEvent = counter.Key;
                     var timesOccured = counter.Value.Exchange(0);
-                    Log.Info($"Event {agentHealthEvent} has occurred {timesOccured} times in the last {_timeBetweenExecutions.TotalSeconds} seconds");
+                    events.Add(string.Format("{0} {1} {2}", timesOccured, agentHealthEvent, (timesOccured == 1) ? "event" : "events"));
```
tippmar-nr, line 78: Totally a nitpick, but `string.Format()` went out of vogue years ago... ;-)

chynesNR: Are you talking about string interpolation or is there some other formatter I'm not thinking of? I don't like using interpolation with conditionals, I think it's harder to read. But maybe that's just me. :)

tippmar-nr: Yeah, string interpolation is where I was headed. Pretty easy to write this one as 
```
events.Add($"{timesOccured} {agentHealthEvent}, {(timesOccured == 1 ? "event" : "events")}
```
The trick being to wrap the ternary in parentheses.

Oh - one more nit: `timesOccured` should be `timesOccurred` ...

jaffinito: If all the things your are putting into the string are strings, the performance is close to String.Format, but if they are not strings, older versions of .NET have issues with the new interpolation style.  Calling ToString on the non-strings resolves it however.

nrcventura (approved):

Issue #1857, opened 2023-08-22
Incorrect transaction name change reported in logs when SetTransactionName is called multiple times


nrcventura:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

**Description**

FINEST level logs generated by `CandidateTransactionName.TrySet` are not accurate if the `SetTransactionName` API method is called multiple times in a single transaction. Currently the agent logs indicate that the transaction name specified in the second call to the `SetTransactionName` API is ignored, but the code does not ignore that name and instead keeps it. The log message is incorrect.

**Expected Behavior**

If the current transaction name, and the new transaction name both have a priority of `User`, the log message should indicate that the transaction name is changed to the new transaction name. Doing this will make the log message consistent with the actual logic.

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/diagnostics-cli-nrdiag/diagnostics-cli-nrdiag/) results**

**Steps to Reproduce**

1. Create a transaction
2. Call the `SetTransactionName` API with a value such as `NameFromFirstCall`
3. Call the `SetTransactionName` API with a new value such as `NameFromSecondCall`
4. Observe that the logs indicate that the transaction name `NameFromSecondCall` is ignored, even though the transaction data shows that the transaction name is `NameFromSecondCall` is used for the transaction event and the scoped metrics.

**For Maintainers Only or Hero Triaging this bug**
*Suggested Priority (P1,P2,P3,P4,P5):* P4
*Suggested T-Shirt size (S, M, L, XL, Unknown):* S

PR #1886, opened 2023-09-05
fix: Fix misleading log message on transaction name change. (#1857)

chynesNR:
Calling `SetTransactionName()` more than once would change the name, but the log would incorrectly report that it hadn't changed. I would have liked to combine the setting and logging logic, but the current structure seems deliberate for performance reasons.


tippmar-nr (approved):
Looks good!

jaffinito (approved):

PR #1888, opened 2023-09-06
Fix: Prevent broken traces when HttpClient content headers contain tracing headers. (#1843)

### LINKED ISSUE #1843: HttpClient instrumentation does not handle case when tracing headers are content headers
The HttpClient instrumentation currently handles the case when the distributed tracing and CAT headers are defined in the request headers but it does not handle the case when the tracing headers are already defined as content headers.

**Description**

The HttpClient instrumentation currently handles the case when the distributed tracing and CAT headers are defined in the request headers. It handles this by first removing the existing headers, and then adding the expected headers for the request. However, if these tracing headers are present in the content headers collection, the agent does not detect them and it results in 2 tracing headers being present in the external request. This results in broken traces.

**Expected Behavior**

The agent should replace any tracing headers defined on both the request and the content, and the appropriate headers should be added to the request (not the content).

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/diagnostics-cli-nrdiag/diagnostics-cli-nrdiag/) results**

**Steps to Reproduce**

1. Add a tracing header to an outbound HttpClient request using code like the following.
`request.Content.Headers.TryAddWithoutValidation("newrelic", "newrelic2");`
2. Perform the outbound call in an instrumented app.
3. Observe in the downstream request that there are multple `newrelic` header values. One added by the agent and the other added using the above code.

Note that `request.Headers.Add("newrelic", $"newrelic1");` does not result in multiple header values being observed in the downstream request.

**Your Environment**

**Additional context**

*Suggested Priority (P1,P2,P3,P4,P5):* P5
*Suggested T-Shirt size (S, M, L, XL, Unknown):* M (most of the effort is probably in the integration tests for the different tracing types -- W3C, New Relic DT, CAT)

nrcventura:

## Description

Fixes #1843.

Prevents duplicate outgoing tracing headers, for HttpClient calls,  by ensuring that the headers are only present as request headers, and that they do not also appear as content headers. An existing Integration Test was updated to ensure that the distributed trace is not broken when tracing headers are part of the Content Headers collection.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/HttpClient/SendAsync.cs
@@ -135,6 +135,9 @@ namespace NewRelic.Providers.Wrapper.HttpClient
         {
             var setHeaders = new Action<HttpRequestMessage, string, string>((carrier, key, value) =>
             {
+                // Content headers should not contain the expected header keys, but sometimes they do,
+                // and their presence can cause problems downstream by having 2 values for the same key.
+                carrier.Content?.Headers?.Remove(key);
                 // "Add" will throw if value exists, so we must remove it first
                 carrier.Headers?.Remove(key);
                 carrier.Headers?.Add(key, value);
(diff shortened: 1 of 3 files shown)


tippmar-nr (approved):
Nicely done!

PR #1891, opened 2023-09-07
chore: add new methods in VisbilityBypasser to simplify "this or that" usage pattern.

### LINKED ISSUE #1797: Refactor VisibilityBypasser
We have a few places in the code where we either want to find property/method A *or* property/method B, depending on which one is available. Right now the only way to do that is with a try/catch on the VisibilityBypasser call. It would be cleaner if we could either have methods to test if it exists first, or have alternate versions that return null/a status code if the field isn't available. Or maybe even a function that says "give me A or B, whichever exists".

### Acceptance Criteria
New functionality and integration or unit tests.

### Estimates
S/M

tippmar-nr:
Adds a couple of `TryXxx()` methods in `VisibilityBypasser` that simplify calling patterns where we attempt to get a particular property or method, then fall back to getting a different property or method. 

Resolves #1797.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Log4NetLogging/Log4netWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Log4NetLogging/Log4netWrapper.cs
@@ -59,32 +59,18 @@ namespace NewRelic.Providers.Wrapper.Logging
             // Older versions of log4net only allow access to a timestamp in local time
             var getTimestampFunc = _getTimestamp ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<DateTime>(logEventType, "TimeStamp");
 
-            Func<object, Exception> getLogExceptionFunc;
-
-            try
+            if (_getLogException == null
+                && !VisibilityBypasser.Instance.TryGeneratePropertyAccessor<Exception>(logEventType, "ExceptionObject", out _getLogException)
+                // Legacy property, mainly used by Sitecore
+                && !VisibilityBypasser.Instance.TryGeneratePropertyAccessor<Exception>(logEventType, "m_thrownException", out _getLogException))
             {
-                getLogExceptionFunc = _getLogException ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<Exception>(logEventType, "ExceptionObject");
-            }
-            catch
-            {
-                try
-                {
-                    // Legacy property, mainly used by Sitecore
-                    getLogExceptionFunc = _getLogException ??= VisibilityBypasser.Instance.GenerateFieldReadAccessor<Exception>(logEventType, "m_thrownException");
-                    _legacyVersion = true;
-                }
-                catch
-                {
-                    _getLogException = (x) => null;
-                    getLogExceptionFunc = _getLogException;
-                }
-
+                _getLogException = (x) => null;
             }
 
             // This will either add the log message to the transaction or directly to the aggregator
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Log4NetLogging/Log4netWrapper.cs
```diff
@@ -59,32 +59,18 @@ private void RecordLogMessage(object logEvent, Type logEventType, IAgent agent)
...
-
-            try
-            {
-                getLogExceptionFunc = _getLogException ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<Exception>(logEventType, "ExceptionObject");
-            }
-            catch
+            if (_getLogException == null
+                && !VisibilityBypasser.Instance.TryGeneratePropertyAccessor<Exception>(logEventType, "ExceptionObject", out _getLogException)
+                // Legacy property, mainly used by Sitecore
+                && !VisibilityBypasser.Instance.TryGeneratePropertyAccessor<Exception>(logEventType, "m_thrownException", out _getLogException))
```
nrcventura, line 65: Even though this is more compact and it reduces nesting, I find this harder to read. It is easier for me to understand if there is only 1 method with an out parameter being called in the conditional.

chynesNR:
What I was imagining when I wrote the ticket was to avoid the overhead of a caught exception at all, since we're the ones throwing it, if we can't find the given method. Here we're still relying on `GetMethodInfo` to throw when it can't find the first one in our "this or that" scenario. That's the refactoring I was talking about, but if it ends up being a lot of rework it's probably not worth it.

tippmar-nr:
@chynesNR Yeah, I see what you're saying - but it gets pretty complicated to unravel the "throw an exception if we can't find the thing" logic, as that's how every method in VisibilityBypasser was originally implemented. So my refactor at least helps isolate the ugliness from the outside world, I guess... ;-)

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Log4NetLogging/Log4netWrapper.cs
tippmar-nr: Would you prefer something like this? 
```
            if (_getLogException == null)
            {
                if (!VisibilityBypasser.Instance.TryGeneratePropertyAccessor<Exception>(logEventType, "ExceptionObject", out _getLogException))
                {
                    // Legacy property, mainly used by Sitecore
                    if (!VisibilityBypasser.Instance.TryGeneratePropertyAccessor<Exception>(logEventType, "m_thrownException", out _getLogException))
                    {
                        _getLogException = (x) => null;
                    }
                }
            }
```

nrcventura: I think that makes it clearer.

nrcventura:
> @chynesNR Yeah, I see what you're saying - but it gets pretty complicated to unravel the "throw an exception if we can't find the thing" logic, as that's how every method in VisibilityBypasser was originally implemented. So my refactor at least helps isolate the ugliness from the outside world, I guess... ;-)

I think that the refactor could be done in multiple steps. This first PR cleans up the usage. The follow up PR could update the code to not rely on the exceptions.

jaffinito (approved):
I don't care for using exceptions as logic, but I am not seeing a way not to here.  Looks good!

PR #1903, opened 2023-09-11
feat: Use structured logging for internal logs

### LINKED ISSUE #1667: Refactor string formatting in log messages
Serilog provides full support for structured logging, but our existing usage of methods like `Log.Debug()`, etc. generally either performs an in-line string formatting or sometimes calls `Log.DebugFormat()` etc. to format the string. In either case, there are potential performance implications in allocating and formatting a string, only to have the string dropped later because the log level was below the current "minimum" log level.

We need to refactor the method signatures in our `Log` class to more closely match the underlying Serilog methods that we call, and then update our usage of those methods to use structured logging "best practices" in our message templates. This should considerably reduce usages of string formatting in log message calls.

Practices to consider:
* Use message templates instead of messages:
`Log.Info("The time is " + DateTime.Now);` would be better as `Log.Info("The time is {Now}", DateTime.Now);`
* Log exceptions without modification: 
Instead of 
```            catch (Exception e)
            {
                var msg = new StringBuilder(e.Message);
                if (e.InnerException != null)
                {
                    msg.Append("; ");
                    msg.Append(e.InnerException.Message);
                }
                Log.ErrorFormat("Exception performing thread profiling data aggregation: {0}", msg);
            }
```
do this instead
```
catch (Exception e) 
{
    Log.Error(e, "Exception performing thread profiling data aggregation");
}
```

Serilog is smart enough to avoid the overhead of generating a log message when the log level is disabled, so we should only be using the log "guard" methods (like `Log.IsDebugEnabled`) in cases where the work required to gather the parameters for the log message is expensive enough to warrant checking the log level enabled status first.

### Acceptance Criteria
All Unit and Integration tests should pass and log output should be generally identical after the change.

### Estimates
M

chynesNR:
This gives us:
* Full exception info and stack trace for all errors
* No longer need separate "Format" versions of logging functions (e.g. `DebugFormat`, `FinestFormat`, etc) or separate exception logging functions

I also removed some redundant `if [Debug]Enabled Log.[Debug]` checks.

Two areas that could still use some cleanup:
* I didn't convert any string interpolation to use a structured log line, only explicit calls to the Format methods
* `Transaction.LogFinest()` could probably use a little restructuring to avoid all the checks if Finest is enabled

Resolves #1667


src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -236,7 +236,7 @@ public void ReportWrapperShutdown(IWrapper wrapper, Method method)
             }
 
             Log.Error($"Wrapper {wrapperName} is being disabled for {method.MethodName} due to too many consecutive exceptions. All other methods using this wrapper will continue to be instrumented. This will reduce the functionality of the agent until the agent is restarted.");
-            _recurringLogDatas.Add(new RecurringLogData(Log.Debug, $"Wrapper {wrapperName} was disabled for {method.MethodName} at {DateTime.Now} due to too many consecutive exceptions. All other methods using this wrapper will continue to be instrumented. This will reduce the functionality of the agent until the agent is restarted."));
+            _recurringLogDatas.Add(new RecurringLogData((s) => Log.Debug(s), $"Wrapper {wrapperName} was disabled for {method.MethodName} at {DateTime.Now} due to too many consecutive exceptions. All other methods using this wrapper will continue to be instrumented. This will reduce the functionality of the agent until the agent is restarted."));
```
nrcventura, line 239: Why does this need a lambda now? If we could continue to use `Log.Debug` there will be fewer allocations, we might just need to use a specific delegate type to disambiguate which overload should be used.

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs
```diff
@@ -30,10 +30,10 @@ public class ConnectionHandler : ConfigurationBasedService, IConnectionHandler
...
         private static readonly Dictionary<string, Action<string>> ServerLogLevelMap = new Dictionary<string, Action<string>>
         {
-            {"INFO", Log.Info},
-            {"WARN", Log.Warn},
-            {"ERROR", Log.Error},
-            {"VERBOSE", Log.Finest}
+            {"INFO", (s) => Log.Info(s)},
+            {"WARN", (s) => Log.Warn(s)},
+            {"ERROR", (s) => Log.Error(s)},
+            {"VERBOSE", (s) => Log.Finest(s)}
```
nrcventura, line 36: Is there another way to reference the correct overload so that we don't have to allocate a lambda?

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs
```diff
@@ -276,7 +276,7 @@ private void LogConfigurationMessages(ServerConfiguration serverConfiguration)
                 if (string.IsNullOrEmpty(message.Text))
                     continue;
 
-                var logMethod = ServerLogLevelMap.GetValueOrDefault(message.Level) ?? Log.Info;
+                var logMethod = ServerLogLevelMap.GetValueOrDefault(message.Level) ?? new Action<string>((s) => Log.Info(s));
```
nrcventura, line 279: Is there a way to avoid allocating an action instance and a lambda here?

src/Agent/NewRelic/Agent/Core/Samplers/GcSampler.cs
```diff
@@ -222,19 +222,7 @@ private int CreatePerfCounterProxies(string processInstanceName)
...
-
-            if (ex == null)
-            {
-                Log.LogMessage(level, $"GC Performance Counters: {message}");
-            }
-            else
-            {
-                Log.LogMessage(level, $"GC Performance Counters: {message}, {ex}");
-            }
+            Log.LogMessage(level, ex, $"GC Performance Counters: {message}");
```
nrcventura, line 225: The previous code only incurred the string format/interpolation cost if the log level was enabled, this version will always incur that cost.

src/Agent/NewRelic/Agent/Core/Wrapper/WrapperMap.cs
```diff
@@ -60,10 +60,7 @@ public WrapperMap(IEnumerable<IWrapper> wrappers, IDefaultWrapper defaultWrapper
                 Log.Error("No wrappers were loaded.  The agent will not behave as expected.");
             }
 
-            if (Log.IsFinestEnabled)
-            {
-                Log.Finest($"WrapperMap has NonDefaultWrappers: {string.Join(", ", _nonDefaultWrappers)}");
-            }
+            Log.Finest("WrapperMap has NonDefaultWrappers: {0}", string.Join(", ", _nonDefaultWrappers));
```
nrcventura, line 63: The previous code only incurs the cost of the string.Join call and allocations if finest logging is enabled, but this version will always incur that cost.

src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
chynesNR: There a few places where logging functions are passed as `Action`s, and in the moment I was thinking we couldn't use `params` with an `Action`. Which is mostly true, but can be worked around. Fixed.

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs
chynesNR: Done

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs
chynesNR: Done

src/Agent/NewRelic/Agent/Core/Samplers/GcSampler.cs
chynesNR: Refactored

src/Agent/NewRelic/Agent/Core/Wrapper/WrapperMap.cs
chynesNR: Reverted

src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -63,9 +63,9 @@ public override void Dispose()
 
         private void LogRecurringLogs()
         {
-            foreach (var data in _recurringLogDatas)
+            foreach (var logMessage in _recurringLogData)
             {
-                data?.LogAction(data.Message);
+                Log.Info(logMessage);
```
nrcventura, line 68: It seems like this log data was previously logged at the Debug level.

nrcventura: This also seems like an agent debugging feature that was introduced to make certain types of log messages repeat with every single harvest. I'm not sure that this is a feature that we need to continue to support.

chynesNR: Yeah I can see how "Hey this wrapper is busted" is something you might want to report periodically so it doesn't get missed, but hopefully that's not something that happens often any more. I'd also be ok with removing it.

nrcventura: I'm fine with either removing it, or updating the log level.

nrcventura (approved):

PR #1915, opened 2023-09-14
Log .NET Runtime Version at info level

neilpdx:

## Description

It would be useful information for Support to ensure the application's .NET Runtime Version being used is compatible with the agent version being used, since we have deprecated support for many .NET Framework and .NET Core versions. Having this information in existing logs without needing to elevate the log level would be ideal.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentManager.cs
@@ -282,7 +282,7 @@ namespace NewRelic.Agent.Core
                     }
                 }
 
-                Log.Debug($".NET Runtime Version: {RuntimeInformation.FrameworkDescription}");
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/AgentManager.cs
```diff
@@ -282,7 +282,7 @@ private void LogInitialized()
                     }
                 }
 
-                Log.Debug($".NET Runtime Version: {RuntimeInformation.FrameworkDescription}");
+                Log.Info($".NET Runtime Version: {RuntimeInformation.FrameworkDescription}");
```
nrcventura, line 285: I don't think that this log line will give you the information that you want for all of the versions of .net (especially .net framework). You might want to look at the changes in #1852 which call out when a framework is unsupported and is already at a higher log level.

PR #1916, opened 2023-09-14
Chore: Resolve warnings for all tests

### LINKED ISSUE #1883: Restructure Integration Test dependencies

The current set of integration test dependencies is an interdependent tangled collection of dependencies, which has led to unexpected projects being used in both test applications and the test runners. We need to enforce a stricter separation of direct and transient dependencies between the test applications and the test runners. The integration tests that we write for the agent are typically designed to work against a specific set of library versions being used in a test app. We do not want to accidentally change those versions by updating a dependency in a test runner.

For example, with the current dependency structure if we update the IntegrationTestHelpers project to include a new dependency, that dependency can bring other dependencies with, such as grpc and the Microsoft.Extensions.* libraries. This change has a cascading effect on the MultiConsoleApplication project, which is primarily intended to be used within test applications (and not the test runner), that will now cause that project to use a different version of the grpc and Microsoft.Extensions.* libraries, which in turn leads to our test apps using unexpected library versions.

We need to restructure our test projects and their dependencies so that the can enforce a strict separation of test runner code and test application code. The one exception to this is the Secrets project, but we might need to update that project so that it only contains code related to Secret management.

As part of this the RandomPortGenerator should not be used from directly within test apps. Any dependent ports need to be passed into the application. The current usage from within the test apps can lead to unexpected port conflicts.

### Acceptance Criteria

1. Integration Test runner projects are separated from Test application projects
2. Code is not directly or transitively shared between the Test runner projects and the Test application projects
3. Secrets can be used by both the Test runner and Test application projects.
4. The RandomPortGenerator should not be used by the SerilogWebAdapter from within a test application
5. The Integration Tests, Unbounded Integration Tests, and the Container Tests meet the above acceptance criteria

### Dependencies

N/A

N/A

### Additional context

The current solution also makes it difficult to update dependencies in our integration tests, because of the unexpected cascading effects.

### Estimates

t-shirt size: M

nrcventura:

## Description

Resolves #1883.

This PR includes fixes for the remaining warnings for all of the tests. This includes resolving all of the remaining dependency mismatch warnings.

To help prevent unexpected changes in dependencies, this PR also includes enabling the setting to treat warnings as errors.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/tests/Agent/IntegrationTests/Applications/ConsoleAsyncApplication/AsyncFireAndForgetUseCases.cs
+++ b/tests/Agent/IntegrationTests/Applications/ConsoleAsyncApplication/AsyncFireAndForgetUseCases.cs
@@ -108,7 +108,7 @@ namespace ConsoleAsyncApplication
         {
             var transactionName = UpdateTransactionName("SF");
 
-            Task.Run(() => AsyncMethod(DELAY_TIME, transactionName, waitHandle));
+            _ = Task.Run(() => AsyncMethod(DELAY_TIME, transactionName, waitHandle));
 
             Thread.Sleep(SHORT_DELAY_TIME);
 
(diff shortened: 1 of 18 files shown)


jaffinito (approved):

chynesNR (approved):
Looks good!

PR #1934, opened 2023-09-26
ci: Misc workflow refactoring

tippmar-nr:
* Removes the (unmaintained) `southpolesteve/cosmos-emulator-github-action` action from `all_solutions.yml` and replaces it with direct calls to Powershell to start the CosmosDB emulator. Tested and verified that the emulator starts as expected.
* Refactors `deploy_agent.yml` to make all relevant input options true booleans rather than strings, which makes the UI for manual invocation of the workflow quite a bit cleaner: 

image

Note that none of the comparisons in the workflow were modified - they *should* all continue to work as-is. I haven't been able to test, of course, because it's not trivial to do so in this workflow.
* Changes the defaults for "Deploy Build Artifacts" and "Deploy APT/YUM Packages to Production Repository" to `true` instead of `false`. This is the most common use case for this workflow, so it seems reasonable to use these defaults.

### DIFF
--- a/.github/workflows/deploy_agent.yml
+++ b/.github/workflows/deploy_agent.yml
@@ -11,37 +11,37 @@ on:
         required: true
       deploy:
         type: boolean
-        description: 'If "true", deploy the artifacts. If "false", do everything except deploy.'
+        description: Deploy Build Artifacts. (If not selected, do everything except deploy.)'
         required: true
         default: true
       downloadsite:
         type: boolean
-        description: 'If "true", will run the deploy-downloadsite job. If "false", will not.'
+        description: 'Deploy MSI / ZIP / Tarball to the Download Site'
         required: true
         default: true
       nuget:
         type: boolean
-        description: 'If "true", will run the deploy-nuget job. If "false", will not.'
+        description: 'Deploy Nuget Packages'
         required: true
         default: true
       linux:
         type: boolean
-        description: 'If "true", will run the deploy-linux job. If "false", will not.'
+        description: 'Build APT / YUM Packages'
         required: true
         default: true
       linux-deploy-to-production:
         type: boolean
-        description: 'If "true", will deploy Linux packages to the real apt/yum.newrelic.com. If "false", will deploy to a test repository.'
+        description: 'Deploy APT / YUM Packages to Production Repository. (If not selected, will deploy to a test respository.)'
         required: true
         default: true
       indexdownloadsite:
         type: boolean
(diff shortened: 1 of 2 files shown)


chynesNR (approved):

.github/workflows/deploy_agent.yml
```diff
@@ -10,33 +10,40 @@ on:
...
         required: true
-        default: 'true'
+        default: true
       linux-deploy-to-production:
-        description: 'If "true", will deploy Linux packages to the real apt/yum.newrelic.com. If "false", will deploy to a test repository.'
+        type: boolean
+        description: Deploy APT / YUM Packages to Production Repository. (If not selected, will deploy to a test respository.)
         required: true
-        default: 'false'
+        default: true
```
nrcventura, line 36: This default used to be `false`, is there a reason we are changing it to `true`?

.github/workflows/deploy_agent.yml
```diff
@@ -10,33 +10,40 @@ on:
         description: 'Run ID of the Release Workflow (all_solutions.yml) that was triggered by creating a Release in GitHub.  ID can be found in URL for run.'
         required: true
       deploy:
-        description: 'If "true", deploy the artifacts. If "false", do everything except deploy.'
+        type: boolean
+        description: Deploy Build Artifacts. (If not selected, do everything except deploy.)
         required: true
-        default: 'false'
+        default: true
```
nrcventura, line 16: This used to default to `false`, is there a reason we are changing it to `true`?

tippmar-nr: Because we always end up setting it to `true` when we run the workflow during a release. Might as well default everything to the value it's normally going to be anyway.

.github/workflows/deploy_agent.yml
tippmar-nr: Same reasoning as for the other setting that was twiddled.

.github/workflows/deploy_agent.yml
chynesNR: We never run Deploy automatically, right? That would be my only concern with changing the defaults.

tippmar-nr: No, it is only ever invoked manually. Note that the only trigger in the workflow is `workflow_dispatch`

nrcventura: I'm fine with changing the defaults if we think it is safe, but I would like the change called out in the PR description so that we know it was done intentionally.

PR #1935, opened 2023-09-27
test: EventWaitHandle error handling attempt

tippmar-nr:
Attempts to work around an intermittent test failure when creating an `EventWaitHandle` by retrying with `EventWaitHandle.OpenExisting()`. No idea if it will help or not, but I don't think it can hurt anything. 

The error we see in CI runs looks like this:
```
 [9/27/2023 9:51:19 AM 1208-1] HostedWebCore: HostedWebCore.exe failed: Reason unknown.: System.IO.FileLoadException: The process cannot access the file because it is being used by another process. (Exception from HRESULT: 0x80070020)
    at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
    at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode)
    at HostedWebCore.HostedWebCore.Run() in D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\tests\Agent\IntegrationTests\HostedWebCore\HostedWebCore.cs:line 62
    at HostedWebCore.Program.Main(String[] args) in D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\tests\Agent\IntegrationTests\HostedWebCore\Program.cs:line 51
```

### DIFF
--- a/tests/Agent/IntegrationTests/HostedWebCore/HostedWebCore.cs
+++ b/tests/Agent/IntegrationTests/HostedWebCore/HostedWebCore.cs
@@ -63,12 +63,14 @@ namespace HostedWebCore
             EventWaitHandle eventWaitHandle;
             try
             {
+                Program.Log($"Creating EventWaitHandle named {eventWaitHandleName}");
                 eventWaitHandle = new EventWaitHandle(false, EventResetMode.ManualReset, eventWaitHandleName);
             }
             catch (Exception ex)
             {
-                Program.Log($"EventWaitHandle ctor failed with exception {ex}. Re-trying with OpenExisting().");
+                Program.Log($"EventWaitHandle creation failed with exception: {ex}. Re-trying with EventWaitHandle.OpenExisting().");
                 eventWaitHandle = EventWaitHandle.OpenExisting(eventWaitHandleName);
+                Program.Log($"EventWaitHandle.OpenExisting() succeeded!");
             }
 
             using (eventWaitHandle)
--- a/tests/Agent/IntegrationTests/HostedWebCore/HostedWebCore.cs
+++ b/tests/Agent/IntegrationTests/HostedWebCore/HostedWebCore.cs
@@ -58,9 +58,20 @@ namespace HostedWebCore
         public void Run()
         {
             StartWebServer();
-            //The HWC creates this shutdown event and waits for the test runner to set so that it can shutdown.  
-            using (var eventWaitHandle = new EventWaitHandle(false, EventResetMode.ManualReset,
-                       "app_server_wait_for_all_request_done_" + _port.ToString()))
+            //The HWC creates this shutdown event and waits for the test runner to set so that it can shutdown.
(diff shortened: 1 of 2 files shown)


tests/Agent/IntegrationTests/HostedWebCore/HostedWebCore.cs
```diff
@@ -58,9 +58,22 @@ public HostedWebCore(string port)
...
+            EventWaitHandle eventWaitHandle;
+            try
+            {
+                Program.Log($"Creating EventWaitHandle named {eventWaitHandleName}");
+                eventWaitHandle = new EventWaitHandle(false, EventResetMode.ManualReset, eventWaitHandleName);
+            }
+            catch (Exception ex)
+            {
+                Program.Log($"EventWaitHandle creation failed with exception: {ex}. Re-trying with EventWaitHandle.OpenExisting().");
+                eventWaitHandle = EventWaitHandle.OpenExisting(eventWaitHandleName);
```
nrcventura, line 72: Should we wait a bit before attempting to reopen the wait handle?

tippmar-nr: I'm not sure there's any reason to wait after the exception gets thrown - presumably, it's thrown because the handle (somehow) already exists and can't be created - so attempting to open an existing handle should succeed immediately.

I do wonder, however, if it might be prudent to put a small delay before the first attempt to create the wait handle? Not entirely sure what happens when `StartWebServer()` is called, but maybe a short delay after calling that might be helpful?

nrcventura: I missed that difference the first time I looked. The delay might not help. I'm wondering if the problem is related to a retry because HostedWebCore is shutting down with a forced Exit so the event wait handle is not closed properly?

tippmar-nr: I suppose that's possible - in which case, the `OpenExisting()` call should succeed. 

Also makes me wonder whether, if we're doing a retry, we should use the same port number or get a new one? (Not sure what we do currently.)

nrcventura: It looks like we reuse the port during a retry, so this approach may actually work around the problem for HostedWebCore apps. It looks like we use `Environment.Exit` in multiple test apps, which may also be experiencing the same problem.

I think we can either go with this approach or stop using `Environment.Exit` in our test apps. Since the test app can also be forcefully terminated via test runner it might be better to leverage the approach already being used in this PR. You might be able to add a helper method in the SharedApplicationHelpers project that can be reused in multiple test apps to make this easier (assuming this works).

nrcventura (approved):

tippmar-nr:
After further investigation, it turns out the error was actually getting thrown from `StartWebServer()` - a separate PR will be created to address that issue.

PR #1943, opened 2023-10-02
fix: Detect Docker container id on cgroup v2.

tippmar-nr:
Updates our Docker container id lookup to support the newer cgroup v2 standard, parsing `/proc/self/mountinfo` to obtain the container id. 

The new lookup is executed first and if not successful, we fall back to the original (cgroup v1) lookup.

Adds a unit test to verify the parsing logic for `/proc/self/mountinfo`.

Tested in staging and saw the container id being reported as expected:
image

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
+++ b/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
@@ -280,21 +280,18 @@ namespace NewRelic.Agent.Core.Utilization
 
             if (isLinux)
             {
-                var vendorModel = TryGetDockerCGroupV1(subsystemsIndex, controlGroupIndex);
-                if (vendorModel == null)
+                try
                 {
-                    try
-                    {
-                        var fileContent = File.ReadAllText("/proc/self/mountinfo");
-                        vendorModel = TryGetDockerCGroupV2(fileContent);
-                    }
-                    catch
-                    {
-                        return null;
-                    }
-                }
+                    var fileContent = File.ReadAllText("/proc/self/mountinfo");
+                    var vendorModel = TryGetDockerCGroupV2(fileContent)
+                                      ?? TryGetDockerCGroupV1(subsystemsIndex, controlGroupIndex);
 
-                return vendorModel;
+                    return vendorModel;
+                }
+                catch
+                {
+                    return null;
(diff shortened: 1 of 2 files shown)


jaffinito (approved):

src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
```diff
@@ -273,50 +274,83 @@ private string GetProcessEnvironmentVariable(string variableName)
...
+                {
+                    var fileContent = File.ReadAllText("/proc/self/mountinfo");
+                    var vendorModel = TryGetDockerCGroupV2(fileContent)
+                                      ?? TryGetDockerCGroupV1();
+
+                    return vendorModel;
+                }
+                catch
+                {
+                    return null;
```
chynesNR, line 291: Maybe we could add a Finest-level log line here? I imagine this would be tough to troubleshoot otherwise if something goes wrong

chynesNR: Oh wait, does this always execute? Maybe just log if we find one of the files but fail to parse them?

tippmar-nr: Actually, I need to rework the logic a bit - it won't fall back to calling the V1 method if `/proc/self/mountinfo` doesn't exist.

tippmar-nr: @chynesNR How does my latest change look?

nrcventura (commented):
Is now a good time to refactor the docker detection code so that we can have a test that ensures that we fallback to v1 detection if v2 detection fails?

chynesNR:
> Is now a good time to refactor the docker detection code so that we can have a test that ensures that we fallback to v1 detection if v2 detection fails?

I'll leave it up to @tippmar-nr -- if he doesn't think now is a good time we can add a ticket to the backlog.

tippmar-nr:
@nrcventura @chynesNR I'll see about refactoring and adding some more tests. Shouldn't be too hard.

nrcventura (approved):

chynesNR (approved):
Very nice!

PR #1945, opened 2023-10-03
ci: Adjust metric harvest cycle and introduce a delay for linux smoke tests

tippmar-nr:
Changes the metric harvest override to 10 seconds instead of 5 and introduces a delay after exercising the app but before looking for metric harvest. These changes should help resolve flakiness in the linux smoke tests.


chynesNR (commented):
Is our goal to test the harvest cycles, or would it be better to use `SendDataOnExit?`

nrcventura:
> Is our goal to test the harvest cycles, or would it be better to use `SendDataOnExit?`

The test is already using the `SendDataOnExit` setting. In one of the recent failures, the shutdown process was able to send all of the data except for the metrics, which caused the test fail.

nrcventura (approved):

PR #1954, opened 2023-10-05
ci: Update markdown check workflow to only run on schedule

tippmar-nr:
Modifies `markdowncheck.yml` to run only on a daily schedule. 

I'll subscribe the `dotnet-agent-github` channel to get notifications about the scheduled run.


nr-ahemsath (approved):
👍

PR #1968, opened 2023-10-10
Feature work/kafka consumer wrapper

tippmar-nr:


chynesNR (approved):

src/Agent/NewRelic/Agent/Core/Transactions/TransactionName.cs
```diff
@@ -71,6 +71,8 @@ public static TransactionName ForBrokerTransaction(MessageBrokerDestinationType
             var trxName = new StringBuilder(vendor)
                 .Append(MetricNames.PathSeparator)
                 .Append(EnumNameCache<MessageBrokerDestinationType>.GetName(type))
+                .Append(MetricNames.PathSeparator)
+                .Append(MetricNames.MessageBrokerConsume)
```
nrcventura, line 75: Will this affect NServiceBus and RabbitMQ instrumentation?

tippmar-nr: I haven't any idea; @jaffinito your thoughts please?

tippmar-nr: @nrcventura @jaffinito Looking at the unbounded integration tests for RabbitMQ and NServiceBus, it appears this change will affect instrumentation of the consume side - quite a few tests fail since they're not aware of the change. 

I'm not sure where to go from here - the [Kafka spec](https://source.datanerd.us/agents/agent-specs/blob/main/Kafka.md#transaction-name) requires the transaction name to include `/Consume` - I don't know what spec(s) governed the implementation of transaction names for RabbitMQ and NServiceBus.

nrcventura: You might just need to create a different overload of SetMessageBrokerTransactionName to leave the other instrumentation working as is.

tippmar-nr: I reverted this change and created separate overloads for doing things in a Kafka-specific way.

chynesNR: Good catch! I was not reading closely and missed that.

PR #1969, opened 2023-10-11
feat: Enable startup logging to Event Log for all applications on Windows.

tippmar-nr:
## Description

Enables configuration of the Event Log sink for Serilog during Agent startup when running on Windows in both .NET Framework and .NET 6+ applications. 

Note that the Event Log sink will only write to the event log if the `New Relic .NET Agent` Event Source already exists *or* if the application is running with administrator privileges. 

This change supersedes a previous commit (#4310e982) that was reverted.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
+++ b/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
@@ -11,8 +11,9 @@ using Serilog.Formatting;
 using Logger = NewRelic.Agent.Core.Logging.Logger;
 using NewRelic.Agent.Core.Logging;
 using Serilog.Templates;
-#if NETFRAMEWORK
 using Serilog.Events;
+#if NETSTANDARD2_0
+using System.Runtime.InteropServices;
 #endif
 
 namespace NewRelic.Agent.Core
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
```diff
@@ -107,33 +108,46 @@ private static LoggerConfiguration ConfigureInMemoryLogSink(this LoggerConfigura
...
+        /// 
         /// </summary>
         /// <param name="loggerConfiguration"></param>
         private static LoggerConfiguration ConfigureEventLogSink(this LoggerConfiguration loggerConfiguration)
         {
-#if NETFRAMEWORK
-            const string eventLogName = "Application";
-            const string eventLogSourceName = "New Relic .NET Agent";
-            try
+#if NETSTANDARD2_0
```
nrcventura, line 124: You can reduce the amount of compiler directives if you did something like the following instead.
```
#if NETSTANDARD2_0
    var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
#else
    var isWindows = true;
#endif

    if (isWindows)
```

chynesNR (approved):

nrcventura (approved):

PR #1973, opened 2023-10-12
ci: Run artifact builder on every all_solutions CI run.

tippmar-nr:
Removes previous changes that would only run Artifact Builder during a release. It now runs with every all_solutions CI run.


chynesNR:
Maybe we should see how much time this adds to each build before merging? We may want it to run more selectively

tippmar-nr:
@chynesNR We discussed in pairing and concluded that Artifact Builder would always be done before all of the integration tests complete, so it shouldn't add any extra time at all, just a bit more load on the github runner.

PR #1975, opened 2023-10-13
ci: Remove API call to update dotnet agent version in EU.

tippmar-nr:
As per the recent thread in #agent-dev, we no longer need to call the EU endpoint when updating the dotnet agent version.


nr-ahemsath (approved):
🧹 🎉

PR #1976, opened 2023-10-13
feat: Add new span attributes to more closely match OTel spec

### LINKED ISSUE #1942: Add Span attributes to move closer to OTel spec
Attributes we're looking to add:
* thread.id
* db.system
* server.address
* server.port

See the Agent spec for more details

### Acceptance Criteria
Attributes should be added with corresponding unit/integration tests.

### Estimates
S/M

chynesNR:
This was trickier than first expected for a few reasons:
* For database calls, we want the port, but what we store is the "port, path, or ID" and we don't record which it is. I had to split out the port at creation time.
* The "known values" for database providers that OTel uses are close to, but not exactly the same as, the ones we use
* The request for managed Thread ID doesn't really align with the way .NET handles async work. For now at least, we'll only provide the Thread ID if all Segments occur on the same thread.
* Both database calls and HTTP requests now have `server.address` and `server.port` attributes, but they're optional for database calls (making them Agent attributes) and required for HTTP requests (making them Intrinsic attributes)

Resolves #1942 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
+++ b/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
@@ -43,7 +43,9 @@ namespace NewRelic.Agent.Core.Attributes
         AttributeDefinition<float, double> DatabaseDuration { get; }
         AttributeDefinition<string, string> DbCollection { get; }
         AttributeDefinition<string, string> DbInstance { get; }
+        AttributeDefinition<string, string> DbOperation { get; }
         AttributeDefinition<string, string> DbStatement { get; }
+        AttributeDefinition<string, string> DbSystem { get; }
         AttributeDefinition<string, string> DistributedTraceId { get; }
         AttributeDefinition<TimeSpan, double> Duration { get; }
         AttributeDefinition<bool, bool> IsErrorExpected { get; }
@@ -91,6 +93,8 @@ namespace NewRelic.Agent.Core.Attributes
         AttributeDefinition<string, string> RequestUri { get; }
         AttributeDefinition<int?, string> ResponseStatus { get; }
         AttributeDefinition<bool, bool> Sampled { get; }
+        AttributeDefinition<string, string> ServerAddress { get; }
+        AttributeDefinition<long, long> ServerPort { get; }
         AttributeDefinition<SpanCategory, string> SpanCategory { get; }
         AttributeDefinition<string, string> SpanErrorClass { get; }
         AttributeDefinition<string, string> SpanErrorMessage { get; }
@@ -102,6 +106,7 @@ namespace NewRelic.Agent.Core.Attributes
         AttributeDefinition<string, string> SyntheticsMonitorIdForTraces { get; }
         AttributeDefinition<string, string> SyntheticsResourceId { get; }
         AttributeDefinition<string, string> SyntheticsResourceIdForTraces { get; }
+        AttributeDefinition<long, long> ThreadId { get; }
         AttributeDefinition<DateTime, long> Timestamp { get; }
         AttributeDefinition<DateTime, long> TimestampForError { get; }
         AttributeDefinition<TimeSpan, double> TotalTime { get; }
@@ -476,6 +481,12 @@ namespace NewRelic.Agent.Core.Attributes
                 .AppliesTo(AttributeDestinations.SpanEvent)
                 .Build(_attribFilter));
 
+        private AttributeDefinition<long, long> _threadId;
+        public AttributeDefinition<long, long> ThreadId => _threadId ?? (_threadId =
+            AttributeDefinitionBuilder.CreateLong("thread.id", AttributeClassification.Intrinsics)
+                .AppliesTo(AttributeDestinations.SpanEvent)
+                .Build(_attribFilter));
+
         private AttributeDefinition<string, string> _component;
         public AttributeDefinition<string, string> Component => _component ?? (_component =
             AttributeDefinitionBuilder.CreateString("component", AttributeClassification.Intrinsics)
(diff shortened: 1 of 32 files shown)


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Providers/Wrapper/Constants.cs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Collections.Concurrent;
+using System.Runtime.CompilerServices;
```
tippmar-nr, line 5: Is this `using` necessary?

tippmar-nr (approved):
Looks good!

src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
```diff
@@ -476,6 +483,12 @@ public AttributeDefinitions(IAttributeFilter attribFilter)
                 .AppliesTo(AttributeDestinations.SpanEvent)
                 .Build(_attribFilter));
 
+        private AttributeDefinition<long, long> _threadId;
+        public AttributeDefinition<long, long> ThreadId => _threadId ?? (_threadId =
+            AttributeDefinitionBuilder.CreateLong("thread.id", AttributeClassification.Intrinsics)
```
nrcventura, line 488: Is this really an Intrinsic attribute, since it will not always be added?

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
```diff
@@ -441,5 +441,26 @@ public string GetCategory()
         {
             return EnumNameCache<SpanCategory>.GetName(Data.SpanCategory);
         }
+
+        public static int? CheckAllSameThread(ICollection<Segment> segments)
```
nrcventura, line 445: Would it be safer to check if a segment was created from an instrumented method where the IsAsync flag is set? I think that the exclusive time calculation is mostly complicated when async-await is used, and when async-await is used the IsAsync flag should be set.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Providers/Wrapper/Constants.cs
```diff
@@ -82,6 +83,24 @@ public enum DatastoreVendor
...
+        {
+            switch (vendor)
+            {
+                case DatastoreVendor.Other:
+                    return "other_sql";
+                case DatastoreVendor.IBMDB2:
+                    return "db2";
+                // The others match our enum name
+                default:
+                    return vendor.ToString().ToLower();
```
nrcventura, line 99: Calling ToString on an enum involves reflection. I think we already have code that can be used to cache the results so that we don't need to incur the reflection cost each time. We should probably leverage that code here.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Providers/Wrapper/Constants.cs
chynesNR: Huh, not sure why that got auto-added.

src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
chynesNR: Perhaps not, but that's how it's defined in the spec. We have other Intrinsic attributes that are conditional, so I don't think it will be a problem.

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
chynesNR: That was my first approach, but we don't keep track of `IsAsync` beyond the initial check, and I had to touch a lot of code to get it to propagate down to where the attributes are set. This way I thought was at least similar to how we track parent/child relationships. Maybe we can brainstorm some alternatives.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Providers/Wrapper/Constants.cs
chynesNR: Ah I wasn't aware of `EnumNameCache`. I can do that.

chynesNR: It's literally the next line in that file. 🤦

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
nrcventura: This will act as a heuristic because it is just checking that each segment started on the same thread. It will not successfully ignore all of the async-await scenarios, because the thread will not change until there is an await. This is one of the reasons why the exclusive time calculation is comparing the the starting thread of the parent to the ending thread id of the child.

chynesNR: Ok, I think I'm with you now. I misread the conversation in the spec PR and also misunderstood how we were tracking thread IDs for Segments. Is this more in line with what you were expecting?

src/Agent/NewRelic/Agent/Core/Wrapper/AgentWrapperApi/Data/MethodCallData.cs
```diff
@@ -8,12 +8,14 @@ public class MethodCallData
         public readonly string TypeName;
         public readonly string MethodName;
         public readonly int InvocationTargetHashCode;
+        public readonly bool IsAsync;
 
-        public MethodCallData(string typeName, string methodName, int invocationTargetHashCode)
+        public MethodCallData(string typeName, string methodName, int invocationTargetHashCode, bool isAsync = false)
```
nrcventura, line 13: Would it be hard to make the isAsync parameter required? I find that it is easier to track down cases where `isAsync` is `false` when the parameter is required.

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
nrcventura: Yes, I think that your changes will better detect the case where code is async even if the async methods started on the same thread.

src/Agent/NewRelic/Agent/Core/Wrapper/AgentWrapperApi/Data/MethodCallData.cs
chynesNR: That was my first thought, too, but there are 133 places in unit tests that create a `MethodCallData` that would need updating. The good news is that there are only two places in Agent code that create a `MethodCallData`. The primary one takes a `MethodCall`, and it's required in that structure.

nrcventura: We could use .ctor chaining to handle that case if you think it would be helpful.
```
public MethodCallData(string typeName, string methodName, int invocationTargetHashCode) : this(typeName, methodName, invocationTargetHashCode, false) {}

public MethodCallData(string typeName, string methodName, int invocationTargetHashCode, bool isAsync)
{
   //Initialize the type here
}
```

nrcventura (approved):

src/Agent/NewRelic/Agent/Core/Wrapper/AgentWrapperApi/Data/MethodCallData.cs
chynesNR: Ah, I see what you mean now. I thought you were concerned with catching all the correct uses on this initial implementation, but you want to be able to find the synchronous cases more easily. Neither of the two (non-testing) places where the `MethodCallData` constructor is called use the default param, so I think I'll leave it as-is.

PR #1978, opened 2023-10-16
ci: improve Dotty error handling and output

nr-ahemsath:
## Description

This PR does the following:
1. Improves Dotty's handling of array indexing issues when looking for current and previous release catalog entries.
2. Improves Dotty's error output to indicate which package is causing the problem.
3. Prevents Dotty from sending a Slack message to the team if an exception is caught while in test mode.


tippmar-nr (approved):
Looks good!

PR #1980, opened 2023-10-17
Feature work/kafka -- Add more metrics

tippmar-nr:
Adds additional metrics related to message count and size, as well as a custom attribute for message size, as per the Kafka spec.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Kafka/KafkaConsumerWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Kafka/KafkaConsumerWrapper.cs
@@ -6,6 +6,7 @@ using System.Collections.Concurrent;
 using System.Text;
 using Confluent.Kafka;
 using NewRelic.Agent.Api;
+using NewRelic.Agent.Api.Experimental;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 using NewRelic.Reflection;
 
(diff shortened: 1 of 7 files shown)


chynesNR (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Kafka/KafkaConsumerWrapper.cs
```diff
@@ -80,15 +88,69 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
+
+            if (totalSize > 0)
+            {
+                transaction.AddCustomAttribute("kafka.consume.byteCount", totalSize);
+            }
+
+            // Add metrics for bytes received and messages received
+            var agentExp = agent.GetExperimentalApi();
+            agentExp.RecordCountMetric($"Message/Kafka/Topic/Named/{topic}/Received/Messages", 1);
+            agentExp.RecordCountMetric($"Message/Kafka/Topic/Named/{topic}/Received/Bytes", totalSize);
```
nrcventura, line 125: I think that this Bytes metric should not be a Count metric, because it will only report that bytes value in the count field, which will prevent the UI from performing useful functions on it. This should probably be reported as a summary metric (which will allow usage of all of the functions in the UI).

tippmar-nr: @nrcventura I'd be happy to make that change if I understood a bit more about how metric payloads work. I'll dig around and see what I can figure out and then ask for help if I get stuck... ;-)

PR #1986, opened 2023-10-17
ci: fix propagation delay in  YUM / Apt validation

tippmar-nr:
The conditional that determines whether to delay 5 minutes before attempting to validate Apt / Yum deployment was broken. Fixed it using a pattern that we've used in other workflows.


nr-ahemsath (approved):
👍

PR #1987, opened 2023-10-17
Feature work/kafka - Revise byte count metric

tippmar-nr:
Exposes a `RecordByteMetric()` method on `IAgentExperimental` that reports a "proper" byte count metric that allows the NR UI to do it's magic for avg/sum/etc.

Also adds Kafka to the ArtifactBuilder components lists.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -57,14 +57,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      #- name: Add msbuild to PATH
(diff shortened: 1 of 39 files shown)


tippmar-nr:
@nrcventura Appreciate if you'd take a look when you get time!

nrcventura (commented):
This looks like a better way to track the bytes metric so that all of the timeslice fields are populated. We should probably add some unit tests to cover the new code.

nrcventura (approved):

PR #1989, opened 2023-10-18
ci: Adjust project dependencies, use dotnet build in all_solutions workflow.

### LINKED ISSUE #1981: Intermittent build failures
We're seeing more frequent build errors in `Build FullAgent and MSIInstaller` where a binary that should exist is missing:

```
    51>PostBuild:
         Setting up params
         Resolve-Path : Cannot find path 'D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\src\Agent\NewRelic\Agent\Extensions\P
         roviders\Wrapper\StackExchangeRedis2Plus\bin\Release' because it does not exist.
         At D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\Build\build_home.ps1:59 char:29
         + ... eworkPath = Resolve-Path "$wrapperDirPath\bin\$Configuration\net4*")  ...
         +                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             + CategoryInfo          : ObjectNotFound: (D:\a\newrelic-d...lus\bin\Release:String) [Resolve-Path], ItemNotFoundE 
            xception
             + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.ResolvePathCommand
          
    51>D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\src\Agent\NewRelic\Home\Home.csproj(12,5): error MSB3073: The command "powershell.exe -ExecutionPolicy Bypass -NoProfile -NonInteractive -File .\build_home.ps1 -Configuration Release" exited with code 1.
    51>Done Building Project "D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\src\Agent\NewRelic\Home\Home.csproj" (default targets) -- FAILED.
```

It usually, but I don't think always, fails to find `StackExchangeRedis2Plus`, even though it's clearly built earlier in the build process. This may be a problem with our project dependencies. Triggering a rebuild usually fixes it.

### Estimates
S/M

tippmar-nr:
Two changes of note: 
* The `Home` project needs to have dependencies on every other (non-test) project in the solution, to ensure that it builds after all other projects are complete. It was missing dependencies on multiple recently added wrapper projects.
* I updated the `all_solutions` workflow to use `dotnet build` instead of `MSBuild.exe` - the net effect of this change is that the build process now uses a newer version of `MSBuild.exe` which seems a bit less verbose and maybe runs a little faster. I don't really think this specific change has any effect positive or negative on resolving the intermittent build failures, but I also don't think it will cause any harm. 
* Note that we still are required to run `MSBuild` for the MSI build (because Wix v3 doesn't work with `dotnet build`) and for the Intgegration test builds, because many of the helper apps are in the old `.csproj` format, which isn't compatible with `dotnet build`. 

Resolves #1981


nrcventura (approved):
🥇

chynesNR (approved):
Good work!

PR #1990, opened 2023-10-18
feat: Auto-instrument Confluent.Kafka

tippmar-nr:
## Description

Implements auto-instrumentation for `Confluent.Kafka`. 

Supports versions 1.4.0+

Instruments the following methods:
`IProducer.Produce`
`IProducer.ProduceAsync`
`IConsumer.Consume`

Also instruments serialization for the built-in serializers included in `Confluent.Kafka`.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
```diff
@@ -291,6 +291,13 @@ private void GenerateAndCollectMetrics(ImmutableTransaction immutableTransaction
                 MetricBuilder.TryBuildClientApplicationMetric(referrerCrossProcessId, catResponseTime, catResponseTime, txStats);
             }
 
+            // Capture required Kafka metrics
```
chynesNR, line 294: Does this need to be done here specifically? It feels a bit out of place.

tests/Agent/IntegrationTests/ContainerIntegrationTests/LinuxKafkaTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        _topicName = GenerateTopic();
+        var brokerName = "broker" + _topicName;
+        ((ContainerApplication)_fixture.RemoteApplication).DockerDependencies.Add(brokerName);
+
+        _fixture.Actions(setupConfiguration: () =>
+            {
+                var configModifier = new NewRelicConfigModifier(_fixture.DestinationNewRelicConfigFilePath);
+                configModifier.SetLogLevel("debug");
+                configModifier.ConfigureFasterMetricsHarvestCycle(10);
+                configModifier.LogToConsole();
```
chynesNR, line 38: Was `LogToConsole` for debugging or does it need to be on for the test?

tests/Agent/IntegrationTests/ContainerApplications/KafkaTestApp/KafkaTestApp.sln
nrcventura, line 1: Do we need this solution? Is it sufficient to just have the projects be part of one of the other testings solutions?

tests/Agent/IntegrationTests/ContainerApplications/docker-compose.yml
```diff
@@ -76,8 +78,49 @@ services:
             service: smoketestapp
         build:
             dockerfile: SmokeTestApp/Dockerfile.centos
+
+    UbuntuX64Kafka1TestApp:
+        extends:
+            service: smoketestapp
```
nrcventura, line 84: Instead of extending the smoketestapp, maybe it would be better to extract a common service definition that this service and the smoketestapp service can extend? Taking that approach will probably make it easier to understand which things are unique to each app.

tests/Agent/SmokeTest/SmokeTest.sln
nrcventura, line 1: Is this solution necessary? It only contains a single project.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -257,6 +257,26 @@ public ISegment StartMessageBrokerSegment(MethodCall methodCall, MessageBrokerDe
             return segment;
         }
 
+        public ISegment StartMessageBrokerSerializationSegment(MethodCall methodCall, MessageBrokerDestinationType destinationType, MessageBrokerAction operation, string brokerVendorName, string destinationName, string kind)
```
tippmar-nr, line 260: @jaffinito Did we miss anything with the serialization instrumentation changes? This method and `CreateMessageBrokerSerializationSegmentData()` below are not used anywhere in this branch. 

If we don't need these methods, we can remove them (here, `ITransaction`, `NoOpTransaction`).

src/Agent/NewRelic/Agent/Core/Segments/MessageBrokerSerializationSegmentData.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using NewRelic.Agent.Core.Aggregators;
+using NewRelic.Agent.Core.Metrics;
+using NewRelic.Agent.Core.Time;
+using static NewRelic.Agent.Core.WireModels.MetricWireModel;
+using NewRelic.Agent.Configuration;
+
+namespace NewRelic.Agent.Core.Segments
+{
+    public class MessageBrokerSerializationSegmentData : AbstractSegmentData
```
tippmar-nr, line 14: If we end up removing the two methods in `Transaction`, we can remove this class as well.

src/Agent/NewRelic/Agent/Core/Transformers/TransactionTransformer/TransactionTransformer.cs
jaffinito: I had trouble finding a good spot for this since we needed the full name.  Since all that was finalized here, it seemed like a good spot.

tests/Agent/IntegrationTests/ContainerIntegrationTests/LinuxKafkaTests.cs
jaffinito: I believe it is required.  The smoke tests also use it.

tests/Agent/IntegrationTests/ContainerApplications/docker-compose.yml
jaffinito: Yeah, that would be a good idea.  I think it should be a ticket for later though.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
jaffinito: I'll need to take a look and see.  I was getting the metrics in the test so I would expect it to be used.

tests/Agent/IntegrationTests/ContainerApplications/KafkaTestApp/KafkaTestApp.sln
jaffinito: Ug. I deleted this at least a dozen times.  VS Code really wants it to exist.  Deleting again....

tests/Agent/SmokeTest/SmokeTest.sln
jaffinito: Another that keeps getting created.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
jaffinito: That is being used [here](https://github.com/newrelic/newrelic-dotnet-agent/blob/5d44c667b8d5a926a642f6bc05cae8b9e5e35760/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Kafka/KafkaSerializerWrapper.cs#L26).

jaffinito: This is being called from [here](https://github.com/newrelic/newrelic-dotnet-agent/blob/5d44c667b8d5a926a642f6bc05cae8b9e5e35760/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Kafka/KafkaSerializerWrapper.cs#L29).

tippmar-nr: Yep, I don't know why VS wouldn't find that reference when I searched for it though...

src/Agent/NewRelic/Agent/Core/Transactions/NoOpTransaction.cs
```diff
@@ -81,6 +81,14 @@ public ISegment StartMessageBrokerSegment(MethodCall methodCall, MessageBrokerDe
             return Segment.NoOpSegment;
         }
 
+        public ISegment StartMessageBrokerSerializationSegment(MethodCall methodCall, MessageBrokerDestinationType destinationType, MessageBrokerAction operation, string brokerVendorName, string destinationName, string kind)
+        {
+#if DEBUG
+            Log.Finest("Skipping StartMessageBrokerSegment outside of a transaction");
```
chynesNR, line 87: Do you think this will be useful to keep in?

chynesNR (approved):

tests/Agent/IntegrationTests/ContainerApplications/docker-compose.yml
nrcventura: If you decide to take this on later can you submit a story for it?

src/Agent/NewRelic/Agent/Core/Transactions/NoOpTransaction.cs
nrcventura: I think that keeping it in makes it consistent with the other Start*Segment methods. It may make sense to remove all of these logs in a separate story though. At finest level logs we should already be able to see that the current transaction is a noop transaction.

nrcventura (approved):

PR #1992, opened 2023-10-19
feat: Add a supportability metric (Supportability/Logging/Forwarding/Empty) when a log event with no message and no exception is received. 

### LINKED ISSUE #1472: Add Supportability Metric for "empty" log messages
### Feature Description
We drop log messages where the message is empty and the exception is null. This occasionally leads to support issues where it takes time to figure out that the user was sending empty messages. 

Adding a supportability metric when we drop empty messages would help with troubleshooting.

### Priority
[Nice to Have]

tippmar-nr:
Resolves #1472

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Agent.cs
+++ b/src/Agent/NewRelic/Agent/Core/Agent.cs
@@ -439,14 +439,17 @@ namespace NewRelic.Agent.Core
 
                 var logMessage = getLogMessage(logEvent);
                 var logException = getLogException(logEvent);
+                var logContextData = _configurationService.Configuration.ContextDataEnabled ? getContextData(logEvent) : null;
 
                 // exit quickly if the message and exception are missing
+                // We would also check logContextData here, but behavior differs by logging framework as to whether they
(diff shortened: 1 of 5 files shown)


src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -439,14 +439,15 @@ public void RecordLogMessage(string frameworkName, object logEvent, Func<object,
 
                 var logMessage = getLogMessage(logEvent);
                 var logException = getLogException(logEvent);
+                var logContextData = _configurationService.Configuration.ContextDataEnabled ? getContextData(logEvent) : null;
 
-                // exit quickly if the message and exception are missing
-                if (string.IsNullOrWhiteSpace(logMessage) && logException is null)
+                // exit quickly if the message, exception and context data are missing
+                if (string.IsNullOrWhiteSpace(logMessage) && logException is null && (logContextData is null || logContextData.Count == 0))
```
nrcventura, line 445: We need to ensure that the corresponding agent spec is updated to account for this change.
We should also ensure that this change is called out in the changelog.

nrcventura (approved):

PR #2001, opened 2023-10-23
Adding alternate instrumentation point and first pass at tests

chynesNR:
A few things to note, as this is still a work in progress:
* The preprocessor directives are needed because of the dependency on `Microsoft.Extensions.Hosting`, which we already have a dependency on elsewhere and the versions are tricky to align
* Naming the queue doesn't work yet, so the metric checks in the tests won't pass. We may need to give up on that approach.
* I'm not sure we need two different wrappers? The code seems to work and is identical in both.
* The tests call `Publish` but our `Send` filter is being invoked (I know they're all called "Send" but it's the one with a `SendContext` when I was expecting the one with a `PublishContext`). But it's getting the action as "Publish" so I'm not sure if that's correct or not.


nrcventura (approved):

nr-ahemsath (approved):
🥇 Great work!

PR #2002, opened 2023-10-23
Feature work: ASP.NET Core 6+ - RUM injection logic

tippmar-nr:
Implements logic for injecting the RUM script in ASP.NET Core 6+ web pages, using rules identical to the existing ASP.NET wrapper. 

Adds unit tests for cross-agent browser injection using this new logic.

A later task will consider how to refactor common code to reduce duplication.


src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -292,60 +293,25 @@ public void HandleWrapperException(Exception exception)
...
-                {
-                    return null;
-                }
-
-                var shouldInject = _browserMonitoringPrereqChecker.ShouldAutomaticallyInject(transaction, requestPath, contentType);
-                if (!shouldInject)
-                {
-                    return null;
-                }
+            var script = TryGetRUMScriptInternal(contentType, requestPath);
```
nrcventura, line 296: To maintain the existing behavior, we may want to keep the stream and encoding checks in this method.

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -292,60 +293,25 @@ public void HandleWrapperException(Exception exception)
...
+            return script == null ? null : new BrowserMonitoringStreamInjector(() => script, stream, encoding);
```
nrcventura, line 297: Do we need to use a lambda to pass the script to the stream injector? It seems like we could reduce allocations if we just pass the script directly.

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -292,60 +293,25 @@ public void HandleWrapperException(Exception exception)
...
+        }
 
-                // Once the transaction name is used for RUM it must be frozen
-                transaction.CandidateTransactionName.Freeze(TransactionNameFreezeReason.AutoBrowserScriptInjection);
-                var script = _browserMonitoringScriptMaker.GetScript(transaction, null);
-                if (script == null)
-                {
-                    return null;
-                }
+        public async Task TryInjectBrowserScriptAsync(string contentType, string requestPath, byte[] buffer, Stream baseStream, ITransaction transaction)
```
nrcventura, line 300: Should baseStream be checked for null? Or is the null possibility just an asp.net thing?

src/Agent/NewRelic/Agent/Core/Agent.cs
tippmar-nr: The lambda function is to accommodate unit testing.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore6Plus/ResponseStreamWrapper.cs
```diff
@@ -36,18 +35,6 @@ public override void Flush()
             _baseStream.Flush();
         }
 
-        public override Task FlushAsync(CancellationToken cancellationToken)
-        {
-            if (!_isContentLengthSet && IsHtmlResponse())
-            {
-                _context.Response.Headers.ContentLength = null;
-                _isContentLengthSet = true;
-            }
```
nrcventura, line 45: Is this code not necessary anymore?

src/Agent/NewRelic/Agent/Core/Agent.cs
nrcventura: Is there another way to write the tests so that we do not need the extra allocations and overhead in the agent code?

src/Agent/NewRelic/Agent/Core/Agent.cs
tippmar-nr: Good catch.

src/Agent/NewRelic/Agent/Core/Agent.cs
tippmar-nr: I honestly doubt `null` was even a possibility in asp.net. Looks to me like some unnecessary defensive coding.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore6Plus/ResponseStreamWrapper.cs
tippmar-nr: No, it was from the example code I originally was working with, that was for a slightly different purpose than what we're doing.

src/Agent/NewRelic/Agent/Core/Agent.cs
tippmar-nr: I'm not sure; I was just replicating the way `BrowserMonitoringWriter` and its unit tests (in `RumLoaderInsertionLocationTests`) was written. The only real difference between the classes is that mine is `static`.

nrcventura: It just seems like complexity that we don't need. I could see us passing a lambda or local function if we needed to lazy load, or defer execution of some logic. However, in this case, we already executed the code to get the script, so it doesn't seem necessary to wrap it in a lambda or local function.

tippmar-nr: Turns out it was pretty easy to refactor and just pass the byte array directly.

nrcventura (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore6Plus/ResponseStreamWrapper.cs
```diff
@@ -129,17 +124,14 @@ private bool IsHtmlResponse(bool forceReCheck = false)
                 return false;
 
             // Requirements for script injection:
-            // * has to have result body
-            // * 200 or 500 response
             // * text/html response
             // * UTF-8 formatted (explicit or no charset)
 
             _isHtmlResponse =
-                _context.Response.StatusCode is 200 or 500 &&
```
chynesNR, line 138: Do we not need to check for these cases?

tippmar-nr: I don't think so; there are other valid status codes that we might want to instrument.

PR #2011, opened 2023-10-25
Feature work (ASP.NET Core 6+ Browser Injection): Response Compression

tippmar-nr:
Adds instrumentation for `Microsoft.AspNetCore.ResponseCompression.ResponseCompressionBody.OnWrite()` to wrap the compression stream with our browser injecting wrapper. Achieves reliable browser injection when response compression is used, and works in conjunction with the wrapper we add at the beginning of the pipeline.

I will add integration tests in a separate PR.


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore6Plus/BrowserInjectingStreamWrapper.cs
```diff
@@ -65,36 +66,43 @@ public override void SetLength(long value)
...
         {
-            if (IsHtmlResponse())
-            {
-                var curBuf = buffer.AsMemory(offset, count).ToArray();
-                _agent.TryInjectBrowserScriptAsync(_context.Response.ContentType, _context.Request.Path.Value, curBuf,
-                        _baseStream)
-                    .GetAwaiter().GetResult();
-            }
-            else
+            if (_context.Items[InjectingRUM] == null) // pass through without modification if we're already in the middle of injecting
```
chynesNR, line 69: Is it possible to get here without this key being set?

tippmar-nr: This implementation is probably going to be worthy of some explanation at some point - there are some odd "recursion" cases that we have to deal with, necessitating the `_context.Items[]` usage.

nrcventura: I think it's fine to use `_context.Items`, but I'm more concerned about causing a crash if the key doesn't exist in that collection. In other words, can we safely use `[InjectingRUM]` to do the check, or do we need to use a different method that will not throw an exception?

tippmar-nr: `.ContainsKey()` I guess is an alternative. Though the current code doesn't throw if the key doesn't exist (it, probably, creates a key with a `null` default value).

tippmar-nr: Reworked the code a bit.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore6Plus/BrowserInjectingStreamWrapper.cs
```diff
@@ -65,36 +66,43 @@ public override void SetLength(long value)
...
         {
-            if (IsHtmlResponse())
-            {
-                var curBuf = buffer.AsMemory(offset, count).ToArray();
-                _agent.TryInjectBrowserScriptAsync(_context.Response.ContentType, _context.Request.Path.Value, curBuf,
-                        _baseStream)
-                    .GetAwaiter().GetResult();
-            }
-            else
+            if (!_context.Items.ContainsKey(InjectingRUM))  // pass through without modification if we're already in the middle of injecting
```
nrcventura, line 69: Since this check is used in a couple of place it might be good to extract it into a method, such as `CurrentlyInjecting()`. Then the code could read like `if (!CurrentlyInjecting())`.

tippmar-nr: Reasonable suggestion. I refactored a bit.

chynesNR (approved):

nrcventura (approved):

PR #2012, opened 2023-10-26
chore: Reduce profiler iscoreclr checks and use correct tokenizer

nrcventura:

## Description

Reduces the number of times we need to check the isCoreClr flag to determine which assembly to reference in an instruction.

This also includes a fix to ensure that the correct tokenizer is always used.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Profiler/MethodRewriter/ApiFunctionManipulator.h
+++ b/src/Agent/NewRelic/Profiler/MethodRewriter/ApiFunctionManipulator.h
@@ -50,15 +50,7 @@ namespace NewRelic { namespace Profiler { namespace MethodRewriter
                     
                     _instructions->Append(_X("ldnull"));
                     BuildObjectArrayOfParameters();
-
-                    if (_isCoreClr)
-                    {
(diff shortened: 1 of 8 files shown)


chynesNR (approved):
Very nice, this is much cleaner. As an optimization, have you considered using a helper function based on `sprintf`? I think that would end up being fewer string allocations but might be more trouble than it's worth.

nrcventura:
> Very nice, this is much cleaner. As an optimization, have you considered using a helper function based on `sprintf`? I think that would end up being fewer string allocations but might be more trouble than it's worth.

I thought about doing that, but I think that required moving away from our xplat string helpers, but I didn't spend much time trying it out. I really wanted to do something similar to c#'s interpolated strings, and `sprintf` was the closest thing that I found to that.

With newer versions of the C++ standard  the `+` operator becomes a `constexpr` which could help with the performance.

I also considered using a string stream, but that approach also seemed like more trouble than it was worth.

These methods should only be called when we are injecting byte code into a method, which doesn't happen frequently. So  I'm ok with slightly worse performance for better readability in this code. I can submit another story to reconsider using `sprintf`, because I think that could make the code more readable depending on if we need to manage a string buffer or not.

PR #2017, opened 2023-10-27
chore: Reduce duplicate methods for handling .net framework and .net/.net core injection.

nrcventura:

## Description

Mostly replaces the duplicate methods that handled slight differences between .net framework and .net/.net core byte code injection.

Some other changes related to this cleanup include.

- Removes the code that created an instance of the ModuleInjector, because it only contains static methods.
- Removes the OverrideEventMask method because it no longer did anything.
- Moved some of the logic from the ModuleInjector into the Module class
- Removed unnecessary methods and fields from the Module class

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Profiler/Profiler/CorProfilerCallbackImpl.h
+++ b/src/Agent/NewRelic/Profiler/Profiler/CorProfilerCallbackImpl.h
@@ -69,7 +69,6 @@ namespace NewRelic { namespace Profiler {
 
     private:
         std::atomic<int> _referenceCount;
-        std::shared_ptr<ModuleInjector::ModuleInjector> _moduleInjector;
 
     public:
         CorProfilerCallbackImpl()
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Profiler/Profiler/CorProfilerCallbackImpl.h
```diff
@@ -306,18 +298,6 @@ namespace NewRelic { namespace Profiler {
             return S_OK;
         }
 
-
-        virtual DWORD OverrideEventMask(DWORD eventMask)
-        {
-#ifndef PAL_STDCPP_COMPAT
-            if (!_isCoreClr)
-            {
-                _moduleInjector.reset(new ModuleInjector::ModuleInjector());
```
chynesNR, line 315: I'm not sure I follow what this was accomplishing before, but it's no longer needed?

chynesNR (approved):
Looking good!

src/Agent/NewRelic/Profiler/Profiler/CorProfilerCallbackImpl.h
nrcventura: It created an instance of a class that didn't have any instance level methods or data, so there was no state that needed to be managed for that class which means that we were replacing the current instance of the injector with a new one without a reason to do that. That class only contains static methods, and with this change the usage of those methods were updated to use a static invocation instead of an instance-level invocation style. More specifically `_moduleInjector->InjectIntoModule(*module)` was changed to `ModuleInjector::ModuleInjector::InjectIntoModule(*module, _isCoreClr)` because InjectIntoModule was a static method and we do not need an instance of ModuleInjector to invoke it.  The previous code is probably a remnant of some older behavior that has not been in place for a while.

PR #2018, opened 2023-10-31
fix: Fix a segfault when using sendDataOnExit with Linux on Docker.

jaffinito:
## Description

The repro provided by the customer demonstrates one possible way the issue can occur.  In the repro, sendDataOnExit is enabled and StartAgent API call is used to get the agent running.  Immediately after starting the SetApplicationName API is called and then the app shuts down.  This triggers a segfault in Docker running Linux (kubernetes as well). In the case of the repro, the agent gets connected and then the name change triggers the reconnect so that agent is not connected anymore.  sendDataOnExit forces Connect to be synchronous as well. 

~~I narrowed down the trouble to the clean shutdown code that attempts to send data and then disconnect - this is called by default.  The aggregators then check for sendDataOnExit and if it is present they try and harvest and send data.   By adding a boolean field to track the very basic connected state, the clean shutdown only occurs if we are connected.~~

I narrowed the issue down to the GCSamplerNetCore class, at least indirectly.  When the agent is shutting down we make config changes as well as requesting various services to stop, include the types derived from AbstractSampler (a ConfigurationBasedService). The problem I was seeing is that GCSamplerNetCore was stopped, but then it was started agent moments later.  The log files ends with "EnableEvents" begin called and a segfault occurs. It looks like the service is starting again due to the base class (AbstractSampler) still returning "true" for Enabled.  Disabling the samplers via config did not help with this in my repro.

To address this, I updated the AbstractSampler.Enabled property to also check "Agent.IsAgentShuttingDown" which is set immediately after Shutdown is called.  This way, even if the config changes have not made it to every sampler, it will still return disabled and not attempt to start agent.  This worked and I was able to remove my previous "connected" changes.

- Updates AbstractSampler.Enabled to also check Agent.IsAgentShuttingDown as part of its conditional.
- Adds a finest level log message in AgentManager to note that a "clean" shut down is occurring (I liked the extra - but its optional).
- Updates the Scheduler.ExecuteEvery "existing timer" message to specify the timer it is stopping.
- Updates MultiFunctionApplicationHelpers.csproj to not run the NugetAudit.  It was blocking my builds.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
@@ -6,6 +6,7 @@
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <DisableImplicitNamespaceImports>true</DisableImplicitNamespaceImports>
     <LangVersion>default</LangVersion>
+    <NuGetAudit>false</NuGetAudit>
   </PropertyGroup>
  
   <ItemGroup>
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
@@ -6,7 +6,6 @@
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <DisableImplicitNamespaceImports>true</DisableImplicitNamespaceImports>
     <LangVersion>default</LangVersion>
-    <NuGetAudit>false</NuGetAudit>
   </PropertyGroup>
  
   <ItemGroup>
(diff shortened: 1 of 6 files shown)


tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
```diff
@@ -6,6 +6,7 @@
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <DisableImplicitNamespaceImports>true</DisableImplicitNamespaceImports>
     <LangVersion>default</LangVersion>
+    <NuGetAudit>false</NuGetAudit>
```
nrcventura, line 9: Should this be added to the props for all of the projects in the integration tests directory? If so I think we already have a props file that we use for all of our tests.

jaffinito: Yeah, I should have done that.  Let me push an update...

nrcventura (approved):
This seems like a safe way to handle the shutdown. Good job narrowing the problem down to the one sampler. I'm guessing that the problem occurs when trying to subscribe to the EventPipe after the process has started to shutdown. If you can reproduce the problem without the New Relic agent, it may be worth opening a ticket with Microsoft to see if they can add more protections to the EventPipe.

chynesNR (approved):
Looks good! I suspect there are other places where we can start something up during shutdown that we should keep an eye out for.

nrcventura (approved):

PR #2020, opened 2023-10-31
Feature work (ASP.NET Core 6 Plus): Browser Injection Integration Tests

tippmar-nr:
Adds integration tests to exercise browser injection with and without response compression. 

Adds a new Razor web app test harness also.


chynesNR (approved):
Very nice!

tests/Agent/IntegrationTests/IntegrationTests/AgentFeatures/BrowserAgentAutoInjection6Plus.cs
```diff
@@ -0,0 +1,87 @@
...
+            (
+                setupConfiguration: () =>
+                {
+                    var configPath = fixture.DestinationNewRelicConfigFilePath;
+
+                    var configModifier = new NewRelicConfigModifier(configPath);
+                    configModifier.AutoInstrumentBrowserMonitoring(true);
+                    configModifier.BrowserMonitoringEnableAttributes(true);
+
+                    configModifier.BrowserMonitoringLoader("rum");
```
nrcventura, line 34: Should we add tests for the different browser agents that are available? For the Asp.Net browser tests, we are testing that the different browser agents work.

tippmar-nr: I don't think so - the logic for loading a specific browser agent is unchanged for the ASP.NET Core 6+ browser injection.

nrcventura: We received requests from the browser team a couple of times where they asked us to confirm if a certain type of browser agent was causing problems. We have test coverage for different string formats, and we somewhat recently added coverage for the 3 browser agent types in an aspnet app (which was the only thing we supported). It might save us some time if we have a similar set of tests for aspnetcore.

tippmar-nr: I can add some tests, but I don't think they'll do anything useful - injecting the script is exactly the same, regardless of which browser agent type gets loaded.

nrcventura: Right, I see this as more of an end to end test that has been requested from time to time, and it automates what we have been asked to do manually.

PR #2025, opened 2023-11-01
Support logging to stdout in the Profiler

chynesNR:
As part of the change to support read-only filesystems, we would like the option to redirect Profiler logging to the console. The Agent already has this option.

Changes include:
* The existing "console" setting in the Agent now applies to the Profiler as well. When enabled, the Profiler will write log lines to the console instead of a file. The new environment variable `NEW_RELIC_LOG_CONSOLE` can also be used.
* Logging can be disabled entirely in the Profiler by setting the new value `enabled` to `false` under the existing `log` node in the Agent config file. Logging is of course still enabled by default. The new environment variable `NEW_RELIC_LOG_ENABLED` can also be used.
* If an environment variable toggle is set to an invalid value (i.e. not 0/1/true/false), it is ignored, rather than treated as 0/false.
* If file logging is enabled but the Profiler is unable to write to disk, logging is disabled and no further attempts to log will be made. The Profiler no longer treats this as a fatal error.
* Unit tests for the new settings.

Observations:
* If you have logging set to Finest and Console logging is enabled, even the most basic application is essentially rendered useless due to the amount of resources spent writing to the console.
* Previously, we initialized the logger then read the config file. But this meant even if logging was disabled, an empty log file would be created. I had to rearrange some of the startup logic to make sure disabling logging resulted in no attempted writes to disk.


src/Agent/NewRelic/Profiler/MethodRewriterTest/MockSystemCalls.h
chynesNR, line 1: This class wasn't being used and couldn't be instantiated in its current state.

src/Agent/NewRelic/Profiler/Logging/Logger.h
```diff
@@ -86,6 +88,32 @@ namespace NewRelic {
...
+                }
+                void SetEnabled(bool enabled)
+                {
+                    _enabled = enabled;
+                }
+                bool GetEnabled()
+                {
+                    return _enabled;
+                }
+                void SetIniitalized()
```
nr-ahemsath, line 108: Minor misspelling

chynesNR: Whoops!

src/Agent/NewRelic/Profiler/Logging/Logger.h
```diff
@@ -127,15 +158,15 @@ namespace NewRelic {
...
+                if (log.GetInitialized() && log.GetEnabled() && (level >= log.GetLevel()))
                 {
                     //each thread will have these on the stack...
                     std::tm  tstruct;
                     std::time_t now;
                     (void)time(&now);
                     (void)gmtime_s(&tstruct, &now);
                     auto levelstr = nrlog::GetLevelString(level);
-                    auto& strm = log.ostr();
+                    std::wostream& strm = log.GetConsoleLogging() ? std::wcout : log.ostr();
```
nrcventura, line 169: Will `wcout` work across platforms for us? I think there are cases where the `L` prefix is used strings that we log, no prefix is used for the strings, on Linux some of the strings will be prefixed with `u`, and then we also log strings of the type xstring_t which are different between windows and linux.

src/Agent/NewRelic/Profiler/ConfigurationTest/MockSystemCalls.h
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+#include <functional>
+#include <string>
+#include <memory>
+#include <unordered_map>
+#include "../Profiler/SystemCalls.h"
```
nrcventura, line 9: If you want this to build on Linux, you will need to conditionally include the UnixSystemCalls.

src/Agent/NewRelic/Profiler/Logging/Logger.h
chynesNR: That's a good question, I'll have to look into it.

src/Agent/NewRelic/Profiler/ConfigurationTest/MockSystemCalls.h
chynesNR: I didn't realize they built at all on Linux. Updated.

chynesNR:
As we discussed this morning, it will now clamp console logging to Info level or higher.

src/Agent/NewRelic/Profiler/Logging/Logger.h
chynesNR: I think we're fine. The stream insertion operator `<<` should already be handling the conversion so that everything is the same string type, and Alex and I just ran a quick test and all the output was correct. We can also do some additional testing at the end of the milestone.

nrcventura (approved):

PR #2036, opened 2023-11-08
ci: Use actual release version for release notes branch

nr-ahemsath:
Fix a minor issue with the workflow that generates the release notes PR for the docs website; it was using a hardcoded version string of 10.13.0 to create the branch instead of the actual release version.


tippmar-nr (approved):
Good catch!

PR #2045, opened 2023-11-09
Agent support for read-only file systems

chynesNR:
This adds support on the Agent side to better handle read-only file systems:
* A config setting (`log:enabled`) and environment variable (`NEW_RELIC_LOG_ENABLED`) to disable logging completely. No file writes will be attempted.
* If file logging is enabled but fails to write a file, disable logging.
* An environment variable to use console logging (`NEW_RELIC_LOG_CONSOLE`) -- there is already an existing config setting.
* Supportability metrics for when logging is disabled by the user or by error.


src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -829,5 +830,17 @@ protected override void OnConfigurationUpdated(ConfigurationUpdateSource configu
             // Some one time metrics are reporting configured values, so we want to re-report them if the configuration changed
             _oneTimeMetricsCollected = false;
         }
+
+        private void ReportIfLoggingDisabled()
+        {
+            if (!_configuration.LoggingEnabled)
+            {
+                ReportSupportabilityCountMetric(MetricNames.SupportabilityLoggingDisabled);
```
nrcventura, line 838: Is the name of this metric and corresponding methods clear enough to differentiate them from log forwarding and decoration? I'm not sure if referring to these things as Agent Logs helps with disambiguating the different "Logs".

chynesNR: Yeah I struggled a bit with the names. If anyone has a better suggestion I'd be glad to hear it. "Internal logs"?

tests/Agent/UnitTests/Core.UnitTest/AgentHealth/AgentHealthReporterTests.cs
```diff
@@ -45,19 +45,21 @@ public void TearDown()
...
 
-        private static IConfiguration GetDefaultConfiguration()
+        private static IConfiguration GetDefaultConfiguration(bool enableLogging)
         {
             var configuration = Mock.Create<IConfiguration>();
             Mock.Arrange(() => configuration.LogEventCollectorEnabled).Returns(true);
             Mock.Arrange(() => configuration.LogDecoratorEnabled).Returns(true);
             Mock.Arrange(() => configuration.LogMetricsCollectorEnabled).Returns(true);
             Mock.Arrange(() => configuration.InfiniteTracingCompression).Returns(true);
+            Mock.Arrange(() => configuration.LoggingEnabled).Returns(enableLogging);
```
nrcventura, line 55: You can probably avoid removing the SetUp attribute and needing to manually call SetUp from each test if you update this Mock to something more like the following if you make _enableLogging a field.
```suggestion
            Mock.Arrange(() => configuration.LoggingEnabled).Returns(() => _enableLogging);
```
Using that syntax will delay the evaluation of that field until the test needs to access configuration.LoggingEnabled.

src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
nr-ahemsath: I think "AgentLogging" is clear enough; we refer to the instrumentation for customers' application logs as "ApplicationLogging" in several places (in config, anyway, perhaps not in supportability metrics).

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
```diff
@@ -594,11 +599,46 @@ private string GetLogFileName()
             return "newrelic_agent_" + Strings.SafeFileName(name) + ".log";
         }
 
+        private bool GetOverride(string name, bool fallback)
```
nrcventura, line 602: Would it be better to make the `EnvironmentOverrides` method in `DefaultConfiguration` `public static` so that it can be used here?

tests/Agent/UnitTests/Core.UnitTest/AgentHealth/AgentHealthReporterTests.cs
chynesNR: Oh that's clever. I'll do that.

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
chynesNR: Hmm... maybe. I thought `DefaultConfiguration` got initialized after `ConfigurationLoader` but that's not the case. I'll give it a try.

nrcventura: DefaultConfiguration is initialized later, but the static methods on that class don't require it to be initialized.

tests/Agent/UnitTests/Core.UnitTest/AgentHealth/AgentHealthReporterTests.cs
```diff
@@ -42,24 +44,24 @@ public void SetUp(bool enableLogging = true)
         [TearDown]
         public void TearDown()
         {
+            _enableLogging = true;
```
nrcventura, line 47: It might be clearer if you moved this line into the SetUp method, and if you do that you don't need to use the field initializer to set an initial value either.

chynesNR: Fair enough.

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
chynesNR: That's technically true but it needs an `IEnvironment` passed in (currently in the constructor) to actually get the values, which will be harder to do for the `ConfigurationLoader` tests. Or am I misunderstanding your suggestion?

nrcventura: Ok, that seems like it might be a bigger refactor than it is worth. Unless we do something like...
```
// In DefaultConfiguration
private returnType EvironmentOverrides(.....)
{
   return GetEnvironmentOverrides(_enviornment, ....);
}

public static returnType GetEnvironmentOverrides(IEnvironment, ....)
{
    //The current logic for getting environment overrides
}

// In logging configuration
private returnType GetOverrides(....)
{
   // _environment can be initialized in the .ctor manually since the DI container is not available yet
   return DefaultConfiguration.GetEnvironmentOverrides(_environment, ....);
}
```

The main thing that I'm trying to accomplish is to keep the environment variable parsing and overriding logic consistent between the two locations. There are many other ways to accomplish that.

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
```diff
@@ -194,6 +187,10 @@ private static LoggerConfiguration ConfigureConsoleSink(this LoggerConfiguration
         /// <param name="config">The configuration for the appender.</param>
         private static LoggerConfiguration ConfigureFileSink(this LoggerConfiguration loggerConfiguration, ILogConfig config)
         {
+            if (!config.Enabled)
```
tippmar-nr, line 190: Do we need to add similar code in `ConfigureAuditLogSink()`?

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
chynesNR: Yeah, I don't like the duplicate code either. I should have said that I was trying to avoid a big refactor. Your idea would work but I'm not sure `DefaultConfiguration` is the right place for it to live. I think we need a common helper class. I'll put a tech debt ticket in the backlog so we can address it properly. Sound good?

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
chynesNR: I meant to ask the team this but forgot about it. Is logging disabled but audit log enabled a valid configuration? What should happen?

tippmar-nr: Seems to me that if logging is disabled (because of read-only FS), audit logging has to be disabled as well. But as long as the app doesn't crash when the audit log sink tries to create the log file, I guess it isn't critical.

chynesNR: I've explicitly disabled it now if you have logging disabled; I think that makes sense.

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
nrcventura: That sounds good to me.

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
nrcventura: That seems like a scenario that it would be great to warn a customer about, but unfortunately with logging disabled there is no way to provide a warning about it.

nrcventura (approved):

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
chynesNR: True, we'll have to rely on the supportability metric.

src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
chynesNR: https://github.com/newrelic/newrelic-dotnet-agent/issues/2049

PR #2063, opened 2023-11-13
chore: Deleted code samples.

tippmar-nr:
Code samples are now housed in the (soon-to-be-public) `newrelic-dotnet-examples` repo.


samples/README.md
chynesNR, line 1: Maybe we should leave the readme in place, directing people to the new repo?

tippmar-nr: Seems reasonable, after the new repo goes public. For now we can't really reference it.

jaffinito (approved):

PR #2075, opened 2023-11-15
test: Eliminate firewall pop-ups when running integration tests locally

### LINKED ISSUE #1921: Fix firewall popups when running integration tests locally
Running the integration tests locally triggers a number of firewall popups, because exclusions are path-specific, and our test applications are dynamically generated at run time. They don't affect the outcome of the tests, but are annoying.

There is a POC branch that may already be working: https://github.com/newrelic/newrelic-dotnet-agent/commit/9992b81bf0f9507ec9ec3e00d307101974f71fb2

### Acceptance Criteria
Should be able to run all integration tests locally without any user input.

### Estimates
M

tippmar-nr:
Updates Integration tests so that the annoying firewall pop-ups don't occur when running locally. Much of that issue was caused by "wildcard" bindings in some of the test services. There were also some places where using `localhost` caused the popup, for unexplained reasons. So the goal was to globally use `127.0.0.1` everywhere - but a few odd cases caused me to fall just short of reaching that goal. 

In general, we'll use `127.0.0.1` now instead of `localhost` - but if a test is using `localhost` there's almost certainly a reason for it. 

Highlight of changes:
* Changed `RemoteApplication.DestinationServerName` to return `127.0.0.1` instead of doing a DNS lookup of the hostname of the local machine.
* Changed nearly all usages of `localhost` in the integration test applications to `127.0.0.1` instead.
* Changed nearly all usages of `localhost` in integration tests to use `RemoteApplication.DestinationServerName`
* Modified "wildcard" (`*:{port}`) bindings for several services to use a specific address (usually `127.0.0.1`) instead.
* Had to leave some tests (WCF, I'm looking at you) referencing `localhost` - though even some of the WCF tests had to reference `127.0.0.1` - it's a complex web of interdependencies...

The only test that fails consistently on my local machine is `BasicAspWebService` - but I think that test also fails locally on the current `main` branch, so maybe it's not a big deal. 

Resolves #1921

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatDisabledChainedRequestsWebRequest.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatDisabledChainedRequestsWebRequest.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Net.Http.Headers;
 using NewRelic.Agent.IntegrationTestHelpers;
 using NewRelic.Agent.IntegrationTestHelpers.Models;
+using NewRelic.Agent.IntegrationTestHelpers.RemoteServiceFixtures;
 using NewRelic.Agent.IntegrationTests.CatInbound;
 using NewRelic.Testing.Assertions;
 using Xunit;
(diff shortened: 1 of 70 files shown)


jaffinito (approved):
I don't think the smoketests failure is related.

tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatDisabledChainedRequestsWebRequest.cs
```diff
@@ -62,15 +63,15 @@ public void Test()
             {
                 new Assertions.ExpectedMetric { metricName = @"External/all", callCount = 1 },
                 new Assertions.ExpectedMetric { metricName = @"External/allWeb", callCount = 1 },
-                new Assertions.ExpectedMetric { metricName = $@"External/{_fixture.RemoteApplication.DestinationServerName}/all", callCount = 1 },
-                new Assertions.ExpectedMetric { metricName = $@"External/{_fixture.RemoteApplication.DestinationServerName}/Stream/GET", callCount = 1 },
-                new Assertions.ExpectedMetric { metricName = $@"External/{_fixture.RemoteApplication.DestinationServerName}/Stream/GET", metricScope = @"WebTransaction/MVC/DefaultController/ChainedWebRequest", callCount = 1 }
+                new Assertions.ExpectedMetric { metricName = $@"External/{RemoteApplication.DestinationServerName}/all", callCount = 1 },
```
nr-ahemsath, line 66: Is this change (and all the others like it) related to the popups, or does it improve something else?

chynesNR (approved):

tests/Agent/IntegrationTests/IntegrationTests/CatOutbound/CatDisabledChainedRequestsWebRequest.cs
nr-ahemsath: Never mind, I figured it out for myself.

nr-ahemsath (approved):
Looks good.  I pulled the branch and built/ran the tests locally.  Much better experience without all the popups!

PR #2076, opened 2023-11-16
Add integration tests for file logging disabled and console logging enabled

nr-ahemsath:
Adds two sets of integration tests, one for enabling console logging and one for disabling file logging.


chynesNR (approved):
Looks good!

PR #2082, opened 2023-11-17
fix: Remove dependency on Serilog.Expressions.

### LINKED ISSUE #2083: Agent v10.11+ fails to report data or create a log file
**Description**

When running a .NET Framework application that has a binding redirect for `System.ValueTuple`, referencing an old version of that assembly, the .NET agent v10.11 and later produces a profiler log file but does not create an agent log file and does not report data to New Relic.

Example:
```
      <dependentAssembly>
        <assemblyIdentity name="System.ValueTuple" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
      </dependentAssembly>
```

If the binding redirect for `System.ValueTuple` is removed from the .NET Framework application, the .NET agent starts and reports data as expected.

**Expected Behavior**

The .NET agent should start and report data without regard to any binding redirects in a customer application.

**Steps to Reproduce**

Create a .NET Framework console application targeting Framework v4.8 and add a binding redirect to the `app.config` file similar to the following: 
```
  <runtime>
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="System.ValueTuple" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
      </dependentAssembly>
    </assemblyBinding>
  </runtime>
```

Run the application and note that a `NewRelic.Profiler.99999.log` file is created in the New Relic home folder, but an Agent logfile is not created. Also note that the application fails to report data to New Relic. 

Remove the binding redirect and run the application. Note that Profiler and Agent log files are created and the application reports data to New Relic

P1
M

tippmar-nr:

## Description

Removes the dependency on the `Serilog.Expressions` package, which resolves issues experienced by some customers where the agent does not report data and does not create a log file. 

Added a new log enricher for the required timestamp format in the log file. Updated existing log enrichers to improve performance where possible.

An integration test was added to validate that the logfile output is in the expected format.

Resolves #2083 

Props to @chynesNR for sleuthing this one!

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
+++ b/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
@@ -23,9 +23,9 @@ namespace NewRelic.Agent.Core
         //private static ILayout AuditLogLayout = new PatternLayout("%utcdate{yyyy-MM-dd HH:mm:ss,fff} NewRelic %level: %message\r\n");
         //private static ILayout FileLogLayout = new PatternLayout("%utcdate{yyyy-MM-dd HH:mm:ss,fff} NewRelic %6level: [pid: %property{pid}, tid: %property{threadid}] %message\r\n");
 
-        private const string AuditLogLayout = "{UTCTimestamp} NewRelic Audit: {Message}\n";
+        private const string AuditLogLayout = "{UTCTimestamp} NewRelic Audit: {Message:l}\n";
 
-        private const string FileLogLayout = "{UTCTimestamp} NewRelic {NRLogLevel,6}: [pid: {pid}, tid: {tid}] {Message}\n{Exception}";
+        private const string FileLogLayout = "{UTCTimestamp} NewRelic {NRLogLevel,6}: [pid: {pid}, tid: {tid}] {Message:l}\n{Exception:l}";
 
         private static LoggingLevelSwitch _loggingLevelSwitch = new LoggingLevelSwitch();
 
--- a/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
+++ b/src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
@@ -272,7 +272,6 @@ namespace NewRelic.Agent.Core
                     .WriteTo
                     .File(path: fileName,
                             outputTemplate:outputFormat,
-                            //formatter: textFormatter,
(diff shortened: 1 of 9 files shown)


src/Agent/NewRelic/Agent/Core/Logging/ThreadIdEnricher.cs
```diff
@@ -8,13 +8,24 @@
...
-            logEvent.AddPropertyIfAbsent(propertyFactory.CreateProperty(
-                "tid", Thread.CurrentThread.ManagedThreadId));
+            var threadId = Thread.CurrentThread.ManagedThreadId;
+
+            var prop = _tidProperty;
+
+            if (prop == null || (int?)((ScalarValue)prop.Value).Value != threadId)
+                _tidProperty = prop = propertyFactory.CreateProperty("tid", threadId);
+
+            logEvent.AddPropertyIfAbsent(prop);
```
nrcventura, line 28: Does this actually improve performance? Or is the simpler original code good enough? The only thing this code might save time with is creating a property. If doing that is relatively expensive you might be able to use a thread local cache for it in order to simplify the code.

tippmar-nr: We can avoid unnecessary allocations by re-using the thread id property. The property is re-created anytime there's a change in the thread id, so maybe you're right that caching by thread id would be a better solution. I'll look into that.

src/Agent/NewRelic/Agent/Core/Logging/LoggerBootstrapper.cs
```diff
@@ -272,7 +271,8 @@ private static LoggerConfiguration ConfigureRollingLogSink(this LoggerConfigurat
                 return loggerConfiguration
                     .WriteTo
                     .File(path: fileName,
-                            formatter: textFormatter,
+                            outputTemplate:outputFormat,
+                            //formatter: textFormatter,
```
jaffinito, line 275: nitpick: Stray comment

jaffinito (approved):

nr-ahemsath (approved):
Looks good.

nrcventura (approved):

PR #2088, opened 2023-11-20
chore: update linux smoke tests to only remove networks in debug build

tippmar-nr:
Updated linux smoke tests so that we don't try to prune the networks at the end of every container execution -- it looks like there are occasional concurrency issues where a network gets deleted out from under a container that's firing up.


nr-ahemsath (approved):
Looks good. 👍

PR #2095, opened 2023-11-22
tests: consolidate sync and async variations of MsSql integration tests

nr-ahemsath:
To help reduce execution time of the MsSql integration tests, this PR combines the sync and async versions of tests.  A few overall points to note:

- Sql traces are kept based on the `SqlId` which is based on the command/statement text.  In order to get traces for both sync and async operations, a different table name is used for some sync vs. async operations, and the SELECT command text is altered (query by first vs last name) for the operations against the pre-existing `teammembers` table
- Some assertions are varied depending on whether the sync or async transaction was sampled
- An unused web-based test application was deleted
- Some repeated constants of SQL text in the exercisers were refactored into a base class


nrcventura (approved):
Looks like a nice time savings of 8 minutes!

PR #2104, opened 2023-11-29
test: Refactor kafka integration tests to reduce flickers

tippmar-nr:
Refactors the Kafka integration tests so they can run in parallel and have no conflicts with the other integration tests. 

* Moves the `smoketestapp` service to a separate docker compose file, which is then referenced by each service that extends that service.
* Move the Kafka test services to a separate docker compose file. This eliminates the need for a unique network name each time these tests run and allows the `kafka-broker` service to be shared between the two tests. 
* Added an additional parameter to specify the docker-compose file to use when running `docker compose ....` during the test. 
* Refactored so that each test runs in a separate container "project", rather than starting different services in the same container. This makes the tests truly isolated from each other and simplifies cleanup of resources by allowing for a single `docker compose -p projectName down`.

All container tests passed when running locally. Hopefully they pass in CI as well.

### DIFF
--- a/tests/Agent/IntegrationTests/ContainerIntegrationTests/LinuxKafkaTests.cs
+++ b/tests/Agent/IntegrationTests/ContainerIntegrationTests/LinuxKafkaTests.cs
@@ -18,7 +18,7 @@ public abstract class LinuxKafkaTest<T> : NewRelicIntegrationTest<T> where T : L
 {
     private const int TopicNameLength = 15;
 
-    internal string _topicName;
+    private readonly string _topicName;
     private readonly T _fixture;
 
     protected LinuxKafkaTest(T fixture, ITestOutputHelper output) : base(fixture)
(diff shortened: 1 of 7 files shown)


tests/Agent/IntegrationTests/ContainerIntegrationTests/LinuxKafkaTests.cs
```diff
@@ -27,8 +27,8 @@ protected LinuxKafkaTest(T fixture, ITestOutputHelper output) : base(fixture)
         _fixture.TestLogger = output;
 
         _topicName = GenerateTopic();
-        var brokerName = "broker" + _topicName;
-        ((ContainerApplication)_fixture.RemoteApplication).DockerDependencies.Add(brokerName);
+        var brokerName = "kafka-broker";
+        //((ContainerApplication)_fixture.RemoteApplication).DockerDependencies.Add(brokerName);
```
chynesNR, line 31: Can we remove the commented out bits if we don't need them any more?

chynesNR:
Very nice!

tests/Agent/IntegrationTests/ContainerIntegrationTests/LinuxKafkaTests.cs
tippmar-nr: Yeah, I missed that one...

chynesNR (approved):

nrcventura (approved):

PR #2109, opened 2023-11-30
fix: Fix a crash that can occur when the profiler logs certain characters. (#1982)

### LINKED ISSUE #1982: Instrumenting a method with non-ASCII characters causes app crash
**Description**

If an application contains an instrumented method that has non-ascii characters (in the assembly name, namespace, type name, or method name) it can cause an application running on Linux to crash.

The problem appears to be related to the way we log things within the profiler. On Linux, if we attempt to log a message that contains a non-ascii character it causes the app to crash. On Windows, it partially writes the log message (up until the non-ascii character) and then everything continues.

Current workarounds include:

- Changing the log level so that the profiler doesn't attempt to log non-ascii characters
- Removing the non-ascii characters from the code

**Expected Behavior**

The agent should not crash and the method should be instrumented normally. The log messages should support non-ascii characters.

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/diagnostics-cli-nrdiag/diagnostics-cli-nrdiag/) results**

The issue doesn't appear if the log level is reduced to not include the problematic log lines.
The issue doesn't appear on Windows.

**Steps to Reproduce**

1. Create a new aspnetcore 7 application that is dockerized and runs in Linux container.
2. Add a controller endpoint that contains a non-ascii character (such as `\u0435`)
3. Enable finest level logs for the agent (to ensure that the profiler attempts to log the non-ascii character).
4. Send a request to the endpoint with the non-ascii character
5. Notice that the docker container terminates with a non-successful exit code (which indicates that the app crashed)

**Your Environment**

This was reproduced in a Linux docker container but it can probably be reproduced on any Linux system. This can probably also be reproduced when using the Transaction or Trace attributes (or xml instrumentation) on a method containing non-ascii characters.

*Suggested Priority (P1,P2,P3,P4):* P4
*Suggested T-Shirt size (S, M, L, XL, Unknown):* Unknown

nrcventura:

## Description

Updates the profiler log file initialization to imbue a locale with a codecvt facet so that character codes can be mapped more appropriately. 

This change was tested locally on Windows and Linux. For Linux this was manually tested using the same distros and architectures used by our Linux smoke tests.

Fixes #1982 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -408,7 +408,7 @@ jobs:
           path: src/Agent
 
       - name: Run test commands
-        uses: uraimo/run-on-arch-action@4ed76f16f09d12e83abd8a49e1ac1e5bf08784d4 # v2.5.1
+        uses: uraimo/run-on-arch-action@4803fad82c266c7860218e93d3024b5810c7866a # v2.6.0
         with:
           arch: aarch64
           distro: ubuntu18.04
--- a/.github/workflows/build_profiler.yml
+++ b/.github/workflows/build_profiler.yml
@@ -246,7 +246,7 @@ jobs:
           rm -rf ${{ github.workspace }}/src/Agent/_profilerBuild/x86-Release || true
         shell: bash
 
-      - uses: uraimo/run-on-arch-action@4ed76f16f09d12e83abd8a49e1ac1e5bf08784d4 # v2.5.1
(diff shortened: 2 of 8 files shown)


chynesNR (approved):
Good work!

tippmar-nr (commented):
Looks good!

PR #1542, opened 2023-04-13
chore: Obfuscate license keys in logs.

nr-ahemsath:
## Description

There were a couple of cases where the agent would log your entire New Relic license key value to a local log file:

- When audit logging is enabled
- When Infinite Tracing is enabled, and the agent's log level is set to FINEST

This PR fixes this by only logging the first 8 characters of the key for troubleshooting purposes, and replacing the rest with asterisks.

# Author Checklist
- [X] Unit tests, Integration tests, and Unbounded tests completed

# Reviewer Checklist
- [ ] Perform code review
- [ ] Pull request was adequately tested (new/existing tests, performance tests)

PR #2110, opened 2023-11-30
fix: obfuscate logged appSettings values

nr-ahemsath:
## Description

Apply the same obfuscation from https://github.com/newrelic/newrelic-dotnet-agent/pull/1542 to the debug-level logging of values read from appSettings.

This PR also makes the names of keys being read by the agent from appsettings constants. 🏕️ 

# Author Checklist
- [X] Unit tests, ~Integration tests, and Unbounded tests~ completed

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
@@ -5,6 +5,7 @@
 using System;
 using System.IO;
 using Microsoft.Extensions.Configuration;
+using NewRelic.Core;
 using NewRelic.Core.Logging;
 
 namespace NewRelic.Agent.Core.Configuration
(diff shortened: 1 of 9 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
```diff
@@ -79,6 +80,10 @@ public static string GetAppSetting(string key)
                 }
                 else
                 {
+                    if (key.Equals(Constants.AppSettingsLicenseKey))
+                    {
+                        value = Strings.ObfuscateLicenseKey(value);
```
nrcventura, line 85: Would it be better for us to log a similar message to the one when the license key is found in an environment variable?

nr-ahemsath: I think that logging at least part of the license key is useful for troubleshooting; it confirms to the customer that the right key is being set (unless they mangle some part of the key after the first 8 chars).

chynesNR (approved):
I like the new constants. Very nice!

src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
chynesNR: We should probably be consistent across the different ways we encounter the license key, but I think either a message or obfuscation is fine.

PR #2139, opened 2023-12-13
*DRAFT** ci: Revert linux deploy changes

tippmar-nr:
* Reverts the changes made in #2113, as we discovered that there is some sort of conflict with the aws cli when running on `debian:stable` in a container on a Github runner. 
* Modifies the `deploy_agent` workflow so that it checks out the repo and uses the files in the `deploy` folder rather than downloading those files from the build artifact uploaded from the `all_solutions` workflow. This was necessary to allow for testing changes to the linux deploy Dockerfile and scripts. 
* Removes the `deploy` folder from `deploy-artifacts`  in the `all_solutions` workflow.
* Attempts (yet again) to fix a bug in the `post_deploy_agent` workflow where we want the workflow to delay 5 minutes when called from another workflow but not when invoked manually.
* Removes the (unnecessary) `get-external-artifacts` job from the `publish_release_notes` workflow and refactors the primary job to download the artifacts instead.

### DIFF
--- a/.github/workflows/publish_release_notes.yml
+++ b/.github/workflows/publish_release_notes.yml
@@ -30,36 +30,7 @@ env:
   DOTNET_NOLOGO: true
 
 jobs:
-
-  get-external-artifacts:
-    name: Get and Publish Deploy Artifacts Locally
-    runs-on: ubuntu-latest
-    steps:
-      - name: Harden Runner
-        uses: step-security/harden-runner@eb238b55efaa70779f274895e782ed17c84f2895 # v2.6.1
-        with:
-          disable-sudo: true
-          egress-policy: audit
-
-      - name: Download Deploy Artifacts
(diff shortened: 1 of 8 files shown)


chynesNR (approved):
Thanks for tracking this down!

.github/workflows/publish_release_notes.yml
```diff
@@ -74,10 +45,14 @@ jobs:
           fetch-depth: 0
       
       - name: Download Deploy Artifacts
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        uses: dawidd6/action-download-artifact@268677152d06ba59fcec7a7f0b5d961b6ccd7e1e # v2.28.0
```
jaffinito, line 48: Was this downgrade intentional?

jaffinito (approved):

tippmar-nr:
Abandoning this PR in favor of #2142

PR #2142, opened 2023-12-13
ci: Deploy workflow fixes and other goodies

### LINKED ISSUE #2140: Update linux package deploy Dockerfile to debian:stable
A previous attempt (#2113) to update the `Dockerfile` and related scripts that are used to deploy the APT and YUM packages had to be reverted because we found that the `deploy_agent` workflow would fail even though local testing was successful. 

The workflow fails right after starting the Docker container, with
```
>>> synchronizing local copy /data/s3/production with latest changes from ***

<botocore.awsrequest.AWSRequest object at 0x7f8c02e1eed0>
Exiting.
255
Error: Process completed with exit code 255.]
```

[Some preliminary research](https://github.com/aws/aws-cli/issues/5623) suggests that there is an environment variable that can be set that will prevent that failure, but brief attempts to follow that guidance were met with failure. It may be that the environment variable wasn't set correctly or in the right location. 

We need to re-apply the changes from #2113 and test the `deploy_agent` workflow (with options set so that  only deploys APT and YUM, and deploys them to the "test" repo) and figure out how to resolve the failure.

tippmar-nr:
* Modifies `deploy.bash` to export a magic environment variable that keeps the aws cli from failing when running on a github hosted runner. Resolves #2140 
* Modifies the `deploy_agent` workflow so that it checks out the repo and uses the files in the deploy folder rather than downloading those files from the build artifact uploaded from the `all_solutions` workflow. This was necessary to allow for testing changes to the linux deploy Dockerfile and scripts.
* Removes the `deploy` folder from `deploy-artifacts` in the `all_solutions` workflow.
* Attempts (yet again) to fix a bug in the `post_deploy_agent` workflow where we want the workflow to delay 5 minutes when called from another workflow but not when invoked manually.
* Removes the (unnecessary) `get-external-artifacts` job from the `publish_release_notes` workflow and refactors the primary job to download the artifacts instead.

### DIFF
--- a/deploy/linux/deploy.bash
+++ b/deploy/linux/deploy.bash
@@ -24,6 +24,8 @@ fi
 AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-west-2}
 AWS_DEFAULT_OUTPUT=${AWS_DEFAULT_OUTPUT:-text}
 
+# a magic incantation that somehow prevents aws cli failures when running in Github. See https://github.com/aws/aws-cli/issues/5623
+AWS_EC2_METADATA_DISABLED=true
 
 ## Actions
 # 'release' => release new packages
--- a/deploy/linux/deploy.bash
+++ b/deploy/linux/deploy.bash
@@ -25,7 +25,7 @@ AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-west-2}
 AWS_DEFAULT_OUTPUT=${AWS_DEFAULT_OUTPUT:-text}
 
 # a magic incantation that somehow prevents aws cli failures when running in Github. See https://github.com/aws/aws-cli/issues/5623
-AWS_EC2_METADATA_DISABLED=true
+export AWS_EC2_METADATA_DISABLED=true
 
 ## Actions
 # 'release' => release new packages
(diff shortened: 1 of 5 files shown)


nrcventura (approved):

deploy/linux/deploy.bash
```diff
@@ -24,6 +24,8 @@ fi
 AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-west-2}
 AWS_DEFAULT_OUTPUT=${AWS_DEFAULT_OUTPUT:-text}
```
nrcventura, line 25: Should these be exported too?

chynesNR (approved):

deploy/linux/deploy.bash
tippmar-nr: I have no idea. I assume the values aren't really being used -- there's no other reference to them anywhere in our scripts. But maybe the aws cli looks for them? All I know at the moment is that the change is working and I'm not inclined to keep messing with it... ;-)

nrcventura: Those variables are standard awscli variables. If those variables were previously exported, we might not have had the problem in the first place. But, yes, I'm also fine with leaving this as is since it is working.

jaffinito (approved):
Magic LGTM!

PR #2145, opened 2023-12-14
test: Update CSP tests to look for "Conflict" instead of "Gone" in connect response body

### LINKED ISSUE #2144: Update CSP integration tests
A recent server-side change related to HSM configuration validation resulted in connect responses returning an HTTP 409 Conflict response where we previously expected an HTTP 410 Gone response, causing several of our CSP integration tests to fail.

Update the tests to look for "Conflict" instead of "Gone" and verify that all CSP tests pass now.

tippmar-nr:
Resolves #2144


nrcventura (approved):

chynesNR (approved):

nr-ahemsath (approved):
👍

PR #2157, opened 2023-12-20
fix: Path to agentinfo.json was built incorrectly, leading to file not found errors when running on Linux. (#2156)

### LINKED ISSUE #2156: Agent does not report install type correctly when running on Linux
**Description**
At startup, the agent tries to read from the `agentinfo.json` file in the New Relic "home" folder. If the file does not exist or can't be read, the agent reports `["Install Type", "Unknown"]` in the connect request. 

When running on Linux, the connect request always reports "Unknown" even though an `agentinfo.json` file exists in the New Relic "home" folder.

**Expected Behavior**
The correct install type, read from `agentinfo.json` should be reported when running on Linux.

P3
S

tippmar-nr:
Update the hard-coded pathing in `GetAgentInfo()` to use `Path.Combine()` instead for Linux compatibility. 

Fixes #2156.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
@@ -217,11 +217,11 @@ namespace NewRelic.Agent.Core
                 catch (Exception e)
                 {
                     Log.Debug(e, $"Could not deserialize agent info from {agentInfoPath}.");
-                    return null;
                 }
             }
+            else
+                Log.Debug($"Could not get agent info from {agentInfoPath}. File does not exist.");
 
-            Log.Debug($"Could not get agent info from {agentInfoPath}. File does not exist.");
             return null;
         }
     }
--- a/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
@@ -217,6 +217,7 @@ namespace NewRelic.Agent.Core
                 catch (Exception e)
                 {
                     Log.Debug(e, $"Could not deserialize agent info from {agentInfoPath}.");
+                    return null;
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
```diff
@@ -200,19 +200,21 @@ private static string GetNewRelicLogLevel()
...
             {
                 try
                 {
                     return JsonConvert.DeserializeObject<AgentInfo>(File.ReadAllText(agentInfoPath));
                 }
-                catch
+                catch (Exception e)
                 {
-                    // Fail silently
+                    Log.Debug(e, $"Could not get agent info from {agentInfoPath}.");
```
tippmar-nr, line 213:
```suggestion
                    Log.Debug(e, $"Could not deserialize agent info from {agentInfoPath}.");
```

chynesNR (approved):
Good catch!

src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
```diff
@@ -200,19 +200,21 @@ private static string GetNewRelicLogLevel()
...
                 }
-                catch
+                catch (Exception e)
                 {
-                    // Fail silently
+                    Log.Debug(e, $"Could not deserialize agent info from {agentInfoPath}.");
                 }
             }
 
+            Log.Debug($"Could not get agent info from {agentInfoPath}. File does not exist.");
```
nr-ahemsath, line 217: Won't we end up logging this if the file does exist but there was a deserialization error?  Should it be in an `else` clause for the `if (File.Exists)...`?

tippmar-nr: Yep, good catch. I've pushed a change.

nr-ahemsath (approved):
LGTM! 👍

PR #2159, opened 2023-12-21
ci: Configure CodeQL Workflow, Unify <langVersion>

tippmar-nr:
This PR introduces a new workflow to run CodeQL, superseding the "default" configuration that was recently enabled for all repos in the New Relic org. 

~~For the time being, we will only scan C# projects, as we can (apparently) have only a single CodeQL workflow per repo, and our C++ projects have to build on Windows 2019, while our C# projects have to build on Windows 2022.~~
I was able to get the scan to run on both C# and C++. Yay! 
Results for the analysis on this PR are at: https://github.com/newrelic/newrelic-dotnet-agent/security/code-scanning?query=tool%3ACodeQL+pr%3A2159 -- there doesn't seem to be anything for C++, so I guess that code is solid. The other items reported for C# are either used in tests or false positives.

The other (somewhat huge) set of changes in this PR is to eliminate explicit use of `<langVersion>` attributes in .csproj files (which specified wildly different versions), in favor of a single, unified `<langVersion>default</langVersion>` in the base `Directory.Build.props` file. Aside from ensuring that all projects build with the correct language version (based on the tfm), this also helps alleviate some issues with the CodeQL scanning tool, as it was unhappy about using certain hard-coded language versions.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -44,6 +44,11 @@ jobs:
       uses: github/codeql-action/init@v3
       with:
         languages: ${{ matrix.language }}
+        config: |
+          paths: 
+            - src/Agent/NewRelic
+            - src/Agent/NewRelic.Api.Agent
+            - src/NewRelic.Core
 
     - name: Build .NET Agent Solution
       if: ${{ matrix.language == 'csharp' }}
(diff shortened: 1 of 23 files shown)


chynesNR (approved):

tippmar-nr:
Holding this PR until @nrcventura has a chance to review and chime in on the `<langVersion>` changes.

nr-ahemsath (approved):
Looks good.  I like standardizing on the default language per target framework version. 👍

PR #2164, opened 2024-01-02
ci: Update to artifacts v4

nr-ahemsath:
## Description

Fix the main agent CI workflow after merging breaking artifacts changes (v3->v4) in Dependabot PRs.  The main issue is that you can't upload multiple artifacts to the same artifact name.  This is only really a problem for the matrixed integration test jobs.  The fix is to use a unique artifact name based on the matrix namespace value.  Additionally, the workflow is modified to only archive the test results on failure since it's highly unlikely anybody wants to look at them if everything passes.

An additional breaking change from artifacts v3->v4 is a limit on the number of artifacts that can be uploaded per job (10).  This means that if more than ten namespaces have integration tests failures, we would have failures uploading some of the test working directories and results files.  However, nothing downstream of the integration test jobs need those artifacts, and if more than ten namespaces are failing we probably have bigger problems to deal with.

As a side note, these changes do seem to improve the speed of the workflow by a significant amount.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -502,7 +502,7 @@ jobs:
           path: src/Agent
 
       - name: Download Unbounded Integration Test Artifacts
-        uses: actions/download-artifact/@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        uses: actions/download-artifact@f44cd7b40bfd40b6aa1cc1b9b5b7bf03d3c67110 # v4.1.0
         with:
           name: unboundedintegrationtests
           # Should not need a path because the integration test artifacts are archived with the full directory structure
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -364,19 +364,19 @@ jobs:
         if: ${{ failure() }}
         uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
         with:
-          name: integration-test-artifacts-${{ matrix.namespace }}
+          name: integration-test-working-directory-${{ matrix.namespace }}
           path: |
             C:\IntegrationTestWorkingDirectory\**\*.log
             C:\IntegrationTestWorkingDirectory\**\*.config
           if-no-files-found: error
 
-      # - name: Archive Test Artifacts
-      #   if: ${{ always() }}
-      #   uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
-      #   with:
-      #     name: integration-test-artifacts
-      #     path: C:\IntegrationTestWorkingDirectory\TestResults\**\*TestResults.trx
-      #     if-no-files-found: error
+      - name: Archive Test Artifacts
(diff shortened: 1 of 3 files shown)


chynesNR:
I had originally thought we'd remove our use of `action-download-artifact` at the same time, but since that's mostly used in our deploy workflows, should we pull that out into a separate issue?

.github/workflows/all_solutions.yml
```diff
@@ -364,17 +364,17 @@ jobs:
...
         uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
         with:
-          name: integration-test-artifacts
+          name: integration-test-working-directory-${{ matrix.namespace }}
           path: |
             C:\IntegrationTestWorkingDirectory\**\*.log
             C:\IntegrationTestWorkingDirectory\**\*.config
           if-no-files-found: error
 
       - name: Archive Test Artifacts
```
tippmar-nr, line 373: How about we combine the artifact upload in the previous step with this one and just upload a single artifact per test matrix namespace?

.github/workflows/all_solutions.yml
```diff
@@ -441,15 +441,15 @@ jobs:
         if: ${{ failure() }}
         uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
         with:
-          name: integration-test-artifacts-linux-arm64
+          name: integration-test-working-directory-linux-arm64
           path: ${{ env.test_path }}/**/*
           if-no-files-found: error
 
-      - name: Archive Test Artifacts
+      - name: Archive Test Results on failure
```
tippmar-nr, line 448: Same here

.github/workflows/all_solutions.yml
```diff
@@ -576,17 +576,17 @@ jobs:
...
-          name: integration-test-artifacts
+          name: unbounded-test-working-directory-${{ matrix.namespace }}
           path: |
             C:\IntegrationTestWorkingDirectory\**\*.log
             C:\IntegrationTestWorkingDirectory\**\*.config
           if-no-files-found: error
 
-      - name: Archive Test Artifacts
-        if: ${{ always() }}
+      - name: Archive Test Results on failure
```
tippmar-nr, line 585: ... and here

.github/workflows/all_solutions.yml
nr-ahemsath: You're reading my mind, I had the same thought a few minutes ago while looking at something else.  I'll give it shot.

nr-ahemsath:
> I had originally thought we'd remove our use of `action-download-artifact` at the same time, but since that's mostly used in our deploy workflows, should we pull that out into a separate issue?



> I had originally thought we'd remove our use of `action-download-artifact` at the same time, but since that's mostly used in our deploy workflows, should we pull that out into a separate issue?

I was going to do that as a separate PR for the deploy workflows, but I am going to make sure we can make it work for the standalone run-integration-tests workflow as part of this one.

tippmar-nr (approved):

chynesNR (approved):

PR #2165, opened 2024-01-02
ci: use official download-artifact everywhere

nr-ahemsath:
## Description

Replace use of `dawidd6/action-download-artifact` with the GitHub provided version.  We previously used the third-party action because it supported downloading artifacts from other workflows, but the "official" action supports this now.

Some additional changes:

- In the Lambda tracer and Azure Site Extension deploy workflows, I eliminated the "download external artifacts, then upload locally, then redownload from local" complication.  I tested these workflows (by running them with the `deploy` input set to false so they wouldn't actually deploy) and they worked as expected as far as I can tell.
- Some tweaks to the action permissions (I'm not sure why the third-party action wanted to be able to read commit hashes, but it's not necessary for the official one - again, tested in the Lambda and site extension deploy workflows)
- Deleted the `get_release_checksums.yml` workflow - this was unused, the `ReleaseNotesBuilder` code gets the checksums directly from the build artifacts in the release notes workflow.

I was able to successfully test the changes to the `run_linux_container_tests.yml` workflow here: https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/7391280783/job/20107780705


.github/workflows/deploy_awslambda.yml
```diff
@@ -21,8 +21,8 @@ jobs:
 
   get-external-artifacts:
     permissions:
-      actions: read  # for dawidd6/action-download-artifact to query and download artifacts
-      pull-requests: read  # for dawidd6/action-download-artifact to query commit hash
+      actions: read  # for download-artifact to query and download artifacts (unsure if needed since moving from dawidd6/actions-download-artifact?)
```
chynesNR, line 24: Could we test this somewhere else so we can know for sure?

nr-ahemsath: Looking closer at that workflow it does have a test mode, so I can test it out.  I think it only needs `actions: read`.  I'm also going to try and eliminate the "download external and upload locally, then redownload" business.

chynesNR (approved):
Looks good!

PR #2167, opened 2024-01-03
test: Fix flicker in SpanStreamingServiceTests

tippmar-nr:
Multiple threads were trying to add to the (unsynchronized) `List<T>`. A `ConcurrentBag<T>` fixed that right smart quick...


nr-ahemsath (approved):
Good catch!

chynesNR (approved):
Good catch!

PR #2184, opened 2024-01-11
fix: Eliminate intermittent deadlock on agent startup. (#2183)

### LINKED ISSUE #2183: Agent freezes customer application intermittently at startup
**Description**
In certain circumstances, a .NET 6+ app can deadlock at startup when an exception is thrown while attempting to connect to New Relic. 

Core dump analysis has shown that agent initialization starts in the static `AgentShim()` constructor. If an exception is thrown, the exception handling logic attempts to make a GET request to https://google.com as a way to determine if there is a networking issue or some other problem. But when the GET request is made (via `HttpClient.GetAsync()`) on a new thread, the profiler attempts to instrument that call but is blocked because the static `AgentShim()` constructor has not completed.

This issue only affects the .NET Standard build target of the .NET agent; the .NET Framework build target performs a "deferred" initialization, based on specific methods called during initialization that would normally be instrumented. We can leverage that solution in the .NET Standard build target instead of the static constructor to avoid this problem.

**Expected Behavior**
The agent should never cause a customer application to freeze

**Steps to Reproduce**
1. Create a very simple .NET 6+ Console App and add the `NewRelic.Agent.Api` Nuget package. 
```
    static class Program
    {
        [Transaction(Web = false)]
        static void Main()
        {
            Console.WriteLine("Main enter");

            var stopWatch = new Stopwatch();
            stopWatch.Start();

            NewRelic.Api.Agent.NewRelic.SetTransactionName("category", "name");

            stopWatch.Stop();

            Console.WriteLine("Main exit");
        }
    }
```
2. Configure the .NET agent to use an invalid port when connecting to New Relic:
`NEW_RELIC_PORT=1234`
3. Enable profiling and start the application. The application will hang, and the last INFO-level line in the profile logfile will read 
```
[Info ] 2024-01-10 22:50:29 Instrumenting method: (Module: C:\Program Files\dotnet\shared\Microsoft.NETCore.App\6.0.26\System.Net.Http.dll, AppDomain: clrhost)[System.Net.Http]System.Net.Http.SocketsHttpHandler.SendAsync(System.Net.Http.HttpRequestMessage,System.Threading.CancellationToken)
```
P2
S

tippmar-nr:

## Description
This PR changes how the agent is initialized when running on the .NET Standard 2.0 build target so that it uses the same "deferred" initialization pattern that the .NET Framework build target currently uses. This will eliminate an intermittent deadlock that has been observed at agent startup which has the unfortunate side effect of freezing the customer application.

A new integration test was added that verifies the agent does not deadlock. This test was run against the agent from the current `main` branch and failed; the same test was run against this PR branch and succeeded.

Fixes #2183

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentShim.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentShim.cs
@@ -20,12 +20,6 @@ namespace NewRelic.Agent.Core
             AgentInitializer.InitializeAgent();
         }
 
-#if NETSTANDARD2_0
-		static AgentShim()
-		{
-			Initialize();
-		}
-#else
         private static bool _initialized = false;
         private static object _initLock = new object();
 
@@ -49,6 +43,7 @@ namespace NewRelic.Agent.Core
         private static HashSet<string> _deferInitializationOnTheseMethods = new HashSet<string>
         {
             "System.Net.Http.HttpClient.SendAsync",
+            "System.Net.Http.SocketsHttpHandler.SendAsync",
             "System.Net.HttpWebRequest.SerializeHeaders",
             "System.Net.HttpWebRequest.GetResponse"
         };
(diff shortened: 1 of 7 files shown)


nr-ahemsath (approved):
Based on our team discussion yesterday, and the addition of the test, this looks good to me. 🚢

nrcventura (approved):

PR #2186, opened 2024-01-12
fix: Prevent multiple SessionCache instances from being created. (#2180)

### LINKED ISSUE #2180: StackExchangeRedis2PlusWrapper can create multiple SessionCache instances
**Description**
Recent customer-supplied memory dumps have revealed that the agent is holding 2 instances of `SessionCache` when we only intend for there to be a single instance. This would appear to be a race condition when `StackExchange.Redis.ConnectionMultiplexer.CreateMultiplexer` is called multiple times in quick succession, or it may be related to an agent restart. 

In any case, we should consider adding some locking in the `RegisterProfiler()` callback in `ConnectWrapperV2.cs`

**Expected Behavior**
The agent should use a single instance of the `SessionCache` 

P3
S

tippmar-nr:

## Description
Prevents an edge case where multiple `SessionCache` instances could get created if `StackExchange.Redis.ConnectionMultiplexer.CreateMultiplexer()` is called multiple times. 

Improves resiliency in `SessionCache.Cleanup()` and adds logging.

Fixes #2180 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 New Relic, Inc. All rights reserved.
+// Copyright 2020 New Relic, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
@@ -10,6 +10,7 @@ using NewRelic.Agent.Api.Experimental;
 using NewRelic.Agent.Extensions.Helpers;
 using NewRelic.Agent.Extensions.Parsing;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
+using NewRelic.Core.Logging;
 using NewRelic.Parsing.ConnectionString;
 using StackExchange.Redis.Profiling;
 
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/StackExchangeRedis2Plus/SessionCache.cs
```diff
@@ -107,10 +108,11 @@ private void CleanUp()
                         }
                     }
                 }
+
+                Log.Finest($"SessionCache.Cleanup() removed {cleanedSessions}");
```
chynesNR, line 112: Can we do an `if (cleanedSessions > 0)` here to cut down on noise? Or are you concerned that we'll miss the thread ending?

tippmar-nr: I was kind of looking for that message to be a heartbeat for the thread, but I can go either way with it.

chynesNR: Once a minute is probably fine.

chynesNR (approved):
I don't think this will hurt anything, but I also suspect this won't address the issue we're seeing.

nrcventura (approved):

PR #2194, opened 2024-01-16
chore(deps): Update NuGet packages for Core, etc.

tippmar-nr:
Updates NuGet packages for the Core project, as well as unit test projects where necessary. Doesn't change any package references for instrumentation wrappers.

Added new container integration tests for infinite tracing across a subset of Linux distros, which will hopefully catch any issues with updates to `Google.Protobuf` and/or `Grpc.Tools` packages.

Also added Dependabot configuration to automate scanning for NuGet updates in selected projects.


chynesNR (approved):
gRPC and Protobuf updates could in theory affect Infinite Tracing behavior, but as long as those tests still pass we should be ok.

src/Agent/NewRelic/Agent/Core/Core.csproj
```diff
@@ -33,21 +33,25 @@
...
     </PackageReference>
-    <PackageReference Include="Serilog" Version="3.0.1" />
+    <PackageReference Include="Serilog" Version="3.1.1" />
     <PackageReference Include="Serilog.Sinks.Async" Version="1.5.0" />
-    <PackageReference Include="Serilog.Sinks.Console" Version="4.1.0" />
+    <PackageReference Include="Serilog.Sinks.Console" Version="5.0.1" />
     <PackageReference Include="Serilog.Sinks.Debug" Version="2.0.0" />
     <PackageReference Include="Serilog.Sinks.EventLog" Version="3.1.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="8.0.0" />
```
nrcventura, line 47: What required this dependency? I think that ILRepacking this dependency may have some unexpected side effects. This will result in the possibility of 2 different versions of diagnostic source being loaded into an app, which is often what we want. However, if there were assumptions about that instance of diagnostics source getting information from the rest of the application, it will not work. For example, Activity.Current will potentially have 2 different values because they are essentially different types because of the ILRepacking.

src/Agent/NewRelic/Agent/Core/Core.csproj
```diff
@@ -33,21 +33,25 @@
 
   <ItemGroup>
     <PackageReference Include="Autofac" Version="4.5.0" />
-    <PackageReference Include="Google.Protobuf" Version="3.21.12" />
-    <PackageReference Include="Grpc.Tools" Version="2.28.1">
+    <PackageReference Include="Google.Protobuf" Version="3.25.2" />
+    <PackageReference Include="Grpc.Tools" Version="2.60.0">
```
nrcventura, line 37: This is a very large jump, and I don't know if our integration test coverage is adequate for it.

src/Agent/NewRelic/Agent/Core/Core.csproj
tippmar-nr: Adding that dependency resolved a failing integration test around W3C TraceID tests -- it was specifically because `Activity.Current` had an unexpected value in the test. Once the package reference and ILRepack was added, the test passes.

nrcventura: We should take a closer look at it during a team pairing session.

tippmar-nr: Looks like the need for `DiagnosticSource` is driven by a [dependency on it from Serilog v3.1.1](https://www.nuget.org/packages/Serilog/3.1.1), for .NET Framework targets. So maybe I can scope it down to just the Framework build and (maybe) not need to ILRepack it?

tippmar-nr: For now, I'm reverting this change and the Serilog change.

src/Agent/NewRelic/Agent/Core/Core.csproj
```diff
@@ -105,14 +109,15 @@
       <ILRepackInclude Include="@(PossibleRefsForILRepack)" Condition="'%(FileName)' == 'Serilog.Sinks.Debug'" />
       <ILRepackInclude Include="@(PossibleRefsForILRepack)" Condition="'%(FileName)' == 'Serilog.Sinks.EventLog'" />
       <ILRepackInclude Include="@(PossibleRefsForILRepack)" Condition="'%(FileName)' == 'Serilog.Sinks.File'" />
+      <ILRepackInclude Include="@(PossibleRefsForILRepack)" Condition="'%(FileName)' == 'System.Memory'" />
```
nrcventura, line 112: Is this now required for the netstandard2.0 target? I thought it was part of .net and only required for .net framework?

tippmar-nr: Integration tests pass after reverting.

jaffinito (approved):

src/Agent/NewRelic/Agent/Core/Core.csproj
tippmar-nr: Added infinite tracing tests to the container integration tests to verify protobuf is working across Linux distros.

tests/Agent/IntegrationTests/ContainerIntegrationTests/Tests/LinuxContainerTests.cs
```diff
@@ -42,9 +58,20 @@ protected LinuxContainerTest(T fixture, ITestOutputHelper output) : base(fixture
...
+
+        // verify 8T metrics
+        var expectedSeenCount = 4;
+
+        var expectedMetrics = new List<Assertions.ExpectedMetric>
+        {
+            new Assertions.ExpectedMetric { metricName = @"Supportability/InfiniteTracing/Span/Seen", callCount = expectedSeenCount },
+            new Assertions.ExpectedMetric { metricName = @"Supportability/InfiniteTracing/Span/Sent", callCount = ExpectedSentCount },
+        };
+
```
nrcventura, line 73: Should these new 8T tests be a separate set of tests instead of using 8T as our default set of tests for Linux?

tippmar-nr: I'm not sure it makes much difference either way - I would kind of expect to see the container integration tests evolve over time to include specific things we want to test on Linux, and adding separate tests for every one of them would explode the number of containers that have to spin up and run.

tippmar-nr: (maybe I wasn't clear, but for tests we want to run across all distros, it seems to make more sense to put them all in a single test class rather than separate - that would lead to the explosion of containers if we have multiple tests running on all distros...)

nrcventura: My main concern about including 8T as the default in these tests, is that most customers do not use 8T, and it results in us not testing normal span events.

tippmar-nr: I can certainly split these out to separate tests, but it's going to essentially double the runtime for the container tests.

jaffinito: Given that 8T is a major feature, it might be good to test more thoroughly and often.

nrcventura (approved):

chynesNR (approved):

PR #2227, opened 2024-02-02
feat: Report ignored instrumentation usage

nrcventura:

## Description

Reports the configured ignored instrumentation with the agent settings and generates a supportability metric to when any ignored instrumentation is configured.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -828,5 +829,13 @@ private void ReportIfLoggingDisabled()
                 ReportSupportabilityCountMetric(MetricNames.SupportabilityLoggingFatalError);
             }
         }
+
+        private void ReportIfInstrumentationIsDisabled()
+        {
+            if (_configuration.IgnoredInstrumentation.Any())
+            {
+                ReportSupportabilityCountMetric(MetricNames.SupportabilityIgnoredInstrumentation);
```
chynesNR, line 837: Do you think it would be more valuable to report the number of items being ignored?

chynesNR (approved):
Looking good!

src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
nrcventura: I considered doing that, but I felt that information was not as valuable as knowing that the feature was used. I think the count was less informative because having 5 elements that specify both an assembly and class name make affect the same amount of things as 1 item that just lists the assembly.

I think one of the useful scenarios that this enables (without the count) is that when we are troubleshooting problems this supportability metric will let us know that we should look at the environment values so that we can see which instrumentations were configured to be ignored. 

I was less certain that capturing the count would help us with answering questions. For example, if we captured the count we might be able to see how many elements are typically specified, which can help with validating our assumptions about how many things will typically be configured, but I don't think that knowing that information would change our code structure. It might be more interesting to know which assemblies are typically ignored, but doing that would result in many more metrics being generated, and it would be harder to track in angler.

chynesNR: I agree that it probably wouldn't make troubleshooting easier, I was thinking more about insight as to how customers are using it. Are they disabling problem points, or disabling everything except what they absolutely need? But It's still probably a minor value add, so I'm fine either way.

nrcventura: I thought of a potential use case for having the count available. I can help us and customers detect changes to this configuration (or at least when the amount of items configured changed). I'll make the change.

PR #2244, opened 2024-02-14
chore: Change debug type to "embedded" instead of "PDB"

tippmar-nr:
This change is to enabled embedded PDBs in our debug builds of ILRepacked assemblies, to facilitate local debugging. 

This is intended to be a temporary change, pending resolution of an [issue where the DLL is written with an incorrect path to the .PDB file](https://github.com/gluck/il-repack/issues/352). 

There is no change to the release build at all.

### DIFF
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/SharedApplicationHelpers/ReflectionHelpers/ReflectionUtil.cs
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/SharedApplicationHelpers/ReflectionHelpers/ReflectionUtil.cs
@@ -79,7 +79,18 @@ namespace NewRelic.Agent.IntegrationTests.Shared.ReflectionHelpers
 
             foreach (var assembly in assemblies)
             {
-                foreach (var type in assembly.GetTypes())
+                Type[] types;
+                try
+                {
+                    types = assembly.GetTypes();
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine($"Exception calling .GetTypes on assembly {assembly.FullName}");
+                    throw;
+                }
+
+                foreach (var type in types)
                 {
                     if (type.IsDefined(attribType, true))
                     {
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/SharedApplicationHelpers/ReflectionHelpers/ReflectionUtil.cs
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/SharedApplicationHelpers/ReflectionHelpers/ReflectionUtil.cs
@@ -84,7 +84,7 @@ namespace NewRelic.Agent.IntegrationTests.Shared.ReflectionHelpers
                 {
                     types = assembly.GetTypes();
                 }
-                catch (Exception e)
+                catch (Exception)
                 {
                     Console.WriteLine($"Exception calling .GetTypes on assembly {assembly.FullName}");
                     throw;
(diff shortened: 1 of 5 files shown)


chynesNR (approved):

nr-ahemsath (approved):
Good job figuring this out and filing the issue in their repo. 💯

tippmar-nr:
> Good job figuring this out and filing the issue in their repo. 💯

@nr-ahemsath it was actually @chynesNR who figured out the pathing problem. I just did the paperwork... ;-)

tests/Agent/IntegrationTests/SharedApplications/Common/SharedApplicationHelpers/ReflectionHelpers/ReflectionUtil.cs
```diff
@@ -79,7 +79,18 @@ public static List<Type> GetClassesByType<T>(Assembly assembly)
...
-                foreach (var type in assembly.GetTypes())
+                Type[] types;
+                try
+                {
+                    types = assembly.GetTypes();
+                }
+                catch (Exception)
+                {
+                    Console.WriteLine($"Exception calling .GetTypes on assembly {assembly.FullName}");
+                    throw;
```
chynesNR, line 90: Don't we want to continue if we hit a bad assembly? Or do you want it to fail out?

nrcventura:
Are we sure that we want to do this? We found a workaround of manually adding the build folders as a symbol location. We don't know when or if a fix will be delivered. I think that I'd rather keep our dependencies more up-to-date.

tests/Agent/IntegrationTests/SharedApplications/Common/SharedApplicationHelpers/ReflectionHelpers/ReflectionUtil.cs
tippmar-nr: I'd rather have it die, as it's a sign of a bigger problem.

tippmar-nr:
> Are we sure that we want to do this? We found a workaround of manually adding the build folders as a symbol location. We don't know when or if a fix will be delivered. I think that I'd rather keep our dependencies more up-to-date.

@nrcventura I think I've found a better way to deal with the issue - the author of ILRepack [acknowledged the bug and suggests a workaround](https://github.com/gluck/il-repack/issues/352#issuecomment-1945267824) of setting `<DebugType>` to `embedded` for our ILRepack'ed assemblies, which will avoid the issue with finding PDBs entirely. 

I'm testing that now but it looks promising.

nrcventura (approved):

chynesNR (approved):

KirillOsenkov:
I recommend that you always use embedded pdbs, even in Release. There are numerous benefits and no known downsides. You'll get line numbers in exception stack traces in production, and never have to worry about locating the pdb files (which might be in a different location).

The problem with https://github.com/gluck/il-repack/issues/352 is that even if ILRepack fixes it somehow, the path to the pdb embedded in the dll is just a hint to the debugger anyway. It won't work if you're on a different machine, or if you otherwise move the dll and the pdb from the original location.

KirillOsenkov:
Can you catch me up on what the actual problem was? The VS debugger should be able to pick up the .pdb if it's next to the .dll. Were you not copying the .pdb next to the .dll? ILRepack copies both the .dll and the .pdb from the temporary folder that gets deleted into the output directory.

chynesNR:
Hi @KirillOsenkov  -- thanks for checking in. The issue was with local debugging. We build our binaries, then do an ILRepack, then have a copy step that places the DLLs in a central location (without the PDBs). When attaching Visual Studio to a process using those libraries, the hint path was wrong, pointing to a directory that didn't exist, and we had to manually locate the PDBs in the ILRepack folder. Certainly not the end of the world, but it cost us some time that day as we tried to figure out what had changed, and discovered it was ILRepack. And since the hint path was pointing to a directory that didn't exist, and everything was run locally, it felt like a bug.

KirillOsenkov:
Makes sense! Yes, just embed pdbs always and this class of issues will go away.

PR #2250, opened 2024-02-16
feat: Allows instrumentation to be disabled from a newrelic.config file.

nrcventura:

## Description

Allows customers to control which assemblies and optionally classes should not be instrumented by specifying instrumentation ignore rules within a newrelic.config file. 

Previous techniques required customers to directly remove or modify the installed instrumentation xml files. This new method should make it easier for these customers to upgrade the agent and keep certain instrumentations disables. It also enables the ability to have instrumentations disabled on a per application basis when a newrelic.config file is deployed with each application.

If you do not want a particular assembly instrumented you can add the following snippet to the [instrumentation element in the newrelic.config file](https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#instrumentation-element).

```
<rules>
  <ignore assemblyName="NameOfAssemblyToIgnore" />
</ rules>
```

The `ignore` rule instructs the profiler to not instrument any methods defined in the specified assembly. Methods defined in other assemblies will still be instrumented.

The `ignore` rule allows you to optionally define a class name. In the following example, only methods defined in `MyNamespace.MyClass` in the assembly `MyAssembly` will be ignored. Other methods in other classes both inside that assembly and within other assemblies will not be ignored by the profiler.

```
<rules>
  <ignore assemblyName="MyAssembly" className="MyNamespace.MyClass" />
</ rules>
```

More than one ignore rule can be specified. The following example disables both the Confluent Kafka and StackExchange Redis instrumentations.

```
<rules>
  <ignore assemblyName="Confluent.Kafka" />
  <ignore assemblyName="StackExchange.Redis" />
</ rules>
```

CAVEAT - Custom instrumentation added by using the `Transaction` or `Trace` attributes are not affected by this new configuration option. It only affects xml based custom instrumentation or the default instrumentation packaged with the .net agent.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -828,5 +829,14 @@ private void ReportIfLoggingDisabled()
                 ReportSupportabilityCountMetric(MetricNames.SupportabilityLoggingFatalError);
             }
         }
+
+        private void ReportIfInstrumentationIsDisabled()
+        {
+            var ignoredCount = _configuration.IgnoredInstrumentation.Count();
+            if (ignoredCount > 0)
+            {
+                ReportSupportabilityGaugeMetric(MetricNames.SupportabilityIgnoredInstrumentation, ignoredCount);
```
tippmar-nr, line 838: So we're only reporting a count of ignored instrumentation, not a list of what's actually being ignored? I'm wondering if it would be helpful to report the details as supportability metrics?

chynesNR: That would definitely be helpful, but it's not possible with the current Supportability metrics. We would need to create a separate metric for each assembly and class combination.

chynesNR (approved):
Looks good!

tippmar-nr (approved):
Nicely done!

PR #2259, opened 2024-02-21
ci: consolidate test secrets

nr-ahemsath:
Use the same secrets in both kinds of integration tests.

Test run: https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/7993740794


tippmar-nr (approved):
Thanks for digging into this!

chynesNR (approved):

PR #2262, opened 2024-02-22
test: Additional unit tests for the Transaction class, courtesy of Github Copilot.

tippmar-nr:
The results of playing with Github Copilot for about an hour. Increases unit test code coverage significantly in `Transaction.cs` 

It started out pretty dumb, but the more I used it, the smarter it got - even recognizing code changes I made in one unit test an using the same construct in a unit test it generated later. There's some promise to this tool, I think....

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
+++ b/src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
@@ -892,11 +892,6 @@ namespace NewRelic.Agent.Core.Transactions
 
         public void SetQueueTime(TimeSpan queueTime)
         {
-            if (queueTime == null)
-            {
-                throw new ArgumentNullException(nameof(queueTime));
-            }
-
             TransactionMetadata.SetQueueTime(queueTime);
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -892,11 +892,6 @@ public void SetReferrerUri(string uri)
 
         public void SetQueueTime(TimeSpan queueTime)
         {
-            if (queueTime == null)
```
chynesNR, line 895: We're sure this can't be null?

tippmar-nr: Yep. `TimeSpan` is a struct and can never be null. The old code compiled because `TimeSpan` has an `operator ==()` overload. But trying to set a `TimeSpan` to `null` won't compile. 
image

jaffinito:
Very cool!

chynesNR (approved):

PR #2263, opened 2024-02-22
fix: Use original regex to get container id from cgroup v1 in fallback case

### LINKED ISSUE #2256: Fall back to less restrictive Container ID check if not found
As part of [this PR](https://github.com/newrelic/newrelic-dotnet-agent/pull/1943/files#diff-7e62a15b40d1fcece5dd76bbe2c4a835a51c4345785273cf046ed060452c99e0R27) we changed how we pull the the Container ID from cgroups in order to support cgroup v2 and to cut down on false positives. We have a customer where the old Agent (10.17 and below) works for them, but 10.18 and above does not.

We should have a fallback mechanism where if we can't find the Container ID with the existing regex, we should fall back to the "any 64 character alphanumeric string" regex. However, it's important that we _only_ do this for cgroup v1. Attempting to parse `mountinfo` this way will result in false positives.

nr-ahemsath:
Fixes #2256 by adding additional logic to try and parse the container ID out of cgroup v1 data if the newer regex fails to find anything.  Includes a new unit test to verify the intended result.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
+++ b/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
@@ -25,7 +25,7 @@ namespace NewRelic.Agent.Core.Utilization
         private const string ValidateMetadataRegex = @"^[a-zA-Z0-9-_. /]*$";
 #if NETSTANDARD2_0
         private const string ContainerIdV1Regex = @".*:cpu:/docker/([0-9a-f]{64}).*";
-        private const string ContainerIdV1FallbackRegex = @"([0-9a-f]{64})"; // This is the old regex that just looks for any 64-char alphanumeric string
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
```diff
@@ -25,6 +25,7 @@ public class VendorInfo
         private const string ValidateMetadataRegex = @"^[a-zA-Z0-9-_. /]*$";
 #if NETSTANDARD2_0
         private const string ContainerIdV1Regex = @".*:cpu:/docker/([0-9a-f]{64}).*";
+        private const string ContainerIdV1FallbackRegex = @"([0-9a-f]{64})"; // This is the old regex that just looks for any 64-char alphanumeric string
```
tippmar-nr, line 28: Minor nit in the comment: It's actually looking for a 64-char hexadecimal string... ;-)

tippmar-nr (approved):
Looks good!

chynesNR (approved):

PR #2266, opened 2024-02-26
Added ServerlessModeEnabled to config and refactored as needed.

tippmar-nr:
Added a `serverlessModeEnabled` attribute to the `<configuration>` element in the XSD, regenerated code and then added a `ServerlessModeEnabled` property to `IConfiguration` and implemented logic as specified in the Lambda spec for determining whether the agent is running in serverless mode or not. 

Refactored a number of `TODO`'s related to behavior specific to serverless mode, including a minor refactor of the agent initialization process since we need access to the `ServerlessModeEnabled` property prior to registering services. in the IoC container.

Added unit tests for serverlessMode configuration


chynesNR (approved):
Looks good!

src/Agent/NewRelic/Agent/Core/Config/configurationPartial.cs
nrcventura, line 1: Previously, all of our configuration logic was contained in DefaultConfiguration. Now, we have some of it in this partial class. Is that the direction we want to take, or should we find a way to be able to use the DefaultConfiguration class during the bootstrap process?

tippmar-nr: That's a fair question. It seems really difficult to get access to the `IConfigurationService` instance before registering everything else (which is where we need to know whether we're in Serverless mode or not). 

I experimented (briefly) with doing a 2-part container registration where I would register enough services to get `IConfigurationService`, figure out if we're in Serverless mode, and the register the remaining services as appropriate.
 But because we've abstracted the use of AutoFac, it got too difficult to manage in the time I was going to spend working on it. (I'd be totally in favor of refactoring to get rid of the `IContainer` abstraction and just use AutoFac directly...)

nrcventura: That seems too complex. I can be ok with having the two locations, especially since one is clearly marked for bootstrapping.

Exploring removing the IContainer may be good thing to explore separately. I'm still questioning whether the container makes things easier or harder to maintain in the agent.

src/Agent/NewRelic/Agent/Core/Config/configurationPartial.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    // The configuration class is partial.  Part of it is implemented here,
+    // and another part is autogenerated from Configuration.xsd into Configuration.cs.
+    //
+    // Beware the tarpit surrounding the case of property names, as different cases reflect
+    // different origins of the property names.
+    //
+    // Property names such as "agentEnabled" come to us from Configuration.xsd via Configuration.cs.
+    // Property names such as "AgentEnabled" are added in here or inherited from BootstrapConfig.
+    public partial class configuration : IBootstrapConfig
```
nrcventura, line 18: What if this wasn't a partial class, and this instead took a local configuration in its .ctor. Then the DefaultConfiguration class could take a reference to the Bootstrap configuration. That would make it clearer within the DefaultConfiguration class, that accessing `ServerlessModeEnabled` does more than just grab the local newrelic config value.

tippmar-nr: Maybe? I didn't invent the partial class implementation, just moved it to it's own file. I really think for our purposes at the moment, it's fine as-is. But we could add a tech debt ticket to come back to it at some point.

PR #2285, opened 2024-02-28
Add serverless mode support to Profiler

chynesNR:
This adds the ability to add an instrumentation point to the Profiler via an environment variable, `AWS_LAMBDA_FUNCTION_NAME`, in the format `assembly::class::method`. If the variable is not in the correct format it will log a Warning but continue running. It does not attempt to verify that the method exists.

The environment variable can be added at any point during the startup sequence and the Profiler will pick it up (as long as the method hasn't been invoked yet)..

When the named method is invoked it will call the wrapper `NewRelic.Providers.Wrapper.AwsLambda.HandlerMethod` in the managed Agent.

Unit tests are included.

This also removes the hard-coded method name in the `instrumentation.xml` that was used in the POC branch.

### DIFF
--- a/src/Agent/NewRelic/Profiler/MethodRewriter/Instrumentors.h
+++ b/src/Agent/NewRelic/Profiler/MethodRewriter/Instrumentors.h
@@ -29,6 +29,7 @@ namespace NewRelic { namespace Profiler { namespace MethodRewriter
     {
         bool Instrument(IFunctionPtr function, InstrumentationSettingsPtr instrumentationSettings) override
         {
+            instrumentationSettings->GetInstrumentationConfiguration()->CheckForEnvironmentInstrumentationPoint();
             auto instrumentationPoint = instrumentationSettings->GetInstrumentationConfiguration()->TryGetInstrumentationPoint(function);
             if (instrumentationPoint == nullptr)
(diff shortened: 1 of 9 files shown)


tippmar-nr (approved):
Looks good!

src/Agent/NewRelic/Profiler/MethodRewriter/Instrumentors.h
```diff
@@ -29,6 +29,7 @@ namespace NewRelic { namespace Profiler { namespace MethodRewriter
     {
         bool Instrument(IFunctionPtr function, InstrumentationSettingsPtr instrumentationSettings) override
         {
+            instrumentationSettings->GetInstrumentationConfiguration()->CheckForEnvironmentInstrumentationPoint();
```
nrcventura, line 32: It would be nice if we did not need to make this call every time, but that problem can be solved later.

chynesNR: Agreed. I briefly tried to find which internal method was setting it, so we could key off that, but no luck. As we do more testing I'm hoping we can find a cleaner place for it.

PR #2286, opened 2024-02-28
fix: Update regex parsing for CGroup v1 to match agent spec.

tippmar-nr:
Updates our container id parsing logic for CGroup v1 to match the agent spec and removes the "fallback" lookup we recently added. The spec requires that we find a line that has `cpu` in it and contains a 64-character hexadecimal string. The container ID is the 64-character string.

The logic should find a container id of `47cbd16b77c50cbf71401c069cd2189f0e659af17d5a2daca3bddf59d8a870b2` when parsing any of the following lines in `/proc/self/cgroup`:

```
1:cpuset:/docker/47cbd16b77c50cbf71401c069cd2189f0e659af17d5a2daca3bddf59d8a870b2
2:cpu:/docker/47cbd16b77c50cbf71401c069cd2189f0e659af17d5a2daca3bddf59d8a870b2
3:cpuacct,cpu:/docker/47cbd16b77c50cbf71401c069cd2189f0e659af17d5a2daca3bddf59d8a870b2
4:memory,cpu,cpuset:/docker/47cbd16b77c50cbf71401c069cd2189f0e659af17d5a2daca3bddf59d8a870b2
5:cpu,cpuacct:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod04f9c4b4_5e71_4a0a_aa3a_f62f089e3f73.slice/cri-containerd-47cbd16b77c50cbf71401c069cd2189f0e659af17d5a2daca3bddf59d8a870b2.scope
6:cpuset:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod04f9c4b4_5e71_4a0a_aa3a_f62f089e3f73.slice/cri-containerd-47cbd16b77c50cbf71401c069cd2189f0e659af17d5a2daca3bddf59d8a870b2.scope
```

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
+++ b/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
@@ -24,8 +24,7 @@ namespace NewRelic.Agent.Core.Utilization
     {
         private const string ValidateMetadataRegex = @"^[a-zA-Z0-9-_. /]*$";
 #if NETSTANDARD2_0
-        private const string ContainerIdV1Regex = @".*:cpu:/docker/([0-9a-f]{64}).*";
-        private const string ContainerIdV1FallbackRegex = @"([0-9a-f]{64})"; // This is the old regex that just looks for any 64-char hexadecimal string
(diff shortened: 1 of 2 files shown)


nr-ahemsath (approved):
Looks good.

PR #2289, opened 2024-02-28
Don't register IConnectionHandler or IConnectionManager in Serverless mode.

tippmar-nr:
Remove Autofac registration of `IConnectionManager` and `IConnectionHandler` when in Serverless mode. This ensures that there is no way to attempt to connect or reconnect to the collector.


chynesNR (approved):
We may want to revisit later to see if there are other components we can leave turned off. Just off the top of my head, we probably don't need `AgentHealthReporter` since Supportability metrics get discarded, or `SpanStreamingService` since Infinite Tracing probably doesn't make sense in a Lambda.

nrcventura (commented):
I like how this simplifies the code.

PR #2292, opened 2024-02-29
Miscellaneous config updates for serverless mode

tippmar-nr:
* Disables CAT when in Serverless mode
* Adds support for `NEW_RELIC_APDEX_T` environment variable to set ApdexT when in Serverless mode

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -817,6 +817,10 @@ namespace NewRelic.Agent.Core.Configuration
 
         private bool IsCatEnabled()
         {
+            // CAT must be disabled in serverless mode
+            if (ServerlessModeEnabled)
+                return false;
+
             var localenabled = _localConfiguration.crossApplicationTracingEnabled;
             //If config.crossApplicationTracingEnabled is true or default then we want to check the
             //config.crossApplicationTracer, if that object is not null then use it's default or value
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1691,7 +1695,16 @@ public virtual bool TransactionEventsTransactionsEnabled
         #region Transaction Tracer
 
         public virtual TimeSpan TransactionTraceApdexF { get { return TransactionTraceApdexT.Multiply(4); } }
-        public virtual TimeSpan TransactionTraceApdexT { get { return TimeSpan.FromSeconds(ServerOverrides(_serverConfiguration.ApdexT, 0.5)); } }
+        public virtual TimeSpan TransactionTraceApdexT
+        {
+            get
+            {
+                if (ServerlessModeEnabled) // get apdex_t from environment variable if running in serverless mode
```
chynesNR, line 1702: I think we can allow the environment variable override even if we aren't in serverless mode, right?

tippmar-nr: It normally comes from the connect response, so I don't know if we want to allow an override locally or not, especially without having a proper config setting for it.

chynesNR: We have some other server side config option overrides, but we can revisit when we do our eventual "overrides for everything" work.

chynesNR (approved):

PR #2316, opened 2024-03-07
ci: Fix issues in post-deploy sub-workflows

### LINKED ISSUE #2284: Agent deploy workflow issues
The 10.21.0 release on 2/27 ran into a number of failures during the [`deploy_agent` workflow](https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/8070280834), most of which appear to be an issue with how we're passing arguments to nested workflows. 

The `publish_release_notes` workflow failed at the [`Deploy Download Artifacts` step](https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/8070280834/job/22048061242#step:6:4), apparently because the `run-id` parameter was not available. However, looking at the [inputs for the `publish_release_notes` workflow](https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/8070280834), the `run_id` value was specified. 

There were several issues in the `post_deploy_agent` workflow:
* The `Wait for APT to update` and `Wait for YUM to update` steps are supposed to delay for 5 minutes when called from another workflow, but those steps were skipped because the `if` condition was not met, causing those jobs to fail.
* The [`validate-download-site-s3` job failed](https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/8070280834/job/22048060550#step:5:22) because of an apparent issue with the `AGENT_VERSION` parameter that was specified when running the `S3Validator` tool. But the inputs to the workflow indicate that `agent_version` was specified as `10.21.0`
* The [`validate-nuget-packages` job failed](https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/8070280834/job/22048060028#step:5:23) for what appears to be the same reason. 
* The `report-deprecated-nuget-packages` failed due to some kind of error in the `OctoKit` package we use, but that issue has been addressed in a separate PR.

nr-ahemsath:
Fix #2284 by using the `inputs.` input values instead of `github.event.inputs`.  See: https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-inputs-context-in-a-reusable-workflow

Apparently the "github event context" of a called workflow is that of the _caller_, not the _callee_.  That's why `github.event.inputs` was working in the sub-workflows when we ran the deploy workflow and manually specified the agent version and run id (because the agent version and run_id would be inputs to the `deploy_agent` workflow), but not when the agent version and run id were determined automatically.

TL;DR use `inputs` not `github.event.inputs` when working with reusable workflows.

### DIFF
--- a/.github/workflows/deploy_agent.yml
+++ b/.github/workflows/deploy_agent.yml
@@ -438,7 +438,7 @@ jobs:
     needs: [get-release-info]
     # if: true
     name: Create and Publish Release Notes
-    uses: newrelic/newrelic-dotnet-agent/.github/workflows/publish_release_notes.yml
+    uses: newrelic/newrelic-dotnet-agent/.github/workflows/publish_release_notes.yml@ci/release-workflow-issues
     with:
       agent_version: ${{ needs.get-release-info.outputs.release_version }}
       run_id: ${{ needs.get-release-info.outputs.workflow_run_id }}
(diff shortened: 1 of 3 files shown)


nrcventura (approved):
🥇

tippmar-nr:
Nice catch! I wonder if we've got other workflows that may have the same issue?

PR #2320, opened 2024-03-08
feat:Add a new API method to time currently unsupported datastore method calls.

jaffinito:
## Description

Adds a new public API for timing unsupported datastore calls called `RecordDatastoreSegment`,  To support this, a new IDisposable type has been added to the API called SegmentWrapper that creates and ends segments. It has a private constructor and uses static methods to create segments, currently only Datastore segments, to allow it to be extended to other segment types in the future.

The RecordDatastoreSegment API call requires a vendor name, model, and operation since this are needed to make useful datastore segments. There are four optional parameters to capture more of the rest of data used to make sql traces.  Not included were query parameters since that would require the other data to be in a very specific format that I don't think would be intuitive for out customers.

Includes unit tests for the added TransactionBridgeApi method and integration tests for the rest of the work.

API:
```
SegmentWrapper? RecordDatastoreSegment(string vendor, string model, string operation,
    string? commandText = null, string? host = null, string? portPathOrID = null, string? databaseName = null)
```

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic.Api.Agent/ITransaction.cs
+++ b/src/Agent/NewRelic.Api.Agent/ITransaction.cs
@@ -80,5 +80,20 @@ namespace NewRelic.Api.Agent
         /// </summary>
         /// <param name="userid">The User Id for this transaction.</param>
         void SetUserId(string userid);
+
+        /// <summary>
+        /// Records a datastore segment.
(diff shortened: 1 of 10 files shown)


chynesNR (approved):
Looks good!

PR #2322, opened 2024-03-08
ci: Improve post-deploy workflow

nr-ahemsath:
This PR makes the following improvements to the post-deploy workflow:
1. Add a test-mode input that lets us run this workflow without creating duplicate GitHub issues for deprecated Nuget package versions
2. Switch from `dotnet build` to `dotnet publish` for the two Nuget validation/deprecation tools, so we don't have to hardcode the .NET version (TFM) in the execution path.
3. (Finally!) fix the logic for the `wait_for_apt_and_yum` option.  Thanks @tippmar-nr for figuring this one out.


tippmar-nr (approved):
Good work!

chynesNR (approved):

nrcventura (approved):

PR #2323, opened 2024-03-08
ci: add the linux container test job to the list of jobs checked in `integration-test-status`

tippmar-nr:
Based on a suggestion by @nrcventura, we will add the `run-linux-container-tests` job to the list of jobs that are checked for successful status in `integration-test-status`, which will allow us to remove the `run-linux-container-tests` job as a required status check. 

This (hopefully) will resolve the strange issues we've seen with PRs expecting a status from the container test job, even though that job was skipped.

Also adds a "harden runner" step to that job, which was overlooked when we initially created it.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -136,8 +136,9 @@ jobs:
           if-no-files-found: error
 
   run-linux-container-tests:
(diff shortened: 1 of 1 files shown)


nr-ahemsath (approved):
Looks good!

PR #2326, opened 2024-03-08
Add unit tests for agent services registered based on serverless mode

nr-ahemsath:
Add a unit test with two testcases, serverless mode either enabled or disabled, to verify which services are registered in the agent (or not) DI container based on serverless mode.


tippmar-nr (approved):
Nice!

nrcventura (approved):

PR #2334, opened 2024-03-12
test: Modify SimpleCache to allow direct call to MaintainCache(), update unit tests.

tippmar-nr:
Addresses a unit test flicker in `SimpleCacheTests` by making `SimpleCache.MaintainCache()` public and calling it directly in the unit tests rather than relying on `Thread.Sleep(1000)` to let the timer make the call.


nrcventura (commented):
This looks good to me, but do we need to also add a test for the timer part of the code?

nrcventura (approved):

chynesNR (approved):

PR #2335, opened 2024-03-12
tests: Speed up execution time of AgentLogs integration test namespace

nr-ahemsath:
There were a couple of cases where the integration test harness was waiting for three minutes for an agent log to appear in the default location, that wasn't ever going to be created because the test case involved changing the log file name or directory.  This PR fixes that, as well as tweaking some names to be make more sense (to me, anyway).

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogFile.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogFile.cs
@@ -9,7 +9,6 @@ using System.Diagnostics.Contracts;
 using System.IO;
 using System.Linq;
 using System.Threading;
-using NewRelic.Agent.IntegrationTestHelpers.RemoteServiceFixtures;
 using Xunit.Abstractions;
 
 namespace NewRelic.Agent.IntegrationTestHelpers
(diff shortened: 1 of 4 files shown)


tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogFile.cs
```diff
@@ -49,10 +49,12 @@ public AgentLogFile(string logDirectoryPath, ITestOutputHelper testLogger, strin
                 }
 
                 Thread.Sleep(TimeSpan.FromSeconds(1));
-            } while (timeTaken.Elapsed < timeout);
+            } while (timeTaken.Elapsed < timeout && logFileExpected);
```
tippmar-nr, line 52: I wonder if you couldn't wrap the entire `do...while` and `throw` with `if (logfileExpected)`? Might be a tad cleaner.

tests/Agent/IntegrationTests/IntegrationTestHelpers/ProfilerLogFile.cs
```diff
@@ -48,7 +48,7 @@ public ProfilerLogFile(string logDirectoryPath, TimeSpan? timeoutOrZero = null,
                 Thread.Sleep(TimeSpan.FromSeconds(1));
             } while (timeTaken.Elapsed < timeout);
```
tippmar-nr, line 49: Does this need the same logic as in `AgentLogFile.cs`?

tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogFile.cs
chynesNR, line 52: I realize the scenario we're fixing is unlikely to fail, but if we were to create a log file when we weren't expecting to, I'm not sure 1 second would be long enough to find it. Maybe we could just pass in a shorter timeout?

tests/Agent/IntegrationTests/IntegrationTestHelpers/ProfilerLogFile.cs
nr-ahemsath: No current test cases need it, but it makes sense to make it the same between the two classes. 👍

tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogFile.cs
nr-ahemsath: Well, the logic in AgentLogFile.cs doesn't throw an exception if it finds a log when it wasn't expecting to.  Some of the tests in this namespace have assertions that specifically make sure that log files aren't created when/where they shouldn't be based on config.

tippmar-nr (approved):
Looks good! Thanks for digging in!

PR #2336, opened 2024-03-12
test: MySql test performance improvements

tippmar-nr:
* Refactor `MySqlStoredProcedureTests` to combine the with and without @ signs tests into a single test, reducing the number of tests by half. 
* Adjusted the harvest cycle overrides for MySql tests from 45 seconds down to 10 (with one reduced to 20), significantly improving total execution time. 
* Added timing log messages for application startup, exercise application and application shutdown. 
```
    Remote application build/startup time: 22.2470 seconds
...
    ExerciseApplication execution time: 21.5051 seconds
...
    Remote application shutdown time: 0.4504 seconds
```
* Added timing to log messages in `WaitForLogLines()` to show how long the test had to wait for the given match.
```
2024-03-13 14:04:15,945 WaitForLogLines  Matched expression: ^.*?NewRelic\s+FINEST: \[pid: \d+, tid: .+\] Transaction (.*) \((.*)\) transform completed. in 5.8s.
```
* Tweaked the `all_solutions` workflow so that it only installs the MSMQ feature when running the MSMQ tests in the unbounded matrix.
* Removed some unnecessary package references
* Modified the initialization script for MySql to insert 5 rows instead of a few thousand.


chynesNR (approved):
Nice!

PR #2337, opened 2024-03-12
chore: Refactor serverless bootstrap config into a separate bootstrap config.

nrcventura:

## Description

The agent has a need for a separate bootstrap config that can be used before the configuration service is available. The serverless config is a current example of some of the settings that can be moved. With this change, the logic for the serverless config is now in a separate object from the local config.

This is the start of a path that we can take so that there are only 2 places in the code that control the logic of which "configuration source" takes precedence for all of the agent settings (DefaultConfiguration and BootstrapConfiguration). The BootstrapConfiguration can also be used by DefaultConfiguration so that the agent does not need to look in multiple locations outside of the Bootstrap process (before an instance of DefaultConfiguration is available).

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentManager.cs
@@ -95,9 +95,11 @@ namespace NewRelic.Agent.Core
         {
             // load the configuration
             configuration config = null;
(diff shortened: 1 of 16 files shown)


chynesNR (approved):
Very nice!

PR #2340, opened 2024-03-14
Bedrock integration tests

chynesNR:
This adds integration tests for our Bedrock instrumentation to verify:
* `LlmChatCompletionSummary`, `LlmChatCompletionMessage`, `LlmEmbedding`, `LlmFeedbackMessage` types
* LLM Errors
* The token calculation callback
* Custom attributes added to LLM events
* Support for various models: `meta13`, `ai21`, `amazonembed`, `amazonexpress`, `cohere`, and `anthropic`

This also includes some fixes for issues I found during testing:
* `Response.model` is not available for all models
* Embedded model responses were not being parsed correctly
* Error events could be dropped if the transaction was too short


tippmar-nr (commented):
Weren't you going to add a log message somewhere when config changes and causes events to be dumped?

tippmar-nr (approved):
Looks solid to me!

chynesNR:
> Weren't you going to add a log message somewhere when config changes and causes events to be dumped?

I absolutely was and then I saw this:
```
        protected override void OnConfigurationUpdated(ConfigurationUpdateSource configurationUpdateSource)
        {
            // It is *CRITICAL* that this method never do anything more complicated than clearing data and starting and ending subscriptions.
            // If this method ends up trying to send data synchronously (even indirectly via the EventBus or RequestBus) then the user's application will deadlock (!!!).
```

Writing to the log _should_ be fine but I want to put some more thought into it. I'll put in a ticket to come back.

nrcventura (approved):

PR #2344, opened 2024-03-18
chore: More bootstrap config refactoring

nrcventura:

## Description

Moves the code in the configuration partial class into the bootstrap config (it is only used at bootstrap time and not used after that so dynamic updates do not matter).

This results in the following:
- A more consistent management of the AgentEnabled and AgentEnabledAt settings.
- Elimination of the unused Xml property so that we do not need to keep the original xml string in memory
- Elimination of the LogConfig property, which is replaced with the original `log` field
- Elimination of the configurationPartial partial class

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/Config/BootstrapConfiguration.cs
```diff
@@ -56,5 +134,55 @@ private bool CheckServerlessModeEnabled(configuration localConfiguration)
...
+            {
+                var stringValueWithProvenance = getStringValueWithProvenance(Constants.AppSettingsAgentEnabled);
+                if (stringValueWithProvenance != null && stringValueWithProvenance.Value != null && bool.TryParse(stringValueWithProvenance.Value, out var value))
+                {
+                    return new ValueWithProvenance<bool>(value, stringValueWithProvenance.Provenance);
+                }
+            }
+            catch (Exception ex)
+            {
+                Log.Error($"Failed to read {Constants.AppSettingsAgentEnabled} from local config.", ex);
```
chynesNR, line 182: I think you have this syntax backwards: the exception should be the first parameter.

chynesNR (approved):
Looks good!

src/Agent/NewRelic/Agent/Core/Config/BootstrapConfiguration.cs
nrcventura: Good catch! I didn't look closely enough at the copilot output.

nrcventura:
The MySql test failure is a known flicker, and will be addressed separately.

PR #2348, opened 2024-03-19
test: Fix a MySql flicker by introducing an intentional delay

tippmar-nr:
Added an intentional delay in one of the MySql queries to ensure that we always know it will be the slowest, which should address a recent frequently-occurring failure. This same change may need to be more widely applied if other flickers show up, but for now it's just focused on a single test.


chynesNR (approved):
Looks like a good way to make it consistent.

nrcventura (approved):

nr-ahemsath (approved):
Clever!

nrcventura (approved):

PR #2349, opened 2024-03-19
Use custom var and fallback for lambda instrumentation point

chynesNR:
The previous solution for auto-detecting the lambda instrumentation point in the profiler would not work in many cases. Now we try two environment variables:
* `_HANDLER` which should be auto-populated but may not be available in all environments (notably the test tool)
* `NEW_RELIC_LAMBDA_FUNCTION_HANDLER` which the user must provide

This is now only checked on startup instead of prior to each method's compilation.

### DIFF
--- a/src/Agent/NewRelic/Home/Home.csproj
+++ b/src/Agent/NewRelic/Home/Home.csproj
@@ -13,7 +13,7 @@
   </Target>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


tippmar-nr (approved):
Quick work!

nrcventura (approved):

PR #2351, opened 2024-03-19
More Lambda / DT configuration tweaks

tippmar-nr:
* Adds an override for `SamplingTarget` and `SamplingTargetPeriodInSeconds` when in Serverless mode, as per the spec. 
* Minor cleanup in the function handler wrapper.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -929,13 +929,13 @@ namespace NewRelic.Agent.Core.Configuration
             EnvironmentOverrides(_localConfiguration.distributedTracing.account_id, "NEW_RELIC_ACCOUNT_ID")
             : _serverConfiguration.AccountId;
 
-        public int? SamplingTarget => _serverConfiguration.SamplingTarget;
+        public int? SamplingTarget => ServerlessModeEnabled ? 10 : _serverConfiguration.SamplingTarget;
 
         // Faster Event Harvest configuration rules apply here, which is why ServerOverrides takes precedence over EnvironmentOverrides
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -929,13 +929,13 @@ private bool IsDistributedTracingEnabled()
             EnvironmentOverrides(_localConfiguration.distributedTracing.account_id, "NEW_RELIC_ACCOUNT_ID")
             : _serverConfiguration.AccountId;
 
-        public int? SamplingTarget => _serverConfiguration.SamplingTarget;
+        public int? SamplingTarget => ServerlessModeEnabled ? 10 : _serverConfiguration.SamplingTarget;
```
nrcventura, line 932: Should this be more like the following to avoid the conditional check with each access?
```suggestion
        public int? SamplingTarget { get; } = ServerlessModeEnabled ? 10 : _serverConfiguration.SamplingTarget;
```

tippmar-nr: We use this same pattern in about a zillion places already; I can't imagine there'll be much of a performance improvement by making this change.

nrcventura: The main reason I bring it up is that I believe that Sampling is invoked more frequently than other settings.

tippmar-nr: `AdaptiveSampler` stores the current value of `SamplingTargetPeriodInSeconds` and `SamplingTarget` anytime there's a configuration change. So this property is only referenced once at a configuration change.

nrcventura (approved):

PR #2352, opened 2024-03-19
Minor cleanup of integration tests

nr-ahemsath:
Removed an unused local variable, as well as some unused usings.


chynesNR (approved):
Thanks, I forgot to do a cleanup pass when I was done.

nrcventura (approved):

PR #2353, opened 2024-03-20
Bedrock - HSM updates

tippmar-nr:
Disable AI Monitoring when HSM is enabled. Updates tests accordingly.


tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -3742,11 +3744,6 @@ public void AiMonitoringStreamingDisabledByEnvironmentVariableWhenAiMonitoringEn
             Assert.That(_defaultConfig.AiMonitoringStreamingEnabled, Is.False);
         }
 
-        [Test]
```
chynesNR, line 3745: Isn't this still a valid test? Though now that I look at the spec again, isn't it supposed to be enabled by default?

tippmar-nr: This test was essentially a duplicate of `AiMonitoringRecordContentDisabledWhenAiMonitoringDisabled`

nrcventura (changes requested):
We need to review the latest spec changes regarding this before we can merge this PR.

nr-ahemsath (approved):
Looks solid! 👍

jaffinito (approved):

nrcventura (approved):

PR #2361, opened 2024-03-21
Fix profiler bug with tracer flags not being set properly

chynesNR:
This could result in (among other issues) a non-async method being flagged as async, and therefore not instrumenting properly


src/Agent/NewRelic/Profiler/Configuration/InstrumentationConfiguration.h
```diff
@@ -163,6 +163,7 @@ namespace NewRelic { namespace Profiler { namespace Configuration
             instrumentationPoint->ClassName = segments[1];
             instrumentationPoint->MethodName = segments[2];
             instrumentationPoint->Parameters = nullptr;
+            instrumentationPoint->TracerFactoryArgs = 0;
```
nrcventura, line 166: Should we also set some default values for when this type is constructed?

chynesNR: Yeah good idea

nrcventura (approved):

PR #2364, opened 2024-03-22
Lambda: Event Support & Tests

tippmar-nr:
Adds support for all event types called out in the Lambda spec, with unit tests for each.

There are still some TODOs that need to be worked out eventually but are not blockers for this PR.

It's also not clear (and won't be until we have working integration tests) how much additional null checking might be required - the current code assumes the dynamic properties it references exist and are non-null in almost all cases.


chynesNR (approved):
Looks good, at least until we can start integration testing

PR #2378, opened 2024-03-26
Move JSON deserialization and models to NewRelic.Core

nr-ahemsath:
This PR moves the Bedrock request/response models and the deserialization of their corresponding JSON from the Bedrock wrapper project to `NewRelic.Core`, which ILRepacks `Newtonsoft.Json`, so it will always be available for use.

This prevents assembly load exceptions I found while testing the Bedrock instrumentation with a standalone test app that, unlike our integration test applications, did not include a reference to Newtonsoft.Json.


nrcventura (approved):

chynesNR (approved):
Good find!

PR #2387, opened 2024-04-04
Feature work/lambda missing data

chynesNR:
Allow cases where the lambda context is missing, or data within it is missing.

Also force `CompleteTransactionsOnThread` to be enabled in serverless mode.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
@@ -48,8 +48,8 @@ namespace NewRelic.Agent.Core.DataTransport
         private object _writeLock = new object();
         private readonly IDateTimeStatic _dateTimeStatic;
         private DateTime _lastMetricSendTime;
-        private static string _arn;
-        private static string _functionVersion;
+        private static string _arn = null;
+        private static string _functionVersion = null;
         private static string _executionEnvironment = null;
         private string _outputPath = $"{Path.DirectorySeparatorChar}tmp{Path.DirectorySeparatorChar}newrelic-telemetry";
 
@@ -213,7 +213,10 @@ namespace NewRelic.Agent.Core.DataTransport
         public static void SetMetadata(string functionVersion, string arn)
         {
             _arn = arn; // It doesn't seem like there's a fallback way to get the ARN without the ILambdaContext
-            _functionVersion = functionVersion ?? System.Environment.GetEnvironmentVariable("AWS_LAMBDA_FUNCTION_VERSION") ?? "$LATEST";
+            _functionVersion = _functionVersion ??
+                functionVersion ??
+                System.Environment.GetEnvironmentVariable("AWS_LAMBDA_FUNCTION_VERSION") ??
+                "$LATEST";
             _executionEnvironment ??= System.Environment.GetEnvironmentVariable("AWS_EXECUTION_ENV");
         }
 
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
@@ -48,10 +48,8 @@ namespace NewRelic.Agent.Core.DataTransport
         private object _writeLock = new object();
         private readonly IDateTimeStatic _dateTimeStatic;
         private DateTime _lastMetricSendTime;
-        private static string _arn = null;
-        private static string _functionVersion = null;
-        private static string _executionEnvironment = null;
         private string _outputPath = $"{Path.DirectorySeparatorChar}tmp{Path.DirectorySeparatorChar}newrelic-telemetry";
+        private static IDictionary<string, object> _metadata = null;
 
         public ServerlessModeDataTransportService(IDateTimeStatic dateTimeStatic)
         {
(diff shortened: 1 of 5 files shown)


src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -170,7 +170,7 @@ public void End(bool captureResponseTime = true)
             };
 
             // The completion of transactions can be run on thread or off thread. We made this configurable.  
-            if (Agent.Configuration.CompleteTransactionsOnThread)
+            if (Agent.Configuration.CompleteTransactionsOnThread || Agent.Configuration.ServerlessModeEnabled)
```
nrcventura, line 173: Would it be better to move the ServerlessModeEnabled check into the `CompleteTransactionsOnThread` property in DefaultConfiguration?

nrcventura: If we move the logic int the CompleteTransactionsOnThread property, we can more easily unit test that logic.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -36,46 +36,102 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
+                    functionVersion = functionVersionGetter(lambdaContext);
+                }
+                catch (Exception ex)
+                {
+                    agent.Logger.Log(Agent.Extensions.Logging.Level.Error, $"Failed to read lambda context: {ex.Message}");
+                }
+            }
+            else
+            {
+                agent.Logger.Log(Agent.Extensions.Logging.Level.Warn, $"No Lambda context information found");
```
nrcventura, line 97: We might want to ensure that this is only logged once, otherwise every invocation can trigger this log message.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -36,46 +36,102 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
+            object inputObject = null;
+            AwsLambdaEventType eventType = AwsLambdaEventType.Unknown;
 
-            var functionNameGetter = _getFunctionNameFromLambdaContext ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(lambdaContext.GetType(), "FunctionName");
-            var functionVersionGetter = _getFunctionVersionFromLambdaContext ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(lambdaContext.GetType(), "FunctionVersion");
-            var requestIdGetter = _getAwsRequestIdFromLambdaContext ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(lambdaContext.GetType(), "AwsRequestId");
-            var functionArnGetter = _getInvokedFunctionArnFromLambdaContext ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(lambdaContext.GetType(), "InvokedFunctionArn");
+            foreach (var arg in instrumentedMethodCall.MethodCall.MethodArguments)
+            {
+                // TODO: I don't like this approach, but our options are limited without access to the interface
```
nrcventura, line 45: One improvement we could make is to only do some of this work on the initial invocation, because some of this data should not change with each invocation. That could be done in a separate PR if we think it is worthwhile.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -36,46 +36,102 @@
...
+                }
+                else
+                {
+                    var fullName = arg.GetType().FullName;
+                    agent.Logger.Log(Agent.Extensions.Logging.Level.Debug, $"input object fullname = {fullName}");
+                    eventType = fullName.ToEventType();
+                    if (eventType != AwsLambdaEventType.Unknown)
+                    {
+                        inputObject = arg;
+                        agent.Logger.Log(Agent.Extensions.Logging.Level.Debug, $"eventType = {eventType}");
```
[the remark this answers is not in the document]
nrcventura: This seems like a false positive, but I did not look deep enough to fully confirm it.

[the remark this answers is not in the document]
chynesNR: This came up in a prior PR and I'm still puzzled by it. It's just an enum value and no different than the others in that enum. I'm not sure what it thinks is "private" data. Maybe because it has "email" in the name?

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
chynesNR: I considered that but I was concerned it would hide the fact that serverless mode was behaving differently. I don't feel too strongly though if you think the override is cleaner.

src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
```diff
@@ -211,8 +212,12 @@ public bool FlushData(string transactionId)
 
         public static void SetMetadata(string functionVersion, string arn)
         {
-            _arn = arn;
-            _functionVersion = functionVersion;
+            _arn = arn; // It doesn't seem like there's a fallback way to get the ARN without the ILambdaContext
```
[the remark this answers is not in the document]
chynesNR: That's a good idea, I'll do that.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
chynesNR: Sure, we can do that

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
chynesNR: I think you're probably right but I'd like to see more real data to make sure they can't actually change on the fly. :)

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
nrcventura: Either approach works for me (I don't think that we have consistency in the code base either way), but I think the logic should be covered in a unit test.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
chynesNR: I ended up doing as you suggested. I realized we would have to handle more race conditions if we expect some of the values to change.

PR #2395, opened 2024-04-11
Async lambda methods were not being instrumented correctly

chynesNR:


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -224,10 +230,19 @@ void InvokeTryProcessResponse(Task responseTask)
                     {
                         return;
                     }
+
+                    if (responseTask.Status == TaskStatus.Faulted)
+                    {
+                        transaction.NoticeError(responseTask.Exception);
+                    }
```
nrcventura, line 237: Should this be moved before the previous if statement so that we notice the error before the early return?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -224,10 +230,19 @@ void InvokeTryProcessResponse(Task responseTask)
...
+
                     var responseGetter = _getRequestResponseFromGeneric ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(responseTask.GetType(), "Result");
                     var response = responseGetter(responseTask);
                     if (_functionDetails.EventType is AwsLambdaEventType.APIGatewayProxyRequest or AwsLambdaEventType.ApplicationLoadBalancerRequest)
                         CaptureResponseData(transaction, response);
+
+                    segment.End();
+                    transaction.End();
```
nrcventura, line 245: This should probably also happen in the error case too, before the early return.

tippmar-nr: Maybe put it in a `finally` block?

nrcventura (approved):

PR #2399, opened 2024-04-15
test: Test latest versions of Elastic.Clients.Elasticsearch and MongoDB.Driver

chynesNR:
The latest `Elastic.Clients.Elasticsearch` appears to have removed a helper extension that we were using to test the library. I found nothing in the release notes about it going away. The new test code supports both old and new versions. The new release also marks some methods we're using as deprecated, which was causing a warning-as-error build failure.


nr-ahemsath (approved):
Looks good 👍

nrcventura:
Should we add a tech debt story to address the obsolete api calls used for the newer library versions?

nrcventura (approved):

chynesNR:
> Should we add a tech debt story to address the obsolete api calls used for the newer library versions?

I figured we'd just wait until they broke, but being proactive about it is probably a good idea. Done.

PR #2404, opened 2024-04-16
Refactoring and tests for serverless mode payload

tippmar-nr:
* Refactors `ServerlessModeDataTransport` to separate the payload management (building / writing) to a separate class (`ServerlessModePayloadManager`) to allow for unit testing. 
* Adds unit tests to verify that the payload created in serverless mode is correct, including metadata and verification of the compressed portion of the payload. 

Additional tests will be added for `ServerlessModeDataTransport` in a separate PR.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Utilities/FileWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Core/Utilities/FileWrapper.cs
@@ -1,0 +1,31 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.IO;
+
+namespace NewRelic.Agent.Core.Utilities
+{
+    /// <summary>
+    /// Wraps some File methods to allow for unit testing
+    /// </summary>
+    public interface IFileWrapper
+    {
+        bool Exists(string path);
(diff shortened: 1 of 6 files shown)


src/Agent/NewRelic/Agent/Core/Utilities/FileWrapper.cs
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System.IO;
+
+namespace NewRelic.Agent.Core.Utilities
+{
+    /// <summary>
+    /// Wraps some File methods to allow for unit testing
+    /// </summary>
+    public interface IFileWrapper
```
chynesNR, line 11: I was going to suggest that we have a single interface for mocking file system calls, but it doesn't seem like we need that in too many places, so this is probably fine. :)

chynesNR (approved):

src/Agent/NewRelic/Agent/Core/Utilities/FileWrapper.cs
tippmar-nr: Yeah, I thought maybe we already had one somewhere but didn't see anything. This is easy enough to extend if/when we need it elsewhere.

PR #2409, opened 2024-04-17
Composite tests for serverless mode data transport

tippmar-nr:
* Enables serverless mode in CompositeTest agent.
* Adds test to verify ServerlessDataTransport correctly receives and reports the expected data in a number of different use cases.


src/Agent/NewRelic/Agent/Core/Config/ConfigurationLoader.cs
```diff
@@ -434,6 +434,12 @@ public static void PublishDeserializedEvent(configuration config)
             EventBus<ConfigurationDeserializedEvent>.Publish(new ConfigurationDeserializedEvent(config));
         }
 
+        // primarily for testing in CompositeAgent
```
nrcventura, line 437: Can we make this a `///` style comment so that intellisense can help us to understand that this shouldn't be used in production code, and it is only present for the tests?

tests/Agent/UnitTests/CompositeTests/CompositeTestAgent.cs
```diff
@@ -261,6 +311,9 @@ public void Dispose()
             var transaction = _primaryTransactionContextStorage.GetData();
             transaction?.Finish();
 
+            // reset the bootstrap configuration in case this was a serverless mode test
+            ConfigurationLoader.UseBootstrapConfigurationForTesting(BootstrapConfiguration.GetDefault());
```
nrcventura, line 315: Instead of restoring the result of GetDefault, perhaps the composite test agent should capture the original value before swapping it out, and then restore that original value here?

src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
```diff
@@ -161,8 +161,11 @@ public DataTransportResponseStatus Send(LogEventWireModelCollection loggingEvent
             return DataTransportResponseStatus.RequestSuccessful;
         }
 
-        // Not supported in serverless mode
-        public DataTransportResponseStatus Send(LoadedModuleWireModelCollection loadedModules, string transactionId) => throw new NotImplementedException();
+        public DataTransportResponseStatus Send(LoadedModuleWireModelCollection loadedModules, string transactionId)
+        {
+            // Not supported in serverless mode
+            return DataTransportResponseStatus.RequestSuccessful;
```
chynesNR, line 167: Was this causing a problem? If we're generating module events in serverless mode something is probably wrong, so we may want to keep that as an exception.

tests/Agent/UnitTests/CompositeTests/CompositeTestAgent.cs
tippmar-nr: Done

src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
tippmar-nr: This was probably a leftover from my early testing where the agent wasn't really in serverless mode like I expected. I reverted the change.

src/Agent/NewRelic/Agent/Core/DataTransport/ServerlessModeDataTransportService.cs
```diff
@@ -161,7 +151,6 @@ public DataTransportResponseStatus Send(LogEventWireModelCollection loggingEvent
             return DataTransportResponseStatus.RequestSuccessful;
         }
 
-        // Not supported in serverless mode
         public DataTransportResponseStatus Send(LoadedModuleWireModelCollection loadedModules, string transactionId) => throw new NotImplementedException();
```
tippmar-nr, line 154:
```suggestion
       // Not supported in serverless mode
        public DataTransportResponseStatus Send(LoadedModuleWireModelCollection loadedModules, string transactionId) => throw new NotImplementedException();
```

chynesNR (approved):

PR #2413, opened 2024-04-18
Added some logging for unrecognized request / response types

tippmar-nr:


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -155,6 +158,11 @@ private void InitLambdaData(InstrumentedMethodCall instrumentedMethodCall, IAgen
                     {
                         agent.Logger.Log(Agent.Extensions.Logging.Level.Debug, $"Supported Event Type found: {_functionDetails.EventType}");
                     }
+                    else if (!_unsupportedInputTypes.Contains(name))
```
chynesNR, line 161: Just FYI, `Add` returns false if the item is already present, so you could simplify this with `if (_unsupportedInputTypes.Add(name))`  (and similarly below with `_unexpectedResponseTypes`)

chynesNR (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -120,6 +120,9 @@ public object GetInputObject(InstrumentedMethodCall instrumentedMethodCall)
         public bool IsTransactionRequired => false;
 
         private static bool _coldStart = true;
+        private HashSet<string> _unexpectedResponseTypes = new();
+        private HashSet<string> _unsupportedInputTypes = new();
```
nrcventura, line 124: Do we need to use thread safe collections here?

tippmar-nr: I don't think so; worst case, we log the warning a couple of times instead of just once.

nrcventura: I'm ok if we warn a few times, but I'm more concerned about the internal state of the collection being in a bad state if add is called concurrently.

nrcventura (approved):

PR #2415, opened 2024-04-18
Feature work/lambda agent attributes variant

jaffinito:
## Description

Variant of https://github.com/newrelic/newrelic-dotnet-agent/pull/2414 (closed in favor of this PR)

- Adds new attribute definition and method for creating lambda specific attributes as agent attributes.
- Adds `AddLambdaAttribute(string name, string value)` method to `API/ITransaction` for adding Lambda specific attributes.
- Removes dictionary for attributes and instead adds them directly to the transaction, changing `LambdaAttributeExtensions` to support only one method.
- Updates `TransactionTests`, `TransactionAttributeMakerTests`, and `SpanEventMakerTests` to test that the new agent attribute is working properly.
- Updates `LambdaAttributeExtensionsTests` to call `AddLambdaAttribute` instead of `AddCustomAttribute`.

# Author Checklist
- [ ] Unit tests, Integration tests, and Unbounded tests completed
- [ ] Performance testing completed with satisfactory results (if required)

# Reviewer Checklist
- [ ] Perform code review
- [ ] Pull request was adequately tested (new/existing tests, performance tests)

PR #2418, opened 2024-04-19
Clean up TODOs in Lambda feature branch

nr-ahemsath:
Clean up (most) TODOs in the Lambda feature branch:

* `Lambda` is fine for the transaction category.  This doesn't appear to be used in the Lambda UI anywhere.
* `webReqEvent.Path` is fine to use for `transaction.SetUri()`
* `DateTime().ToString()` is fine to send for the timestamp event source attributes for the event types where it is required (cloudwatch, S3 and SNS)
* Code already modified to send required Lambda attribute data as agent attributes instead of custom attributes, see https://github.com/newrelic/newrelic-dotnet-agent/pull/2415
* Distributed tracing header parsing
  * We were correctly parsing DT headers for web requests, SNS and SQS events, but we weren't actually processing them with `transaction.AcceptDistributeTraceHeaders()`.  This is fixed.
  * There is one additional TODO for also parsing W3C trace context headers from SQS events (which is complicated, since we have to be able to parse headers out of the message body text).  This will be handled in a follow-on PR.


nrcventura (approved):

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs
```diff
@@ -78,7 +75,6 @@ public static void AddEventTypeAttributes(IAgent agent, ITransaction transaction
                     transaction.AddEventSourceAttribute("length", (string)s3Event.Records.Count.ToString());
                     transaction.AddEventSourceAttribute("region", (string)s3Event.Records[0].AwsRegion);
                     transaction.AddEventSourceAttribute("eventName", (string)s3Event.Records[0].EventName);
-                    // TODO: Figure out if the eventTime value should be in some specific format. The spec doesn't say.
```
chynesNR, line 81: I'm a bit surprised that these aren't called out anywhere. There's a big difference between UTC/local time, date localization, etc.

chynesNR (approved):

PR #2421, opened 2024-04-22
Fix: Refactors AddLambdaAttribute to support any object type as value.

jaffinito:
## Description

Refactors AddLambdaAttribute to support any object type as value.  Updates AttributeDefs, unit tests, and integration tests to allow or test the same.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
+++ b/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
@@ -126,7 +126,7 @@ namespace NewRelic.Agent.Core.Attributes
         AttributeDefinition<object, object> GetCustomAttributeForSpan(string name);
         AttributeDefinition<object, object> GetCustomAttributeForTransaction(string name);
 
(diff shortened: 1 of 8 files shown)


src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
```diff
@@ -126,7 +126,7 @@ public interface IAttributeDefinitions
         AttributeDefinition<object, object> GetCustomAttributeForSpan(string name);
         AttributeDefinition<object, object> GetCustomAttributeForTransaction(string name);
 
-        AttributeDefinition<string, string> GetLambdaAttribute(string name);
+        AttributeDefinition<object, object> GetLambdaAttribute(string name);
```
nrcventura, line 129: Shouldn't this be AttributeDefinition<string, object>?

nrcventura: I just got clarification that this refers to the type going in and the type going out, so the generic types are fine.

chynesNR (approved):

PR #2428, opened 2024-04-23
chore: Deprecate .NET 7 usage in tests

tippmar-nr:
* Updated build targets to `net8.0` as appropriate. 
* Removed tests that were specific to .NET 7 that can't be run under .NET 8 (mostly `Microsoft.Data.SqlClient`)
* Removed .NET 7.0 package inclusion rules for integration tests (which will reduce integration test runtime. Yay!)


chynesNR (approved):

jaffinito (approved):

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/MySql/MySqlExerciser.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020 New Relic, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-#if !NET7_0 // there is no .NET 7 test for MySql
+#if !NET8_0 // there is no .NET 7 test for MySql
```
nrcventura, line 4: I think this is breaking our MySql tests.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MsSql/MsSqlQueryParamTests.cs
```diff
@@ -205,34 +205,6 @@ public MsSqlQueryParamTests_SystemData_NoAtSigns_FWLatest(ConsoleDynamicMethodFi
 
     #region System.Data.SqlClient (.NET Core/5+ only)
 
-    // Set to net7 since the package this relies on does not currently support net8.
-    [NetCoreTest]
-    public class MsSqlQueryParamTests_SystemDataSqlClient_CoreLatest : MsSqlQueryParamTestsBase<ConsoleDynamicMethodFixtureCore70>
-    {
-        public MsSqlQueryParamTests_SystemDataSqlClient_CoreLatest(ConsoleDynamicMethodFixtureCore70 fixture, ITestOutputHelper output)
```
nrcventura, line 212: Shouldn't these core latest tests point to the net8 fixture instead of removing them?

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MsSql/MsSqlQueryParamTests.cs
```diff
@@ -205,34 +205,6 @@ public MsSqlQueryParamTests_SystemData_NoAtSigns_FWLatest(ConsoleDynamicMethodFi
 
     #region System.Data.SqlClient (.NET Core/5+ only)
 
-    // Set to net7 since the package this relies on does not currently support net8.
```
nrcventura, line 208: We should try to figure out why we can test ms sql on net8 so that we can keep these tests.

tests/Agent/UnitTests/Core.UnitTest/CrossAgentTests/CatMapTests.cs
```diff
@@ -124,7 +124,11 @@ public void Test(TestCase testCase)
             {
                 var transactionName = GetTransactionNameFromString(request.OutboundTxnName);
                 _transaction.CandidateTransactionName.TrySet(transactionName, (TransactionNamePriority)namePriority++);
+#if NETFRAMEWORK
                 var outboundHeaders = _agent.CurrentTransaction.GetRequestMetadata().ToDictionary();
+#else
+                var outboundHeaders = Enumerable.ToDictionary(_agent.CurrentTransaction.GetRequestMetadata());
```
nrcventura, line 130: Why do we need something different here? Did the ToDictionary method go away?

tippmar-nr: netstandard doesn't have `.ToDictionary()`, so we use the `MoreLinq` package to provide it. 

net8 does have `.ToDictionary()` but it can't be invoked as an extension method because of widespread use of other `MoreLinq` extensions throughout this test. 

My preference would be to dump `MoreLinq` entirely but I wasn't going to include that in this set of changes.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/MySql/MySqlExerciser.cs
tippmar-nr: Yeah, it was a leftover from my last round of test refactoring. We were building a .net 6, 7 and 8 version of the library but never using the .net 7 build. I should have removed the #if rather than updating it to net 8. Which I just did... ;-)

tests/Agent/IntegrationTests/UnboundedIntegrationTests/MsSql/MsSqlQueryParamTests.cs
tippmar-nr: It's all coming back to me now... System.Data.SqlClient isn't usable in .NET 8... https://github.com/dotnet/runtime/issues/86969

PR #2430, opened 2024-04-23
chore: Update S3Validator to use dotnet publish in post-deploy workflow

tippmar-nr:
Modifies the S3Validator job in the post_deploy workflow to use `dotnet publish` so we can avoid a hard-coded output path that changes with .NET version updates.

Also updates the publish_release_notes workflow to follow the same pattern.


nr-ahemsath (approved):
💯

nr-ahemsath (approved):

PR #2433, opened 2024-04-23
fix: Fix duplicate key error when using Kafka with DT enabled

chynesNR:
There were actually two bugs when using Kafka with Distributed Tracing:
* If headers were not present, we weren't creating them (due to a scoping error)
* We were not checking to see if the keys already existed before adding them. This did not cause an immediate problem because the headers are kept in a list rather than a dictionary, but can cause a validation failure later.

The first bug kept us from finding the second problem in our integration tests. Both have been addressed.


nrcventura (approved):

tippmar-nr (approved):
Good fix!

PR #2435, opened 2024-04-24
Parse W3C trace context payloads from SQS message body

nr-ahemsath:
Update SQS tracing header processing to handle W3C trace context data as well as the New Relic proprietary DT data.

Update unit tests to cover both cases alone and combined.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs
@@ -130,8 +130,11 @@ public static class LambdaEventHelpers
     }
 
     private const string NEWRELIC_TRACE_HEADER = "newrelic";
+    private const string W3C_TRACEPARENT_HEADER = "traceparent";
+    private const string W3C_TRACESTATE_HEADER = "tracestate";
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs
```diff
@@ -130,27 +130,39 @@ public static void AddEventTypeAttributes(IAgent agent, ITransaction transaction
...
+    private const string SQS_MSG_ATTR_VALUE_PREFIX = @"Value"":""";
 
-    // TODO: Also handle W3C trace context headers
     private static void TryParseSQSDistributedTraceHeaders(dynamic sqsEvent, ITransaction transaction)
     {
         // We can't pass anything dynamic to AcceptDTHeaders, so we have to copy the sqs
         // message attributes to a new <string,string> dict and then pass that to AcceptDTHeaders
         IDictionary<string, string> sqsHeaders = new Dictionary<string, string>();
 
         var record = sqsEvent.Records[0];
```
tippmar-nr, line 144: Do we need to add some defensive code here in case `.Records` is null or empty?

jaffinito (approved):

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs
nr-ahemsath: Good idea, done.

PR #2436, opened 2024-04-24
Use a more bullet-proof method of checking for the lambda context.

tippmar-nr:
Uses `Type.GetInterface()` to look for `ILambdaContext`. Thanks @nrcventura for the idea!


chynesNR (approved):
Very nice!

PR #2439, opened 2024-04-25
Handle SNS message attributes in SQS message body with JSON deserialization

nr-ahemsath:
The logic that handles extracting tracing data from an SNS notification inside an SQS message body was initially ported from the Lambda OpenTracing solution.  However, I manually tested setting the relevant scenario up in AWS (a lambda function that is triggered by an SQS queue, which in turn is subscribed to an SNS notification topic), the format of the SNS message attributes embedded in the SQS message body was different than what the existing logic expected.

This PR:

1. Changes the logic in `TryParseSqsDistributedTraceHeaders` to use `System.Text.Json` to deserialize an SNS event (just the MessageAttributes part) from the SQS message body if it looks like an SNS notification.
2. Updates unit tests accordingly
3. Simplifies the unit tests (now that we're not using our own brittle string parsing logic, I don't see the need to test three different trace data scenarios)
4. Adopts @nrcventura's test case pattern for testing the null and empty records handling tests.


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/NewRelic.Agent.Extensions.csproj
```diff
@@ -17,6 +17,9 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(RootProjectDirectory)\src\NewRelic.Core\NewRelic.Core.csproj" />
+  </ItemGroup>
```
nrcventura, line 22: Does this introduce a circular dependency between NewRelic.Core.csproj and NewRelic.Agent.Extensions.csproj?

tippmar-nr: It does not. `NewRelic.Core` has no project dependencies.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs
```diff
@@ -183,18 +184,23 @@ private static void TryParseSQSDistributedTraceHeaders(dynamic sqsEvent, ITransa
...
+                var snsMessage = WrapperHelpers.DeserializeObject<SnsMessage>((string)record.Body);
+                foreach (var messageAttribute in snsMessage.MessageAttributes)
                 {
-                    sqsHeaders.Add(tracingKey, headerValue);
+                    sqsHeaders.Add(messageAttribute.Key, messageAttribute.Value.Value);
                 }
             }
+            catch
+            {
+                // do nothing
```
tippmar-nr, line 200: Might want to log the exception at Debug level just so we can see what the problem was - might also include `record.Body` in the message, as it'll help diagnosing the failure.

tests/Agent/UnitTests/NewRelic.Agent.Extensions.Tests/Lambda/LambdaEventHelpersTests.cs
```diff
@@ -29,6 +29,39 @@ public class LambdaEventHelpersTests
     private const string W3CTraceParentPayload = "00-da8bc8cc6d062849b0efcf3c169afb5a-7d3efb1b173fecfa-01";
     private const string W3CTraceStatePayload = "33@nr=0-0-33-2827902-7d3efb1b173fecfa-e8b91a159289ff74-1-1.23456-1518469636035";
 
+    // I don't like hardcoding the same values for the tracing data in this JSON string as are already defined above, but
+    // having to escape everything else to make string interpolation possible is gross.
+    private const string SnsBodyJson = @"{
```
tippmar-nr, line 34: I learned a new thing today while working on the integration tests -- you can convert this to a "raw string" and get rid of all the double quotes. 

There's a context menu fix available for me, but it might be coming from ReSharper. 

If you change this line to 
```
private const string SnsBodyJson = """
```` 
and change the closing (down around line 63) to `""";` you can drop all the double quotes. 

And if you want to get really fancy, you can, indeed, use string interpolation pretty cleanly here -- use 
```
private const string SnsBodyJson = $$"""
```
and then use double-curly for the string interpolation, eg. 
```
"{{W3CTraceParentKey}}": {
```

Refer to [this](https://github.com/newrelic/newrelic-dotnet-agent/blob/feature/lambda-support/tests/Agent/IntegrationTests/IntegrationTests/RemoteServiceFixtures/AwsLambda/LambdaApplicationLoadBalancerRequestTriggerFixture.cs#L62) from my recent PR for more details.

File: src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs (hunk 2 above)
Comment (nr-ahemsath) [reply src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs]: Good idea, done.

File: src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs [hunk 4]
```diff
@@ -197,9 +197,9 @@ private static void TryParseSQSDistributedTraceHeaders(dynamic sqsEvent, ITransa
                     sqsHeaders.Add(messageAttribute.Key, messageAttribute.Value.Value);
                 }
             }
-            catch
+            catch (Exception e)
             {
-                // do nothing
+                agent.Logger.Log(Logging.Level.Debug, $"Caught exception in TryParseSQSDistributedTraceHeaders. Exception: {e.Message}, record.Body={(string)record.Body}");
```
nrcventura, line 202: I don't think that we should record the message body, because it could result in customer data being stored in an unexpected location. I think that recording the exception is still valuable though.

nrcventura: We need to be careful about flooding the logs with this message though. We should probably restrict the number of times it is recorded. For example, just log the first 3 errors (of any type).

tippmar-nr: Good point. I hadn't thought about data leakage.

tests/Agent/UnitTests/NewRelic.Agent.Extensions.Tests/Lambda/LambdaEventHelpersTests.cs
nr-ahemsath: Wow, that's really great.  Thanks!  And done.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Lambda/LambdaEventHelpers.cs
nr-ahemsath: It's hard to see how useful the log message could be without seeing the body of what's being deserialized.  I feel like we should have some kind of umbrella disclaimer that the agent may write logs that contain sensitive data and customers should be responsible for securing access to the log file on their system.  However, I'd rather get this PR put to bed.

nrcventura: Well, there can be problems when customers have requirements that all data at rest must be encrypted. It's one thing to log metadata about what's happening. And it's another thing to log actual payloads.

nrcventura (approved):

PR #2447, opened 2024-04-29
Lambda - Add tests for `Stream` return type

tippmar-nr:
Adds tests for APIGatewayProxy and ApplicationLoadBalancer requests where the Lambda function returns `Stream` instead of the actual response type.


jaffinito (approved):

tests/Agent/IntegrationTests/Applications/LambdaSelfExecutingAssembly/LambdaSelfExecutingAssembly.csproj
```diff
@@ -17,6 +17,7 @@
     <PackageReference Include="Amazon.Lambda.Serialization.SystemTextJson" Version="2.4.1" />
     <PackageReference Include="Amazon.Lambda.SimpleEmailEvents" Version="3.1.0" />
     <PackageReference Include="Amazon.Lambda.SNSEvents" Version="2.1.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
```
nrcventura, line 20: You shouldn't need to add this to the test app. The test app already has a serialization library included "Amazon.Lambda.Serialization.SystemTextJson".

nrcventura: There are already examples in the test app of how to use that library to manually deserialize and serialize requests and responses.

PR #2448, opened 2024-04-29
summary: Add AWS Lambda support

tippmar-nr:
feat: Add preliminary support for AWS Lambda instrumentation. For details on how to enable this functionality, please contact your New Relic representative.

notice: Due to an ongoing issue with the WiX Toolset, the Windows MSI installer must be run as admin. This will be addressed in an upcoming release.


src/NewRelic.Core/JsonConverters/BedrockHelpers.cs
tippmar-nr, line 1: This file should be renamed to match the class name

tests/Agent/IntegrationTests/ContainerApplications/LambdaFunctionTestApp/payload.json
tippmar-nr, line 1: Do we need this file?

tests/Agent/IntegrationTests/ContainerApplications/LambdaFunctionTestApp/Function.cs
tippmar-nr, line 1: Pretty sure we don't need this file.

src/NewRelic.Core/JsonConverters/LambdaPayloads/SnsMessage.cs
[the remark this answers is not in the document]
nrcventura: That payload is used to extract sns contents from an sqs event's message body. So far that's the only case we we look for another nested event type within the event body of the triggering event. This was functionality that we ported over from the open tracing solution, but we ported it over with a more robust parsing mechanism (so that DT headers could be extracted more reliably).

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -0,0 +1,332 @@
...
+
+                string name = arg.GetType().FullName;
+                agent.Logger.Log(Agent.Extensions.Logging.Level.Debug, $"Checking parameter: {name}");
+                if (_functionDetails.SetEventType(name, idx))
+                {
+                    agent.Logger.Log(Agent.Extensions.Logging.Level.Debug, $"Supported Event Type found: {_functionDetails.EventType}");
+                }
+                else if (!_unsupportedInputTypes.Contains(name))
+                {
+                    agent.Logger.Log(Agent.Extensions.Logging.Level.Warn, $"Unsupported input object type: {name}. Unable to provide additional instrumentation.");
```
tippmar-nr, line 160: Should this be a warning? Customers using "unsupported" types will get flooded with messages.

chynesNR: In the short term, I think we want to know when customers are using types we don't support, but I'd also be ok with reducing it to Debug. You'll only see it once, though -- that's what `_unsupportedInputTypes` is tracking.

tippmar-nr: It's fine as-is for now.

jaffinito (approved):

nrcventura (approved):

PR #2456, opened 2024-04-30
Test a range of lambda event library versions

chynesNR:
I rolled everything back as far as I could without hitting version conflicts, and all the tests still pass.

This also adds the event libraries to Dotty so we can watch for breaking changes.


nr-ahemsath (approved):
Looks good!

nrcventura (approved):

PR #2457, opened 2024-04-30
fix: Fix bug in distributed tracing when `excludeNewrelicHeader` is set to true

nr-ahemsath:
When the agent is [configured to exclude the `newrelic` proprietary distributed tracing header](https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#excludeNewrelicHeader)  on outbound traces (and rely solely on the W3C trace context headers instead), an exception could occur which would show up in agent logs like this:

```
NewRelic  ERROR: [pid: 5412, tid: 73] InsertDistributedTraceHeaders() failed
System.InvalidOperationException: Nullable object must have a value.
at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
at NewRelic.Agent.Core.DistributedTracing.DistributedTracePayloadHandler.BuildTraceParent(IInternalTransaction transaction)
at NewRelic.Agent.Core.DistributedTracing.DistributedTracePayloadHandler.GetOutboundHeader(DistributedTraceHeaderType headerType, IInternalTransaction transaction, DateTime timestamp)
at NewRelic.Agent.Core.DistributedTracing.DistributedTracePayloadHandler.InsertDistributedTraceHeaders[T](IInternalTransaction transaction, T carrier, Action`3 setter)
```

This is because the value of `Transaction.Sampled` was not being set in this scenario.

This PR fixes that issue and adds integration tests to test both configuration options.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs
+++ b/src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs
@@ -74,6 +74,10 @@ namespace NewRelic.Agent.Core.DistributedTracing
                         setter(carrier, Constants.DistributedTracePayloadKeyAllLower, distributedTracePayload);
                     }
                 }
+                else
+                {
+                    transaction.SetSampled(_adaptiveSampler);
+                }
 
                 var createOutboundTraceContextHeadersSuccess = false;
                 try
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/NewRelicConfigModifier.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/NewRelicConfigModifier.cs
@@ -243,6 +243,25 @@ namespace NewRelic.Agent.IntegrationTestHelpers
             }
         }
 
+        /// <summary>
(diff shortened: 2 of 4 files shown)


src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs
```diff
@@ -74,6 +74,10 @@ public void InsertDistributedTraceHeaders<T>(IInternalTransaction transaction, T
                         setter(carrier, Constants.DistributedTracePayloadKeyAllLower, distributedTracePayload);
                     }
                 }
+                else
+                {
+                    transaction.SetSampled(_adaptiveSampler);
```
nrcventura, line 79: Do we need to check if sampled is not already set first?

chynesNR: Nope, `SetSampled()` only does anything if it hasn't been set yet

chynesNR (approved):
Thanks for finishing this up!

nrcventura (approved):

PR #2470, opened 2024-05-07
test: Parallelization for AwsLambda tests

tippmar-nr:
Breaks the AwsLambda integration tests into separate namespaces to allow for parallelization during the all_solutions workflow. 

Also adds a previously-omitted namespace (`Logging.AuditLog`) to the list of namespaces being tested.


tippmar-nr:
> Change looks fine, but I'd like to see how much of a difference it makes before we merge.

Previous runs in the `feature/lambda-support` branch without parallization took more than an hour; this run took about 40 minutes. I think it's a good change.

nr-ahemsath (approved):
👍

chynesNR:
> > Change looks fine, but I'd like to see how much of a difference it makes before we merge.
> 
> Previous runs in the `feature/lambda-support` branch without parallization took more than an hour; this run took about 40 minutes. I think it's a good change.

Looks like the Lambda tests dropped from 43 minutes down to 17, not bad!

PR #2481, opened 2024-05-14
feat: Add support for capturing container id from AWS ECS.

jaffinito:
On hold while we wait for more details about the missing linkage with infra.

## Description

* Updates GetDockerVendorInfo to check for both the v4 and v3 ECS metadata endpoint URI environment variables and if present parse out the docker from JSON obtain from the endpoint.
* Includes unit tests for both v3 and v4.

Notes:
* Matches docker id captured by the ECS and Docker AWS integrations.
* Environment variables are only read if the CGroup v1 and v2 checks fail (always with ECS Fargate).
* Environment variables are only present in ECS (non-fargate and fargate).  No additional web calls will be made outside of that environment.
* Metadata v4 is checked first and if present, v3 will not be used.  v3 is a fallback since it will eventually disappear.
* No integration tests since that would either require running in ECS Fargate or faking the environment variables with a fake locally hosted endpoint.  The unit test does basically the same thing without the extra effort.

Once/If this is approved, I will update the agent specs with our implementation.

# Author Checklist

# Reviewer Checklist
- [x] Perform code review


src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
```diff
@@ -279,37 +282,51 @@ private string GetProcessEnvironmentVariable(string variableName)
...
-                            Log.Finest("Found /proc/self/cgroup but failed to parse Docker container id.");
-                    }
-                    catch (Exception ex)
-                    {
-                        Log.Finest(ex, "Failed to parse Docker container id from /proc/self/cgroup.");
-                        return null;
-                    }
+                    vendorModel = TryGetEcsFargateDockerId();
+                    if (vendorModel == null)
+                        Log.Finest("Found ECS_CONTAINER_METADATA_URI_V4 but failed to parse Docker container id.");
```
chynesNR, line 320: Won't `vendorModel` be null in most cases? Such as if neither environment variable exists?

jaffinito: Your right, I see the problem.  I'll get it fixed.

chynesNR (approved):
Looks good!

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
```diff
@@ -97,13 +100,10 @@ public VendorInfo(IConfiguration configuration, IAgentHealthReporter agentHealth
             if (_configuration.UtilizationDetectDocker)
             {
 #if NETSTANDARD2_0
-                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                var dockerVendorInfo = GetDockerVendorInfo(new FileReaderWrapper(), RuntimeInformation.IsOSPlatform(OSPlatform.Linux));
```
nrcventura, line 103: We might also need to make this call available to .net framework applications so that the ECS metadata checks can be made from .net framework windows containers.

jaffinito: I was thinking about that and while it would be possible to run a .NET Framework app, I am not sure how many folks would actually do that.

I will take another pass at the code.

jaffinito: Latest commit has the changes to allow ECS checks on Framework.

tippmar-nr (approved):
Nice work!

nr-ahemsath (approved):
Makes sense to me. 👍

nrcventura (approved):

PR #2487, opened 2024-05-16
feat: Enable configuration of ignored and expected HTTP status code errors with environment variables

### LINKED ISSUE #2181: Support NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES and NEW_RELIC_ERROR_COLLECTOR_IGNORE_ERROR_CODES environment variables
We can only use either .NET configuration providers and environment variables to configure the New Relic agent. Right now, the error collector can only be configured through a newrelic.config file.

### Feature Description
I would like to see support the NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES and NEW_RELIC_ERROR_COLLECTOR_IGNORE_ERROR_CODES environment variables. Those environment variables are already supported by the Node.js agent:

https://docs.newrelic.com/docs/apm/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration/

### Describe Alternatives
I mean, if the error collector could be configure through any .NET configuration provider it would also work. We could set the configuration through either environment variables, KeyVault, file configuration (appsettings.json), etc. and to be honest that's the way we configure everything through .NET configuration providers except New Relic :(. Sadly that alternative is NOT supported either:

https://forum.newrelic.com/s/hubtopic/aAX8W0000008ci1WAA/net-configure-net-agent-error-collector-with-appsettingsjson

### Additional context
I hope everything is clear and no additional context is needed.

### Priority
[Really Want]

nr-ahemsath:
## Description

This PR resolves #2181 by introducing two new environment variables for configuring agent behavior:
`NEW_RELIC_ERROR_COLLECTOR_IGNORE_ERROR_CODES`
`NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES`

Both env vars support defining multiple HTTP status codes in a comma-separated list, e.g.:
`NEW_RELIC_ERROR_COLLECTOR_IGNORE_ERROR_CODES="401,503"`

The env var for expected error codes also supports defining a range of codes like this:
`NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES="401,500-505"`

The internal logic of the error collector doesn't currently support ranges for ignored error codes, hence the difference.

I will update the description of this PR with a link to the docs site update PR once that is created.

# Author Checklist
- [x] Unit tests, ~Integration tests, and Unbounded tests~ completed

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -2628,10 +2628,10 @@ namespace NewRelic.Agent.Core.Configuration
             var expectedStatusCodesArrayLocal = _localConfiguration.errorCollector.expectedStatusCodes?.Split(StringSeparators.Comma, StringSplitOptions.RemoveEmptyEntries);
             var expectedStatusCodesArrayServer = _serverConfiguration.RpmConfig.ErrorCollectorExpectedStatusCodes;
 
-            var expectedStatusCodesArray = ServerOverrides(expectedStatusCodesArrayServer, expectedStatusCodesArrayLocal);
+            var expectedStatusCodesArray = EnvironmentOverrides(ServerOverrides(expectedStatusCodesArrayServer, expectedStatusCodesArrayLocal)?.ToList(), @"NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES");
 
             ExpectedStatusCodes = ParseExpectedStatusCodesArray(expectedStatusCodesArray);
-            ExpectedErrorStatusCodesForAgentSettings = expectedStatusCodesArray ?? new string[0];
+            ExpectedErrorStatusCodesForAgentSettings = expectedStatusCodesArray ?? new List<string>();
 
             ExpectedErrorsConfiguration = new ReadOnlyDictionary<string, IEnumerable<string>>(expectedErrorInfo);
             ExpectedErrorMessagesForAgentSettings = new ReadOnlyDictionary<string, IEnumerable<string>>(expectedMessages);
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -2628,7 +2628,7 @@ namespace NewRelic.Agent.Core.Configuration
             var expectedStatusCodesArrayLocal = _localConfiguration.errorCollector.expectedStatusCodes?.Split(StringSeparators.Comma, StringSplitOptions.RemoveEmptyEntries);
             var expectedStatusCodesArrayServer = _serverConfiguration.RpmConfig.ErrorCollectorExpectedStatusCodes;
 
-            var expectedStatusCodesArray = EnvironmentOverrides(ServerOverrides(expectedStatusCodesArrayServer, expectedStatusCodesArrayLocal)?.ToList(), @"NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES");
+            var expectedStatusCodesArray = EnvironmentOverrides(ServerOverrides(expectedStatusCodesArrayServer, expectedStatusCodesArrayLocal)?.ToList(), "NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES");
 
             ExpectedStatusCodes = ParseExpectedStatusCodesArray(expectedStatusCodesArray);
             ExpectedErrorStatusCodesForAgentSettings = expectedStatusCodesArray ?? new List<string>();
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -2628,10 +2628,10 @@ private void ParseExpectedErrorConfigurations()
             var expectedStatusCodesArrayLocal = _localConfiguration.errorCollector.expectedStatusCodes?.Split(StringSeparators.Comma, StringSplitOptions.RemoveEmptyEntries);
             var expectedStatusCodesArrayServer = _serverConfiguration.RpmConfig.ErrorCollectorExpectedStatusCodes;
 
-            var expectedStatusCodesArray = ServerOverrides(expectedStatusCodesArrayServer, expectedStatusCodesArrayLocal);
+            var expectedStatusCodesArray = EnvironmentOverrides(ServerOverrides(expectedStatusCodesArrayServer, expectedStatusCodesArrayLocal)?.ToList(), "NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES");
```
nrcventura, line 2631: Is the ToList call necessary?

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -999,19 +999,43 @@ public string IgnoreErrorsAndIgnoreClassesCombineTests(string[] ignoreClasses, s
...
-        [TestCase("401", null, ExpectedResult = new[] { "401" })]
-        public string[] ExpectedStatusCodesSetFromLocalAndServerOverrides(string local, string[] server)
+        [TestCase("401", new[] { "405" }, null,       ExpectedResult = new[] { "405" })]
+        [TestCase("401", new string[0],   null,       ExpectedResult = new string[0])]
+        [TestCase("401", null,            null,       ExpectedResult = new[] { "401" })]
+        [TestCase(null,  null,            "401",      ExpectedResult = new[] { "401" })]
+        [TestCase(null,  new[] { "405" }, "401",      ExpectedResult = new[] { "401" })]
+        [TestCase("402", new string[0],   "401",      ExpectedResult = new[] { "401" })]
+        [TestCase("402", new string[0],   "401, 503", ExpectedResult = new[] { "401", "503" })]
+        public string[] ExpectedStatusCodesSetFromLocalServerAndEnvironmentOverrides(string local, string[] server, string env)
```
nrcventura, line 1009: Should there be tests for ranges in the environment variable?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
nr-ahemsath: Yes, because of the method signature of the overload of `EnvironmentOverrides` being used: https://github.com/newrelic/newrelic-dotnet-agent/blob/ccde882ae97011eb878698c23163741c66851c39/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs#L2385

Is it worth exploring having that take and return an `IEnumerable<string>` instead?

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
nr-ahemsath: There could be, but in this testcase it would only validate that any given string with a hyphen in it isn't mangled by the EnvironmentOverrides method.  We're asserting on the value of `ExpectedErrorStatusCodesForAgentSettings` which is just the same string array returned by the `EnvironmentOverrides(ServerOverrides(.....))` call.

We'd have to do something more like the logic in the `ExpectedStatusCodesParserTests` test case right below these in order to assert on the actual `ExpectedStatusCodes` MatchRule values.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
nrcventura: If expectedStatusCodesArrayLocal, or expectedStatusCodesArrayLocal were already lists, then the ToList call would not be necessary.

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
nrcventura: I just want to be sure that we have a test case that covers that we can handle status code ranges in the environment variable. I'll defer finding an appropriate testing point to you.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
nr-ahemsath: Actually, it wasn't that hard to do. 👍

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
nr-ahemsath: Put in a testcase with a range string, for what it's worth.

chynesNR (approved):

nrcventura (approved):

PR #2488, opened 2024-05-17
fix: Accept inbound tracing headers in Kafka consume method instrumentation

nr-ahemsath:
## Description

This PR fixes a bug in our instrumentation for `Confluent.Kafka`.  On the message `Consume` side, we were not property processing inbound distributed tracing headers.  This is now fixed.

I verified with manual testing of two separate test applications, one a Kafka producer and the other a consumer, that before this change, the two application entities would not be shown as linked in the Distributed Tracing UI.  After this change, they are.

The Kafka containerized integration test is updated to assert for the relevant supportability metrics for trace context payload create and accept.

# Author Checklist
- [X] ~Unit tests~, Integration tests, ~and Unbounded tests~ completed
- [X] ~Performance testing completed with satisfactory results (if required)~ N/A

# Reviewer Checklist
- [x] Perform code review


chynesNR (approved):
Thanks for tracking this down. I was too focused on the exception being thrown last time and missed this.

nrcventura (approved):

jaffinito (approved):

PR #2489, opened 2024-05-17
ci: Deploy a new Lambda layer on Agent release

chynesNR:
When we deploy a new version of the .NET Agent, this will create a tag in the [New Relic Lambda Layers repo](https://github.com/newrelic/newrelic-lambda-layers/), which in turn will trigger creation of a new .NET Lambda layer. The tag needs to be in the format `vX.Y.Z_dotnet`. This PR is waiting for [that workflow](https://github.com/newrelic/newrelic-lambda-layers/pull/228) to be merged.

### DIFF
--- a/.github/workflows/deploy_agent.yml
+++ b/.github/workflows/deploy_agent.yml
@@ -44,6 +44,11 @@ on:
         description: Re-Index the Download Site S3 container
         required: true
         default: true
+      lambdalayer:
+        type: boolean
+        description: Deploy Lambda Layer
+        required: true
+        default: true
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_agent.yml
```diff
@@ -443,12 +448,23 @@ jobs:
       run_id: ${{ needs.get-release-info.outputs.workflow_run_id }}
     secrets: inherit
 
+  deploy-lambda-layer:
+    permissions:
+      contents: write
```
nrcventura, line 453: Do we need to move this permission into the job that is ultimately called? Or do we not need write access?

chynesNR: Yeah I think you're right, since the tag isn't getting created here.

nrcventura (approved):

nr-ahemsath (approved):
Looks correct to me! 👍

chynesNR:
Putting this aside for now. We're going to try using the [auto-layer](https://source.datanerd.us/aws-lambda/auto-layer-releases/pull/8).

PR #2497, opened 2024-05-21
chore: update S3Validator config to expect different file sizes for MSIs, etc

nr-ahemsath:
Update the config of the S3Validator tool:

1. The MSIs are slightly smaller after the WIX version upgrade.
2. We have new versionless zips and tarballs to validate.


chynesNR (approved):
Thanks for fixing these!

nrcventura (approved):

PR #2510, opened 2024-05-28
chore: Adds a supportability metric when CSP is enabled

### LINKED ISSUE #2265: Add supportability metric for when LASP/CSP is enabled
All public docs for LASP/CSP were removed, but the agent still supports the configuration.  We would like to deprecate this feature (all agents), but to do so we will need data about who is using it.  To help with this we should add a supportability metrics when it is enabled and update angler to capture the metric.

### Acceptance Criteria
- Agent has a supportability metric for LASP/CSP enabled
- Angler is capturing the metric

### Dependencies
- [ ] Dependencies on UI
- [ ] Dependencies on Backend
- [ ] Other dependencies ....

Any other way to get usage for this feature?

### Estimates
S

tippmar-nr:
Adds a new supportability metric, `Supportability/FeatureEnabled/CSP` when `IConfiguration.SecurityPoliciesTokenExists` is `true`, indicating that CSP is enabled. 

Also adds an assert in cross-agent tests for CSP to verify the metric is emitted.

Resolves #2265

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs
@@ -88,6 +88,8 @@ namespace NewRelic.Agent.Core.DataTransport
 
                     var securityPoliciesConfiguration = new SecurityPoliciesConfiguration(preconnectResult.SecurityPolicies);
                     EventBus<SecurityPoliciesConfigurationUpdatedEvent>.Publish(new SecurityPoliciesConfigurationUpdatedEvent(securityPoliciesConfiguration));
+
+                    _agentHealthReporter.ReportSupportabilityCountMetric(MetricNames.GetSupportabilityFeatureEnabled("LASP"));
                 }
 
                 var serverConfiguration = SendConnectRequest();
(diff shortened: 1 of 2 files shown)


chynesNR (approved):

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs
```diff
@@ -88,6 +88,8 @@ public void Connect()
 
                     var securityPoliciesConfiguration = new SecurityPoliciesConfiguration(preconnectResult.SecurityPolicies);
                     EventBus<SecurityPoliciesConfigurationUpdatedEvent>.Publish(new SecurityPoliciesConfigurationUpdatedEvent(securityPoliciesConfiguration));
+
+                    _agentHealthReporter.ReportSupportabilityCountMetric(MetricNames.GetSupportabilityFeatureEnabled("LASP"));
```
nrcventura, line 92: We may need to change this to `CSP`, because if I'm remembering things correctly, the original documentation and announcement referred to the feature as `CSP`, and `LASP` was the name used during the original development.

chynesNR: According to the spec it's still LASP internally and CSP externally. Since this is a Supportability metric we may want the internal name but either is probably fine.

nrcventura: I called it out because customers can see supportability metrics. An alternative approach is to consider if we can extract this information in BMDS instead.

tippmar-nr: I'll change it to CSP instead. Easy peasy.

nrcventura (approved):

PR #2536, opened 2024-06-10
feat: Add an option to disable the instrumentation file system watcher.

### LINKED ISSUE #1446: Option to disable FileSystemWatcher in InstrumentationWatcher
Using the Agent in a docker container with read-only filesystem no new extension files are added after startup. Agent monitors extensions folder using FileSystemWatcher that starts a dedicated thread.

### Feature Description
Configuration option to disable FileSystemWatcher from watching extensions folder in InstrumentationWatcher

### Additional context
Each FileSystemWatcher creates dedicated thread. This would provide a way to reduce the number of threads (and inotify instances) used by the application for file watchers in case instrumentation changes are not required after startup. 

### Priority
Nice to Have

jaffinito:
## Description

Adds a config option to disable the instrumentation file system watcher.  This is meant to be used with docker and other read-only file systems where new instrumentation files cannot be added post-startup.

Also checks if logging is enabled or if servelessmode is enabled.

Does not cache the result in a field since this will only be read once on startup.

Option:
`<instrumentation disableFileSystemWatcher="false" />`

Defaults to "false".
Includes a new env var: `NEW_RELIC_DISABLE_FILE_SYSTEM_WATCHER`

Includes integration tests via some changes to the existing Rejit tests.

Closes #1446 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


tippmar-nr (approved):
Nicely done!

chynesNR:
The customer requested the ability to disable file watching, so this implementation works for that. I feel like we have three related scenarios that I had hoped we could consolidate:
1. In read-only mode we don't want to watch for dynamic changes, or write to the log
2. We can reasonably assume we're in read-only mode if we try to write to the log and fail (There's overlap here with the "disable logging" config option)
3. We can also probably assume that we're in read-only mode if we're in serverless mode

But I didn't explain my thought process clearly in the ticket. Sorry about that. I'd be interested in others' input on if this is something we want to continue to work on now or maybe revisit in the future.

nrcventura:
>  I'd be interested in others' input on if this is something we want to continue to work on now or maybe revisit in the future.

I'm ok with merging this PR, but not closing the ticket until we take care of the other things that were mentioned.

jaffinito:
What you listed makes sense for this PR.  I will update things and push a new commit.

nrcventura (changes requested):
The ConfigurationTracker is another place in the code where we watch for updates on the file system.

nrcventura (approved):

chynesNR (approved):
Looks good, thanks!

PR #2541, opened 2024-06-11
chore(deps): Bump Serilog packages to latest version for net481 and net8.0 targets

tippmar-nr:
Updates Serilog packages to latest version for the net481 and net8.0 build targets. These should have been caught by Dotty but were missed somehow. 

Also updates Dotty config to scan for additional serilog package updates.

### DIFF
--- a/.github/workflows/nuget_slack_notifications.yml
+++ b/.github/workflows/nuget_slack_notifications.yml
@@ -95,6 +95,10 @@ jobs:
                  rabbitmq.client
                  restsharp
                  serilog
+                 serilog.extensions.logging
+                 serilog.aspnetcore
+                 serilog.sinks.file
+                 serilog.sinks.console
                  stackexchange.redis
                  system.data.sqlclient"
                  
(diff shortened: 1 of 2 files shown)


.github/workflows/nuget_slack_notifications.yml
```diff
@@ -95,6 +95,10 @@ jobs:
                  rabbitmq.client
                  restsharp
                  serilog
+                 serilog.extensions.logging
+                 serilog.aspnetcore
```
chynesNR, line 99: Aren't these secondary dependencies, rather than things we instrument?

tippmar-nr: They are, but when they update, I think there's a chance our instrumentation could be broken even if indirectly. `Serilog.Extensions.Logging` seems like one we definitely need to keep up with. The sinks I could go either way, but unless someone remembers to update the sinks while applying other Serilog updates, we'll fall behind again.

chynesNR (approved):

.github/workflows/nuget_slack_notifications.yml
chynesNR: We can try it :)

nrcventura (approved):

PR #2545, opened 2024-06-12
chore: Address profiler warnings

chynesNR:
This addresses a number of compiler and linter warnings in the Profiler:
* Moved some large XML objects from stack to heap. Since XML parsing isn't done during instrumentation it feels like a safe change.
* Made a number of member functions `const`
* Addressed a number of uninitialized variables (They were all set before use; this is just to make the compiler happy)
* Disabled warnings for a couple of the CoreCLR headers that we include but are not owned by the profiler.
* Changed a lot of loop iterators from `auto` to `auto&`, to avoid an unnecessary copy. This could in theory result in a behavior change but I don't think any of them were deliberate.

There are still a few linter warnings that I don't want to change (e.g. in the CoreClr headers), but Visual Studio doesn't have a way to exclude files from lint.

All unit and integration tests are passing with the new Profiler.

### DIFF
--- a/src/Agent/NewRelic/Profiler/Common/CorStandIn.h
+++ b/src/Agent/NewRelic/Profiler/Common/CorStandIn.h
@@ -3,6 +3,7 @@
 * SPDX-License-Identifier: Apache-2.0
 */
 #pragma warning(push)
-#pragma warning(disable : 4458)
+#pragma warning(disable: 4458)
+#pragma warning(disable: 26495)
 #include <corhlpr.h>
 #pragma warning(pop)
--- a/src/Agent/NewRelic/Profiler/Common/CorStandIn.h
+++ b/src/Agent/NewRelic/Profiler/Common/CorStandIn.h
@@ -3,7 +3,8 @@
 * SPDX-License-Identifier: Apache-2.0
 */
 #pragma warning(push)
-#pragma warning(disable: 4458)
-#pragma warning(disable: 26495)
+// Since this isn't our code, we don't want to mess with it. These warnings can be safely ignored.
+#pragma warning(disable: 4458) // Scope hides class member with same name.
+#pragma warning(disable: 26495) // Uninitialized member variable, even if it's always set before use.
 #include <corhlpr.h>
 #pragma warning(pop)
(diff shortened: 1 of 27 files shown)


src/Agent/NewRelic/Profiler/Configuration/Configuration.h
```diff
@@ -47,24 +47,24 @@ namespace NewRelic { namespace Profiler { namespace Configuration {
...
                 }
 
                 auto appliedNewRelicConfigurationNode = globalNewRelicConfigurationNode;
+                auto localNewRelicConfigurationDocument = std::make_shared<rapidxml::xml_document<xchar_t>>();
 
                 if (localNewRelicConfiguration.second)
                 {
                     try
                     {
-                        rapidxml::xml_document<xchar_t> localNewRelicConfigurationDocument;
```
chynesNR, line 66: This may have been an actual bug. We were using elements from `localNewRelicConfigurationDocument` after it went out of scope. When I moved it to the heap it started to crash, which is why I moved its allocation.

src/Agent/NewRelic/Profiler/Common/CorStandIn.h
```diff
@@ -3,6 +3,7 @@
 * SPDX-License-Identifier: Apache-2.0
 */
 #pragma warning(push)
-#pragma warning(disable : 4458)
+#pragma warning(disable: 4458)
```
tippmar-nr, line 6: Could you add a comment to describe the warning that's being disabled?

src/Agent/NewRelic/Profiler/Profiler/stdafx.h
```diff
@@ -33,7 +33,10 @@
 
 
 // Profiler Header Files:
+#pragma warning(push)
+#pragma warning(disable: 26495)
```
tippmar-nr, line 37: Here also

tippmar-nr (approved):

src/Agent/NewRelic/Profiler/Common/CorStandIn.h
chynesNR: Thanks, I meant to do that but forgot.

nrcventura (approved):

PR #2552, opened 2024-06-13
chore: Update core tech version scan tool dependencies

nr-ahemsath:
This PR:

1. Fixes a bug in recent runs of this tool by updating to the latest Oktokit
2. Upgrades the tool to net8.0
3. Updates other dependencies
4. Updates the workflow to publish the tool and run from that location rather than hardcoding the output path including the .NET version TFM

Here's a successful run of the workflow from this branch: https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/9503813056/job/26195081810


chynesNR (approved):
Thanks for tracking this down!

PR #2554, opened 2024-06-13
chore: Move Parsing and NewRelic.Core types into other agent projects.

jaffinito:
Beware clicking on Files Changed.... its a bit large and slows down the browser.

## Description

Both the Parsing and NewRelic.Core projects were created to support additional semi-independent libraries living inside our repo, like the old OpenTracing LambdaTracer.  These additional projects no longer exist.  

This returns the various types back to either NewRelic.Agent.Core or NewRelic.Agent.Extensions depending on what the types dependencies are. As for new locations, using existing Agent.Core and Extensions folders were preferred where possible/logical.  In some cases this flattened the namespaces.

NewRelic.Core includes types that depended on Newonsoft.Json rather than in the wrapper projects, to avoid the wrapper having a dependency on Newtonsoft.Json. NewRelic.Core ILRepacks Newtonsoft.Json so it is available here for sure.  To account for this, NewRelic.Agent.Extensions is now includes a reference to NewRelic.Agent.Extensions and ILRepacks it, just like NewRelic.Core did.  This also moves the output folders from inside the project folder to the src/_build folder with the other libraries.

HomeBuilder was updated to remove the Parsing and NewRelic.Core libraries and change the search location for NewRelic.Agent.Extensions .

MsiInstaller was updated to remove the Parsing and NewRelic.Core libraries from its items.  I did not add an action to remove the older libraries since they should not interfere with anything just sitting on disk,

ArtifactBuilder was updated to account for the removals.

Any unit tests were moved to the matching projects.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


tippmar-nr (approved):
If the solution builds and all tests pass, I'd say we're in good shape.

chynesNR:
This looks like a nice bit of cleanup. Can you verify that an MSI update-then-uninstall works? That's the only thing I can think of that might be an issue

jaffinito:
I ran a check of the MSI early on, but not after the final.  I will double check it!

jaffinito:
I was able to test the MSI on a server and it worked as expected.  I also ran a quick test app and it was fine.

chynesNR (approved):

PR #2559, opened 2024-06-14
test: Refactor tests using large single-line JSON strings

chynesNR:
This ended up being more tedious than expected. Using raw string literals makes the code much more readable, but the tests are expecting an exact match, and I couldn't just strip the whitespace because some values contain spaces. For that I added an "unformatting" helper function.


tippmar-nr (approved):
Nice. Should really help readability.

PR #2564, opened 2024-06-18
chore: Eliminate "No matching wrapper" message when MEL is disabled

tippmar-nr:
When MEL instrumentation has been disabled in favor of another logging provider, we will no longer emit the (mildly confusing) message: 
```
No matching wrapper found for Microsoft.Extensions.Logging.Logger.Log(Microsoft.Extensions.Logging.LogLevel,Microsoft.Extensions.Logging.EventId,!!0,System.Exception,System.Func`3[!!0,System.Exception,System.String]) in assembly [Microsoft.Extensions.Logging, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60] (requested wrapper name was MicrosoftLogging). This usually indicates misconfigured instrumentation. This method will be ignored.
```

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Wrapper/WrapperMap.cs
+++ b/src/Agent/NewRelic/Agent/Core/Wrapper/WrapperMap.cs
@@ -66,26 +66,32 @@ namespace NewRelic.Agent.Core.Wrapper
 
         public TrackedWrapper Get(InstrumentedMethodInfo instrumentedMethodInfo)
         {
+            var suppressDefaultWrapperDebugMessage = false; 
+
             //Then, see if there's a standard wrapper supporting this method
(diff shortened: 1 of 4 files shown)


nrcventura:
Would the changes be simpler and more testable if the MEL wrapper was still enabled, but behaved as a noop instead? The changes would at least be isolated to a single wrapper.

tippmar-nr:
> Would the changes be simpler and more testable if the MEL wrapper was still enabled, but behaved as a noop instead? The changes would at least be isolated to a single wrapper.

So just have `BeforeWrappedMethod()` return `Delegates.NoOp` when it's "disabled"? That does seem like a much simpler solution...

chynesNR (approved):

PR #2569, opened 2024-06-20
tests: Skip testing deprecated non-async client methods in latest Elastic.Clients.Elasticsearch versions

nr-ahemsath:
In the latest Elastic.Clients.Elasticsearch versions, non-async methods are deprecated.  We previously ignored the deprecation warnings and continued testing these methods.  This PR:

1. Uses preprocessor directives to avoid compiling the non-async methods in the latest ConsoleMF TFMs (net481 and net8.0), which are tied to the latest Elastic.Clients.Elasticsearch library versions.
2. Adds logic to the test code to skip testing the non-async methods for the latest version of Elastic.Clients.Elasticsearch
3. Modifies the `Connect` and `GenerateError` methods to be "async all the way" in order to use `PingSync()` instead of `Ping()` for those methods.

This also introduces the use of a 3rd-party library `XUnit.SkippableFact` to skip tests based on runtime conditions.  I wanted to be explicit that we were skipping test cases rather than just having them pass without actually testing anything.

### DIFF
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchElasticClient.cs
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchElasticClient.cs
@@ -188,7 +188,7 @@ namespace MultiFunctionApplicationHelpers.NetStandardLibraries.Elasticsearch
         }
 
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
-        public override void GenerateError()
+        public override async Task GenerateError()
         {
             // This isn't the password, so connection should fail, but we won't get an error until the Ping
             var settings = new ElasticsearchClientSettings(Address)
@@ -198,7 +198,7 @@ namespace MultiFunctionApplicationHelpers.NetStandardLibraries.Elasticsearch
 
             var client = new ElasticsearchClient(settings);
 
-            var response = _ = client.PingAsync().Result;
+            var response = _ = await client.PingAsync();
 
             if (response.IsSuccess())
             {
--- a/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchElasticClient.cs
+++ b/tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchElasticClient.cs
@@ -47,7 +47,7 @@ namespace MultiFunctionApplicationHelpers.NetStandardLibraries.Elasticsearch
         }
 
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
-        public override async Task Connect()
+        public override async Task ConnectAsync()
         {
             var settings = new ElasticsearchClientSettings(Address)
                     .Authentication(new BasicAuthentication(Username, Password)).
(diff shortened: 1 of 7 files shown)


tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchElasticClient.cs
```diff
@@ -204,9 +198,7 @@ public override void GenerateError()
 
             var client = new ElasticsearchClient(settings);
 
-#pragma warning disable CS0618 // obsolete usage is ok here
-            var response = client.Ping();
-#pragma warning restore CS0618
+            var response = _ = await client.PingAsync();
```
tippmar-nr, line 201: Kind of odd syntax, not sure you need the discard here.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchTests.cs
```diff
@@ -82,27 +100,31 @@ protected ElasticsearchTestsBase(TFixture fixture, ITestOutputHelper output, Cli
             _fixture.Initialize();
         }
 
-        [Fact]
+        [SkippableFact]
         public void Index()
         {
+            Skip.IfNot(_syncMethodsOk, SyncMethodSkipReason);
```
tippmar-nr, line 106: Neat!

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/Elasticsearch/ElasticsearchElasticClient.cs
nr-ahemsath: Ooops, not sure how that happened.  I'll fix it.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Elasticsearch/ElasticsearchTests.cs
```diff
@@ -33,13 +33,27 @@ protected enum ClientType
...
+            // non-async methods are deprecated in the latest Elastic.Clients.Elasticsearch versions
+            if (_clientType != ClientType.ElasticClients ||
+                (_fixture.GetType() != typeof(ConsoleDynamicMethodFixtureCoreLatest) && _fixture.GetType() != typeof(ConsoleDynamicMethodFixtureFWLatest)))
+            {
+                _syncMethodsOk = true;
+            }
+            else
+            {
+                _syncMethodsOk = false;
+            }
```
nrcventura, line 55: It might be simpler to move `_syncMethodsOk` to a be a property on the fixture. You can set the property to false in the fixtures that should not have sync methods. Then the skippable facts can check the property on the fixture to determine if the fact should run or not.

nr-ahemsath: It's not just a matter of the fixture (latest Core or FW) but also which library we're testing (`ClientType.ElasticClients`), so I don't think this would be enough of a simplification to be worth it.  I also like keeping the special-case logic in the test class.

nrcventura: The type checking that we're doing here is my biggest concern, because that's where we defeating the purpose of having an inheritance hierarchy.

It seems like the better place for that setting is within the test subclasses. In those subclasses you can set the appropriate value for syncMethodsOk to false for the couple of cases where we shouldn't test the sync methods.

PR #2591, opened 2024-07-02
SQS: Distributed Tracing support

tippmar-nr:
Adds distributed tracing support to SQS instrumentation.

Tests to follow in a separate PR after this one completes.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text;
 using NewRelic.Agent.Api;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
```diff
@@ -48,25 +54,57 @@ public SqsAttributes(string url)
...
 
-                if (headers == null)
-                {
-                    headers = new Dictionary<string, object>();
-                    wr.Headers = headers;
-                }
+                var messageAttributes = getMessageAttributes(smr);
+
+                // SQS is limited to no more than 10 attributes; if we can't add up to 3 attributes, don't add any
+                if ((messageAttributes.Count + 3 - headersInserted) > 10)
```
chynesNR, line 76: Please make sure there's a test case for this.

tippmar-nr: Yep, working on tests for `SqsHelper` now, will add them to this PR.

nrcventura: We should also confirm that CAT will also add no more than 3 attributes. It may also be possible to add a method that returns how many headers the current configuration would return.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
```diff
@@ -74,19 +82,67 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
-            return Delegates.GetDelegateFor(segment);
+            // modify the request to ask for DT headers in the response message attributes
+            if (action == MessageBrokerAction.Consume)
+            {
+                if (request.MessageAttributeNames == null)
+                    request.MessageAttributeNames = new List<string>();
+
+                request.MessageAttributeNames.Add(NEWRELIC_TRACE_HEADER);
+                request.MessageAttributeNames.Add(W3C_TRACESTATE_HEADER);
+                request.MessageAttributeNames.Add(W3C_TRACEPARENT_HEADER);
```
nrcventura, line 105: We might want to add an experimental extensions API that allows us to get the actual list of configured headers, because we won't always have 3. Sometimes it might just be 1 or 2. Or CAT could be enabled, and none of these 3 will be used.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
tippmar-nr: @nrcventura Do we really need to worry about CAT since it's supposed to be deprecated soon?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
tippmar-nr: Good idea. I'll look into that in a separate PR.

PR #2597, opened 2024-07-08
ci: Update RPM dockerfile to hack around mirrorlist EOL

tippmar-nr:
Updates RPM dockerfile to adjust urls for yum repos as per https://serverfault.com/a/1161847. This is a temporary fix until we can rebuild the RPM packaging process.


chynesNR (approved):
Thanks for the quick fix!

PR #2599, opened 2024-07-08
chore: Exclude pre-compiled modules from update_loaded_modules payload.

jaffinito:
## Description

Our vulnerability management backend uses the assemblies from update_loaded_modules to locate which nuget packages the customer is using.  Pre-compiled assemblies are made from local files and are not part of any nuget package. Likewise, dynamically created assemblies are also not going to be found in nuget packages.

The Assembly type already have a property for IsDynamic that we can read, but there is no such property for pre-compiled assemblies (from [ASP.NET Precompilation](https://learn.microsoft.com/en-us/previous-versions/aspnet/bb398860(v=vs.100))).  I noticed that all of my test pre-compiled assemblies reported 0.0.0.0 as the version unlike any of the other assemblies in my apps.  While it is possible that non pre-compiled  assemblies could have a zerod version, such an assembly would not be useful when trying to link the assembly to a versioned nuget, which is what the data is supposed to enable. We will be using this to exclude assemblies.

The payload sent up via update_loaded_modules is not normally visible to customers.  The environment page for .NET still uses the data sent up during connect and it is not updated.

Changes:
- Excludes any "zerod" version assemblies since they are highly likely to be pre-compiled.
- Excludes assemblies marked as Assembly.IsDynamic = true
- Adds a pair of unit tests for the above
- Modifies the existing unit tests to take into account the changes in logic

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/WireModels/LoadedModuleWireModelCollection.cs
+++ b/src/Agent/NewRelic/Agent/Core/WireModels/LoadedModuleWireModelCollection.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 New Relic, Inc. All rights reserved.
+// Copyright 2020 New Relic, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
@@ -28,8 +28,7 @@ namespace NewRelic.Agent.Core.WireModels
             {
                 if (!TryGetAssemblyName(assembly, out var assemblyName))
                 {
-                    // no way to properly track this assembly
-                    continue;
+                    continue; // we don't want to report this assembly
                 }
 
                 var assemblyDetails = assembly.GetName();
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/WireModels/LoadedModuleWireModelCollection.cs
```diff
@@ -71,11 +70,17 @@ public static LoadedModuleWireModelCollection Build(IList<Assembly> assemblies)
...
+            // We skip any assemblies that cannot be linked to a nuget package.
+
             try
             {
-                if (assembly.IsDynamic)
+                if (assembly.IsDynamic) // skip dynamic assemblies
+                {
+                    assemblyName = null;
+                }
+                else if (assembly.GetName().Version.ToString() == "0.0.0.0") // skip assemblies that have a version of 0.0.0.0 - these are precompiled assemblies
```
chynesNR, line 81: Since we could be checking many thousands of these, it might be faster to use the `==` operator on the `Version`  rather than a string conversion?

jaffinito: Yeah, your right. I will fix it up.

nrcventura (approved):

PR #2602, opened 2024-07-09
SQS: Rework distributed trace tests

tippmar-nr:
Reworks the SQS container test app to use a background service to receive messages, rather than directly receiving them from an API endpoint. The solution is (mildly) convoluted, but works well. 

Updated tests to verify the trace id and parent id match in the produce and consume sides.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Agent.cs
+++ b/src/Agent/NewRelic/Agent/Core/Agent.cs
@@ -477,6 +477,23 @@ namespace NewRelic.Agent.Core
             _customEventTransformer.Transform(eventType, attributes, transaction.Priority);
         }
 
+        public IEnumerable<string> GetConfiguredDTHeaders()
+        {
+            List<string> headers = [];
+            if (_configurationService.Configuration.DistributedTracingEnabled)
+            {
+                headers.Add(Constants.TraceParentHeaderKey);
+                headers.Add(Constants.TraceStateHeaderKey);
+
+                if (!_configurationService.Configuration.ExcludeNewrelicHeader)
+                {
+                    headers.Add(Constants.DistributedTracePayloadKeyAllLower);
+                }
+            }
+
+            return headers;
+        }
+
         public ISimpleSchedulingService SimpleSchedulingService
         {
             get { return _simpleSchedulingService; }
--- a/src/Agent/NewRelic/Agent/Core/Agent.cs
+++ b/src/Agent/NewRelic/Agent/Core/Agent.cs
@@ -477,7 +477,7 @@ namespace NewRelic.Agent.Core
             _customEventTransformer.Transform(eventType, attributes, transaction.Priority);
         }
 
-        public IEnumerable<string> GetConfiguredDTHeaders()
+        public List<string> GetConfiguredDTHeaders()
         {
             List<string> headers = [];
             if (_configurationService.Configuration.DistributedTracingEnabled)
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
@@ -62,19 +62,20 @@ namespace NewRelic.Agent.Extensions.AwsSdk
             return segment;
         }
 
-        public static void InsertDistributedTraceHeaders(ITransaction transaction, object sendMessageRequest, IEnumerable<string> dtHeaders)
+        // SQS allows a maximum of 10 message attributes
+        private const int MaxSQSMessageAttributes = 10;
+
+        public static void InsertDistributedTraceHeaders(ITransaction transaction, object sendMessageRequest, int dtHeaderCount)
         {
             var headersInserted = 0;
-
-            var requiredHeadroom = dtHeaders.Count();
 
             var setHeaders = new Action<object, string, string>((smr, key, value) =>
             {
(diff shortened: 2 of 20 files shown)


tests/Agent/IntegrationTests/ContainerApplications/AwsSdkTestApp/AwsSdkExerciser/AwsSdkExerciser.cs
```diff
@@ -36,6 +36,7 @@ private AmazonSQSClient GetSqsClient()
             var awsCredentials = new Amazon.Runtime.BasicAWSCredentials("dummy", "dummy");
             var config = new AmazonSQSConfig
             {
+                //ServiceURL = "http://localhost:4566",
```
nrcventura, line 39: Do we want to keep this comment around?

tippmar-nr: Good catch. Probably not.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
```diff
@@ -115,6 +115,7 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
             // modify the request to ask for DT headers in the response message attributes
             if (action == MessageBrokerAction.Consume)
             {
+                // TODO: update the distributed trace API to give the set of headers to accept based on current configuration
```
tippmar-nr, line 118: Didn't mean to include this in my last commit. It's a reminder for the next PR that I'm working on...

nr-ahemsath (commented):
We should consider adding assertions on the DT supportability metrics.  See https://github.com/newrelic/newrelic-dotnet-agent/blob/82f601d922f67176cdb30b47fc283c0875a61d91/tests/Agent/IntegrationTests/ContainerIntegrationTests/Tests/KafkaTests.cs#L86-L87 as an example.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
```diff
@@ -61,17 +62,19 @@ public static ISegment GenerateSegment(ITransaction transaction, MethodCall meth
...
             var headersInserted = 0;
 
+            var requiredHeadroom = dtHeaders.Count();
+
             var setHeaders = new Action<object, string, string>((smr, key, value) =>
             {
                 var getMessageAttributes = _getMessageAttributes.GetOrAdd(smr.GetType(), t   => VisibilityBypasser.Instance.GeneratePropertyAccessor<IDictionary>(t, "MessageAttributes"));
                 var messageAttributes = getMessageAttributes(smr);
 
                 // SQS is limited to no more than 10 attributes; if we can't add up to 3 attributes, don't add any
```
chynesNR, line 76: Maybe we should update the comment since it isn't necessarily `3`? And maybe make `10` a `const` so we remember not to mess with it?

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -477,6 +477,23 @@ public void RecordLlmEvent(string eventType, IDictionary<string, object> attribu
             _customEventTransformer.Transform(eventType, attributes, transaction.Priority);
         }
 
+        public IEnumerable<string> GetConfiguredDTHeaders()
```
nrcventura, line 480: We should probably return either List<string> or IList<string> here so that when we evaluate the count, we don't have to enumerate the IEnumerable.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/SqsHelper.cs
```diff
@@ -61,17 +62,19 @@ public static ISegment GenerateSegment(ITransaction transaction, MethodCall meth
             return segment;
         }
 
-        public static void InsertDistributedTraceHeaders(ITransaction transaction, object sendMessageRequest)
+        public static void InsertDistributedTraceHeaders(ITransaction transaction, object sendMessageRequest, IEnumerable<string> dtHeaders)
         {
             var headersInserted = 0;
 
+            var requiredHeadroom = dtHeaders.Count();
```
nrcventura, line 69: This call enumerates the IEnumerable it might be more efficient if we just had access to the List or IList instead.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
```diff
@@ -106,24 +103,24 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
+            // we can't know which header(s) the message might contain, so we have to ask for all of them
             if (action == MessageBrokerAction.Consume)
             {
                 if (request.MessageAttributeNames == null)
                     request.MessageAttributeNames = new List<string>();
 
-                request.MessageAttributeNames.Add(NEWRELIC_TRACE_HEADER);
-                request.MessageAttributeNames.Add(W3C_TRACESTATE_HEADER);
-                request.MessageAttributeNames.Add(W3C_TRACEPARENT_HEADER);
+                request.MessageAttributeNames.Add(Constants.DistributedTracePayloadKeyAllLower);
```
nrcventura, line 119: Should we just add all of the headers returned by the GetConfiguredDTHeaders call here? We are already calling it for the produce action.

nrcventura (approved):

PR #2618, opened 2024-07-11
chore: Dotty refactor

tippmar-nr:
Significant refactoring of Dotty: 
* Disables nullable value types, as they're annoying and mostly used incorrectly
* Introduces the use of the `Nuget.Protocol` package to handle queries to the Nuget API rather than direct http client calls
* Eliminates a number of no longer needed classes
* Adds `packageInfo.json` to contain the set of core tech nuget packages we want to scan rather than passing the list of packages as command-line arguments from the workflow
* Introduces configuration options in `packageInfo.json` to ignore patch and/or minor version updates. 
* Updates `elastic.clients.elasticsearch` to ignore patch version updates, but leaves all other packages as-is (only looking for newly published, non pre-release versions). 
* Adds logic to check for ignored patch and minor versions, with appropriate logging
* Updates the body of Github issues to be a tad more verbose
image
* Changes the workflow to run once a week and passes the timestamp from the previous successful run to Dotty, to use as the lower bound for searching. That ensures we don't miss anything, given the vagaries of what "today" means in different contexts.


chynesNR (approved):
Fancy!

.github/workflows/scripts/nugetSlackNotifications/packageInfo.json
```diff
@@ -0,0 +1,32 @@
...
+    { "packageName": "amazon.lambda.cloudwatchevents", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "amazon.lambda.dynamodbevents", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "amazon.lambda.kinesisevents", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "amazon.lambda.kinesisfirehoseevents", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "amazon.lambda.s3events", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "amazon.lambda.simpleemailevents", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "amazon.lambda.snsevents", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "amazon.lambda.sqsevents", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "elasticsearch.net", "ignorePatch": false, "ignoreMinor": false },
+    { "packageName": "elastic.clients.elasticsearch", "ignorePatch": true, "ignoreMinor": false },
```
nrcventura, line 13: Should we add an optional json field that allows us to explain why certain releases are ignored?

tippmar-nr: Sure!

chynesNR (approved):

PR #2620, opened 2024-07-11
feat: Instrumentation for Amazon Simple Queuing Service (AWSSDK.SQS)

tippmar-nr:
Adds instrumentation for Amazon SQS, including distributed tracing support (Cross-Application Tracing (CAT) is not supported).


chynesNR (approved):
Nice work!

nrcventura (approved):

PR #2621, opened 2024-07-11
chore: Refactor AwsSDKPipelineWrapper to encapsulate SQS handling

tippmar-nr:
Refactors `AwsSdkPipelineWrapper` to encapsulate handling for SQS messages and ensure we don't accidentally process a non-SQS request that happens to match one of the names we're looking for.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
@@ -19,6 +19,8 @@ namespace NewRelic.Providers.Wrapper.AwsSdk
 
         private const string WrapperName = "AwsSdkPipelineWrapper";
         private static readonly ConcurrentDictionary<Type, Func<object, object>> _getRequestResponseFromGeneric = new();
+        private static HashSet<string> _unsupportedRequestTypes = new();
+        private static HashSet<string> _unsupportedSQSRequestTypes = new();
 
         public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
         {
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
```diff
@@ -54,12 +54,25 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
...
+            string requestType = request.GetType().FullName;
+            agent.Logger.Finest($"AwsSdkPipelineWrapper: Request type is {requestType}");
 
-            agent.Logger.Finest("AwsSdkPipelineWrapper: Request type is " + requestType);
+            if (requestType.StartsWith("Amazon.SQS"))
+            {
+                return HandleSQSRequest(instrumentedMethodCall, agent, transaction, request, isAsync, executionContext);
+            }
+
+            agent.Logger.Debug($"AwsSdkPipelineWrapper: Unsupported request type: {requestType}. Returning NoOp delegate.");
```
chynesNR, line 65: I'm not sure we need this and the above log line. This could be pretty verbose if they're using something besides SQS in the AWSSDK.

tippmar-nr: You mean get rid of both lines or just one of them? I'm inclined to keep this one, though I could add some logic to only log the message once per type.

chynesNR: Yeah, I think just one or the other would be fine.

nrcventura (approved):

PR #2623, opened 2024-07-12
ci: Remove CentOS from Linux smoke tests

tippmar-nr:
CentOS is EOL, so there's no need for us to include that distro in our Linux smoke tests.


tippmar-nr:
I'm an idiot. These tests are using CentOS 9...

PR #2632, opened 2024-07-16
feat:  Add Kafka node metrics to enable AWS_MSK_TOPIC entity relationships to be formed

tippmar-nr:
Adds two new metrics that are used to synthesize the entity relationship between AWS MSK and the APM entity:
```
MessageBroker/Kafka/Nodes/<host>:<port>
MessageBroker/Kafka/Nodes/<host>:<port>/<mode>/<topic>
```
where `<host>:<port>` is the Kafka server port,  `<mode>` is either `Produce` or `Consume` and `<topic>` is the Kafka topic name

Updates Kafka container tests to validate the new metrics are present


chynesNR (approved):
Fast work!
I'm not positive collecting just the bootstrap servers is enough to form all the relationships, but we can hash that out during testing.

chynesNR (changes requested):
Actually... now that I think about it, we should probably keep these changes in a feature branch until we're sure they work

PR #2634, opened 2024-07-17
feat: SQS Entity Relationship Updates

tippmar-nr:
Adds additional attributes to enable SQS entity relationships to be formed. Includes unit and integration test updates.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Segments/MessageBrokerSegmentData.cs
+++ b/src/Agent/NewRelic/Agent/Core/Segments/MessageBrokerSegmentData.cs
@@ -24,14 +24,25 @@ namespace NewRelic.Agent.Core.Segments
 
         public MetricNames.MessageBrokerAction Action { get; set; }
 
+        public string MessagingSystemName {get; set;}
+        public string CloudAccountId {get; set;}
+        public string CloudRegion {get; set;}
 
-        public MessageBrokerSegmentData(string vendor, string destination, MetricNames.MessageBrokerDestinationType destinationType, MetricNames.MessageBrokerAction action)
+
+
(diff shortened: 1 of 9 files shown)


src/Agent/NewRelic/Agent/Core/Segments/MessageBrokerSegmentData.cs
```diff
@@ -24,15 +24,26 @@ public class MessageBrokerSegmentData : AbstractSegmentData
...
         public MetricNames.MessageBrokerAction Action { get; set; }
 
+        public string MessagingSystemName {get; set;}
+        public string CloudAccountId {get; set;}
+        public string CloudRegion {get; set;}
 
-        public MessageBrokerSegmentData(string vendor, string destination, MetricNames.MessageBrokerDestinationType destinationType, MetricNames.MessageBrokerAction action)
+
+
+        public MessageBrokerSegmentData(string vendor, string destination, MetricNames.MessageBrokerDestinationType destinationType, MetricNames.MessageBrokerAction action, string messagingSystemName = null, string cloudAccountId = null, string cloudRegion = null)
```
jaffinito, line 33: This is used by RabbitMQ as well.  It looks like these will apply to that instrumentation as well based on the story I am working on.

tippmar-nr: @jaffinito Do you want to merge your branch into this one (or maybe the other way around) so you can leverage the code changes here for RabbitMQ?

jaffinito: I don't even have a branch at this point. I will use this as my base for my work and it should be fine.

PR #2637, opened 2024-07-19
AWS-APM Entity Relationships: RabbitMQ

jaffinito:
## Description

- Adds the required attributes to the RabbitMQ spans for entity relationships.
- Adds missing attributes from latest messaging and span specs
- Do not change the segment or transaction naming, or existing attribute values to reduce impact of existing customers.
- Includes integration tests.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
+++ b/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
@@ -134,12 +134,16 @@ namespace NewRelic.Agent.Core.Attributes
 
         AttributeDefinition<bool, bool> LlmTransaction { get; }
 
-        public AttributeDefinition<string, string> CloudAccountId { get; }
-        public AttributeDefinition<string, string> CloudRegion { get; }
-        public AttributeDefinition<string, string> MessagingSystemName { get; }
+        AttributeDefinition<string, string> CloudAccountId { get; }
+        AttributeDefinition<string, string> CloudRegion { get; }
+         AttributeDefinition<string, string> MessagingSystemName { get; }
         AttributeDefinition<string, string> MessagingDestinationName { get; }
         AttributeDefinition<string, string> BrokerServerAddress { get; }
         AttributeDefinition<int, int> BrokerServerPort { get; }
+        AttributeDefinition<string, string> MessageQueueName { get; }
+        AttributeDefinition<string, string> MessageRoutingKey { get; }
+        AttributeDefinition<string, string> MessagingDestinationRoutingKey { get; }
+        AttributeDefinition<string, string> MessagingDestinationPublishName { get; }
     }
 
 
@@ -1121,11 +1125,37 @@ namespace NewRelic.Agent.Core.Attributes
         public AttributeDefinition<string, string> BrokerServerAddress => _brokerServerAddress ?? (_brokerServerAddress =
             AttributeDefinitionBuilder.CreateString("server.address", AttributeClassification.AgentAttributes)
                 .AppliesTo(AttributeDestinations.SpanEvent)
+                .AppliesTo(AttributeDestinations.TransactionTrace)
                 .Build(_attribFilter));
 
         private AttributeDefinition<int, int> _brokerServerPort;
         public AttributeDefinition<int, int> BrokerServerPort => _brokerServerPort ?? (_brokerServerPort =
             AttributeDefinitionBuilder.CreateInt("server.port", AttributeClassification.AgentAttributes)
+                .AppliesTo(AttributeDestinations.SpanEvent)
+                .AppliesTo(AttributeDestinations.TransactionTrace)
+                .Build(_attribFilter));
+        
+        private AttributeDefinition<string, string> _messageQueueName;
+        public AttributeDefinition<string, string> MessageQueueName => _messageQueueName ?? (_messageQueueName =
+            AttributeDefinitionBuilder.CreateString("message.queueName", AttributeClassification.AgentAttributes)
+                .AppliesTo(AttributeDestinations.All)
+                .Build(_attribFilter));
+
+        private AttributeDefinition<string, string> _messageRoutingKey;
+        public AttributeDefinition<string, string> MessageRoutingKey => _messageRoutingKey ?? (_messageRoutingKey =
+            AttributeDefinitionBuilder.CreateString("message.routingKey", AttributeClassification.AgentAttributes)
+                .AppliesTo(AttributeDestinations.All)
+                .Build(_attribFilter));
+
+        private AttributeDefinition<string, string> _messagingDestinationRoutingKey;
+        public AttributeDefinition<string, string> MessagingDestinationRoutingKey => _messagingDestinationRoutingKey ?? (_messagingDestinationRoutingKey =
+            AttributeDefinitionBuilder.CreateString("messaging.rabbitmq.destination.routing_key", AttributeClassification.AgentAttributes)
+                .AppliesTo(AttributeDestinations.SpanEvent)
+                .Build(_attribFilter));
+
+        private AttributeDefinition<string, string> _messagingDestinationPublishName;
+        public AttributeDefinition<string, string> MessagingDestinationPublishName => _messagingDestinationPublishName ?? (_messagingDestinationPublishName =
+            AttributeDefinitionBuilder.CreateString("messaging.destination_publish.name", AttributeClassification.AgentAttributes)
                 .AppliesTo(AttributeDestinations.SpanEvent)
                 .Build(_attribFilter));
     }
--- a/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
+++ b/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
@@ -142,7 +142,7 @@ namespace NewRelic.Agent.Core.Attributes
         AttributeDefinition<int, int> BrokerServerPort { get; }
         AttributeDefinition<string, string> MessageQueueName { get; }
         AttributeDefinition<string, string> MessageRoutingKey { get; }
-        AttributeDefinition<string, string> MessagingDestinationRoutingKey { get; }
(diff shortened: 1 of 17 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/BasicPublishWrapper.cs
```diff
@@ -22,7 +22,7 @@ public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
 
         public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
         {
-            int rabbitVersion = GetRabbitMQVersion(instrumentedMethodCall);
+            int rabbitVersion = RabbitMqHelper.GetRabbitMQVersion(instrumentedMethodCall);
```
chynesNR, line 25: Can we cache the version here too? Or is there a use case where there might be more than one in use?

jaffinito: I had caching setup, but I must have removed it at one point.  I will re-add it!

tippmar-nr:
You will probably need to add some more unit tests to get our code coverage up to snuff - check the CodeCov report above.

jaffinito:
I will look into more unit tests.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/HandleBasicDeliverWrapper.cs
```diff
@@ -7,13 +7,22 @@
...
 using NewRelic.Agent.Extensions.SystemExtensions;
+using NewRelic.Reflection;
 
 namespace NewRelic.Providers.Wrapper.RabbitMq
 {
     public class HandleBasicDeliverWrapper : IWrapper
     {
         private const string WrapperName = "HandleBasicDeliverWrapper";
 
+        private Func<object, object> _modelGetter;
```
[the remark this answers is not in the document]
jaffinito: I saw a mix of static and instance methods for similar calls. I went with instance fields.  I can change it to static.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/HandleBasicDeliverWrapper.cs
```diff
@@ -70,5 +91,60 @@ IEnumerable<string> GetHeaderValue(IDictionary<string, object> carrier, string k
...
+
+                _hostnameGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(endpoint.GetType(), "HostName");
+                hostname = _hostnameGetter(endpoint);
+
+                _portGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<int>(endpoint.GetType(), "Port");
+                port = _portGetter(endpoint);
+            }
+            catch
+            {
+                _hasGetServerFailed = true;
```
chynesNR, line 144: Nice, this seems much safer. Could we log the error here so we know it failed?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/RabbitMqHelper.cs
```diff
@@ -132,5 +158,74 @@ public static ISegment CreateSegmentForPublishWrappers6Plus(InstrumentedMethodCa
...
+
+                _endpointGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(connection.GetType(), "Endpoint");
+                var endpoint = _endpointGetter(connection);
+
+                _hostnameGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(endpoint.GetType(), "HostName");
+                return _hostnameGetter(endpoint);
+            }
+            catch
+            {
+                _hasGetServerFailed = true;
```
chynesNR, line 198: Here too?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/HandleBasicDeliverWrapper.cs
jaffinito: 👍

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
```diff
@@ -1112,5 +1118,45 @@ private static string IgnoreEmptyAndWhitespaceErrorGroupValues(string errorGroup
...
+
+        private AttributeDefinition<string, string> _messageRoutingKey;
+        public AttributeDefinition<string, string> MessageRoutingKey => _messageRoutingKey ?? (_messageRoutingKey =
+            AttributeDefinitionBuilder.CreateString("message.routingKey", AttributeClassification.AgentAttributes)
+                .AppliesTo(AttributeDestinations.All)
+                .Build(_attribFilter));
+
+        private AttributeDefinition<string, string> _messagingDestinationRoutingKey;
+        public AttributeDefinition<string, string> MessagingDestinationRoutingKey => _messagingDestinationRoutingKey ?? (_messagingDestinationRoutingKey =
+            AttributeDefinitionBuilder.CreateString("messaging.rabbitmq.destination.routing_key", AttributeClassification.AgentAttributes)
```
nrcventura, line 1152: Is this attribute only ever added for RabbitMq? It might surprise us in the future if we add it for another messaging system and see rabbitmq in the name (the property name does not imply that it's a rabbitmq specific attribute).

jaffinito: It is rabbitmq specific, from the messaging spec PR currently in review.  There wasn't anything mentioned about making this for each vendor, however.

I am in favor of making it take a vendor string and using that in the name vs just having this be for rabbitmq.  Thoughts?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/HandleBasicDeliverWrapper.cs
```diff
@@ -70,5 +91,61 @@ IEnumerable<string> GetHeaderValue(IDictionary<string, object> carrier, string k
...
+                port = null;
+            }
+
+            try
+            {
+                _modelGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(instrumentedMethodCall.MethodCall.InvocationTarget.GetType(), "Model");
+                var model = _modelGetter(instrumentedMethodCall.MethodCall.InvocationTarget);
+
+                object connection = null;
+                if (model.GetType().ToString() == "RabbitMQ.Client.Framing.Impl.Model")
```
nrcventura, line 109: Would it be better to see if we can compare the type directly? For example you could have something like the following.
```
ConcurrentDictionary<Type, Action<Type>> _connectionGetter;

modelType = model.GetType()
connectionGetter = _connectionGetter.GetOrAdd(modelType, GetConnectionForType)

if (connectionGetter != null)
{
  connection = connectionGetter(modelType, model)
}

static Action<Type, Object> GetConnectionForType(Type modelType)
{
  if (modelType.ToString() == "framing model")
    return GetConnectionFromFramingModel;
  else if (modelType.ToString() == "autorecovery model")
    return GetConnectionFromAutorecoveryModel;
    
    return null;
}

static object GetConnectionFromFramingModel(Type modelType, object model)
{
   // The content currently in the if blocks for getting the specific connection for that type
}

Doing something like this can avoid the string comparisons every time this instrumentation runs (it would only be run the first time). You would probably need to run a benchmark to see if it's any faster though.
```

src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
nrcventura: Either approach works for me. I think we can either just use a RabbitMQ specific property name here, or make the vendor parameterized. Changing the property name might be the simplest. It can be refactored later if something else needs a similarly named attribute.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/HandleBasicDeliverWrapper.cs
jaffinito: I implemented something like the above and it does work.  Performance is about the same compared to string comparisons.  The initial cost is slightly higher, maybe 2000 ticks more for the above, but the second time its about 500 ticks less.  One advantage of this approach is that it is easier to add new instrumentation logic when new library versions are released.  So far we know that 4, 5, and 6 all need slightly different logic to get at the underlying data.  I am going to push up my commit and we can see what folks think.

nrcventura (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/RabbitMq/RabbitMqHelper.cs
```diff
@@ -195,12 +209,13 @@ public static string GetServerAddress(InstrumentedMethodCall instrumentedMethodC
             }
             catch
             {
+                agent.Logger.Warn("Unable to get RabbitMQ server address/port due to differences in the expected types. Server address/port attributes will not be available.");
```
chynesNR, line 212: Could we pass in the exception to these, too? That should make it easier to fix if we run into an issue in the wild

jaffinito: Done!

chynesNR (approved):

nrcventura (approved):

PR #2639, opened 2024-07-22
chore: Refactor SQS handler to a separate class

tippmar-nr:
Minor refactor to split the SQS request handler out to a separate class.


nr-ahemsath (approved):
👍

PR #2653, opened 2024-07-29
fix: Better Lambda web request input parameter validation.

### LINKED ISSUE #2652: Lambda wrapper should handle invalid payload objects for web requests
When a malformed JSON object is used as input to an AWS Lambda function that takes a web request (`APIGatewayProxyRequest`, `APIGatewayHttpApiV2ProxyRequest` or `ApplicationLoadBalancerRequest`) and the request context is missing the required `Request.Method`  and/or `Request.Path` elements, an exception gets thrown and logged, but it's not obvious why.

Add better handling for this edge case, logging a clear message and returning a `NoOpDelegate` instead of trying to instrument that function invocation.

tippmar-nr:
Adds checking for required parameters (Http Method and Path) for Lambda web requests (`APIGatewayProxyRequest`, `APIGatewayHttpApiV2ProxyRequest` or `ApplicationLoadBalancerRequest`). 

If either or both are missing, we log a message (`Invalid or missing web request parameters. HttpMethod and Path are required for <request type>. Not instrumenting this function invocation.`) and return a NoOp Delegate. 

Resolves #2652


jaffinito (approved):
Looks good!

PR #2675, opened 2024-08-06
feat: Additional metrics and attributes for some instrumented libraries

tippmar-nr:
Adds some additional metrics and attributes for Kafka, Amazon SQS and RabbitMQ libraries.


chynesNR (approved):
Nice work, everyone!

PR #2686, opened 2024-08-08
test: Upgrade to XUnit 2.9.0, fix some test failures

### LINKED ISSUE #2684: Integration tests fail after upgrading to XUnit 2.9.0
**Description**
Several integration tests fail consistently after upgrading to XUnit 2.9.0, with errors similar to the following: 
```
  Error Message:
   System.InvalidOperationException : There is no currently active test.
  Stack Trace:
     at NewRelic.Agent.IntegrationTestHelpers.AgentLogBase.WaitForLogLines(String regularExpression, Nullable`1 timeoutOrZero, Int32 minimumCount) in D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\tests\Agent\IntegrationTests\IntegrationTestHelpers\AgentLogBase.cs:line 169
   at NewRelic.Agent.IntegrationTestHelpers.AgentLogBase.GetReportingAppLink(Nullable`1 timeoutOrZero) in D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\tests\Agent\IntegrationTests\IntegrationTestHelpers\AgentLogBase.cs:line 118
   at NewRelic.Agent.IntegrationTestHelpers.AgentLogBase.GetAccountId(Nullable`1 timeoutOrZero) in D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\tests\Agent\IntegrationTests\IntegrationTestHelpers\AgentLogBase.cs:line 93
   at NewRelic.Agent.IntegrationTests.WCF.Client.WCFClientTestBase.DistributedTracing_Metrics() in D:\a\newrelic-dotnet-agent\newrelic-dotnet-agent\tests\Agent\IntegrationTests\IntegrationTests\WCF\WCFClientTests.cs:line 130
```

The [XUnit release notes for v2.9.0](https://xunit.net/releases/v2/2.9.0) indicate that they fixed a bug related to re-using instances of `TestOutputHelper` between tests. Prior to the fix, reusing an instance would (incorrectly) mix log output between tests; after the fix, the `InvalidOperationException` shown above is thrown. 

The errors occur primarily in the `DistributedTracing` and `WCF` integration test namespaces, indicating that we need to review the implementation of those tests and fix any reuse of `TestOutputHelper` instances between tests. [This comment on the XUnit issue](https://github.com/xunit/xunit/issues/2377#issuecomment-2171891217) may be helpful in our analysis. 

**Expected Behavior**
All integration tests should pass after upgrading to XUnit v2.9.0

P3
M

tippmar-nr:
This PR does the following:
* Re-applies the XUnit 2.9.0 upgrade to all integration test projects
* Modifies `AgentLogBase.GetReportingAppLink()` to use `TryGetLogLine()` instead of `WaitForLogLine()`
* Removes logging that I added in [this commit](https://github.com/newrelic/newrelic-dotnet-agent/commit/cbf4248c1a61c3741dbaf533d4d4793c8ebafd13#diff-6f30640a5ec218e443bcbf334bb43947f57e9f10a1af059525d16eddc5ef3df8R160) several months ago.

These fixes should address the `System.InvalidOperationException : There is no currently active test.` errors we started seeing after the previous upgrade attempt. 

Fixes #2684 (probably...)

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
@@ -164,15 +164,12 @@ namespace NewRelic.Agent.IntegrationTestHelpers
 
             var timeout = timeoutOrZero ?? TimeSpan.Zero;
 
-            //_testLogger?.WriteLine($"{Timestamp} WaitForLogLines  Waiting for expression: {regularExpression}. Duration: {timeout.TotalSeconds:N0} seconds. Minimum count: {minimumCount}");
-
             var timeTaken = Stopwatch.StartNew();
             do
             {
                 var matches = TryGetLogLines(regularExpression).ToList();
                 if (matches.Count >= minimumCount)
                 {
-                    //_testLogger?.WriteLine($"{Timestamp} WaitForLogLines  Matched expression: {regularExpression} in {timeTaken.Elapsed.TotalSeconds:N1}s.");
                     return matches;
                 }
 
@@ -180,7 +177,6 @@ namespace NewRelic.Agent.IntegrationTestHelpers
             } while (timeTaken.Elapsed < timeout);
 
             var message = $"{Timestamp} Log line did not appear a minimum of {minimumCount} times within {timeout.TotalSeconds:N0} seconds.  Expected line expression: {regularExpression}";
-            //_testLogger?.WriteLine(message);
             throw new Exception(message);
         }
 
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
@@ -88,9 +88,9 @@ namespace NewRelic.Agent.IntegrationTestHelpers
 
         public abstract IEnumerable<string> GetFileLines();
 
-        public string GetAccountId(TimeSpan? timeoutOrZero = null)
+        public string GetAccountId()
         {
-            var reportingAppLink = GetReportingAppLink(timeoutOrZero);
+            var reportingAppLink = GetReportingAppLink();
             var reportingAppUri = new Uri(reportingAppLink);
             var accountId = reportingAppUri.Segments[2];
             if (accountId == null)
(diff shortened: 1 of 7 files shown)


tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
```diff
@@ -115,8 +115,8 @@ public string GetCrossProcessId(TimeSpan? timeoutOrZero = null)
 
         public string GetReportingAppLink(TimeSpan? timeoutOrZero = null)
         {
-            var match = WaitForLogLine(AgentReportingToLogLineRegex, timeoutOrZero);
-            return match.Groups[1].Value;
+            var match = TryGetLogLine(AgentReportingToLogLineRegex);
```
nrcventura, line 118: Did the previous call to WaitForLogLine throw an exception previously? Just wondering if we should throw an exception, or return null like you have here?

tippmar-nr: `WaitForLogLine()` does throw an exception after the timeout expires - with the default timeout being 0. 

I might be able to just revert this particular change, though, as I'm thinking now that it was the logging in that I removed from `WaitForLogLines()` that was really causing the exceptions when we initially upgraded to 2.9.0.

tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
```diff
@@ -116,8 +116,7 @@ public string GetCrossProcessId(TimeSpan? timeoutOrZero = null)
         public string GetReportingAppLink(TimeSpan? timeoutOrZero = null)
         {
             var match = WaitForLogLine(AgentReportingToLogLineRegex, timeoutOrZero);
-            return match.Groups[1].Value;
-        }
+            return match.Groups[1].Value;        }
```
nrcventura, line 119: You might want to fix the whitespace on this line.

tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
chynesNR: I'm not sure I follow -- why was logging here the only place triggering the exception?

nrcventura: I think that it was some of the contexts under which the WaitForLogLines method was called, that triggered the problem.

tippmar-nr: The crux of the issue seems to be trying to write log lines (like `WaitForLogLines()` did prior to this PR) after the test execution is over (i.e., while evaluating the results in a `[Fact]` method). 

I'm refactoring this whole chunk of code to use `TryGetLogLine()` instead and throw an exception if not found - that should match the previous behavior.

nrcventura (approved):

PR #2694, opened 2024-08-16
chore: Update Serilog to latest

tippmar-nr:
Updates Serilog usage in `Core.csproj` to latest versions. Major version bumps across almost all Serilog packages we reference. 

The update required an ILRepack of `System.Diagnostics.DiagnosticSource`, which in turn required a refactor of `GuidGenerator` to use reflection to locate the customer-referenced `DiagnosticSource` assembly (if it exists) and use that assembly to get the trace id.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Utilities/GuidGenerator.cs
+++ b/src/Agent/NewRelic/Agent/Core/Utilities/GuidGenerator.cs
@@ -16,6 +16,15 @@ namespace NewRelic.Agent.Core.Utilities
         private static readonly RNGCryptoServiceProvider RngCryptoServiceProvider = new RNGCryptoServiceProvider();
         private static Func<string> _traceGeneratorFunc = GetTraceIdFromCurrentActivity;
 
+        private static bool _initialized;
+        private static object _lockObj = new();
+        private static bool _hasDiagnosticSourceReference;
+
+        private static Func<object, object> _fieldReadAccessor;
+        private static Func<object, object> _valuePropertyAccessor;
+        private static Func<object, object> _traceIdGetter;
+        private static Func<object, object> _idFormatGetter;
+
         /// <summary>
         /// Returns a newrelic style guid.
         /// https://source.datanerd.us/agents/agent-specs/blob/2ad6637ded7ec3784de40fbc88990e06525127b8/Cross-Application-Tracing-PORTED.md#guid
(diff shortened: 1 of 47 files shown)


nr-ahemsath (commented):
Did the newer Serilog package(s) have a new dependency on `System.Diagnostics.DiagnosticSource` and/or `System.Threading.Channels`?

chynesNR:
We don't use version filtering for Serilog, but this might be something to keep in mind: 
`If you're deploying to .NET Framework, note that Serilog's assembly version number has been unpinned from the long-running historical 2.0.0 default, and now matches the package version precisely. If you encounter issues, ensure your build is generating valid assembly binding redirects.`

tippmar-nr:
> Did the newer Serilog package(s) have a new dependency on `System.Diagnostics.DiagnosticSource` and/or `System.Threading.Channels`?

`System.Threading.Channels` is a new dependency, but it's always had a dependency on `System.Diagnostics.DiagnosticSoruce`. I'm not sure why we didn't need to ILRepack DiagnosticSource previously, but it seems that we need to do it now.

nrcventura:
I think that ILRepacking DiagnosticsSource is actually causing breaking behavior in the distributed tracing tests. If I'm remembering correctly, we have some instrumentation that attempts to integrate better with the built in tracing support for AspNetCore and HttpClient, and tries to use the value from Activity.Current. Activity is part of DiagnosticSource, and IlRepacking that library causes us to no longer use the apps reference to Activity.Current.

src/Agent/NewRelic/Agent/Core/Utilities/GuidGenerator.cs
```diff
@@ -63,12 +53,48 @@ private static string GenerateTraceId()
...
+            if (idFormat == null || Enum.GetName(idFormat.GetType(), idFormat) != "W3C") // make sure it's in W3C trace id format
+                return null;
 
-            return GenerateTraceId();
+            // get TraceId property
+            var traceIdGetter = VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(value.GetType(), "TraceId");
+            if (traceIdGetter == null)
+                return null;
+                
+            return traceIdGetter(value).ToString();
```
nrcventura, line 97: This seems like a good start but we probably need to cache these visibility bypasser calls, otherwise we'll continue to pay the reflection and code generation price with every call.

tippmar-nr: Added caching, local testing works well.

nrcventura:
It looks like we need to increase our test coverage for the guid generator changes. Is this something that we can reasonably test in a "unit" test, or do we need to track the coverage for this file in a different manner?

tests/Agent/UnitTests/CompositeTests/DistributedTracingTests.cs
```diff
@@ -285,6 +285,54 @@ public void ExternalHttpSpanEvent_HasExpectedAttributes()
             });
         }
 
+        [Test]
+        public void Transaction_Inside_DiagnosticSourceActivity_GeneratesExpectedTraceId()
```
nrcventura, line 289: Since we have the equivalent of this test in the integration test framework, and it will be difficult to capture all of the branching logic in the guidgenerator class, we might want to consider just adding the guidgenerator to the list of things to exclude from code coverage.

tests/Agent/UnitTests/CompositeTests/CompositeTestAgent.cs
```diff
@@ -405,6 +412,45 @@ public void SetEventListenerSamplersEnabled(bool enable)
             CurrentConfiguration.EventListenerSamplersEnabled = enable;
         }
 
+        public void StartActivity()
+        {
+            _defaultActivityIdFormat = Activity.DefaultIdFormat;
+            Activity.DefaultIdFormat = ActivityIdFormat.W3C;
+            _activitySource = new ActivitySource("CompositeTestAgent");
```
nrcventura, line 419: The activitysource should only need to be created once, and we can probably just dispose it when the compositetestagent is disposed. However, the activity creation and and settings for the activity such as the default id format, should probably be pushed into the test classes so that it will be easier to manage the lifetimes of those activities with the using statement or block.

chynesNR:
The refactor of `GuidGenerator` seems to be adding a lot of complexity to a package update. Do we need to update Serilog now? Did it address the deadlock you were seeing?

nrcventura:
> The refactor of `GuidGenerator` seems to be adding a lot of complexity to a package update. Do we need to update Serilog now? Did it address the deadlock you were seeing?

It did not fix the deadlock. So it's something that I think we can take our time with.

tests/Agent/UnitTests/CompositeTests/CompositeTestAgent.cs
tippmar-nr: No longer relevant since I reverted the composite test changes

nrcventura (approved):

src/Agent/NewRelic/Agent/Core/Utilities/GuidGenerator.cs
```diff
@@ -2,21 +2,31 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using NewRelic.Agent.Extensions.Logging;
+using NewRelic.Reflection;
 using System;
-using System.Diagnostics;
-using System.IO;
+using System.Runtime.InteropServices.WindowsRuntime;
```
nrcventura, line 7: Is this being used by something?

tippmar-nr: Good catch. I'll remove it.

nrcventura (approved):

PR #2695, opened 2024-08-16
feat: Use Lambda function name if application name is not set

chynesNR:
Changes the default behavior in Serverless mode to use the function name rather than "My Application" if no name is set.

Unit tests included.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -276,7 +276,7 @@ namespace NewRelic.Agent.Core.Configuration
             {
                 if (ServerlessModeEnabled)
                 {
-                    string name = GetLambdaFunctionName();
+                    string name = _bootstrapConfiguration.ServerlessFunctionName;
                     if (!string.IsNullOrEmpty(name))
                     {
                         Log.Info("Application name from Lambda Function Name.");
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -41,6 +41,8 @@ namespace NewRelic.Agent.Core.Configuration
         private const int MaxExptectedErrorConfigEntries = 50;
         private const int MaxIgnoreErrorConfigEntries = 50;
 
+        private const string DefaultAppNameInConfigFile = "My Application"; // The name in our default newrelic.config file
+
         private static long _currentConfigurationVersion;
         private readonly IEnvironment _environment = new EnvironmentMock();
         private readonly IProcessStatic _processStatic = new ProcessStatic();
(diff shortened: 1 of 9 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -298,6 +314,11 @@ private IEnumerable<string> GetApplicationNames()
             throw new Exception("An application name must be provided");
         }
 
+        private bool ApplicationNameIsBlankOrDefault() => (_localConfiguration.application.name.Count == 0) ||
+            ((_localConfiguration.application.name.Count == 1) && (_localConfiguration.application.name[0] == DefaultAppNameInConfigFile));
+
+        private string GetLambdaFunctionName() => _environment.GetEnvironmentVariable("AWS_LAMBDA_FUNCTION_NAME");
```
tippmar-nr, line 320: Could we push this down to config as a new property? `BootstrapConfig` already reads the value when checking for serverless mode; it could expose the value as a property that could be propagated up. We reference this same environment variable in the Lambda wrapper - it could be refactored to use the config property.

tippmar-nr (dismissed):

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
chynesNR: Oh you mean `AWS_LAMBDA_FUNCTION_NAME`? Yeah that makes sense. Let me give it a try

chynesNR: How does that look?

tippmar-nr: I was thinking it would be a public property and that the reference to the environment variable in the lambda wrapper would be updated to use the property also.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -270,6 +272,20 @@ private IEnumerable<string> GetApplicationNames()
                 return appName.Split(StringSeparators.Comma);
             }
 
+            if (ApplicationNameIsBlankOrDefault())
```
nrcventura, line 275: Do we even need this check? We could just say that the function name has a higher priority than the new relic config, it looks like we already do that for iisexpress and rolename. If customers want to override the function name default, they still have the option to do that via a .net configuration file or the new relic app name environment variable

chynesNR: Yeah we could do it that way. It felt counterintuitive for a NuGet deploy, since the user wouldn't be able to set the name via the `newrelic.config` as they usually do. But maybe you're right, it might be closer to the IISExpress case.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
chynesNR: Ah I see what you mean now. Moving the env var calls out of the wrapper also sounds like a good idea.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Configuration/IConfiguration.cs
```diff
@@ -15,6 +15,8 @@ public interface IConfiguration
         string AgentEnabledAt { get; }
 
         bool ServerlessModeEnabled { get; }
+        string ServerlessFunctionName { get; }
+        string ServerlessFunctionVersion { get; }
```
nrcventura, line 19: We might need to update the tests with the large json blobs to keep the test coverage numbers up with these additions.

nrcventura (approved):

nrcventura (approved):

PR #2715, opened 2024-08-26
test: Update Oracle server image, restore unbounded tests

tippmar-nr:
* Updates our unbounded services configuration to use the [free Oracle 23ai](https://www.oracle.com/database/free/) container image
* Updates the all_solutions workflow to restore Oracle namespace in the unbounded tests project

The updated services config was deployed to our test EC2 instance and Github secrets were updated appropriately.


nr-ahemsath (approved):
Good find on the free container image!

PR #2721, opened 2024-08-29
feat: Oracle instrumentation update to support latest version

tippmar-nr:
With this PR, the .NET agent now instruments the latest `Oracle.ManagedDataAccess` and `Oracle.ManagedDataAccess.Core` libraries.

Oracle unbounded integration tests were migrated to use the ConsoleMultiFunctionApplication pattern and updated to exercise a range of library versions for both .NET Framework and .NET 6+. Note that the existing `EnterpriseLibraryOracleTests` were not migrated.


chynesNR:
Nice work! I think we need to add Oracle back to the integration tests workflow, right?

tippmar-nr:
> Nice work! I think we need to add Oracle back to the integration tests workflow, right?

No, that was done earlier this week via #2715

chynesNR (approved):

nrcventura (approved):

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Oracle/OracleStoredProcedureTests.cs
```diff
@@ -60,48 +63,85 @@ public void Test()
         {
             var expectedMetrics = new List<Assertions.ExpectedMetric>
             {
-                new Assertions.ExpectedMetric { metricName = $@"Datastore/statement/Oracle/{_procedureName.ToLower()}/ExecuteProcedure", callCount = 1 },
-                new Assertions.ExpectedMetric { metricName = $@"Datastore/statement/Oracle/{_procedureName.ToLower()}/ExecuteProcedure", callCount = 1, metricScope = "WebTransaction/MVC/OracleController/OracleParameterizedStoredProcedure"}
+                new() { metricName = $@"Datastore/statement/Oracle/{_storedProcedureName.ToLower()}/ExecuteProcedure", callCount = 1 },
```
nr-ahemsath, line 66: Is there an advantage to the `new()` style as opposed to being explicit about what kind of Assertions. object we're newing up?

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Oracle/OracleStoredProcedureTests.cs
```diff
@@ -60,48 +63,85 @@ public void Test()
...
 
             var expectedQueryParameters = DbParameterData.OracleParameters.ToDictionary(p => p.ParameterName, p => p.ExpectedValue);
 
-            var expectedTransactionTraceQueryParameters = new Assertions.ExpectedSegmentQueryParameters { segmentName = $"Datastore/statement/Oracle/{_procedureName.ToLower()}/ExecuteProcedure", QueryParameters = expectedQueryParameters };
+            var expectedTransactionTraceQueryParameters = new Assertions.ExpectedSegmentQueryParameters { segmentName = $"Datastore/statement/Oracle/{_storedProcedureName.ToLower()}/ExecuteProcedure", QueryParameters = expectedQueryParameters };
 
             var expectedSqlTraces = new List<Assertions.ExpectedSqlTrace>
             {
-                new Assertions.ExpectedSqlTrace
+                new()
```
nr-ahemsath, line 81: Same thing here

nr-ahemsath (commented):
I have a style question but otherwise this looks good

tests/Agent/IntegrationTests/UnboundedIntegrationTests/Oracle/OracleStoredProcedureTests.cs
nrcventura: It's similar to `var` in that the compiler chooses the natural type. In this case it will be `Assertions.ExpectedMetric` because of its use in the List initializer. If you needed it to be some other type, or a child type, then you would need to use the explicit type constructor.

tippmar-nr: I prefer that style as it's a whole lot less unnecessary typing on my part (and a lot less complaining from ReSharper... :wink:)

PR #2723, opened 2024-08-29
Feature work: Azure Function Mode Config

tippmar-nr:
* Added a config setting (using appSettings in newrelic.config) and a corresponding environment variable (`NEW_RELIC_AZURE_FUNCTION_MODE_ENABLED`) to control whether azure function mode is enabled. The default is to be disabled, so further testing will require setting that environment variable. 
* Added a few unit tests for bits of code where azure function mode detection is relevant. This doesn't cover all of the unit tests that are needed, however.
* Refactored `AzureFunctionHelpers` to move that logic into `IConfiguration`, with implementation in `DefaultConfiguration`. This was necessary to accommodate unit testing.


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -2131,7 +2131,81 @@ public bool AiMonitoringRecordContentEnabled
...
+        #region Azure function support
+
+        public bool AzureFunctionModeDetected => _bootstrapConfiguration.AzureFunctionModeDetected;
+
+        private bool? _azureFunctionModeEnabled;
+        public bool AzureFunctionModeEnabled =>
+            _azureFunctionModeEnabled ??
+            (_azureFunctionModeEnabled = EnvironmentOverrides(TryGetAppSettingAsBoolWithDefault("AzureFunctionModeEnabled", false), "NEW_RELIC_AZURE_FUNCTION_MODE_ENABLED")).Value;
+
+        public string AzureFunctionResourceUri
```
chynesNR, line 2143: This will be needed later I'm guessing?

chynesNR (approved):

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
tippmar-nr: Good catch, that was actually supposed to be the implementation for `AzureFunctionResourceId` - making this property unnecessary.

PR #2725, opened 2024-08-30
Finished removing todos and initial code cleanup.

jaffinito:
## Description

Finishes the initial code cleanup and addressing of todos.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/InvokeFunctionAsyncWrapper.cs
```diff
@@ -39,11 +39,15 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
             }
 
             var functionDetails = new FunctionDetails(functionContext, agent);
-            // TODO: add validation for FunctionDetails? 
+            if (!functionDetails.IsValid())
+            {
+                agent.Logger.Debug($"{WrapperName}: FunctionDetails are invalid, can't instrument this invocation.");
```
tippmar-nr, line 44: Not a huge deal, but this could lead to a lot of logging if the customer's config is somehow messed up and every invocation fails validation. I wonder if we couldn't throw an exception here so that instrumentation will eventually get turned off it it happens too many times?

tippmar-nr (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/InvokeFunctionAsyncWrapper.cs
```diff
@@ -103,64 +104,82 @@ internal class FunctionDetails
...
+                    Trigger = trigger;
                 }
-
-                _functionTriggerCache[FunctionName] = Trigger;
             }
-            else
+            catch(Exception ex)
             {
-                Trigger = trigger;
+                agent.Logger.Error(ex, "Error getting Azure Function details.");
```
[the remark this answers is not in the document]
jaffinito: That makes sense, I'll update it.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/InvokeFunctionAsyncWrapper.cs
jaffinito: I was unsure if we wanted to do that.  There are some other places we could do the same as well.  Let me put up a commit with something.

tippmar-nr: Seems reasonable. I'm sure others will comment on the feature PR if they have a better idea...

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/InvokeFunctionAsyncWrapper.cs
```diff
@@ -114,64 +115,82 @@ internal class FunctionDetails
...
+                    var parameters = _parametersGetter(functionDefinition) as IEnumerable;
 
-                    foreach (var propVal in properties.Values)
+                    // Trigger is normally the first parameter, but we'll check all parameters to be sure.
+                    var foundTrigger = false;
+                    foreach (var parameter in parameters)
                     {
-                        if (propVal.GetType().Name.Contains("Trigger"))
+                        // Properties is an IReadOnlyDictionary<string, object>
+                        _propertiesGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<IReadOnlyDictionary<string, object>>(parameter.GetType(), "Properties");
```
chynesNR, line 139: If the trigger isn't the first parameter, won't `_propertiesGetter` always point to the first parameter?

jaffinito: All the parameters should be of the FunctionParameter type, so I would expect the getter to work for each parameter.

chynesNR: Ah ok, that makes sense

chynesNR (approved):

PR #2726, opened 2024-08-30
Azure functions: Unit tests

tippmar-nr:
Adds unit tests for Azure function code in agent and profiler

### DIFF
--- a/src/Agent/NewRelic/Profiler/ConfigurationTest/ConfigurationTest.cpp
+++ b/src/Agent/NewRelic/Profiler/ConfigurationTest/ConfigurationTest.cpp
@@ -14,793 +14,787 @@
 #include <corerror.h>
 using namespace Microsoft::VisualStudio::CppUnitTestFramework;
 
-namespace NewRelic {
-    namespace Profiler {
-        namespace Configuration {
-            namespace Test
-            {
-                TEST_CLASS(ConfigurationTest)
-                {
-                public:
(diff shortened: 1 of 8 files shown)


jaffinito (approved):

src/Agent/NewRelic/Profiler/ConfigurationTest/ConfigurationTest.cpp
```diff
@@ -14,185 +14,237 @@
...
-            Configuration configuration(_agentDisabledXml , _agentDisabledPair);
-            Assert::IsFalse(configuration.IsAgentEnabled());
-        }
-
-        TEST_METHOD(global_xml_agent_enabled_false_local_xml_agent_enabled_missing)
-        {
-            Configuration configuration(_agentDisabledXml, _missingAgentEnabledConfigPair);
-            Assert::IsFalse(configuration.IsAgentEnabled());
-        }
+namespace NewRelic {
```
chynesNR, line 17: I know VisualStudio does this automatically, but the style convention we typically follow is to have all the namespaces on a single line. I don't know the history behind it though.

tippmar-nr: Yeah, I'm (unfortunately) a serial file reformatter. Wish there was a way to tell VS how we wanted this stuff. I can see about reverting, though it's a bit obnoxious and a pretty artibrary format... ;-)

PR #2737, opened 2024-09-04
Feature work: Azure functions integration test framework

tippmar-nr:
* Adds a testing framework and a very basic implementation of an Azure function test app and corresponding integration tests. 
* Updates the `all_solutions` workflow to install required tools and include the `AzureFunction` workspace in the integration test matrix. 
* Updates the profiler to exclude `FUNC.EXE` (the local Azure test tool) from instrumentation. Note that I had to update a number of profiler unit tests also, since I'm now referencing `_systemCalls` in areas that tests previously didn't expect.


.github/workflows/all_solutions.yml
```diff
@@ -279,6 +280,7 @@ jobs:
       # Make this variable true to enable extra data-gathering and logging to help troubleshoot test failures, at the cost of additional time and resources
       enhanced_logging: false
       NR_DOTNET_TEST_SAVE_WORKING_DIRECTORY: 1
+      azure_func_exe_path: C:\ProgramData\chocolatey\lib\azure-functions-core-tools\tools\func.exe
```
chynesNR, line 283: Is there a variable we can leverage so we don't have to list the full path? I could see it potentially changing with e.g. a new version of Chocolatey

chynesNR (approved):
Looks good!

src/Agent/NewRelic/Profiler/Configuration/Configuration.h
```diff
@@ -630,25 +632,34 @@ namespace NewRelic { namespace Profiler { namespace Configuration {
...
+             
+            bool isAzureFunctionsNetHost = Strings::EndsWith(processPath, _X("FUNCTIONSNETHOST.EXE"));
             if (isAzureFunctionsNetHost)
             {
                 LogInfo(L"FunctionNetHost.exe is a valid Azure function command. This process will be instrumented.");
                 return 1;
             }
 
+            // func.exe is the local test tool and should not be instrumented
+            bool isFuncExe = Strings::EndsWith(processPath, _X("FUNC.EXE"));
```
nrcventura, line 656: Do we think this check is specific enough?

tests/Agent/IntegrationTests/Applications/AzureFunctionApplication/.gitignore
nrcventura, line 1: Is this file necessary?

tests/Agent/IntegrationTests/IntegrationTests/AzureFunction/AzureFunctionTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                new() {metricName = "WebTransactionTotalTime/AzureFunction/Function1", callCount = 1},
+                new() {metricName = "Apdex/AzureFunction/Function1", callCount = 1},
+            };
+
+
+            var metrics = _fixture.AgentLog.GetMetrics().ToList();
+            Assertions.MetricsExist(expectedMetrics, metrics);
+
+            //var transactionSample = _fixture.AgentLog.TryGetTransactionSample("WebTransaction/AzureFunction/Function1");
+            //Assert.NotNull(transactionSample);
```
nrcventura, line 70: Is this a todo for a future test?

src/Agent/NewRelic/Profiler/Configuration/Configuration.h
tippmar-nr: I think so, it'll only be hit if the `FUNCTIONS_WORKER_RUNTIME` environment variable is present, so we know at this point that we're in Azure function mode.

tests/Agent/IntegrationTests/Applications/AzureFunctionApplication/.gitignore
tippmar-nr: It was automatically added when I had VS create the azure function app. Maybe not necessary but doesn't hurt anything.

tests/Agent/IntegrationTests/IntegrationTests/AzureFunction/AzureFunctionTests.cs
tippmar-nr: Kind of. Until I (or someone) figures out the interaction with aspnetcore instrumentation, the sampled transaction won't be the function transaction. But the integration test implementation isn't complete at this point either - I just wanted to get something working.

.github/workflows/all_solutions.yml
tippmar-nr: I'd much rather just rely on the azure tools being in the path but that apparently doesn't happen with chocolatey installs. I don't see any obvious way to determine the base path for chocolatey, so I think we'll just have to live with this for now and deal with any breakage that happens in the future.

tippmar-nr: Actually it looks like there's a `ChocolateyInstall` environment variable I can leverage. Change coming soon..

PR #2751, opened 2024-09-11
chore: Remove System.Net.Http reference on .NET Framework. Resolves #2750

### LINKED ISSUE #2750: Remove remaining references to System.Net.Http on .NET Framework
Because it's very easy to build a .NET Framework app against the wrong version of System.Net.Http, we switched back to using HttpWebRequest here: https://github.com/newrelic/newrelic-dotnet-agent/issues/1844

However, `ConnectionManager` references an Exception that only exists in System.Net.Http, which can cause a `FileNotFoundException`. This was recently observed in a customer environment.

chynesNR:
See #2750 for explanation.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
@@ -10,7 +10,9 @@ using System;
 using System.IO;
 using System.Linq;
 using System.Net;
+#if !NETFRAMEWORK
 using System.Net.Http;
+#endif
 using System.Net.Sockets;
 
 namespace NewRelic.Agent.Core.DataTransport
@@ -107,12 +109,14 @@ namespace NewRelic.Agent.Core.DataTransport
             {
                 HandleHttpErrorResponse(ex);
             }
+#if !NETFRAMEWORK // Only available in System.Net.Http
             // Occurs when the agent is unable to connect to APM. The request failed due to an underlying
             // issue such as network connectivity, DNS failure, server certificate validation or timeout.
             catch (HttpRequestException)
             {
                 ScheduleRestart();
             }
+#endif
             // Occurs when the agent connects to APM but the connection gets aborted by the collector
             catch (SocketException)
             {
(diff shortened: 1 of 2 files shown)


tippmar-nr (dismissed):

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -107,12 +109,14 @@ private void Connect()
             {
                 HandleHttpErrorResponse(ex);
             }
+#if !NETFRAMEWORK // Only available in System.Net.Http
             // Occurs when the agent is unable to connect to APM. The request failed due to an underlying
             // issue such as network connectivity, DNS failure, server certificate validation or timeout.
             catch (HttpRequestException)
```
nrcventura, line 115: It looks like we have a unit test that tests this path that needs to be updated.

chynesNR: We do indeed!

nrcventura (approved):

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -107,12 +109,14 @@ private void Connect()
             {
                 HandleHttpErrorResponse(ex);
             }
+#if !NETFRAMEWORK // Only available in System.Net.Http
```
nr-ahemsath, line 112: What is the fallback behavior if we are instrumenting a .NET Framework app?  Is there a different exception type thrown that is caught with similar behavior (`ScheduleRestart()`), or does it go to the generic exception catch which causes the agent to disconnect?

chynesNR: That's a good question. As far as I can tell there's no equivalent, so it will fall through to the shutdown case. This won't change that behavior, though, since it wasn't possible to generate that exception.

nr-ahemsath: Good point. 👍

nr-ahemsath (approved):

PR #2758, opened 2024-09-12
test: Unit Test TFM update, Integration Test workflow dependency update

tippmar-nr:
* Updates unit test projects to target .NET Framework 4.8.1 rather than 4.6.2
* Updates unit test projects to use the latest version of `JustMock` (previously we couldn't use the latest due to the TFM issue)
* Updates the `all_solutions` workflow to install the `HostableWebCore` and `aiohttp` dependencies only for the specific integration test namespaces that require those. Net effect is a reduction of 1-2 minutes execution time for each test that doesn't need either of those dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,7 +37,7 @@ updates:
           - "*"
     allow:
       - dependency-name: "coverlet.collector"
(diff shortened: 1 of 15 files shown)


chynesNR (approved):
And you made it faster?? Nice work!

nrcventura (approved):

PR #2776, opened 2024-09-18
feat: Parse Azure Function HttpTrigger parameters

tippmar-nr:
Implements parsing for request method, request uri and response status code for Azure Functions that utilize `HttpTrigger`. 

Includes updated integration tests that exercise an Azure Function app that uses the `Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore` package and one that does not, to verify that we can correctly parse the data we need in both scenarios.

(Don't be scared by the ridiculously large commit history - this branch started out as a feature work branch back when the Azure Functions feature was still around...)


src/Agent/NewRelic/Agent/Core/Config/Configuration.cs
```diff
@@ -1,6 +1,3 @@
-// Copyright 2020 New Relic, Inc. All rights reserved.
-// SPDX-License-Identifier: Apache-2.0
-
 // ------------------------------------------------------------------------------
```
tippmar-nr, line 1:
```suggestion
// Copyright 2020 New Relic, Inc. All rights reserved.
// SPDX-License-Identifier: Apache-2.0

// ------------------------------------------------------------------------------
```

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/InvokeFunctionAsyncWrapper.cs
```diff
@@ -96,6 +105,32 @@ void InvokeFunctionAsyncResponse(Task responseTask)
...
                     }
+
+                    // only pull response status code here if it's a web trigger and the Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore assembly is not loaded.
+                    if (functionDetails.IsWebTrigger && functionDetails.HasAspNetCoreExtensionReference != null && !functionDetails.HasAspNetCoreExtensionReference.Value)
+                    {
+                        if (_getInvocationResultMethod == null)
+                        {
+                            // GetInvocationResult is a static extension method
+                            // there are multiple GetInvocationResult methods in this type; we want the one without any generic parameters
+                            Type type = functionContext.GetType().Assembly.GetType("Microsoft.Azure.Functions.Worker.FunctionContextBindingFeatureExtensions");
```
chynesNR, line 116: Looks like we're referencing a few long class/type names. Maybe it would be cleaner to make them constants?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/InvokeFunctionAsyncWrapper.cs
```diff
@@ -179,24 +218,107 @@ public FunctionDetails(dynamic functionContext, IAgent agent)
...
+                }
+            }
+
+            if (_genericFunctionInputBindingFeatureGetter != null)
+            {
+                // Get the input binding feature and bind the input from the function context
+                var inputBindingFeature = _genericFunctionInputBindingFeatureGetter.Invoke(features, []);
+                dynamic valueTask = _bindFunctionInputAsync.Invoke(inputBindingFeature, [functionContext]);
+
+                valueTask.AsTask().Wait(); // BindFunctionInputAsync returns a ValueTask, so we need to convert it to a Task to wait on it
```
chynesNR, line 290: Is this really our best option for getting access to the inputs? Feels a bit clunky.

tippmar-nr: This is, unfortunately, exactly what the GRPC-generated code in an Azure function app does between when the `InvokeFunctionAsync()` method is called and when the actual function `Run()` method is called. It's ugly, but it's all we have to work with. 

```
        public async global::System.Threading.Tasks.ValueTask ExecuteAsync(global::Microsoft.Azure.Functions.Worker.FunctionContext context)
        {
            var inputBindingFeature = context.Features.Get<global::Microsoft.Azure.Functions.Worker.Context.Features.IFunctionInputBindingFeature>();
            var inputBindingResult = await inputBindingFeature.BindFunctionInputAsync(context);
            var inputArguments = inputBindingResult.Values;

            if (string.Equals(context.FunctionDefinition.EntryPoint, "AzureFunctionApp.Function1.Run", StringComparison.Ordinal))
            {
                var instanceType = types["AzureFunctionApp.Function1"];
                var i = _functionActivator.CreateInstance(instanceType, context) as global::AzureFunctionApp.Function1;
                context.GetInvocationResult().Value = await i.Run((global::Microsoft.Azure.Functions.Worker.Http.HttpRequestData)inputArguments[0]);
            }
            else if (string.Equals(context.FunctionDefinition.EntryPoint, "AzureFunctionApp.Function2.Run", StringComparison.Ordinal))
            {
                var instanceType = types["AzureFunctionApp.Function2"];
                var i = _functionActivator.CreateInstance(instanceType, context) as global::AzureFunctionApp.Function2;
                context.GetInvocationResult().Value = i.Run((global::Microsoft.AspNetCore.Http.HttpRequest)inputArguments[0]);
            }
            else if (string.Equals(context.FunctionDefinition.EntryPoint, "AzureFunctionApp.QueueTriggerFunction.Run", StringComparison.Ordinal))
            {
                var instanceType = types["AzureFunctionApp.QueueTriggerFunction"];
                var i = _functionActivator.CreateInstance(instanceType, context) as global::AzureFunctionApp.QueueTriggerFunction;
                i.Run((string)inputArguments[0]);
            }
        }
```

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/AzureFunction.csproj
```diff
@@ -13,6 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.0.1" />
```
nrcventura, line 16: Did we confirm if this dependency is always available for Azure Function apps?

nrcventura: Are we directly using anything from this reference?

tests/Agent/IntegrationTests/IntegrationTests/AzureFunction/AzureFunctionHttpTriggerTests.cs
```diff
@@ -33,19 +40,29 @@ protected AzureFunctionHttpTriggerTestsBase(TFixture fixture, ITestOutputHelper
...
 
             _fixture.Initialize();
         }
 
-        [Fact]
-        public void Test()
+        [SkippableFact]
+        public void Test_SimpleInvocationMode()
         {
+            Skip.IfNot(_testMode == AzureFunctionHttpTriggerTestMode.SimpleInvocation);
```
nrcventura, line 64: Is it possible to use a Skip.If here instead of Skip.IfNot? The negation makes it harder for me to read.

nrcventura: Or maybe it would be better to change it to something like `Skip.IfNot(IsSimpleInvocationTest(_testMode))`? It's just the negation around a conditional that makes it harder to read.

nrcventura: The same approach can be taken for the other skippable facts in this test class.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/AzureFunction.csproj
tippmar-nr: I'm using `HttpContext` in `FunctionsHttpProxyingMiddlewareWrapper` - I suppose it's not strictly required, but it sure simplifies things. I could use `dynamic` just as easily, I guess, if this reference is going to be an issue.

The package reference is an ancient version of the common abstractions, so I don't expect it should be a problem for us to reference it directly. I am correct that we'll use a newer version if the customer references in their app, right?

tippmar-nr: ... and switching to `dynamic` ended up being an easy change. Reference removed.

tests/Agent/IntegrationTests/IntegrationTests/AzureFunction/AzureFunctionHttpTriggerTests.cs
tippmar-nr: I kind of wish the "skip" was more of a "do if" kind of condition. But I refactored it a bit and it's (maybe) a little easier to read now. 

The alternative would have been to implement an entirely separate test that was largely duplicative of this one.

nrcventura (approved):

chynesNR (approved):

PR #2784, opened 2024-09-27
feat: Instrument Lambda invocations in AWS SDK

chynesNR:
Instrument calls to Lambdas and Lambda Aliases from the AWS SDK, with span attributes to allow linking the calls to the actual Lambda (specifically the ARN is what's needed). Includes unit and integration tests. We are able to test without a making a successful call to a lambda, so no live testing resources are necessary.

Note that we are deliberately not creating an External segment, as that would result in an extra entity being created.


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public static AfterWrappedMethodDelegate HandleInvokeRequest(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction, dynamic request, bool isAsync, string region)
+        {
+            string functionName = request.FunctionName;
+            string qualifier = request.Qualifier;
+            if (!string.IsNullOrEmpty(qualifier) && !functionName.EndsWith(qualifier))
+            {
+                functionName = $"{functionName}:{qualifier}";
+            }
+            string arn = AwsSdkHelpers.ConstructArn(agent, functionName, region, "");
```
tippmar-nr, line 52: Is the arn going to be a constant value? If so, could we cache it?

chynesNR: Individual Lambda ARNs are constant, but the user might be calling more than one. We could do a name-to-ARN map, but I can see an edge case where it might fail: if you were calling the same Lambda in more than one region but only specifying the name each time. That's admittedly pretty unlikely, though.

tippmar-nr: It just seems like the parsing logic is so complicated that it may impact performance if we don't cache the result. This will get called on every Lambda invocation, right?

chynesNR: Fair point, done.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -1393,5 +1393,17 @@ public void AddFaasAttribute(string name, object value)
             var faasAttrib = _attribDefs.GetFaasAttribute(name);
             TransactionMetadata.UserAndRequestAttributes.TrySetValue(faasAttrib, value);
         }
+
+        public void AddCloudSdkAttribute(string name, object value)
+        {
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                Log.Debug($"AddCloudSdkAttribute - Unable to set Cloud value on transaction because the key is null/empty");
```
nrcventura, line 1401: Should this say `name` instead of `key`?

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Helpers/AwsSdk.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+                else if (alias == null)
+                {
+                    alias = segment;
+                }
+                else
+                {
+                    if (BadInvocations.Add(invocationName))
+                    {
+                        agent?.Logger.Debug($"Unable to parse function name '{invocationName}'");
```
nrcventura, line 83: 1. It seems like it may be possible to run this code concurrently (two async lambda invocations), and we are not using a thread safe collection. This might cause some problems.
2. Since we do not log this at the default log level, we may want to perform a is log level enabled check before storing that invocationName indefinitely.
3. Do we need to put a cap on this collection so that it doesn't grow too large? Should we just log the first N times we encounter this problem?

nrcventura: This comment applies to the other places where we use this collection and log a message.

nrcventura: Would supportability metrics be useful for debugging these types of problems?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                functionName = $"{functionName}:{qualifier}";
+            }
+            string arn;
+            if (!_arnCache.TryGetValue(functionName, out arn))
+            {
+                arn = AwsSdkHelpers.ConstructArn(agent, functionName, region, "");
+                _arnCache.TryAdd(functionName, arn);
+            }
+            var segment = transaction.StartTransactionSegment(instrumentedMethodCall.MethodCall, "InvokeRequest");
```
nrcventura, line 59: Do we need this to be a leaf segment so that these calls will continue to show up within the Externals UI?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+            segment.GetExperimentalApi().MakeLeaf();
```
nrcventura, line 60: This will suppress the HttpClient instrumentation, so we will no longer get the distributed tracing headers added, or any of the other external call attributes that were previously collected for these calls (prior to this instrumentation).

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            try
+            {
+                dynamic metadata = response.ResponseMetadata;
+                string requestId = metadata.RequestId;
+                transaction.AddCloudSdkAttribute("aws.requestId", requestId);
+            }
+            catch (Exception e)
+            {
+                agent.Logger.Debug(e, "Unable to get RequestId from response metadata.");
```
nrcventura, line 41: Is it possible that this could be logged too frequently at the `Debug` level?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            var getResponse = _getResultFromGenericTask.GetOrAdd(task.GetType(), t => VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(t, "Result"));
+            return getResponse(task);
+        }
+
+        private static void SetRequestIdIfAvailable(IAgent agent, ITransaction transaction, dynamic response)
+        {
+            try
+            {
+                dynamic metadata = response.ResponseMetadata;
```
nrcventura, line 35: Is the generic type on the task always publicly visible? Is it possible for the return type to use an `internal` or `private` type? If so we will not be able to use `dynamic` to simplify the property access.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+        private static ConcurrentDictionary<string, string> _arnCache = new ConcurrentDictionary<string, string>();
+
+        private static object GetTaskResult(object task)
+        {
+            if (((Task)task).IsFaulted)
+            {
+                return null;
+            }
+
+            var getResponse = _getResultFromGenericTask.GetOrAdd(task.GetType(), t => VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(t, "Result"));
```
nrcventura, line 27: If you are using dynamic anyways, do you need to use the concurrent dictionary and visibility bypasser combination here?

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
```diff
@@ -275,6 +275,14 @@
     <PackageReference Include="AWSSDK.BedrockRuntime" Version="3.7.301.35" Condition="'$(TargetFramework)' == 'net6.0'" />
   </ItemGroup>
 
+  <!-- AWS SDK -->
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.Lambda" Version="3.7.404.1" Condition="'$(TargetFramework)' == 'net462'" />
```
nrcventura, line 280: Should we be testing an older version of the SDK here (this is the same version as the latest version?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+        public static AfterWrappedMethodDelegate HandleInvokeRequest(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction, dynamic request, bool isAsync, string region)
+        {
+            string functionName = request.FunctionName;
+            string qualifier = request.Qualifier;
+            if (!string.IsNullOrEmpty(qualifier) && !functionName.EndsWith(qualifier))
+            {
+                functionName = $"{functionName}:{qualifier}";
+            }
+            string arn;
+            if (!_arnCache.TryGetValue(functionName, out arn))
```
nrcventura, line 54: Should the cache key be a Tuple of the functionName and region instead of just the functionName?

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
chynesNR: I was following the pattern in `AddLambdaAttribute` and `AddFaasAttribute` but yeah, "name" is probably clearer.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Helpers/AwsSdk.cs
chynesNR: 1. Ideally we're never going to hit this, but that is technically possible. I can make it thread safe.
2. Sure
3. Since it's a Set, they'd have to be invoking a significant number of unique Lambda names, all of which are un-parse-able. That seems pretty unlikely, but a size check is easy enough.
4. I've replaced all the logging calls with a helper method.
5. I don't think so. This is more for troubleshooting individual cases of "my Lambda calls aren't linking". If we saw a large number of parsing failures, we'd still need to see some concrete examples in order to fix it. AWS provides a regex for what's a valid function identifier, and I've tried to cover all the possibilities in the unit tests, so I'm not expecting many surprises.

nrcventura: If we don't think this is likely, and we just need to capture a few examples, maybe we don't need a collection at all and just log the first 10 times we get something we can't parse?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
chynesNR: That's correct, and the expected behavior. If we leave in the HttpClient segment, an additional Entity is created that can't be linked to the Lambda itself, because the URI is not unique enough. I take your point about the missing attributes, though, and will double check with the other devs on this initiative..

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
chynesNR: It's to avoid the HttpClient instrumentation (details in other comment)

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
nrcventura: Yes, and we need to ensure that we are still generating the expected metrics so that the externals UI will work, and the span.kind is correct for the span that is ultimately generated.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
chynesNR: It's possible, though we don't usually limit error logging in our wrappers. I can make it a one-time thing.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
chynesNR: Yes, `ResponseMetadata` is [publicly available](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/AmazonWebServiceResponse.cs#L39) as is the [RequestId](https://github.com/aws/aws-sdk-net/blob/de9ccc8df5a3152df8dbe990ef92c20e4a6e0eca/sdk/src/Core/Amazon.Runtime/ResponseMetadata.cs#L36) within it. These are both things the caller may want, so it's unlikely they'd be made internal/private.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
chynesNR: That's the pattern we use elsewhere, though now that I look at it I'm not sure why. I can simplify it.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
chynesNR: Ah, I forgot to do that. Good idea.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/LambdaInvokeRequestHandler.cs
chynesNR: That's _probably_ not necessary, but would solve that edge case, so... sure.

tippmar-nr:
Looks like we might want to review the CodeCov report - patch coverage is a bit low...

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
```diff
@@ -38,6 +40,24 @@ private string GetRegion(IAgent agent, dynamic requestContext)
...
+                return _accountId;
+            }
+            _accountId = agent.Configuration.AwsAccountId;
+            if (_accountId == null)
+            {
+                _accountId = "";
+            }
+            else if ((_accountId.Length != 12) || _accountId.Any(c => (c < '0') || (c > '9')))
+            {
+                agent.Logger.Warn("Supplied AWS Account Id appears to be invalid: {0}", _accountId);
```
nrcventura, line 56: Do we need to do something here so that this codeql check does not always get triggered? Maybe we don't need to log the actual id, because it should be available in the settings payload?

chynesNR: I opted for logging it instead of putting it in the reported settings but we don't have to do it that way. The account ID isn't sensitive info so I'd probably just dismiss the warning, but it's also not essential if we just want to skip logging it. We can talk it over in pairing.

nrcventura: Yes, we could dismiss it, but every time this code gets touched, it would trigger yet another warning. So it would just be nice to avoid needing to dismiss it every time.

PR #2791, opened 2024-10-01
chore: Remove SharpZipLib, refactor data compression methods

tippmar-nr:
* Removes reference to `SharpZipLib` NuGet package
* Refactors `DataCompressor` to use `DeflateStream` and `GZipStream` (from `System.IO.Compression`) instead
* Updates unit tests
* Adds conditional compiling around a couple of `System.Net.Http` references that were missed in #2751 -- apparently, `SharpZipLib` included a transient reference to `System.Net.Http`, so we were still referencing that assembly even when we removed in the previous PR


chynesNR:
Since our integration tests are just looking at logs rather than transmitted data, how much test coverage do we think we've got for this?

nrcventura:
There really isn't good end to end coverage of this. The closest thing would actually be tests using the mockcollector, but those just test whether or not another .net process can decode the payloads.

PR #2794, opened 2024-10-02
chore: Ensure all `using System.Net.Http;` are conditionally compiled

tippmar-nr:
Minor cleanup to place the last few `using System.Net.Http;` lines behind conditional compilation for non-NETFRAMEWORK build targets.


chynesNR (approved):
Thanks for finding these!

PR #2797, opened 2024-10-03
feat: Custom instrumentation supportability metrics

chynesNR:
Adds supportability metrics for the first 25 custom instrumentation methods. 
Adds a supportability count metric for the total number of custom instrumentation methods
Includes unit tests.


src/Agent/NewRelic/Agent/Core/Metrics/MetricNames.cs
```diff
@@ -628,6 +628,12 @@ public static string GetSupportabilityLibraryVersion(string assemblyName, string
             return SupportabilityLibraryVersionPs + assemblyName + PathSeparator + assemblyVersion;
         }
 
+        private const string SupportabilityCustomInstrumentationPs = SupportabilityPs + "CustomInst" + PathSeparator;
+        public static string GetSupportabilityCustomInstrumentation(string assemblyName, string className, string method)
+        {
+            return SupportabilityCustomInstrumentationPs + assemblyName + PathSeparator + className + PathSeparator + method;
```
tippmar-nr, line 634: I assume this is what you were referring to when you mentioned the metric name is quite long? I wonder if we have an upper limit on the length of metric names? I don't have any good suggesions, though, for how you could make this any shorter and yet still be descriptive.

chynesNR: Exactly, here's one from our integration tests: `Supportability/CustomInst/DistributedTracingApiApplication/NewRelic.Agent.IntegrationTests.Applications.DistributedTracingApiApplication.Program/CallInsertDTHeaders`
I think having the full class name would help with identifying things we want to instrument, but they are quite long. The spec says the limit is 255 bytes for a single metric, and 1MB for the payload. Most customers should only have a few custom instrumentation points (or usually none), so maybe it's ok.

chynesNR:
Closing this for now, but added a backlog ticket to come back to it

PR #2798, opened 2024-10-03
chore: `Dispose()` refactoring

### LINKED ISSUE #2795: Verify correct Dispose() behavior during application shutdown
We have seen reports of an occasional deadlock during application shutdown that could be related to the `Dispose()` implementation in some of our services. The symptom is that the logs indicate that the agent is shutting down, but the application continues to process one or more timers, sending data to the collector and logging about it. Our analysis indicates that this could only happen if the `Dispose()` call in [`AgentManager.Shutdown()`](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/AgentManager.cs#L387) somehow didn't finish -- if it were to finish, the Serilog logger would have been closed and no further log messages could have been emitted. 

There are a few implementations of `Dispose()` that seem a little bit suspect -- in particular, any `Dispose()` method that calls `base.Dispose()` *before* doing its own work should be analyzed for correctness. In general, the correct pattern is for a derived class to clean up its resources prior to calling the base class `Dispose()` method. 

One location we've been able to identify in crash dumps is [`GCSamplerNetCore.Dispose()`](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Samplers/GCSamplerNetCore.cs#L128) - as well as the base class hierarchy from that class. There are, no doubt, others that should be investigated as well.

### Acceptance Criteria
The .NET agent should successfully shutdown and dispose all services without deadlocks

### Estimates
M

tippmar-nr:
Minor refactoring of our Dispose() pattern to ensure that any base-class `Dispose()` method is called *after* the derived class has cleaned up its own resources. Should have zero impact on agent functionality; this is primarily a cosmetic change. 

Also removed a couple of unnecessary `Dispose()` overrides and modified a couple of log messages. The final "agent has been shutdown" message will only show in the logs when everything else related to shutdown is complete; failure to observe that message is an indicator that the shutdown process did not complete for some reason. 

Resolves #2795

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
@@ -57,8 +57,8 @@ namespace NewRelic.Agent.Core.AgentHealth
 
         public override void Dispose()
         {
-            base.Dispose();
             _scheduler.StopExecuting(LogPeriodicReport);
+            base.Dispose();
         }
 
         private void LogPeriodicReport()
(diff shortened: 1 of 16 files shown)


jaffinito (approved):

chynesNR (approved):
Your logic is sound, but I wouldn't be surprised if this creates some other weird race condition :)

Issue #2811, opened 2024-10-07
SQSRequestHandler.ProcessResponse exceptions


indy-singh:
**Description**
While grepping through logs for [this issue](https://github.com/newrelic/newrelic-dotnet-agent/issues/2809). I noticed that this exception is very prevalent (well 1660 lines in 8,216,840 log lines).

```csharp
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at NewRelic.Providers.Wrapper.AwsSdk.SQSRequestHandler.<>c__DisplayClass2_0.<HandleSQSRequest>g__ProcessResponse|1(Task responseTask)
   at NewRelic.Agent.Extensions.Providers.Wrapper.Delegates.<>c__DisplayClass8_1`1.<OnSuccess>g__EndSegmentWithPossibleException|1()
   at NewRelic.Agent.Extensions.Providers.Wrapper.IAgentWrapperExtensions.HandleExceptions(IAgent agent, Action action)
```

**Expected Behavior**
An exception should not be thrown.

**Steps to Reproduce**
Any app that listens to SQS in the background.

**Your Environment**
Windows Server 2022
dotnet8 
On-prem
Hosted via IIS
NewRelicDotNetAgent_10.30.0_x64.msi (deployed using DSC: ADDLOCAL=NETCoreSupport)
newrelic-infra-amd64.1.57.1.msi (deployed using DSC)

I think the related method is

https://github.com/newrelic/newrelic-dotnet-agent/blob/9673ef1f28a419962a0e50b45b5380fbdaa14528/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/SQSRequestHandler.cs#L125

nrcventura:
Thank you for submitting this bug and for the analysis that you have already done. I think that you are correct about one of the lines where this error can occur (the other line is for the async version of the method).

PR #2816, opened 2024-10-07
Update SQSRequestHandler.cs - Fix 2811

indy-singh:
This quick fixes dynamically pulling out a property see https://github.com/newrelic/newrelic-dotnet-agent/issues/2811

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/SQSRequestHandler.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/SQSRequestHandler.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using NewRelic.Agent.Api;
 using NewRelic.Agent.Extensions.AwsSdk;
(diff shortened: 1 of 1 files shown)


nr-ahemsath:
Thank you for submitting a PR.  At first glance, your changes make sense.  However, on our end we need to do some more research to understand how this scenario can occur, and come up with a reliable reproduction case.  Thank your for your patience while we do the analysis.

indy-singh:
Hi @nr-ahemsath,

It will be an empty `ReceiveMessageResponse`, so the scenario is the queue was polled but it had no messages to action. So an empty response T was received. Depending on how the consumer has configured their AWS SDK; the collection on T is either empty or null.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html#API_ReceiveMessage_ResponseElements

https://github.com/aws/aws-sdk-net/blob/b5c10d5d48299f6c68286ee48039bc4ecbc21019/sdk/src/Services/SQS/Generated/Model/ReceiveMessageResponse.cs#L37

Happy to close the PR, if you guys want to deep dive on it.

Cheers,
Indy

nr-ahemsath:
@indy-singh Thanks for the additional info, that will be very helpful.  We're happy to leave the PR open for now.

nr-ahemsath:
@indy-singh Thanks again for the initial take on this fix.  This PR has a more complete version of the fix, plus testing: https://github.com/newrelic/newrelic-dotnet-agent/pull/2833/

Will close this one now.

PR #2822, opened 2024-10-09
fix: Update Profiler to check whether Azure function mode support is enabled

tippmar-nr:
Updates the Profiler to inspect the `NEW_RELIC_AZURE_FUNCTION_MODE_ENABLED` environment variable. 

If the variable doesn't exist, we don't exclude any Azure function processes from instrumentation, which matches behavior prior to implementing Azure function support. 

If the variable does exist, Azure function mode is only enabled if the variable has a value of `1` or `true`.

### DIFF
--- a/src/Agent/NewRelic/Profiler/Configuration/Configuration.h
+++ b/src/Agent/NewRelic/Profiler/Configuration/Configuration.h
@@ -605,13 +605,23 @@ namespace NewRelic { namespace Profiler { namespace Configuration {
                 return false;
             }
 
-            if (IsAzureFunction()) {
-                auto retVal = ShouldInstrumentAzureFunction(processPath, appPoolId, commandLine);
-                if (retVal == 0) {
-                    return false;
+            // returns -1 if env var isn't specified at all
+            auto isAzureFunctionModeEnabled = IsAzureFunctionModeEnabled();
+
+            if (IsAzureFunction())
+            {
+                if (isAzureFunctionModeEnabled == 1) {
+                    auto retVal = ShouldInstrumentAzureFunction(processPath, appPoolId, commandLine);
+                    if (retVal == 0) {
+                        return false;
+                    }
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Profiler/Configuration/Configuration.h
```diff
@@ -605,13 +605,23 @@ namespace NewRelic { namespace Profiler { namespace Configuration {
...
+                    }
+                    if (retVal == 1) {
+                        return true;
+                    }
                 }
-                if (retVal == 1) {
-                    return true;
+                else if (isAzureFunctionModeEnabled == 0)
+                {
+                    return false;
```
nrcventura, line 624: Is this the behavior we want? Or should this be the same behavior as if the environment variable is not present?

tippmar-nr: If the variable is present with a value of 0, the user has (presumably) explicitly said they don't want to instrument Azure functions -- in the current preview, the user generally won't know about the environment variable and thus, it won't exist. So when it *does* exist, we need to respect it. 

I'm really positioning this bit to remain functional when instrumentation is enabled by default.

nrcventura: I'm mostly calling this out because with the current implementation, the setting acts more like an enumeration than a boolean because it has 3 states. Most of our boolean settings are either on or off, and if the setting is missing we default it to either on or or off. In this case we have:

1. Missing - do not perform the process check
2. On - Perform the process check
3. Off - Never instrument anything in an azure function environment

This behavior is different than the behavior we have within the managed agent, where I think Off and missing have the same behavior.

nrcventura (approved):

chynesNR (approved):

nrcventura (approved):

PR #2833, opened 2024-10-14
fix: Handle null or empty SQS messages and/or message attributes

### LINKED ISSUE #2811: SQSRequestHandler.ProcessResponse exceptions

nr-ahemsath:
Resolves #2811.

The SQS wrapper is updated to check for:

- A null list of messages
- A empty list of messages
- Null message attributes

before attempting to accept tracing headers from the message attributes.

The SQS integration tests are updated to:

- Test configuring `AWSConfigs.InitializeCollections` either true (default) or false (the condition that surfaced this bug)
  - Distributed tracing across SQS queues is not expected to work when this is configured to be false
- Test receiving an empty message
- Verify no wrapper exceptions or application errors


nrcventura (approved):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/SQSRequestHandler.cs
```diff
@@ -107,8 +107,11 @@ public static AfterWrappedMethodDelegate HandleSQSRequest(InstrumentedMethodCall
                         var ec = executionContext;
                         var response = ec.ResponseContext.Response; // response is a ReceiveMessageResponse
 
-                        // accept distributed trace headers from the first message in the response
-                        SqsHelper.AcceptDistributedTraceHeaders(transaction, response.Messages[0].MessageAttributes);
+                        if (response.Messages != null && response.Messages.Count > 0 && response.Messages[0].MessageAttributes != null)
```
tippmar-nr, line 110: Could we refactor this block and the one just like it below into a separate method? Or maybe move the logic to inspect `response.Messages` into the `SqsHelper.AcceptDistributedTraceHeaders()` method?

tests/Agent/IntegrationTests/ContainerApplications/AwsSdkTestApp/AwsSdkExerciser/AwsSdkExerciser.cs
```diff
@@ -117,20 +118,30 @@ public async Task<IEnumerable<Message>> SQS_ReceiveMessageAsync(int maxMessagesT
...
-                // delete message
-                await _amazonSqsClient.DeleteMessageAsync(new DeleteMessageRequest
-                {
-                    QueueUrl = _sqsQueueUrl,
-                    ReceiptHandle = message.ReceiptHandle
-                });
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Caught exception processing response messages: {ex.Message}");
```
tippmar-nr, line 144: Not sure this try/catch is useful. But in any case, it can't write to the console... ;-)

nr-ahemsath: That was left over from previous debugging attempts.  However, when I removed the try/catch, I had to introduce another null check (for `response.Messages`) and return an empty list of messages if it is null, when running the tests in the "don't initialize empty collections" mode.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/SQSRequestHandler.cs
nr-ahemsath: Good idea, done.

nrcventura (dismissed):

tippmar-nr (dismissed):

tippmar-nr (approved):

nrcventura (approved):

PR #2838, opened 2024-10-17
feat: New Garbage Collection Metrics Sampler for .NET 6+

tippmar-nr:
## Description

Adds a new garbage collection metrics sampler that leverages the .NET APIs available in .NET 6 and later, instead of the existing (and somewhat error prone) Event Listener sampler. Available only for applications that target .NET 6 and later.

The new sampler is disabled by default but can be enabled either via `newrelic.config`:
```
<appSettings>
    <add key="GCSamplerV2Enabled" value="true" />
</appSettings>
```
or environment variable:
```
NEW_RELIC_GC_SAMPLER_V2_ENABLED=1
```
When the new sampler is enabled, the Event Listener sampler is disabled. Additionally, a supportability metric, `Supportability/Dotnet/GCSamplerV2/Enabled` is sent when the new sampler is enabled.

The set of garbage collection metrics reported by the new sampler are different from those reported by the Event Listener sampler, as detailed below:
| Metric | Event Listener Sampler | Modern Sampler |
|--------|--------|--------|
| `GC/Handles` | ✔️  |  |
| `GC/Induced` | ✔️  |  |
| `GC/Gen0/Size` | ✔️  | ✔️  | 
| `GC/Gen0/Promoted` | ✔️  |   | 
| `GC/Gen0/Collections` | ✔️  | ✔️  | 
| `GC/Gen0/Fragmentation` |   | ✔️  | 
| `GC/Gen1/Size` | ✔️  | ✔️  | 
| `GC/Gen1/Promoted` | ✔️  |   | 
| `GC/Gen1/Collections` | ✔️  | ✔️  | 
| `GC/Gen1/Fragmentation` |   | ✔️  | 
| `GC/Gen2/Size` | ✔️  | ✔️  | 
| `GC/Gen2/Promoted` | ✔️  |   | 
| `GC/Gen2/Survived` | ✔️  |   | 
| `GC/Gen2/Collections` | ✔️  | ✔️  | 
| `GC/Gen2/Fragmentation` |   | ✔️  | 
| `GC/LOH/Size` | ✔️  | ✔️  | 
| `GC/LOH/Survived` | ✔️  |   | 
| `GC/LOH/Collections` | ✔️  | ✔️  | 
| `GC/LOH/Fragmentation` |   | ✔️  | 
| `GC/POH/Size` |   | ✔️  | 
| `GC/POH/Collections` | ✔️  | ✔️  | 
| `GC/POH/Fragmentation` |   | ✔️  | 
| `GC/Heap/Total` | | ✔️ |
| `GC/Heap/Committed` | | ✔️ |
| `GC/Heap/Allocated` | | ✔️ |

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/Samplers/GCSamplerV2.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#if NETSTANDARD
```
chynesNR, line 4: Should we start using `!NETFRAMEWORK` to make our transition away from Standard easier? We can also just address it later.

tippmar-nr: I count 22 `#if NETSTANDARD2_0` and 19 `#if !NETFRAMEWORK` usages currently. It sure wouldn't hurt if we were consistent everywhere, though.

It's been part of my regular routine to find and update these when updating the .NET 8 prototype branch from `main` so don't think it's a huge issue.

chynesNR:
If I'm following the code correctly, the first time through the `PreviousSample` will be empty, so the first delta report will always be a big spike. Would it be better to skip the first one to get a baseline?

tippmar-nr:
> If I'm following the code correctly, the first time through the `PreviousSample` will be empty, so the first delta report will always be a big spike. Would it be better to skip the first one to get a baseline?

@chynesNR Originally, I had code in place to wait until at least once GC had completed before sending up any data - which pretty much matches what OTel is doing. I took that out a couple days ago because it didn't seem like our other GC samplers behaved that way. I can go either way with it. @nrcventura

chynesNR (dismissed):

src/Agent/NewRelic/Agent/Core/Samplers/GCSamplerV2ReflectionHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    ReflectionFailed = true;
+                }
+                else
+                    GCGetMemoryInfo_Invoker = accessor;
+
+                if (!ReflectionFailed)
+                {
+                    paramType = assembly.GetType("System.Boolean");
+                    returnType = assembly.GetType("System.Int64");
```
nrcventura, line 47: I think this can be simplified to typeof(bool) and typeof(long), which more closely matches the actual signature.

nrcventura:
> > If I'm following the code correctly, the first time through the `PreviousSample` will be empty, so the first delta report will always be a big spike. Would it be better to skip the first one to get a baseline?
> 
> @chynesNR Originally, I had code in place to wait until at least once GC had completed before sending up any data - which pretty much matches what OTel is doing. I took that out a couple days ago because it didn't seem like our other GC samplers behaved that way. I can go either way with it. @nrcventura

I think either approach will ultimately be fine. However, if we do wait to report the first value, the charts might be able to scale themselves better.

src/Agent/NewRelic/Agent/Core/Samplers/GCSamplerV2ReflectionHelper.cs
tippmar-nr: Good callout. Making the change now.

nrcventura (approved):

PR #2839, opened 2024-10-18
Clean up POC

nr-ahemsath:
The only real change is that instead of using a switch statement to convert request type names like "PutItemRequest" and "CreateTableRequest" to operation names like `put_item` and `create_table`, instead use a formula of "remove the trailing `Request` and then convert PascalCase to snake_case".

There was also a minor merge conflict in the new Memcache instrumentation which was created after the original POC was done.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/StringsHelper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/StringsHelper.cs
@@ -119,5 +119,34 @@ namespace NewRelic.Agent.Extensions.Parsing
 
             return value;
         }
+
+        public static string ToSnakeCase(this string text)
+        {
+            if (text == null)
+            {
+                throw new ArgumentNullException(nameof(text));
+            }
+            if (text.Length < 2)
+            {
+                return text.ToLowerInvariant();
+            }
+            var sb = new StringBuilder();
+            sb.Append(char.ToLowerInvariant(text[0]));
+            for (int i = 1; i < text.Length; ++i)
+            {
+                char c = text[i];
+                if (char.IsUpper(c))
+                {
+                    sb.Append('_');
+                    sb.Append(char.ToLowerInvariant(c));
+                }
+                else
+                {
+                    sb.Append(c);
+                }
+            }
+            return sb.ToString();
+        }
+
     }
 }
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/DynamoDbRequestHandler.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/DynamoDbRequestHandler.cs
@@ -1,107 +1,30 @@
 // Copyright 2020 New Relic, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using NewRelic.Agent.Api;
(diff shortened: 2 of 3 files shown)


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/StringsHelper.cs
```diff
@@ -119,5 +119,34 @@ public static string RemoveBracketsQuotesParenthesis(string value)
 
             return value;
         }
+
+        public static string ToSnakeCase(this string text)
```
chynesNR, line 123: You might want to combine this with [ConvertToUpperSnakeCase](https://github.com/newrelic/newrelic-dotnet-agent/blob/a77919e5405283fc57ed0251b731c5c1f3047305/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Providers/Wrapper/Constants.cs#L125)

chynesNR (approved):
Looks good!

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/StringsHelper.cs
nr-ahemsath: I did look at that initially, but that code in that class seems very specific to converting Enum members to strings.  I'll make a note to look at it again before merging the feature branch to main.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/DynamoDbRequestHandler.cs
```diff
@@ -112,5 +35,16 @@ public static AfterWrappedMethodDelegate HandleDynamoDbRequest(InstrumentedMetho
...
         }
+
+        private static string GetOperationNameFromRequestType(string requestType)
+        {
+            if (_operationNameCache.ContainsKey(requestType))
+            {
+                return _operationNameCache[requestType];
+            }
+            var operationName = requestType.Replace("Request", string.Empty).ToSnakeCase();
+            _operationNameCache.Add(requestType, operationName);
```
nrcventura, line 46: Do we need to be worried about concurrency with this cache? It currently uses a regular dictionary which is not thread safe.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Memcached/MemcachedHelpers.cs
```diff
@@ -27,7 +26,7 @@ public static ConnectionInfo GetConnectionInfo(string key, object target, IAgent
         {
             if (_hasGetServerFailed)
             {
-                return new ConnectionInfo(DatastoreVendor.Memcached.ToKnownName(), null, -1, null);
+                return new ConnectionInfo(null, -1, null);
```
nrcventura, line 29: Do we need to undo these changes? These new changes are different than what's on the `main` branch.

chynesNR: No, this is related to a fix I made in the POC branch but isn't in `main` yet. The DB name has been moved somewhere more appropriate.

nr-ahemsath: The POC branch for DynamoDB removed the datastore vendor name from the `ConnectionInfo` class.  The new Memcached instrumentation was created and added to main after the DynamoDB POC branch was created.  This change is making the Memcached code align with the new way of doing things on this feature branch.

PR #2841, opened 2024-10-21
Make operation name cache thread safe

nr-ahemsath:


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/DynamoDbRequestHandler.cs
```diff
@@ -43,8 +43,14 @@ private static string GetOperationNameFromRequestType(string requestType)
                 return _operationNameCache[requestType];
             }
             var operationName = requestType.Replace("Request", string.Empty).ToSnakeCase();
-            _operationNameCache.Add(requestType, operationName);
-            return operationName;
+            if (_operationNameCache.TryAdd(requestType, operationName))
```
nrcventura, line 46: It might be simpler to just use GetOrAdd and replace most of the code in this method. Doing that would also avoid using the methods that are not necessarily thread safe.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/DynamoDbRequestHandler.cs
```diff
@@ -23,7 +23,7 @@ public static AfterWrappedMethodDelegate HandleDynamoDbRequest(InstrumentedMetho
 
             // PutItemRequest => put_item,
             // CreateTableRequest => create_table, etc.
-            operation = GetOperationNameFromRequestType(requestType);
+            operation = _operationNameCache.GetOrAdd(requestType, GetOperationNameFromRequestType(requestType));
```
nrcventura, line 26: When calling `GetOrAdd` in this manner, it will also call `GetOperationNameFromRequestType` every time. In order to prevent the `GetOperationNameFromRequestType` method from being invoked every time, you need to pass a reference to that method. Something like the following should work, because the `requestType` is also the cache key, and the key is passed to the method when an entry needs to be added to the cache.
```suggestion
            operation = _operationNameCache.GetOrAdd(requestType, GetOperationNameFromRequestType);
```

nrcventura (approved):

PR #2842, opened 2024-10-21
fix: Fix potential race condition in AWS SDK, AWS Bedrock, and Elastisearch that could lead to an orphaned Transaction.

jaffinito:
## Description

Discovered that calling ITransaction.DetachFromPrimary when in the middle of a transaction can result in `Transaction was garbage collected without ever ending` errors due to the AsyncLocal context being forward only and the other contexts being cleared.

Removed the call from AWS SDK, AWS Bedrock, and Elastisearch instrumentation where the call was made in the middle of transaction.  Left the method call in all other locations since it was done immediately after CreateTransaction was called.

Updated the summary for ITransaction.DetachFromPrimary to include a warning about this use case.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


chynesNR (approved):
Good find! And a good reminder to be careful when copy-pasting from other wrappers. :)

nrcventura (approved):

nrcventura:
We should probably brainstorm how we should describe this fix in the changelog, because the problem does not apply to all application types that use these libraries.

PR #2844, opened 2024-10-22
chore: Add docker ecosystem to Dependabot config

tippmar-nr:
Adds a specific set of folders for Dependabot scanning in the Docker ecosystem, with a monthly schedule. 

If the recommendations turn out to be unhelpful, we can back this change out later.


chynesNR (approved):
It'll be interesting to see what the report looks like.

PR #2851, opened 2024-10-24
feat: Add config option for providing AWS account ID for linking

chynesNR:
Adds the ability to provide an AWS Account ID, which can be used to help link related entities.

This functionality is needed for multiple feature branches which is why I'm merging to `main` first, even though this won't be used yet.

### DIFF
--- a/tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
+++ b/tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
@@ -4439,7 +4439,89 @@ namespace NewRelic.Agent.Core.Configuration.UnitTest
             Assert.That(result, Is.EqualTo("some-service-name"));
         }
 
+        #endregion Cloud
+        [Test]
+        public void Cloud_Section_Parsing_And_Override()
+        {
+            string xmlString = """
+            <?xml version="1.0"?>
+            <configuration xmlns="urn:newrelic-config" agentEnabled="true">
+              <cloud>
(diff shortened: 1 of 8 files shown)


nrcventura (commented):
Looks good to me. I think we just need a few more test updates to increase the patch coverage.

I'm also assuming that the updates to the instrumentation that will use this setting will be handled in separate PRs.

tippmar-nr (approved):

chynesNR:
> Looks good to me. I think we just need a few more test updates to increase the patch coverage.
> 
> I'm also assuming that the updates to the instrumentation that will use this setting will be handled in separate PRs.

Correct, there are separate tasks for that work and testing

nrcventura (approved):

tippmar-nr (approved):

tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -4439,8 +4439,93 @@ public void AzureFunctionServiceName_ShouldReturnServiceName_WhenEnvironmentVari
             Assert.That(result, Is.EqualTo("some-service-name"));
         }
 
+        #endregion Cloud
```
nr-ahemsath, line 4442: Is this where the `Cloud` region should end?

chynesNR: Well that's weird. Not sure how I swapped the begin and end. Good catch!

nrcventura (approved):

PR #2812, opened 2024-10-07
feat: Environment variable naming consistency

tippmar-nr:
## Description

Unifies all environment variable names used by the .NET agent so that they start with `NEW_RELIC_` everywhere. 

Previously "inconsistent" environment variable names (i.e., those starting with `NEWRELIC_`) are still supported, but that support is expected to be removed in the next major release of the .NET agent.

PR #2852, opened 2024-10-24
fix: Revert environment variable name change in installers and scripts

tippmar-nr:
This PR reverts changes (introduced in #2812) to environment variable names that are set in our installers (MSI, Nuget, etc.) and in various Linux scripts. 

The .NET agent still supports both the "old style" environment variable names (those starting with or containing `NEWRELIC_`) as well as the "new style" environment variable names (those starting with or containing `NEW_RELIC_`) as called out in #2812. This change simply reverts our installers, etc. to use the "old" names so that existing public documentation related to installing the .NET agent remains correct.


build/Linux/build/common/run.sh
```diff
@@ -2,4 +2,4 @@
 
 # This script can be used to run a dotnet application with New Relic monitoring.
 
-CORECLR_NEW_RELIC_HOME=${CORECLR_NEW_RELIC_HOME:-/usr/local/newrelic-dotnet-agent} CORECLR_ENABLE_PROFILING=1 CORECLR_PROFILER={36032161-FFC0-4B61-B559-F6C5D41BAE5A} CORECLR_PROFILER_PATH=$CORECLR_NEW_RELIC_HOME/libNewRelicProfiler.so $@
+NRHOME=${CORECLR_NEWRELIC_HOME:-${CORECLR_NEW_RELIC_HOME:-/usr/local/newrelic-dotnet-agent}} CORECLR_ENABLE_PROFILING=1 CORECLR_PROFILER={36032161-FFC0-4B61-B559-F6C5D41BAE5A} CORECLR_PROFILER_PATH=$NRHOME/libNewRelicProfiler.so $@
```
tippmar-nr, line 5: This is intentional; uses either the "old" or "new" environment variable, whichever is found first

build/Linux/build/common/setenv.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 
-if [ -z "$CORECLR_NEW_RELIC_HOME" ]; then
-    echo "CORECLR_NEW_RELIC_HOME is undefined"
+if [ -z "$CORECLR_NEWRELIC_HOME" ]; && [ -z "$CORECLR_NEW_RELIC_HOME" ]; then
+    echo "CORECLR_NEWRELIC_HOME is undefined"
 else
+    NRHOME=${CORECLR_NEWRELIC_HOME:-${CORECLR_NEW_RELIC_HOME}}
```
tippmar-nr, line 6: Also intentional, use the old or new value, whichever is found first

build/Packaging/AzureSiteExtension/Content/install.ps1
```diff
@@ -320,7 +320,7 @@ try
 
 	if ($env:NEWRELIC_LICENSEKEY -eq $null -and $env:NEW_RELIC_LICENSE_KEY -eq $null)
```
tippmar-nr, line 321: I left the check for both variables just in case the user set the new style at some point

build/Packaging/NugetAzureCloudServices/tools/NewRelicHelper.psm1
```diff
@@ -390,7 +390,7 @@ function cleanup_azure_service_definition([System.__ComObject] $project){
 		
 		$runtimeNode = $modified.Runtime
         if($runtimeNode -ne $null -and $runtimeNode.ChildNodes.Count -gt 0){
-        	$variableNodes = $runtimeNode.Environment.Variable | where { $_.name -eq "COR_ENABLE_PROFILING" -or $_.name -eq "COR_PROFILER" -or $_.name -eq "NEW_RELIC_HOME" -or $_.name -eq "NEW_RELIC_INSTALL_PATH" }
+			$variableNodes = $runtimeNode.Environment.Variable | where { $_.name -eq "COR_ENABLE_PROFILING" -or $_.name -eq "COR_PROFILER" -or $_.name -eq "NEW_RELIC_HOME" -or $_.name -eq "NEW_RELIC_INSTALL_PATH" -or $_.name -eq "NEWRELIC_HOME" -or $_.name -eq "NEWRELIC_INSTALL_PATH" }
```
tippmar-nr, line 393: Added fallbacks to the "new" style names just in case

nr-ahemsath (approved):
This is probably the best thing to do under the circumstances.

jaffinito (approved):

nrcventura (approved):
Do we also need to suppress the warning about using the older environment variable name?

tippmar-nr:
> Do we also need to suppress the warning about using the older environment variable name?

No, I think it's ok to keep it. The old names really are deprecated and we want folks to use the new names; we're just making this change to bridge the gap between our docs and the agent.

build/Linux/build/common/run.sh
MerickOWA: Unfortunately, this doesn't work because AgentInstallConfiguration.cs doesn't look for NRHOME, so if neither CORECLR_NEWRELIC_HOME or CORECLR_NEW_RELIC_HOME is set because the application is depending on run.sh to set these, the profiler/agent ends up not running. See #2887

PR #2854, opened 2024-10-25
DynamoDB integration tests

nr-ahemsath:
This PR adds integration tests for DynamoDB instrumentation.  These tests:

- use the ContainerIntegrationTest pattern
- reuses the AwsSdk test app
- exercise and validate the metrics from all of the DynamoDB operations called out in the agent implementation guide, namely:
  - Create and Delete tables
  - Query and Scan tables
  - CRUD operations on items


tests/Agent/IntegrationTests/ContainerApplications/AwsSdkTestApp/AwsSdkExercisers/AwsSdkDynamoDBExerciser.cs
```diff
@@ -0,0 +1,249 @@
...
+            do
+            {
+                Thread.Sleep(sleepDuration);
+
+                var describeTableResponse = await _amazonDynamoDBClient.DescribeTableAsync(request);
+                status = describeTableResponse.Table.TableStatus;
+
+                Console.Write(".");
+            }
+            while (status != "ACTIVE");
```
chynesNR, line 103: We should probably have a way to break out of this early if something goes wrong?

tests/Agent/IntegrationTests/ContainerApplications/AwsSdkTestApp/AwsSdkExercisers/AwsSdkDynamoDBExerciser.cs
```diff
@@ -0,0 +1,249 @@
...
+        
+
+        private AmazonDynamoDBClient GetDynamoDBClient()
+        {
+
+            AmazonDynamoDBConfig clientConfig = new AmazonDynamoDBConfig();
+            // Set the endpoint URL
+            clientConfig.ServiceURL = "http://dynamodb:8123"; // port must match what is set in docker compose
+            clientConfig.AuthenticationRegion = "us-west-2";
+            var creds = new BasicAWSCredentials("xxx", "xxx");
```
chynesNR, line 32: Is this needed? IIRC there's a constructor that doesn't require a credentials object

nr-ahemsath: When I was getting the plumbing set up earlier this week, I could not get the DynamoDB calls to succeed without providing dummy credentials.  Searching online for the error I got (something about not being able to find credentials) indicated that when working with a local test DynamoDB that you had to specify some creds but it didn't matter what they were.  The same thing is being done in the SQS exerciser, for what it's worth.

tests/Agent/IntegrationTests/ContainerApplications/AwsSdkTestApp/AwsSdkExercisers/AwsSdkDynamoDBExerciser.cs
nr-ahemsath: Agreed.  A two minute timer should do the trick.

tests/Agent/IntegrationTests/ContainerApplications/AwsSdkTestApp/AwsSdkExercisers/AwsSdkDynamoDBExerciser.cs
```diff
@@ -0,0 +1,250 @@
...
+            };
+
+            TableStatus status;
+
+            int sleepDuration = 2000;
+
+            var startTime = DateTime.Now;
+            do
+            {
+                Thread.Sleep(sleepDuration);
```
tippmar-nr, line 97: Since you're already `async`,
```suggestion
                await Task.Delay(sleepDuration);
```

tests/Agent/IntegrationTests/ContainerApplications/AwsSdkTestApp/Controllers/AwsSdkDynamoDBController.cs
```diff
@@ -0,0 +1,122 @@
...
+        private readonly ILogger<AwsSdkDynamoDBController> _logger;
+
+        public AwsSdkDynamoDBController(ILogger<AwsSdkDynamoDBController> logger)
+        {
+            _logger = logger;
+
+            _logger.LogInformation("Created AwsSdkDynamoDBController");
+        }
+
+        // GET: /AwsSdkDynamoDB/CreateTable?tableName=tableName
```
tippmar-nr, line 25: nit: comment doesn't match actual route name. If you're not going to do both sync and async methods in the exerciser, you probably don't need to name the routes with `Async` suffixes.

nr-ahemsath: I fixed the comment rather than changing the name of the routes.  I thought I saw a compiler warning earlier during development that was unhappy when a controller method that is async isn't named something with `Async` in the name, although I just tried to reproduce that and couldn't.

tippmar-nr (approved):

PR #2856, opened 2024-10-26
Fix: remove usage of non-thread safe HashSet in AwsSdk pipeline wrappers

gjunge:
Thank you for submitting a pull request.  Please review our [contributing guidelines](/CONTRIBUTING.md) and [code of conduct](https://opensource.newrelic.com/code-of-conduct/).

## Description

Fixes issue found in #2855. Instead of using the `HashSet<string>` we use the `ConcurrentHashSet<string>`, similar as being used in `NewRelic.Providers.Wrappers.AwsLambda.HandlerMethodWrapper`.

# Author Checklist
- [x] Unit tests, Integration tests, and Unbounded tests completed
- [x] Performance testing completed with satisfactory results (if required)

# Reviewer Checklist
- [ ] Perform code review
- [ ] Pull request was adequately tested (new/existing tests, performance tests)

PR #2857, opened 2024-10-28
fix: Remove usage of non-thread safe HashSet in AwsSdk pipeline wrappers. (#2855)

### LINKED ISSUE #2855: AwsSdkPipelineWrapper: Asynchronous task blocked on a synchronous call
**Description**
After upgrading to 10.31.0 from 10.26.0 we have seen on a number of occasions that the CPU spiked to almost 100% and stayed there. After windbg-ing we have seen that we have a race condition in the new `AwsSdkPipelineWrapper` when inserting a new value into the `HashSet<string> _unsupportedRequestTypes`. 
In this one instance we had around 8 threads locked on this call. 

image

image

**Steps to Reproduce**
This happened when we have a `[Trace]` attribute on a asynchronous method that lists objects from S3.

```
        [Trace]
        public async Task<IEnumerable<S3Object>> GetFilesInternalAsync(string path)
        {
             var results = new List<S3Object>();
            ListObjectsV2Response s3Response = null;
            do
            {
                s3Response = await _s3client.ListObjectsV2Async(new ListObjectsV2Request() {
                    Prefix = path,
                    BucketName = this.Bucket,
                    ContinuationToken = s3Response?.NextContinuationToken
                }).ConfigureAwait(false);

                results.AddRange(s3Response.S3Objects.Where(f => f.Key != path));

            } while (s3Response.IsTruncated);

            return results;
        }
```

**Your Environment**
.NET 4.8.2 , Agent 10.31.0

tippmar-nr:
Use `ConcurrentHashSet<T>` instead of `HashSet<T>` in AWSSdk wrapper to resolve a threading issue. 

Resolves #2855. Replaces #2856.

Co-authored-by: Gideon @gjunge

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using NewRelic.Agent.Api;
+using NewRelic.Agent.Extensions.Collections;
 using NewRelic.Agent.Extensions.Providers.Wrapper;
 
 namespace NewRelic.Providers.Wrapper.AwsSdk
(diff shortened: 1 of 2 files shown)


nr-ahemsath (approved):
LGTM

chynesNR (approved):

PR #2860, opened 2024-10-29
test: Force terminate helper processes still running at dispose time

chynesNR:
An intermittent integration test failure I've noticed: if a `RemoteProcess` is hung, we can still send it shutdown signals without getting an error back, so we never force it to close. Then when we make our final attempt to shut it down in `Dispose`, it throws because the test logger is no longer available.


tippmar-nr (approved):
Good find!

nrcventura (approved):

PR #2883, opened 2024-11-14
chore: Update Dotty project and package lists, other fixes

tippmar-nr:
* Updates list of projects that Dotty monitors, as well as the list of packages
* Fixes a bug in Dotty where multi-line `<PackageReference>` elements in csproj files weren't parsed correctly
* Updates some NuGet packages that had vulnerabilities, as reported by the .NET 9 updated version of nuget.exe.

### DIFF
--- a/.github/workflows/scripts/nugetSlackNotifications/packageInfo.json
+++ b/.github/workflows/scripts/nugetSlackNotifications/packageInfo.json
@@ -42,6 +42,9 @@
         "ignoreReason": "frequent patch releases create too much noise"
     },
     {
+        "packageName": "awssdk.dynamodbv2"
+    },
+    {
         "packageName": "confluent.kafka"
     },
     {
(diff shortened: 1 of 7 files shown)


.github/workflows/scripts/nugetSlackNotifications/packageInfo.json
```diff
@@ -85,10 +106,13 @@
     {
         "packageName": "nest"
     },
+    {
```
chynesNR, line 109: Is this so we update the API reference in our tests? Should we make that part of the Release process?

chynesNR (approved):

.github/workflows/scripts/nugetSlackNotifications/packageInfo.json
tippmar-nr: Yeah, that's the intention. 

I don't know that it needs to be part of the release process - unless we change the API (which is almost never), it doesn't matter if the package gets updated right away or not.

jaffinito (approved):

PR #2886, opened 2024-11-15
chore: Additional exception logging in `ConnectionManager.Connect()`

tippmar-nr:
Adds some logging in the exception handling code in `ConnectionManager.Connect()` which will potentially help us diagnose unexpected exceptions that cause the agent to shutdown and determine if we can handle those specific exceptions in a less ham-fisted manner.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
@@ -112,36 +112,42 @@ namespace NewRelic.Agent.Core.DataTransport
 #if !NETFRAMEWORK // Only available in System.Net.Http
             // Occurs when the agent is unable to connect to APM. The request failed due to an underlying
             // issue such as network connectivity, DNS failure, server certificate validation or timeout.
-            catch (HttpRequestException)
+            catch (HttpRequestException hrex)
             {
+                Log.Info(hrex, "Connection failed.");
                 ScheduleRestart();
             }
 #endif
             // Occurs when the agent connects to APM but the connection gets aborted by the collector
-            catch (SocketException)
+            catch (SocketException sex)
             {
+                Log.Info(sex, "Connection failed.");
                 ScheduleRestart();
             }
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -112,36 +112,42 @@ private void Connect()
...
-            catch (HttpRequestException)
+            catch (HttpRequestException hrex)
             {
+                Log.Info(hrex, "Connection failed.");
                 ScheduleRestart();
             }
 #endif
             // Occurs when the agent connects to APM but the connection gets aborted by the collector
-            catch (SocketException)
+            catch (SocketException sex)
```
chynesNR, line 122: You _might_ want to rethink this naming convention...

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -112,36 +112,42 @@ private void Connect()
 #if !NETFRAMEWORK // Only available in System.Net.Http
             // Occurs when the agent is unable to connect to APM. The request failed due to an underlying
             // issue such as network connectivity, DNS failure, server certificate validation or timeout.
-            catch (HttpRequestException)
+            catch (HttpRequestException hrex)
             {
+                Log.Info(hrex, "Connection failed.");
```
chynesNR, line 117: I think this is the one case where we want to just log the exception message rather than the object, which contains the call stack. Even though it's at `Info` level, a lot of customers see a call stack and think "problem".

tippmar-nr: Are you suggesting this change across all of the exceptions that result in a restart? That's probably reasonable.

chynesNR: Exactly

tippmar-nr: Refactored to move exception handling to a separate method and only log the message for exceptions that will result in a restart.

chynesNR (approved):
Very nice, this looks much cleaner

PR #2891, opened 2024-11-18
ci: disable cosmosdb emulator startup

tippmar-nr:
Disables starting the CosmosDB emulator in `all_solutions.yml`; we are using an Azure instance of CosmosDB now.


nr-ahemsath (approved):
👍 (~assuming the CosmosDB tests pass on the CI run~) Since this was a workflow-only change, no integration tests run.

PR #2901, opened 2024-11-25
feat: Instrument Lambda invocations in AWS SDK

chynesNR:
Instrument calls to Lambdas and Lambda Aliases from the AWS SDK, with span attributes to allow linking the calls to the actual Lambda (specifically the ARN is what's needed). Includes unit and integration tests. We are able to test without a making a successful call to a lambda, so no live testing resources are necessary.

Also includes a new wrapper for `AmazonServiceClient` and logic to parse the AWS account ID from the Access Key credential. Currently supports only a single account ID per application and falls back to using the account ID from configuration if parsing is unsuccessful. We may need to revisit this logic to support multiple account IDs (one per service client instance) and consider whether using the value from configuration should take precedence or remain as a fallback.

Changes since [the last PR](https://github.com/newrelic/newrelic-dotnet-agent/pull/2784):
* Does *not* treat the segment as a Leaf, so we will get an additional HTTP span, which is now expected
* Caps the size of the ARN cache
* Limits most errors to log only once
* Does not log the account ID, only if it is present or not


src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
```diff
@@ -281,6 +283,19 @@ public AttributeDefinition<object, object> GetFaasAttribute(string name)
         }
 
 
+        private AttributeDefinition<object, object> CreateCloudSdkAttribute(string attribName)
```
tippmar-nr, line 286: This is only slightly different than `CreateFaasAttribute()` (which applies the attribute to the transaction as well as trace and span)-- do we really need the distinction?

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Collections/ConcurrentHashSet.cs
```diff
@@ -47,7 +47,13 @@ public void Add(T item)
                 _hashSet.Add(item);
             }
         }
-
+        public bool TryAdd(T item)
```
tippmar-nr, line 50: With the addition of `TryAdd()` do we really need `Add()`?

src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
chynesNR: I think so: `Faas` attributes are `Intrinsic`s, while the Cloud attributes are `Agent` attributes, and aren't supposed to apply to TransactionEvents.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Collections/ConcurrentHashSet.cs
chynesNR: Yep, `ICollection` requires an `Add` that returns void.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/AwsSdk/InvokeLambdaExerciser.cs
```diff
@@ -0,0 +1,96 @@
...
+                Payload = payload
+            };
+
+            // Note that we aren't invoking a lambda that exists! This is guaranteed to fail, but all we care
+            // about is that the agent is able to instrument the call.
+            try
+            {
+                var response = client.Invoke(request);
+            }
+            catch
```
nrcventura, line 35: Should we only catch a specific failure here to help with detecting cases where instrumentation results in an exception that isn't caught?

chynesNR: We can try it. Since we aren't sending a valid request I'm not sure we're guaranteed to get a consistent error back, but local testing seems to work.

nrcventura: It looks like we're seeing a 400 Bad Request error from the CI pipeline, instead of the ResourceNotFound error. Your suspicions about the error not being consistent seem to be valid. Maybe, there's a way to check that the error isn't coming from the agent?

chynesNR: I was hoping it would at least throw something in the `Amazon` namespace so we could check the type, but no such luck. Another option would be to call a real Lambda but that felt unnecessary.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
```diff
@@ -14,12 +18,56 @@ public class AwsSdkPipelineWrapper : IWrapper
...
+                }
+            }
+
+            return new ArnBuilder(partition, systemName, accountId); ;
+        }
+
+        private string GetAccountId(IAgent agent)
+        {
+            string accountId = agent.Configuration.AwsAccountId;
+            if (accountId != null)
```
tippmar-nr, line 57: it would be more performant to check `_reportBadAccountId` here - if it's `false`, none of the rest of the validation needs to be done.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/AwsSdk/InvokeLambdaExerciser.cs
chynesNR: I wrote a helper utility to check the call stack to see if the Exception came from our agent or the integration test app, and it worked, but proved not especially useful. If the Agent throws, we'll miss the Harvest, and everything in the test will fail anyway, and the Agent log will show the Exception.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AwsSdkPipelineWrapper.cs
nrcventura: If that's the case, should the validation be done in the DefaultConfiguration class instead?

chynesNR: I had planned on revisiting this after the ID decoding was done. There were a few things I wanted to consider:
* Having the validation in a central place
* We may want to fall back to the other ID method if one is invalid
* The environment variable for the ID may be set after we've read the config
* Depending on how we found the ID, it may be bad at first but good later

tippmar-nr (approved):

nrcventura (dismissed):

tippmar-nr (approved):
Nice work!

chynesNR:
Moving this back to Draft until we get final word on the External segment discussion

tippmar-nr (approved):

nrcventura (approved):

PR #2924, opened 2024-12-09
Moving cloud SDK attributes to Segments

chynesNR:
Using this separate PR to discuss the pros and cons of moving the new Cloud SDK attributes out of Transaction metadata. The goal would be to support multiple AWS services per transaction.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
+++ b/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
@@ -289,6 +289,7 @@ namespace NewRelic.Agent.Core.Attributes
                 .Create<object, object>(attribName, AttributeClassification.AgentAttributes)
                 .AppliesTo(AttributeDestinations.TransactionTrace)
                 .AppliesTo(AttributeDestinations.SpanEvent)
+                .WithConvert(x => x)
                 .Build(_attribFilter);
         }
 
--- a/src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
+++ b/src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
@@ -449,6 +449,19 @@ namespace NewRelic.Agent.Core.Segments
             return this;
         }
 
+        public ISpan AddCloudSdkAttribute(string key, object value)
+        {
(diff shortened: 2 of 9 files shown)


src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
```diff
@@ -289,6 +289,7 @@ private AttributeDefinition<object, object> CreateCloudSdkAttribute(string attri
                 .Create<object, object>(attribName, AttributeClassification.AgentAttributes)
                 .AppliesTo(AttributeDestinations.TransactionTrace)
                 .AppliesTo(AttributeDestinations.SpanEvent)
+                .WithConvert(x => x)
```
nrcventura, line 292: What does this change enable? Or does it remove some behavior that we don't want it to have?

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
```diff
@@ -449,6 +449,19 @@ public ISpan AddCustomAttribute(string key, object value)
             return this;
         }
 
+        public ISpan AddCloudSdkAttribute(string key, object value)
+        {
+            SpanAttributeValueCollection customAttribValues;
+            lock (_customAttribValuesSyncRoot)
+            {
+                customAttribValues = _customAttribValues ?? (_customAttribValues = new SpanAttributeValueCollection());
```
nrcventura, line 457: Should the collection be renamed since it will no longer just contain the custom attribute values?

src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinitionService.cs
chynesNR: If there's no conversion function defined, some attribute code paths fail silently (like `TrySetValue`). I'm not sure what the reasoning for that is, though. It's something should probably discuss.

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
chynesNR: Yes that's a good idea. If we decide to go this route I'll do a cleanup pass (I think this drops Code Coverage a little as well).

PR #2925, opened 2024-12-09
ci: Remove temporary .NET 9 workarounds

nr-ahemsath:
This PR:
* Removes `setup-dotnet` action to install .NET 9 preview from all actions that run on Windows - `windows-latest` has .NET 9 GA installed now
* Removes a workaround for a Visual Studio build issue
* Re-enables Centos 9 smoke test now that .NET 9 GA is included in that image

For the Linux container tests that run on an Ubuntu runner, we still need to install .NET 9, but we can at least install a GA version instead of a preview.


chynesNR (approved):

nrcventura (approved):
I think the PR is fine as is, but I'm wondering if we should keep the setup-dotnet action in place everywhere and use it to ensure the expected .net versions are always installed?

PR #2928, opened 2024-12-16
Remove references to System.Data.Odbc from wrapper and extensions

nr-ahemsath:
Per team pairing last week, we should not add a reference to `System.Data.Odbc` to the `Sql` wrapper project nor the `Extensions` project with the helper code.

I initially went down a rabbit hole of using visibility bypasser to get the necessary data out of the `OdbcCommand` object.  However, I eventually realized that we can treat the `OdbcCommand` object as an `IDbCommand` interface type and get almost everything we need directly that way.  The one exception is the `Driver` string (needed to get the DatastoreVendor), which is not a part of the `IDbCommand` interface.  However, it can be easily parsed from the ODBC connection string (and cached based on the full connection string).

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/SqlWrapperHelper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/SqlWrapperHelper.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.SqlTypes;
@@ -20,7 +21,7 @@ namespace NewRelic.Agent.Extensions.Parsing
         private const string NullQueryParameterValue = "Null";
 
         private static Regex _getDriverFromConnectionStringRegex = new Regex(@"DRIVER\=\{(.+?)\}");
-        private static Dictionary<string, DatastoreVendor> _vendorNameCache = new Dictionary<string, DatastoreVendor>();
+        private static ConcurrentDictionary<string, DatastoreVendor> _vendorNameCache = new ConcurrentDictionary<string, DatastoreVendor>();
 
         /// <summary>
         /// Gets the name of the datastore being used by a dbCommand.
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Sql/OdbcCommandWrapper.cs
```diff
@@ -2,45 +2,48 @@
...
-		public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
-		{
-			{
-				var odbcCommand = (OdbcCommand)instrumentedMethodCall.MethodCall.InvocationTarget;
-				if (odbcCommand == null)
-					return Delegates.NoOp;
+        public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
+        {
+            {
+                var odbcCommand = (IDbCommand)instrumentedMethodCall.MethodCall.InvocationTarget;
```
tippmar-nr, line 26: A hard cast like this will explode with an exception if the type can't be cast to the destination. Apparently that's never happened, though, as I see the old code did a hard cast to `OdbcCommand` and a null check the same way. Maybe we got lucky. 

In any case, I'd suggest a safer cast with some new-age syntactic dressing: 
```
                if (instrumentedMethodCall.MethodCall.InvocationTarget is not IDbCommand odbcCommand)
                    return Delegates.NoOp;
```

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/SqlWrapperHelper.cs
```diff
@@ -19,6 +19,9 @@ public static class SqlWrapperHelper
     {
         private const string NullQueryParameterValue = "Null";
 
+        private static Regex _getDriverFromConnectionStringRegex = new Regex(@"DRIVER\=\{(.+?)\}");
+        private static Dictionary<string, DatastoreVendor> _vendorNameCache = new Dictionary<string, DatastoreVendor>();
```
chynesNR, line 23: Probably want this to be a `ConcurrentDictionary` to be on the safe side

PR #2931, opened 2024-12-18
feat: Use `SocketsHttpHandler` to configure `HttpClient` in .NET 6+

### LINKED ISSUE #1845: Update HttpClient usage to better support DNS changes on .NET/.NET Core

The dotnet agent's current usage of HttpClient for sending data to New Relic mostly follows MIcrosoft's guidance (this analysis is documented in an internal doc). The one thing that was discovered to be missing is setting the `PooledConnectionLifetime` to an appropriate value (this value is TBD). The agent is already using a single instance of HttpClient.

### Acceptance Criteria

1. The agent uses a single instance of HttpClient to send data to New Relic.
2. The agent configures an appropriate value for the `PooledConnectionLifetime` setting.
3. These changes apply to .net/.net core, and do not apply to .net framework.

### Dependencies

N/A

### Estimates

t-shirt size: S

### LINKED ISSUE #2778: Refactor nested try/catch in ConnectionHandler
There's a nested try/catch in [`SendDataOverWire()`](https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionHandler.cs#L420) that can cause a misleading error message to be emitted indicating a serialization error that didn't really occur, because the nested try/catch throws the exception after logging it. 

We need to refactor so that the call to `_serializer.Serialize()` is in a separate try/catch from the try/catch around `wire.SendData()`. This will eliminate the misleading message. 

### Acceptance Criteria
Error messages in `SendDataOverWire()` are not misleading

### Estimates
S

tippmar-nr:
* Updated `NRHttpClient` to dynamically create an instance of `SocketsHttpHandler` when the app is running in .NET 6 or later, and configured `PooledConnectionTimeout` and `PooledConnectionIdleTimeout` to help the agent respond better to DNS changes. The connection timeout value is hard-coded to 5 minutes, which overrides the default of "never" and is probably sufficient for our needs. I didn't see a compelling reason to add a configuration setting for this value. Resolves #1845.

Additional refactoring:
* Update calls to `HttpClientWrapper.SendAsync()` to add a `.ConfigureAwait(false)` which seems to help prevent deadlocks.
* Replace `lock()` usage with `SemaphoreSlim` in several places, also aimed at helping to prevent deadlocks.
* Added `.ConfigureAwait(false)` in nearly all locations where we're currently doing a `.GetAwaiter().GetResult()` pattern. More deadlock prevention.
* Updated all event aggregators to log the `... harvest finished` message even if no events were sent. This will help in troubleshooting recently reported deadlocks when sending data. 
* Renamed `IHttpClientFactory.CreateClient()` to `GetOrCreateClient()` to better reflect what the method does.
* Removed an unused property from `IHttpClient` and the classes that implement it
* Refactored `ConnectionHandler.SendDataOverWire<T>()` to make exception handling a bit cleaner and more clear. Resolves #2778 
* Cleaned up several log messages in exception handlers so that they only log the exception message instead of the full stack trace if the handler is just going to `throw` the exception again anyway.
* Added logging to allow better tracing when sending data.
* Added a couple `using`s in `NRHttpClient` for objects that are `IDisposable` (`HttpRequestMessage`, `ByteArrayContent`)


src/Agent/NewRelic/Agent/Core/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+        private static readonly TaskFactory _taskFactory =
+            new(CancellationToken.None, TaskCreationOptions.None, TaskContinuationOptions.None, TaskScheduler.Default);
+
+        /// <summary>
+        /// Safely executes an async method synchronously.
+        /// </summary>
+        /// <typeparam name="TReturn"></typeparam>
+        /// <param name="task"></param>
+        /// <returns></returns>
+        public static TReturn RunSync<TReturn>(Func<Task<TReturn>> task)
```
nrcventura, line 24: Marty and I talked about removing this helper method.

src/Agent/NewRelic/Agent/Core/DataTransport/Client/NRHttpClient.cs
```diff
@@ -65,7 +102,9 @@ public override IHttpResponse Send(IHttpRequest request)
                     req.Content.Headers.Add(contentHeader.Key, contentHeader.Value);
                 }
 
-                var response = _httpClientWrapper.SendAsync(req).GetAwaiter().GetResult();
+                Log.Finest($"Request({request.RequestGuid}: Sending");
+                var response = AsyncHelper.RunSync(() => _httpClientWrapper.SendAsync(req));
```
nrcventura, line 106: We may not need this helper method, because ConfigureAwait(false) seems to help resolve the same problem in a simpler manner.

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
```diff
@@ -170,15 +175,7 @@ private void HandleConnectionException(Exception ex)
...
-            _lockSemaphore.Wait();
-            try
-            {
-                _connectionHandler.Disconnect();
-            }
-            finally
-            {
-                _lockSemaphore.Release();
-            }
+            _connectionHandler.Disconnect();
```
nrcventura, line 178: Why does this not need to be guarded anymore?

tests/Agent/IntegrationTests/Applications/CustomAttributesWebApi/Program.cs
```diff
@@ -19,6 +19,8 @@ public class Program
 
         static void Main(string[] args)
         {
+            Debugger.Launch();
```
nrcventura, line 22: Don't forget to remove this line.

src/Agent/NewRelic/Agent/Core/DataTransport/ConnectionManager.cs
tippmar-nr: Turns out you can't call `_lockSemaphore.Wait()` from one method then call it again from a nested method call. `Disconnect()` is being called from `Reconnect()`, which already waits for the semaphore. (`Disconnect()` is also being called from `OnCleanShutdown()` which isn't locked, but I'll throw that in real quick.)

nrcventura: That's a good catch. Yeah, with this change, and the guarded region being particular large makes it harder to track the code that is already guarded, and which code isn't.

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -325,10 +325,10 @@ public async Task TryInjectBrowserScriptAsync(string contentType, string request
             if (rumBytes == null)
             {
                 transaction.LogFinest("Skipping RUM Injection: No script was available.");
-                await baseStream.WriteAsync(buffer, 0, buffer.Length);
+                await baseStream.WriteAsync(buffer, 0, buffer.Length).ConfigureAwait(false);
```
nrcventura, line 328: Are we sure that this is safe? I don't know if there are scenarios that working with this stream will require dealing with the synchronization context.

src/Agent/NewRelic/Agent/Core/BrowserMonitoring/BrowserScriptInjectionHelper.cs
```diff
@@ -25,7 +25,7 @@ public static async Task InjectBrowserScriptAsync(byte[] buffer, Stream baseStre
             {
                 // not found, can't inject anything
                 transaction?.LogFinest("Skipping RUM Injection: No suitable location found to inject script.");
-                await baseStream.WriteAsync(buffer, 0, buffer.Length);
+                await baseStream.WriteAsync(buffer, 0, buffer.Length).ConfigureAwait(false);
```
nrcventura, line 28: Are we sure that this is safe to do? It seems like we would need some additional testing to ensure that this doesn't cause a problem.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AspNetCore6Plus/BrowserInjectingStreamWrapper.cs
```diff
@@ -77,7 +77,7 @@ public override void Write(byte[] buffer, int offset, int count)
                     // Set a flag on the context to indicate we're in the middle of injecting - prevents multiple recursions when response compression is in use
                     StartInjecting();
                     _agent.TryInjectBrowserScriptAsync(_context.Response.ContentType, _context.Request.Path.Value, buffer, _baseStream)
-                        .GetAwaiter().GetResult();
+                            .ConfigureAwait(false).GetAwaiter().GetResult();
```
nrcventura, line 80: Do we need more tests to ensure that this is safe?

src/Agent/NewRelic/Agent/Core/Agent.cs
tippmar-nr: As our instrumentation is essentially "library" code, it is recommended to use `.ConfigureAwait(false)` on all `await`ed async method calls. I can't imagine any scenario where this would be unsafe.

nrcventura (approved):

PR #2941, opened 2025-01-10
ODBC integration tests

nr-ahemsath:
Update the MSSQL integration tests to test the ODBC instrumentation.

* A new `SystemDataOdbcExerciser` is added to exercise the ODBC instrumentation
* The existing `MsSqlTests` are updated to test the ODBC instrumentation
* The existing `MsSqlStoredProcedureUsingOdbcDriverTests` are updated to use the new ODBC exerciser, and account for some changes

Things to note:
* The ODBC instrumentation does not instrument `Connect{Async}` or the DB read iteration methods.  The test logic accounts for this.
* The DB vendor for ODBC is no longer `Other`, but whatever DB we can detect as the underlying destination based on the connection string, in this case `MSSQL`

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/SqlWrapperHelper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/SqlWrapperHelper.cs
@@ -20,7 +20,7 @@ namespace NewRelic.Agent.Extensions.Parsing
     {
         private const string NullQueryParameterValue = "Null";
 
-        private static Regex _getDriverFromConnectionStringRegex = new Regex(@"DRIVER\=\{(.+?)\}", RegexOptions.IgnoreCase);
+        private static Regex _getDriverFromConnectionStringRegex = new Regex(@"DRIVER\=\{(.+?)\}", RegexOptions.IgnoreCase | RegexOptions.Compiled);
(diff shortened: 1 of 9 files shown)


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Parsing/SqlWrapperHelper.cs
```diff
@@ -20,7 +20,7 @@ public static class SqlWrapperHelper
     {
         private const string NullQueryParameterValue = "Null";
 
-        private static Regex _getDriverFromConnectionStringRegex = new Regex(@"DRIVER\=\{(.+?)\}");
+        private static Regex _getDriverFromConnectionStringRegex = new Regex(@"DRIVER\=\{(.+?)\}", RegexOptions.IgnoreCase);
```
nrcventura, line 23: We might want to make this a compiled regex (`RegexOptions.Compiled`) too.

nr-ahemsath: Yeah, I meant to do that but forgot in all the confusion over the connect and iteration metrics.  Good catch!

nrcventura (approved):

jaffinito (approved):

PR #2959, opened 2025-01-22
test: Use arm64 runners for arm64 container integration tests

tippmar-nr:
Updates container integration test workflow to split amd64 and arm64 tests to two separate jobs, each running on an appropriate Github runner. Also removed the use of Qemu in the workflow, as it's no longer required.

Tests were updated to add an "Architecture" trait so it's easy to divide them between the two jobs. 

Local execution is expected to remain unchanged, still needing to leverage Qemu. The magic is in the `BUILD_ARCH` environment variable -- when present, we use the specified architecture to build the container test app; when not present (i.e., running local), we hard-code it to `amd64`.

An example workflow run is at https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/12917552676/job/36024160825


chynesNR (approved):
Nice work!

jaffinito (approved):

PR #2968, opened 2025-01-24
chore: Agent Control health checks

tippmar-nr:
Adds support for writing the Agent Control health check file when enabled via environment variables. Note that this is not a customer-facing feature, as it will only be enabled in certain environments where Agent Control is also running. 

Includes unit tests.


tests/Agent/Shared/TestSerializationHelpers.Test/TestSerializationHelpers.Test.csproj
```diff
@@ -6,6 +6,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="coverlet.collector" Version="6.0.3">
```
jaffinito, line 9: Was this meant to be included in the PR?

jaffinito (approved):

PR #2972, opened 2025-01-30
chore: Log non-successful DataTransportResponseStatus during harvest

tippmar-nr:
Adds logging in the `HandleResponse()` method of all aggregators to indicate whether data was sent successfully, retained or discarded.


chynesNR:
Won't this make the log pretty noisy? Do you have a sample we could look at?

tippmar-nr:
> Won't this make the log pretty noisy? Do you have a sample we could look at?

Yeah, I'll grant you that the start / finish messages are a bit too noisy. Let me re-work it a bit. We probably really only need to log the "non-success" responses (discard / retain / etc.).

chynesNR (approved):

PR #2974, opened 2025-01-30
chore: Add supportabililty metric when Azure Function mode is enabled

tippmar-nr:
Adds `Supportability/Dotnet/AzureFunctionMode/Enabled` metric when Azure Function mode is enabled and the agent is running in an Azure Function.


chynesNR (approved):
Thanks!

PR #2982, opened 2025-02-03
Improve logging and validation for license keys.

### LINKED ISSUE #2783: Do some sanity checking on license keys
We had a customer issue recently where they had set their NR license key in multiple places, and some of them weren't valid (one was set to a blank string). This resulted in inconsistent behavior and confusing agent logs. The only validation we currently do is to check if the key is still set to `REPLACE_WITH_LICENSE_KEY`. At the very least we should log a warning if it doesn't look like a good key. If we want to be extra robust we could use some "best match" logic if we encounter multiple keys.

### Estimates
S

### LINKED ISSUE #2969: Improve logging around invalid license key
The current logging when a license key is invalid is misleading, causing confusion for customers and Support.

Currently, when a license key has a default value of [REPLACE_WITH_LICENSE_KEY](https://github.com/newrelic/newrelic-dotnet-agent/blob/75c2da5861f8394a83d4c0c6802da7ba88e0aca1/src/Agent/NewRelic/Agent/Core/AgentManager.cs#L182) or a blank value, a log message includes the helpful message "Please set your license key."

When an invalid value is set, the agent logs a message similar to the following:

```
NewRelic  ERROR: [pid, tid] Request(): Received HTTP status code Unauthorized with message {}. Request content was: [{"high_security":false}]

NewRelic  ERROR: [pid, tid] Unable to connect to the New Relic service at collector.newrelic.com:443
NewRelic.Agent.Core.DataTransport.HttpException: {}
   at NewRelic.Agent.Core.DataTransport.HttpCollectorWire.ThrowExceptionFromHttpResponseMessage(String serializedData, HttpStatusCode statusCode, String responseText, Guid requestGuid)
   at NewRelic.Agent.Core.DataTransport.HttpCollectorWire.SendData(String method, ConnectionInfo connectionInfo, String serializedData, Guid requestGuid)
   at NewRelic.Agent.Core.DataTransport.ConnectionHandler.SendDataOverWire[T](ICollectorWire wire, String method, Object[] data)
   at NewRelic.Agent.Core.DataTransport.ConnectionHandler.SendNonDataRequest[T](String method, Object[] data)
   at NewRelic.Agent.Core.DataTransport.ConnectionHandler.SendPreconnectRequest()
   at NewRelic.Agent.Core.DataTransport.ConnectionHandler.Connect()
   at NewRelic.Agent.Core.DataTransport.ConnectionManager.Connect()
NewRelic   INFO: [pid, tid] Connection failed: {}
NewRelic   INFO: [pid, tid] Will attempt to reconnect in 60 seconds
```

The mention of "high_security" is misleading, as this is included whether or not the account is a [high security account](https://docs.newrelic.com/docs/apm/agents/manage-apm-agents/configuration/high-security-mode/).

Additionally, it's not clear that the problem is related to the invalid license key or some other network issue.

### Feature Description
Improve logging so that when an invalid license key is being used, that is noted in the message.

For example, something like this would be more helpful:

```
NewRelic  ERROR: [pid, tid] Request(): Received HTTP status code Unauthorized. The current license key with prefix (abcd...) was unable to be validated. Please make sure you've set a valid ingest license key.
```

### Priority
Really Want

jaffinito:
## Description

Adds an additional warning when the agent receives a HTTP 401 Unauthorized response from the collector.  This occurs when the key is not valid (for many reasons).  

Expands the validation logic from the license key in DefaultConfiguration to:
- Still perform the checks once and save the value if one is found. If nothing is found save string.Empty to prevent further checks.
- Check for empty/missing keys
- Check if the value contains "license".  Our default key and keys used in unit test use values that contain that word.
- Check if it is the correct length - 40 characters
- Check if the characters are valid for a key - ASCII characters from 0x21 to 0x7E
- Check each of the different key locations and work through them in priority order to find a key without issues.
  - When a location has a value, but that value is invalid log a Warn message letting you know that a bad key was found and where.
  - If a valid key is found, report a Info message including where. Happens once since value is cached.
  - If no valid key is found log a Warn message to let us know. Happens each time the key is request from config.

Unit tests have been expanded to cover the new logic.

Note: valid in this context means that it is not null or whitespace, exactly 40 ASCII characters from 0x21 to 0x7E, or  string contains "license" (case insensitive)).

Fixes #2969
Fixes #2783 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -227,6 +220,58 @@ public virtual string AgentLicenseKey
...
+                    Log.Finest($"License key from {candidateKey.Key} contains invalid characters. Checking for other license keys.");
+                    continue;
+                }
+
+                Log.Finest($"License key from {candidateKey.Key} appears valid.");
+                return trimmedCandidateKey;
+            }
+
+            // return string.Empty instead of null to allow caching and prevent checking repeatedly
+            Log.Finest("No valid license key found.");
```
tippmar-nr, line 271: Seems like we might want to log this at a higher level -- info or warning?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -227,6 +220,58 @@ public virtual string AgentLicenseKey
...
+                if (string.IsNullOrWhiteSpace(candidateKey.Value))
+                {
+                    continue;
+                }
+
+                // We found something, but is it a valid key?
+
+                // If the key is the default value from newrelic.config, we return the default value
+                // AgentManager.AssertAgentEnabled() relies on this behavior and will throw an exception if the key is the default value
+                if (candidateKey.Value.Equals("REPLACE_WITH_LICENSE_KEY"))
```
chynesNR, line 245: A small suggestion: a lot of our sample code uses variations of `YOUR_LICENSE_KEY` for the environment variable. Maybe if we search for the string "license", it's a good bet that they haven't filled it in? That should cover us for all three key sources, and may be more helpful than the fall-through invalid key log messages.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
jaffinito: After talking with support, I moved most of the message to warning and make some more visible.  They preferred to have the information upfront.

PR #2984, opened 2025-02-04
fix: Improve error handling in AWS account ID parsing logic

tippmar-nr:
Adds some additional null checks and log messages in `AmazonServiceClientWrapper`. 

If an account ID can't be parsed from credentials and no fallback account ID is available in configuration, the wrapper is disabled on future calls.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AmazonServiceClientWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsSdk/AmazonServiceClientWrapper.cs
@@ -11,6 +11,8 @@ namespace NewRelic.Providers.Wrapper.AwsSdk;
 
 public class AmazonServiceClientWrapper : IWrapper
 {
+    private bool _disableServiceClientWrapper;
+
     private const int LRUCapacity = 100;
     // cache the account id per instance of AmazonServiceClient.Config
     public static LRUCache<WeakReferenceKey<object>, string> AwsAccountIdByClientConfigCache = new(LRUCapacity);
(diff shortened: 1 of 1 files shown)


jaffinito (approved):
Nice!

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Caching/WeakReferenceKey.cs
```diff
@@ -11,35 +11,28 @@ namespace NewRelic.Agent.Extensions.Caching
...
         public override bool Equals(object obj)
         {
             if (obj is WeakReferenceKey<T> otherKey)
             {
-                if (WeakReference.TryGetTarget(out var thisTarget) &&
-                    otherKey.WeakReference.TryGetTarget(out var otherTarget))
-                {
-                    return ReferenceEquals(thisTarget, otherTarget);
-                }
+                return otherKey.GetHashCode() == _hashCode;
```
chynesNR, line 27: Doesn't this risk a collision if you aren't checking `ReferenceEquals`?

tippmar-nr: Well... I guess it's possible but in an LRU of 100 items of the same type, it seems unlikely that two different instances would have the same hashcode.

tippmar-nr: See if you like it any better now.

chynesNR (approved):

PR #2948, opened 2025-01-14
feature: Add .NET 8+ support for System.Data.Odbc

nr-ahemsath:
## Description

This PR adds datastore instrumentation for `System.Data.Odbc`, which is Microsoft's implementation of ODBC access for modern .NET.  This resolves #2922.

Note that the instrumentation is less complete than for ODBC operations in .NET Framework applications.  In particular, the following things that are instrumented in .NET Framework are not instrumented for .NET 8+:
* `Connect/ConnectAsync` calls
* Database read iteration calls (note that these aren't enabled by default for .NET Framework either due to performance concerns)
* SQL explain plans

# Author Checklist
- [X] Unit tests, Integration tests, and Unbounded tests completed
- [ ] ~Performance testing completed with satisfactory results (if required)~ N/A

# Reviewer Checklist
- [ ] Perform code review
- [ ] Pull request was adequately tested (new/existing tests, performance tests)

PR #2985, opened 2025-02-04
feat: Add .NET 8+ support for System.Data.Odbc

tippmar-nr:
A PR branch with one empty commit, so we can update the release notes to reference ODBC support that was added in #2948.


chynesNR (approved):
Thanks!

PR #2999, opened 2025-02-19
chore:Add message to NugetVersionDeprecator when there are no packages to deprecate.

jaffinito:
Currently the NugetVersionDeprecator doesn't log a message when there are no packages to deprecate.  This adds one as well as renaming a few things to make it clearer what they do.


chynesNR (approved):
Thanks for making this clearer!

nrcventura (approved):

nr-ahemsath (approved):
👍

PR #3003, opened 2025-02-21
feat: Add instrumentation for In-Process Azure Functions

tippmar-nr:
## Description
* Adds instrumentation for In-Process (Windows only, .NET 6 or .NET 8) Azure Functions
* Adds Distributed Tracing support for In-Process Azure Functions for the following cases:
   * `HttpTrigger` functions will read incoming distributed trace headers if the trigger method has a parameter of type `System.Net.Http.HttpRequest`, `Microsoft.AspNetCore.Http.DefaultHttpRequest` or `Microsoft.AspNetCore.Http.HttpRequest`.
   * `ServiceBusTrigger` functions will read incoming distributed trace headers if the trigger method has a parameter of type `Azure.Messaging.ServiceBus.ServiceBusReceivedMessage`. Trace headers are expected to be in the `ApplicationProperties` collection.
   * Distributed trace headers will be inserted into the `ApplicationProperties` collection for Triggers with return a type of `Azure.Messaging.ServiceBus.ServiceBusMessage`.
* Adds two new supportability metrics for Azure Functions:
   * `Supportability/DotNet/AzureFunction/Worker/InProcess` (or `/Isolated`)
   * `Supportability/DotNet/AzureFunction/Trigger/{TriggerType}` where `{TriggerType}` is the name of the trigger attribute (minus the `TriggerAttribute` suffix). 
 * Existing `Supportability/DotNet/AzureFunctionMode/Enabled` metric has been replaced with `Supportability/DotNet/AzureFunctionMode/enabled` and `Supportability/DotNet/AzureFunctionMode/disabled`.

Includes integration tests. 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/InProcessFunctionDetails.cs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+public  class ServiceBusTriggerDetails
+{
+    public string QueueName { get; set; }
+    public string TopicName { get; set; }
+    public string SubscriptionName { get; set; }
+    public string Connection { get; set; } // not a connection string; just the name of the appSetting that contains the connection string
+
+    public ServiceBusDestinationType DestinationType => !string.IsNullOrEmpty(QueueName) ? ServiceBusDestinationType.Queue : ServiceBusDestinationType.Topic;
```
jaffinito, line 22: I am not seeing any references to this, can you point me in the right direction?

tippmar-nr: Actually, I don't think anything in the `ServiceBusTriggerDetails` class is used at all so I'll get rid of it. Thanks for pointing it out!

jaffinito (dismissed):

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/AzureFunctionInProcessExecuteWithWatchersAsyncWrapper.cs
```diff
@@ -0,0 +1,141 @@
...
+        _fullNameGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(functionDescriptor.GetType(), "FullName");
+        string functionClassAndMethodName = _fullNameGetter(functionDescriptor);
+
+        // cache the function details by function name so we only have to reflect on the function once
+        var inProcessFunctionDetails = _functionDetailsCache.GetOrAdd(functionClassAndMethodName, _ => GetInProcessFunctionDetails(functionClassAndMethodName));
+
+        _idGetter ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<Guid>(functionInstance.GetType(), "Id");
+        string invocationId = _idGetter(functionInstance).ToString();
+
+        agent.Logger.Debug($"Instrumenting in-process Azure Function: {inProcessFunctionDetails.FunctionName} / invocation ID {invocationId} / Trigger {inProcessFunctionDetails.Trigger}.");
```
nrcventura, line 62: We should probably log this at the finest level since it looks like it would happen with every function invocation. We may also want to avoid the string interpolation here so that the string formatting doesn't need to happen unless that logging level is enabled.

src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -985,11 +985,11 @@ private void ReportIfAwsAccountIdProvided()
...
         }
 
-        private void ReportIfAzureFunctionModeIsEnabled()
+        private void ReportIfAzureFunctionModeIsDetected()
         {
-            if (_configuration.AzureFunctionModeEnabled && _configuration.AzureFunctionModeDetected)
+            if (_configuration.AzureFunctionModeDetected)
             {
-                ReportSupportabilityCountMetric(MetricNames.SupportabilityAzureFunctionModeEnabled);
+                ReportSupportabilityCountMetric(MetricNames.SupportabilityAzureFunctionModeEnabled + $"/{(_configuration.AzureFunctionModeEnabled ? "1" : "0")}");
```
nrcventura, line 992: We should see if there are similar metrics for enabled/disabled things just to see if this naming convention is consistent.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/AzureFunctionInProcessInvokeAsyncWrapper.cs
```diff
@@ -0,0 +1,252 @@
...
+
+            if (headers != null)
+            {
+                transaction.AcceptDistributedTraceHeaders(headers, GetHeaderValueFromIEnumerable, TransportType.HTTP);
+            }
+
+            return true;
+        }
+
+        if (argTypeName is "Microsoft.AspNetCore.Http.DefaultHttpRequest" or "Microsoft.AspNetCore.Http.HttpRequest")
```
nrcventura, line 155: In some places this comparison is done using `==` and here it is done using `is`. We should be more consistent in how we do this comparison.

tippmar-nr: `is` is the same thing as `=='; the `is` bit is just syntactic sugar. I'm fine with using whichever makes sense for the line it's on.

src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
tippmar-nr: Looks like we would typically use `/enabled` or `/disabled` -- I'll update to be more consistent.

nrcventura (approved):

PR #3007, opened 2025-02-27
feat: Enable Azure Function instrumentation by default

tippmar-nr:
Azure Function instrumentation is now enabled by default. 

To disable Azure Function instrumentation, set the `NEW_RELIC_AZURE_FUNCTION_MODE_ENABLED=0` environment variable.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
+++ b/src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
@@ -138,7 +138,8 @@ namespace NewRelic.Agent.Core.Utilization
 
         public IVendorModel GetAzureFunctionVendorInfo()
         {
-            if (!_configuration.AzureFunctionModeDetected)
+            // per the spec, only report Azure Function metadata if the agent is running in an Azure Function mode and instrumentation is enabled
+            if (!(_configuration.AzureFunctionModeEnabled &&_configuration.AzureFunctionModeDetected))
                 return null;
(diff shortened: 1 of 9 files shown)


nrcventura (dismissed):

src/Agent/NewRelic/Agent/Core/Utilization/VendorInfo.cs
```diff
@@ -138,7 +138,7 @@ public IDictionary<string, IVendorModel> GetVendors()
 
         public IVendorModel GetAzureFunctionVendorInfo()
         {
-            if (!(_configuration.AzureFunctionModeDetected && _configuration.AzureFunctionModeEnabled))
+            if (!_configuration.AzureFunctionModeDetected)
```
nrcventura, line 141: Shouldn't this still depend on the Azure Function Mode check?

tippmar-nr: I think we want to send the Azure Function vendor info even if azure function mode isn't enabled so that we know it's an azure function. But maybe that vendor info is what lights up the Azure Function-specific UI bits?  @ashishsinghnr can you weigh in?

nrcventura: Correct, that vendor info, is what causes the tags to get added, and the UI uses the tags.

nrcventura (approved):

PR #3015, opened 2025-03-03
fix: Set Profiler logger to initialized when configuration load fails

### LINKED ISSUE #3014: Profiler creates empty log file when global newrelic.config is not found
**Description**
If the "global" `newrelic.config` file is not found in the expected location (`NEW_RELIC_HOME`), the profiler correctly unloads but fails to log the error, producing only a profiler log file with no content. 

**Expected Behavior**
The profiler log file should clearly indicate that the newrelic.config file was not found.

**Steps to Reproduce**
* Delete or rename the `newrelic.config` file in the `NEW_RELIC_HOME` folder
* Start an instrumented application
* Note that the profiler log file (`NewRelic.Profiler.99999.log`) is created, but is empty.
 
P2
S

tippmar-nr:
The profiler log file wasn't being set as "initialized" until after the configuration was successfully loaded. This prevented errors that occurred during configuration load from being written to the log file, with the symptom being a profiler log with no content. 

This fix sets the logger to "initialized" just before the second attempt to load the configuration, ensuring that the exception thrown during that attempt gets logged.

Sample output:
```
[Error] 2025-03-03 21:50:58 The global newrelic.config file was not found at: C:\source\repos\newrelic-dotnet-agent\src\Agent\newrelichome_x64_coreclr
[Error] 2025-03-03 21:51:12 An exception was thrown while initializing the profiler. The profiler will be detached now.
```

Fixes #3014


nrcventura (approved):
🥇

nr-ahemsath (approved):
👍

jaffinito (approved):

src/Agent/NewRelic/Profiler/Profiler/CorProfilerCallbackImpl.h
```diff
@@ -1147,7 +1148,6 @@ namespace NewRelic { namespace Profiler {
             nrlog::StdLog.SetAzureFunctionMode(_systemCalls->IsAzureFunction());
             nrlog::StdLog.SetAzureFunctionLogLevelOverride(_systemCalls->IsAzureFunctionLogLevelOverrideEnabled());
             nrlog::StdLog.SetEnabled(_systemCalls->GetLoggingEnabled(configuration->GetLoggingEnabled()));
-            nrlog::StdLog.SetInitalized();
```
nrcventura, line 1150: It looks like removing this line from here is breaking console logging in the profiler, because the logger is never set to "initialized".

nrcventura (changes requested):
See the comment about the logger not getting initialized if console logging is enabled.

src/Agent/NewRelic/Profiler/Profiler/CorProfilerCallbackImpl.h
tippmar-nr: Good catch. I moved it outside the `if` block.

src/Agent/NewRelic/Profiler/Profiler/CorProfilerCallbackImpl.h
```diff
@@ -1125,7 +1126,6 @@ namespace NewRelic { namespace Profiler {
                     // Imbue with locale and codecvt facet is used to allow the log file to write non-ascii chars to the log
                     nrlog::StdLog.get_dest().imbue(std::locale(std::locale::classic(), new std::codecvt_utf8<wchar_t>));
                     nrlog::StdLog.get_dest().exceptions(std::wostream::failbit | std::wostream::badbit);
-                    LogInfo("Logger initialized.");
```
nrcventura, line 1128: Is this log message accurate? Should we keep this log message?

tippmar-nr: It's accurate but probably not useful enough to warrant a whole cycle of me putting it back, updating the nuget package and re-running the workflows for this PR... ;-)

nrcventura (approved):

nrcventura (approved):

nrcventura (approved):

PR #3033, opened 2025-03-11
fix: Remove unnecessary dependency on `System.Runtime.InteropServices.RuntimeInformation` in .NET 4.6.2  builds.

### LINKED ISSUE #3032: .NET agent fails to load for some .NET Framework apps
**Description**

Multiple customers have reported that the .NET agent newer than version 10.32.0 fails to load (and crashes their .NET Framework 4.8 application) with this error:
```
2025-03-11 07:54:57,328 NewRelic  ERROR: [pid: 7424, tid: 1] There was an error initializing the agent
System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
File name: 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
at NewRelic.Agent.Core.AgentManager.LogInitialized()
at NewRelic.Agent.Core.AgentManager..ctor()
at NewRelic.Agent.Core.AgentManager.AgentSingleton.CreateInstance()
```

**Expected Behavior**
The .NET agent should load and run correctly on .NET Framework 4.8 applications.

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/1

P2
M

tippmar-nr:
Removes an unnecessary dependency on `System.Runtime.InteropServices.RuntimeInformtion` for .NET 4.6.2 builds. This has the helpful side effect of also removing a dependency on `System.Runtime` that was reportedly causing the .NET agent to crash for some users running the agent on a .NET Framework 4.8 application. 

Fixes #3032

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentManager.cs
@@ -350,8 +350,11 @@ namespace NewRelic.Agent.Core
                     }
                 }
 
-
+#if NETSTANDARD
                 Log.Debug($".NET Runtime Version: {RuntimeInformation.FrameworkDescription}");
+#else
+                Log.Debug($".NET Runtime Version: {AppDomain.CurrentDomain.SetupInformation.TargetFrameworkName}");
+#endif
             }
 
 #if NETFRAMEWORK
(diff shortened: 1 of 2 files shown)


chynesNR (approved):
Nice work! Was that a log line we added recently? Do we know what we were trying to catch?

tippmar-nr:
> Nice work! Was that a log line we added recently? Do we know what we were trying to catch?

I think that log line has been in there for quite a while (but I think I originally added it, so it hasn't been there forever. The intent was just to try log out the runtime version that the app is using.

jaffinito (approved):

PR #3048, opened 2025-03-19
feat: Add Couchbase 3.x instrumentation

tippmar-nr:
## Description

Adds instrumentation for `CouchbaseNetClient` v3.2.0 and later. 

Includes integration tests using the ConsoleMFA pattern.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase/Instrumentation.xml
```diff
@@ -7,20 +7,7 @@ SPDX-License-Identifier: Apache-2.0
...
-				<exactMethodMatcher methodName="PerformConcatenate"/>
-				<exactMethodMatcher methodName="PerformMutate"/>
-				<exactMethodMatcher methodName="PerformRemove"/>
-				<exactMethodMatcher methodName="PerformStore" />
-				<exactMethodMatcher methodName="PerformTryGet" />
-				<exactMethodMatcher methodName="PerformTryGetAndTouch"/>
-				<exactMethodMatcher methodName="PerformTryGetWithLock"/>
-				<exactMethodMatcher methodName="Touch" />
-				<exactMethodMatcher methodName="Unlock"/>
-			</match>
```
nrcventura, line 22: Why was this removed? Does removing this break the old instrumentation?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase3/Couchbase3QueryWrapper.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
+    {
+        transaction.AttachToAsync(); // all methods are async
+
+        var operation = instrumentedMethodCall.MethodCall.Method.MethodName;
+        string model = GetBucketName(instrumentedMethodCall.MethodCall.InvocationTarget); 
+
+        // TODO: for SearchAsync / SearchQueryAsync, this is just the name of the index. Should commandText just be null in that case?
+        string commandText = instrumentedMethodCall.MethodCall.MethodArguments[0] as string;
```
nrcventura, line 40: Should we have a separate ticket to follow up on this?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase3/Instrumentation.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+Copyright 2020 New Relic Corporation. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+<extension xmlns="urn:newrelic-extension">
+  <instrumentation>
+
+    <tracerFactory name="Couchbase3CollectionWrapper">
```
nrcventura, line 9: Instrumenting each method is probably an ok short term approach. However we may want to consider leveraging the functionality called out in https://docs.couchbase.com/dotnet-sdk/current/howtos/observability-tracing.html to make things easier going forward. That can be done in a separate follow up ticket.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase/Instrumentation.xml
tippmar-nr: It was removed because it's unused. 

For reasons that long pre-date me, the `CanWrap()` methods in the Couchbase wrappers are looking for specific class and method names, and none of them are looking for `Couchbase.CouchbaseClient`.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase3/Couchbase3QueryWrapper.cs
tippmar-nr: No, I'd like to resolve this question here. Maybe we'll just leave it as-is and consider changing it later if a customer questions it.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase3/Instrumentation.xml
tippmar-nr: Feel free to create that ticket. I looked at that functionality briefly and couldn't see an obvious way we could instrument it.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase3/Couchbase3QueryWrapper.cs
nrcventura: If it's just the index name, maybe we should just treat it the way that we do in the Collection wrapper, and pass null. In that scenario we do not have an actual query to display in the UI.

tippmar-nr: done

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Couchbase3/Instrumentation.xml
nrcventura: I don't think that we'd instrument it. I think we'd need to do something similar to what we do for newer Redis libraries where we integrate with the profiling API early on, and that profiling API integration would be responsible for starting and ending the segments.

nrcventura (dismissed):

PR #3049, opened 2025-03-19
feature: Add AWS SDK instrumentation for Kinesis Data Streams and Kinesis Delivery Streams (Firehose)

nr-ahemsath:
Add instrumentation for AWS SDK operations targeting Kinesis Data Streams and Kinesis Delivery Streams (aka Firehose).

Add unit and integration tests.

Manual testing confirms that entity relationships will now work between an APM application and an infra-agent monitored Kinesis or Firehose stream:
dotnet-kinesis-entity-relationships-kinesis-data-streams-service-map


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/KinesisHelper.cs
```diff
@@ -0,0 +1,103 @@
...
+                // if StreamName is null/unavailable, StreamARN may exist
+                var streamARN = GetDeliveryStreamArnFromRequest(request);
+                if (streamARN != null)
+                {
+                    return GetStreamNameFromArn(streamARN);
+                }
+            }
+            catch
+            {
+            }
```
nrcventura, line 58: This is another path that could use a unit test, similar to the catch block in the other method.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/KinesisHelper.cs
```diff
@@ -0,0 +1,103 @@
...
+
+namespace NewRelic.Agent.Extensions.AwsSdk
+{
+    public static class KinesisHelper
+    {
+
+        private static readonly ConcurrentDictionary<string, string> _streamNameCache = new();
+        private static readonly ConcurrentDictionary<Type, List<string>> _propertyInfoCache = new();
+
+        public static string GetStreamNameFromRequest(dynamic request)
```
nrcventura, line 17: Is the request type always public? Is it possible that it's a Request<T>, or Request<Kinesis<T>, where T is not a public type? If it's potentially not public, we can't use `dynamic`.

nrcventura: This comment applies to the other methods too.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/KinesisHelper.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            return GetPropertyFromDynamic(request, "StreamARN");
+        }
+
+        public static string GetDeliveryStreamArnFromRequest(dynamic request)
+        {
+            return GetPropertyFromDynamic(request, "DeliveryStreamARN");
+        }
+
+        private static string GetPropertyFromDynamic(dynamic request, string propertyName)
```
nrcventura, line 72: We should probably make the `request` parameter an `object` type instead of `dynamic` so that the `GetType` call can be invoked directly instead of using the dynamic runtime.

nrcventura:
Do we also need to add the kinesis libraries to dotty?

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/KinesisHelper.cs
```diff
@@ -0,0 +1,103 @@
...
+                // if StreamName is null/unavailable, StreamARN may exist
+                var streamARN = GetStreamArnFromRequest(request);
+                if (streamARN != null)
+                {
+                    return GetStreamNameFromArn(streamARN);
+                }
+            }
+            catch
+            {
+            }
```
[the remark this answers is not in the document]
nr-ahemsath: Catch block is actually unnecessary, removed.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/KinesisHelper.cs
nr-ahemsath: Refactored to avoid using dynamic entirely.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/KinesisHelper.cs
nr-ahemsath: Done.

src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/AwsSdk/KinesisHelper.cs
nr-ahemsath: Resolved by removing unnecessary catch block

nrcventura (approved):

PR #3058, opened 2025-03-26
chore: Refactor usages of `File.` and `Directory.` to use wrappers

tippmar-nr:
* Refactors usage of `File.` and `Directory.` to use `IFileWrapper` and `IDirectoryWrapper` instead.
* Adds methods to `IFileWrapper` and `IDirectoryWrapper` as needed with default pass-through implementation.
* Adds a static `Instance` property to `FileWrapper` and `DirectoryWrapper` for use in static methods
* Removes `IFileReaderWrapper` and moves implementation to `IFileWrapper` and `FileWrapper`
* Does not alter any unit tests (aside from required argument refactoring); this change simply reduces friction for future unit tests that may need to mock file and directory operations.
* Did not attempt to refactor usages of `Path.` as that usage is ubiquitous throughout the codebase and we haven't previously had a need to mock that functionality for tests.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentInstallConfiguration.cs
@@ -163,7 +163,7 @@ namespace NewRelic.Agent.Core
         private static string GetNewRelicHome()
         {
             var newRelicHome = EnvironmentVariableProxy.GetEnvironmentVariableFromList(NewRelicHomeEnvironmentVariables);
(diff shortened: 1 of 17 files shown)


nrcventura:
Is there value in making these changes if we do not add unit tests to the code, or do not plan to add unit tests to the code? If we do plan on adding unit tests, and this is just the first step in that process, then I think it is fine. Otherwise, I do not think we should make these changes.

jaffinito (approved):

PR #3061, opened 2025-03-27
test: Cleanup / refactor to remove database dependencies in unbounded tests project

tippmar-nr:
* Removes Couchbase, Oracle and MySql library dependencies from UnboundedIntegrationTests
* Refactors Oracle test setup / teardown (table create / drop) to be contained in the Oracle controller of the BasicMVCApplication project
* Changes UnboundedIntegrationTest project tfm to `net9.0` with corresponding all_solutions workflow update


tests/Agent/IntegrationTests/UnboundedIntegrationTests/Oracle/EnterpriseLibraryOracleTests.cs
```diff
@@ -43,15 +43,19 @@ public EnterpriseLibraryOracleTests(RemoteServiceFixtures.OracleBasicMvcFixture
                 },
                 exerciseApplication: () =>
                 {
+                    _fixture.CreateTable();
+
                     _fixture.GetEnterpriseLibraryOracle();
                     _fixture.AgentLog.WaitForLogLine(AgentLogBase.AgentConnectedLogLineRegex, TimeSpan.FromMinutes(1));
                     _fixture.AgentLog.WaitForLogLine(AgentLogBase.SqlTraceDataLogLineRegex, TimeSpan.FromMinutes(1));
+
+                    _fixture.DropTable();
```
nrcventura, line 52: If an exception occurs prior to this line, the table used by the test will not be cleaned up.

nrcventura (changes requested):
The cleanup behavior needs to always happen, even if there is an error while running the test.

nrcventura (approved):

PR #3063, opened 2025-03-27
docs: Update agent development and integration testing documentation

nr-ahemsath:
I went over all three documents in `docs` and updated them to the current state of the world.

Also, as part of this process, I fixed an issue with the Linux profiler build `docker-compose` file where it would bomb out if a certain env var wasn't set, even though that env var is only required for building a debug version of the Linux profiler, which is not relevant to any ongoing development or testing processes.  The fix is to fall back to a default value if the env var isn't set.


docs/integration-tests.md
```diff
@@ -22,8 +22,8 @@ Install the "ASP.NET and web development" workload in the "Web & Cloud" category
 
 Depending on which version of Visual Studio you are using, you may have to install some or all of the following:
 
-* .NET 6
-* .NET 7
+* .NET 8
+* .NET 8
```
tippmar-nr, line 26: Did you mean .NET 9 here or is this just a duplicate line to be removed?
```suggestion
* .NET 9
```

tippmar-nr (dismissed):

docs/integration-tests.md
nr-ahemsath: Good catch, thanks.

nrcventura (approved):

PR #3068, opened 2025-03-31
fix: In-process Azure Function publishes Nuget agent DLLs to the wrong folder

### LINKED ISSUE #3067: Agent Nuget package does not work with in-process Azure Functions
**Description**
When publishing an in-process Azure Function that uses the .NET agent Nuget package to install the agent, all agent .DLL files are moved from the `newrelic` folder to `bin\newrelic`, which causes the agent to fail at startup. 

**Expected Behavior**
Agent DLLs should remain in `newrelic` when an in-process Azure function that uses the .NET agent Nuget package is published.

P2
M

tippmar-nr:
## Description

Add a build target for the agent Nuget package that runs during publish and moves .NET agent DLLs back to `\newrelic`. 

This is to compensate for the [`_GenerateFunctionsAndCopyContentFiles` target](https://github.com/Azure/azure-functions-vs-build-sdk/blob/7a3e505c2382ee9c461985baef1603e980740ca6/src/Microsoft.NET.Sdk.Functions.MSBuild/Targets/Microsoft.NET.Sdk.Functions.Publish.targets#L134) invoked during in-process Azure Function publish that incorrectly moves every .DLL in the build output folder (and subfolders) to the `\bin` folder. 

Fixes #3067.

### DIFF
--- a/build/ArtifactBuilder/Artifacts/NugetAgent.cs
+++ b/build/ArtifactBuilder/Artifacts/NugetAgent.cs
@@ -13,10 +13,11 @@ namespace ArtifactBuilder.Artifacts
         private readonly AgentComponents _coreAgentX86Components;
         private string _nuGetPackageName;
 
-        public NugetAgent(string configuration)
+        public NugetAgent(string configuration, string versionOverride = null)
             : base(nameof(NugetAgent))
         {
             Configuration = configuration;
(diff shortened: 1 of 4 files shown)


nrcventura (approved):

chynesNR (approved):
Nice, this looks like a clean way to handle it

PR #3075, opened 2025-04-01
fix: When sql explain plan generation fails, don't attempt to generate an explain plan on future invocations of the same query.

tippmar-nr:
Adds a cache for tracking queries where explain plan generation attempt throws an exception, so that future execution of the same query will not attempt to generate an explain plan. This fix will help ensure database performance doesn't suffer due to repeated, failed explain plan generation attempts.

Cache capacity is configured to 1000 queries, which should be far more than we ever experience in the wild - it is anticipated that explain plan failures are a very rare occurrence.

Includes unit and integration tests.


src/Agent/NewRelic/Agent/Core/Segments/DatastoreSegmentData.cs
```diff
@@ -120,11 +130,15 @@ internal override IEnumerable<KeyValuePair<string, object>> Finish()
         }
 
 
-        public void ExecuteExplainPlan(SqlObfuscator obfuscator)
+        public bool ExecuteExplainPlan(SqlObfuscator obfuscator)
         {
+            // don't attempt to generate an explain plan if this query has failed previously
+            if (_failedExplainPlanQueryCacheByDatastoreVendor.Contains(DatastoreVendorName, CommandText))
```
nrcventura, line 136: Using this for a design discussion. Should we use the raw `CommandText`? Or should we use the obfuscated SQL?

If we use the raw `CommandText` we could potentially see failures for what is essentially the same sql when someone is not using parameterized sql.

If we use the obfuscated SQL, then we could potentially run into a problem where some commands are fine and others are not depending on the logic in the non-parameterized sql.

I think using the ParsedSqlStatement objects in the agent are too simplistic and would disallow too many sql statements from generating explain plans.

tippmar-nr: Good questions. I considered whether using the obfuscated sql would be better but ultimately decided that this is going to be such a rare occurrence (the explain plan failure) that just using the raw SQL should be fine. 

This solution would be less than optimal only if the customer is using non-parameterized SQL, hits an error generating the explain plan and then calls the same query again with different parameters. But the overall impact on the customer would be essentially no different than it is today -- we'd keep trying (and failing) to generate the explain plan.

src/Agent/NewRelic/Agent/Core/Segments/DatastoreSegmentData.cs
```diff
@@ -239,5 +258,17 @@ public void SetConnectionInfo(ConnectionInfo connInfo)
         {
             _connectionInfo = connInfo;
         }
+
+        // FOR UNIT TESTING ONLY
```
nrcventura, line 262: Should we use the `///` comment style here so that the intellisense will include this message too?

tippmar-nr: Sure.

nrcventura (approved):

PR #3080, opened 2025-04-02
ci: Add a workflow to check pull request title for correct prefix.

nr-ahemsath:
This PR adds a workflow to check the prefix of pull request titles to make sure that the start with a prefix and that the prefix is in the list of known good prefixes: feat, fix, security, notice, ci, chore, test, docs

I initially tested this in my fork.  Unfortunately, the PR title as seen by the workflow logic does not get updated after PR creation.  In other words, you can edit and change the PR title after initial creation, but the workflow only ever gets the initial value.  So, if this check finds a problem with the PR title, the solution is to close the PR with the "bad" title/prefix and open another one with a correct prefix.

### DIFF
--- a/.github/workflows/pr_title_check.yml
+++ b/.github/workflows/pr_title_check.yml
@@ -1,0 +1,45 @@
+name: Ensure PR title has correct prefix
+
+on:
+  pull_request:
+    branches:
+      - main
+  
+permissions:
+  contents: read
+
+jobs:
+  check-pr-summary:
+    name: Check if PR title has a good prefix
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    steps:
+      - name: Check PR title
+        env:
+            PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          echo "Pull request title='${PR_TITLE}'"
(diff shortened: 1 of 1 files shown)


nrcventura:
Does manually re-running the workflow allow it to see the updated title?

.github/workflows/pr_title_check.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  check-pr-summary:
+    name: Check if PR title has a good prefix
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    steps:
+      - name: Check PR title
```
tippmar-nr, line 19: Need to add Harden Runner step here. 
```
          - name: Harden Runner
            uses: step-security/harden-runner@c6295a65d1254861815972266d5933fd6e532bdf # v2.11.1
            with:
              disable-sudo: true
              egress-policy: audit
```

tippmar-nr:
Looks like there's already an action that does the same thing https://github.com/marketplace/actions/pr-title-checker - I wonder if it handles PR title changes?

nr-ahemsath:
Marty found a better solution, closing this hand-rolled attempt.

PR #3081, opened 2025-04-03
ci: Add workflow and configuration file to check PR titles for correct format.

nr-ahemsath:
Take two.  Use an available action to check PR titles for the correct prefix.  Should only apply to PRs into `main`.

### DIFF
--- a/.github/workflows/pr_title_checker.yml
+++ b/.github/workflows/pr_title_checker.yml
@@ -1,0 +1,29 @@
+name: Ensure PR title has a valid prefix
+
+on:
+  pull_request:
+    branches:
+      - main
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - labeled
+      - unlabeled
+
+jobs:
+  check:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_title_checker.yml
```diff
@@ -0,0 +1,29 @@
...
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@c6295a65d1254861815972266d5933fd6e532bdf # v2.11.1
+        with:
+          disable-sudo: true
+          egress-policy: audit
+      - name: Check PR title    
+        uses: thehanimo/pr-title-checker@v1.4.3
```
tippmar-nr, line 25: Need to use the commit hash here or our security scan will get mad.

tippmar-nr (approved):

.github/workflows/pr_title_checker.yml
nr-ahemsath: Already fixed

PR #3100, opened 2025-04-15
feat: Allow incoming W3C traceparent `sampled` flag to influence .NET agent sampling decision.

tippmar-nr:
## Description
- Implements .NET agent support for respecting the W3C traceparent `sampled` flag as per the [Distributed Tracing agent spec](https://source.datanerd.us/agents/agent-specs/blob/main/distributed_tracing/Trace-Context-Payload.md#sampled). 
- Adds configuration to enable customers to control how the agent responds to an incoming traceparent header in both the sampled and not sampled case. Default configuration is to use our existing logic to determine whether to sample spans.
```
  <distributedTracing enabled="true" ... >
    <sampler remoteParentSampled="default|alwaysOn|alwaysOff" remoteParentNotSampled="default|alwaysOn|alwaysOff">
  </distributedTracing>
```
Environment variables can also be used:
```
NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_REMOTE_PARENT_SAMPLED=alwaysOn
NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_REMOTE_PARENT_NOT_SAMPLED=alwaysOff
```
- When enabled, we will always sample (or not sample, depending on the configuration) spans in the agent based on the incoming traceparent `sampled` flag and configuration applied as shown in the following table:

| traceparent Sampled | remoteParentSampled config | remoteParentNotSampled config | agent sampling decision |
|--------|--------|--------|--------|
| missing / invalid | n/a | n/a | Use existing logic |
| 01 | `default` | n/a | Use existing logic |
| 01 | `alwaysOn` | n/a | Sampled = true |
| 01 | `alwaysOff` | n/a | Sampled = false |
| 00 | n/a | `default` | Use existing logic |
| 00 | n/a | `alwaysOn` | Sampled = true |
| 00 | n/a | `alwaysOff` | Sampled = false |

Includes unit and composite agent tests but does not include integration tests.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


chynesNR:
Hey! I've switched this over to Draft for the time being. The Java Agent has run into some unexpected behavior in the field, and it may require a spec change, depending on the root cause.

src/Agent/NewRelic/Agent/Core/DistributedTracing/TracingState.cs
```diff
@@ -148,45 +149,30 @@ public string TransactionId
...
-                }
-
-                if (_newRelicPayload?.Priority != null && _newRelicPayload.Priority.HasValue)
-                {
-                    return _newRelicPayload.Priority;
-                }
-
-                return null;
-            }
+            get => _priority ??
```
nrcventura, line 173: These nested conditionals seem harder to read, and might make it harder to step through with a debugger. I think that it might be better to split up this logic using multiple if blocks and return statements (similar to the previous version of this code).

src/Agent/NewRelic/Agent/Core/DistributedTracing/TracingState.cs
```diff
@@ -254,5 +248,43 @@ public static ITracingState AcceptDistributedTraceHeaders<T>(T carrier, Func<T,
...
+                case RemoteParentSampledBehavior.AlwaysOff:
+                    _priority = 0.0f; // set lowest possible priority
+                    _sampled = false;
+                    break;
+                // don't need a case for Default, as it is handled above
+                default:
+                    throw new ArgumentException($"Invalid {(_traceContext.Traceparent.Sampled ? "remoteParentSampledBehavior" : "remoteParentNotSampledBehavior")} value: {sampledBehavior}.");
+            }
+
+            Log.Finest($"ApplyRemoteParentSampledBehavior:  _traceContext.Traceparent.Sampled={_traceContext.Traceparent.Sampled}, {(_sampled.Value ? "remoteParent" : "remoteParentNot")}SampledBehavior: {sampledBehavior} ==> Sampled: {Sampled}, Priority:{Priority}");
```
nrcventura, line 287: The string formatting for this finest level log message can be deferred if the code is switched to the `Log.Finest("sampled={sampledValue}", sampledValue)` style of log message strings.

src/Agent/NewRelic/Agent/Core/Utilities/HexStringConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+
+namespace NewRelic.Agent.Core.Utilities
+{
+    public static class HexStringConverter
+    {
+        public static byte[] FromHexString(this ReadOnlySpan<char> chars)
+        {
+            if (chars.Length % 2 != 0)
```
nrcventura, line 12: Is this a general requirement, or is this just applicable to the trace flags scenarios?

nrcventura: If it's a general requirement we may want to change the name of this class, because someone might want to use it convert `A` to an array of bytes, because we're naming this class as a "generic" HexStringConverter.

tests/Agent/UnitTests/CompositeTests/RemoteParentSampledBehaviorTests.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    public class RemoteParentSampledBehaviorTests
+    {
+        private CompositeTestAgent _compositeTestAgent;
+        private IAgent _agent;
+
+        [SetUp]
+        public void SetUp()
+        {
+        }
```
nrcventura, line 21: Do we need this?

src/Agent/NewRelic/Agent/Core/DistributedTracing/TracingState.cs
tippmar-nr: I always forget that. We probably should add a ticket to the backlog to scrub "improper" log formatting - it's used in many, many places.

src/Agent/NewRelic/Agent/Core/Utilities/HexStringConverter.cs
tippmar-nr: It takes two hex characters to (fully) represent one byte. I thought about left-padding with `0` if the length was odd, but decided against it.

nrcventura: Should we at least add some `///` comments to call out that limitation? I couldn't come up with a better name.

src/Agent/NewRelic/Agent/Core/DistributedTracing/TracingState.cs
tippmar-nr: I like it the way it is, but I'll change it since the implementation is a bit inconsistent with other properties in this class..

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -1015,6 +1015,12 @@ private bool IsDistributedTracingEnabled()
 
         public bool ExcludeNewrelicHeader => _localConfiguration.distributedTracing.excludeNewrelicHeader;
 
+        // TODO: Do we need to check serverless mode? Need to support environment variable?
```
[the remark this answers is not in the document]
tippmar-nr: Added environment variables, using same names as Ruby and Node (and maybe others)

src/Agent/NewRelic/Agent/Core/Utilities/HexStringConverter.cs
tippmar-nr: Sure.

nrcventura (approved):

PR #3108, opened 2025-04-25
Activity Source configuration

tippmar-nr:
Add `appSettings` configuration support for the list of Activity Sources that should included and excluded. Adds a (placeholder) list of default included activity sources.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -3098,6 +3098,53 @@ namespace NewRelic.Agent.Core.Configuration
         }
         #endregion
 
+        #region Otel Bridge
+
+        // TODO: configure the initial list of activity sources to include
+        private static readonly string[] DefaultIncludedActivitySources = ["NewRelic.Agent", "TestApp activity source"];
+
+        private List<string> _includedActivitySources;
+        public List<string> IncludedActivitySources
+        {
+            get
+            {
+                if (_includedActivitySources == null)
+                {
+
+                    var includedActivitySources = DefaultIncludedActivitySources.ToList();
+                    var appSetting = TryGetAppSettingAsString("OpenTelemetry.ActivitySource.Include");
+                    if (!string.IsNullOrEmpty(appSetting))
+                    {
+                        includedActivitySources.AddRange(appSetting.Split(','));
+                    }
+
+                    _includedActivitySources = includedActivitySources;
+                }
+                return _includedActivitySources;
+            }
+        }
+
+        private List<string> _excludedActivitySources;
+        public List<string> ExcludedActivitySources
+        {
+            get
+            {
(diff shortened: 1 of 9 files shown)


jaffinito (approved):

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -3098,6 +3098,53 @@ public string AwsAccountId
         }
         #endregion
 
+        #region Otel Bridge
+
+        // TODO: configure the initial list of activity sources to include
+        private static readonly string[] DefaultIncludedActivitySources = ["NewRelic.Agent", "TestApp activity source"];
```
nrcventura, line 3104: I think that this list should only include the `NewRelic.Agent` source for now, and that the hybrid agent cross agent tests should use the configuration to add the `TestApp activity source`.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -3098,6 +3098,53 @@ public string AwsAccountId
...
+
+        private List<string> _includedActivitySources;
+        public List<string> IncludedActivitySources
+        {
+            get
+            {
+                if (_includedActivitySources == null)
+                {
+
+                    var includedActivitySources = DefaultIncludedActivitySources.ToList();
```
nrcventura, line 3114: It might be better to do something like the following.
```suggestion
                    var includedActivitySources = new List<string>(DefaultIncludedActivitySources);
```

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
```diff
@@ -83,7 +83,7 @@ private void TryCreateActivityListener()
 
             _activityListener = Activator.CreateInstance(activityListenerType);
 
-            ConfigureShouldListenToCallback(_activityListener, activityListenerType, activitySourceType);
+            ConfigureShouldListenToCallback(this, _activityListener, activityListenerType, activitySourceType);
```
nrcventura, line 86: This method doesn't need to be static so that it can directly access `this` within the method.

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
tippmar-nr: I don't think there's any function difference in those two imnplementations.

nrcventura: .ToLIst usually requires enumerating the IEnumerable, but the .ctor taking the collection can do a bulk copy. For smaller collection sizes it probably doesn't matter. But it may matter as the lists grow. That was also the reason I recommended using `AddRange` to combine the lists.

PR #3109, opened 2025-04-30
feat: Added the ability to load additional appsetting file defined by…

mipa87:
## Description

Added the ability to load additional appsetting file defined by environment variable. This can be used if you want to override some settings when using appsetting configuration.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/AppSettingsConfigResolveWhenUsed.cs
@@ -28,6 +28,8 @@ namespace NewRelic.Agent.Core.Configuration
 
         private static IConfigurationRoot InitializeConfiguration()
         {
+            var env = new SharedInterfaces.Environment();
+
             // Get application base directory, where appsettings*.json will be if they exist
             var applicationDirectory = string.Empty;
             try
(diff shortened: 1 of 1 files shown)


tippmar-nr:
Thanks for your PR, @mipa87 - can you give us a bit more background on what you're trying to accomplish with this change and whether putting the agent config settings in one of the "usual" `appsettings.json` files is sufficient?

mipa87:
We have a service (installed by installer) that uses appsetting.json and appsetting.Production.json, not only for configuring NewRelic. Everything works fine, but there are some servers where we need to change some NewRelic configurations. On these servers, we also have some extra appsettings.json in another folder (is not affected by the installation process), and the service itself knows about it and can load options from it.
So this PR add abbility to use this extra appsettigs.json where we can "override" the NewRelic configuration.

tippmar-nr:
@mipa87 Can you elaborate on which specific New Relic configuration settings you need to override? Have you considered using [environment variables](https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#optional-environment-variables) instead?

mipa87:
@tippmar-nr I don't know which one specifically,  I'd have to look at the code, but Environment variables have a lower priority, so if you are using appsettings, you can't override it with Environment variable.

And yes, we have considered other possibilities, like use newrelic.config and overriding it with Environment variable, but the presented option with addtional appsettings is more common in. NET.

tippmar-nr:
> I don't know which one specifically, I'd have to look at the code, 

There are only 5 config settings that can be specified in `appSettings.json`:
* `NewRelic.ConfigFile` - specifies the location of the `newrelic.config` file. In addition to the path in this setting, the application root folder, execution path, `NEW_RELIC_HOME` and the current working folder are all searched.
* `NewRelic.AgentEnabled` - controls whether the agent is enabled for this specific application. The value from `newrelic.config` is used if this setting isn't specified.
* `NewRelic.LicenseKey` - specifies the license key; `NEW_RELIC_LICENSE_KEY` can be used instead.
* `NewRelic.AppName` - specifies the application name; `NEW_RELIC_APP_NAME` can be used instead
* `NewRelic.Labels` - specifies application labels; `NEW_RELIC_LABELS` can be used instead.

> And yes, we have considered other possibilities, like use newrelic.config and overriding it with Environment variable

Our advice would be to not use the `appSettings.json` settings listed above at all, preferring `newrelic.config` and environment variables. 

At the present time, we are unable to accept your proposed change but we would be happy for you to file a Feature Request that we can add to our backlog for future development. Thanks!

PR #3113, opened 2025-05-02
Added config to enable otel bridge

tippmar-nr:
Disabled by default. 

```
<appSettings>
  <add key="OpenTelemetry.Enabled" value="true" />
</appSettings>
```
`NEW_RELIC_OPEN_TELEMETRY_BRIDGE_ENABLED="true"` (or `1`)


src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd
```diff
@@ -2072,6 +2072,23 @@
                         </xs:all>
                     </xs:complexType>
                 </xs:element>
+
+                <xs:element name="openTelemetryBridge" minOccurs="0" maxOccurs="1">
```
nrcventura, line 2076: Can we avoid making this change in the xsd file? We still need to figure out whether we want customers to be able to turn this off once the feature is GA. For that reason, we should treat this setting as a feature flag, and define an appsetting for it.

nrcventura (changes requested):
We're not ready to have a permanent configuration setting for this yet. This should be treated as a feature flag for now.

nrcventura (approved):

PR #3118, opened 2025-05-08
Feature work/asb followup changes

jaffinito:
Adds a comment with the source of the new message broker actions
Labels the "problem" agent log regex values in the integration tests
Moves the ExecutionTimer class into its own file


tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
```diff
@@ -584,66 +587,17 @@ protected void PostJson(string address, string payload, List<KeyValuePair<string
...
-        }
-
-        /// <summary>
-        /// Executes an asynchronous action and aggregates its run time into the total.
-        /// </summary>
-        /// <param name="asyncAction">The action to measure.</param>
-        public async Task AggregateAsync(Func<Task> asyncAction)
+        // Tests for things like transaction garbage collected, response time already captured, etc.
+        // Works best when logging is at FINEST.
+        private void TestForKnownProblems()
```
tippmar-nr, line 592: How about making this `virtual` so it can be extended/overridden when necessary?

tippmar-nr:
Can we make these changes in a PR to `main` rather than down here in the feature branch? That'll keep the feature branch from getting diluted with changes unrelated to the actual feature.

tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/RemoteApplicationFixture.cs
jaffinito: Will do when I reimplement this in main.

tests/Agent/IntegrationTests/UnboundedIntegrationTests.sln
```diff
@@ -6,16 +6,9 @@ MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "UnboundedIntegrationTests", "UnboundedIntegrationTests\UnboundedIntegrationTests.csproj", "{1E17AA8A-B8BA-4A64-892C-B58151293142}"
 	ProjectSection(ProjectDependencies) = postProject
 		{5350D3F3-CE09-4992-9756-ACCC92A9129C} = {5350D3F3-CE09-4992-9756-ACCC92A9129C}
-		{1D93182F-CF1D-47A1-BC0F-4E27DE9ACEF9} = {1D93182F-CF1D-47A1-BC0F-4E27DE9ACEF9}
```
tippmar-nr, line 9: What are these changes for?

jaffinito: Things work fine locally, but when running in CI they fail with solution errors about not finding projects for the unbounded tests.  I am trying to match my SLN file to main to resolve the problems.  Since they ran this time, it looks like it worked.

PR #3128, opened 2025-05-20
feat: Added transaction renaming requirement for Serverless APM mode

ashishsinghnr:
https://source.datanerd.us/agents/agent-specs/pull/733
Serverless APM Mode is a new agent mode required by the Lambda Fusion project, and which would allow Lambda telemetry to be displayed in the APM UI. Some of the requirements are handled by Serverless Team, as in the case of the Lambda Extension. One requirement, however, will (with some exceptions, such as Python) need to be handled by agents.

For this mode, the Function Name segment of the the transaction name will need to have the event source type added before the function name, separated from the function name by a space character. WebTransaction/Function/myFunction, for example, would be renamed WebTransaction/Function/apiGateway myFunction if triggered by API Gateway.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/ReportedConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/ReportedConfiguration.cs
@@ -742,6 +742,10 @@ namespace NewRelic.Agent.Core.Configuration
         [JsonProperty("agent_control.health.frequency")]
         public int HealthFrequency => _configuration.HealthFrequency;
 
+        [JsonIgnore]
+        public bool AwsLambdaApmModeEnabled => _configuration.AwsLambdaApmModeEnabled;
+        
+
         #endregion
     }
 }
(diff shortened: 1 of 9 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/ReportedConfiguration.cs
```diff
@@ -742,6 +742,10 @@ public IReadOnlyDictionary<string, string> GetAppSettings()
         [JsonProperty("agent_control.health.frequency")]
         public int HealthFrequency => _configuration.HealthFrequency;
 
+        [JsonIgnore]
```
nr-ahemsath, line 745: I think this is a configuration setting we would want to report to New Relic along with the rest of agent config, to help with support troubleshooting.

```suggestion
        [JsonProperty("agent.aws_lambda_apm_mode_enabled")]
```

nrcventura: There's no connect cycle in serverless mode, so it doesn't matter.

tippmar-nr (approved):
Looks good. Thanks for updating the integration tests!

PR #3137, opened 2025-05-23
fix: Add new CreateTransaction api method to allow overriding existing transaction name.

jaffinito:
fix: Add new CreateTransaction api method to allow overriding existing transaction name.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/AzureFunctionInProcessExecuteWithWatchersAsyncWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/AzureFunctionInProcessExecuteWithWatchersAsyncWrapper.cs
@@ -68,7 +68,7 @@ public class AzureFunctionInProcessExecuteWithWatchersAsyncWrapper : IWrapper
             isWeb: inProcessFunctionDetails.IsWebTrigger,
             category: "AzureFunction",
             transactionDisplayName: inProcessFunctionDetails.FunctionName,
-            doNotTrackAsUnitOfWork: true);
+            doNotTrackAsUnitOfWork: false);
 
         if (instrumentedMethodCall.IsAsync)
         {
--- a/tests/Agent/UnitTests/CompositeTests/TransactionNameTests.cs
+++ b/tests/Agent/UnitTests/CompositeTests/TransactionNameTests.cs
@@ -31,15 +31,25 @@ namespace CompositeTests
 
         #region Segment metrics and trace names
 
-        [Test]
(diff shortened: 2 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -130,6 +130,11 @@ public ITransaction CreateKafkaTransaction(MessageBrokerDestinationType destinat
         }
 
         public ITransaction CreateTransaction(bool isWeb, string category, string transactionDisplayName, bool doNotTrackAsUnitOfWork, Action wrapperOnCreate)
+        {
+            return CreateTransaction(isWeb, category, transactionDisplayName, doNotTrackAsUnitOfWork, wrapperOnCreate, (TransactionNamePriority)9999);
```
tippmar-nr, line 134: Is a hardcoded 9999 the best we can do here? Could we introduce a `TransactionNamePriority.Highest` value in the enum? I see that `UserTransactionName` is `int.MaxValue` - so maybe we want to use `int.MaxValue -1` for this priority?

src/Agent/NewRelic/Agent/Core/Agent.cs
```diff
@@ -145,7 +150,21 @@ public ITransaction CreateTransaction(bool isWeb, string category, string transa
...
+            {
+                if (initialTransactionName.IsWeb)
+                {
+                    transaction.SetWebTransactionName(initialTransactionName.Category, initialTransactionName.Name, transactionNamePriority);
+                }
+                else
+                {
+                    transaction.SetOtherTransactionName(initialTransactionName.Category, initialTransactionName.Name, transactionNamePriority);
+                }
+            }
```
nrcventura, line 165: I'm wondering if this logic is better off being handled within the service bus instrumentation directly. Right now that's the only scenario for it, and we might not have other scenarios.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureFunction/AzureFunctionInProcessExecuteWithWatchersAsyncWrapper.cs
nrcventura, line 1: I think we also need to treat the call to CreateTransaction as a unitOfWork, so that ending this transaction will not also end the service bus transaction early in the case of service bus function triggers.

src/Agent/NewRelic/Agent/Core/Agent.cs
jaffinito: I removed all of this and simplified things a bit.

nrcventura (approved):

PR #3148, opened 2025-06-05
Chore : updated the category name of the lambda transaction

ashishsinghnr:
Based on the discussion with PM team, for AWS lambda transactions, Category should be "Function" instead if "Lambda" in the transaction  naming.


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AwsLambda/HandlerMethodWrapper.cs
```diff
@@ -248,7 +248,7 @@ public AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall ins
             // else WebTransaction/Function/myFunction
             transaction = agent.CreateTransaction(
                 isWeb: _functionDetails.EventType.IsWebEvent(),
-                category: "Lambda",
+                category: "Function",
```
nrcventura: Should this change also be gated by the AwsLambdaApmModeEnabled check?

ashishsinghnr: Yes, it would be good idea to keep the new value gated by the AwsLambdaApmModeEnabled check. Updated the code.

nrcventura (approved):

PR #3149, opened 2025-06-05
chore: Remove `Memoizer`, refactor `DefaultConfiguration`

tippmar-nr:
* Removes the `Memoizer` helper class and replaces usage with null-coalescing assignment operator (`??=`).
* Significant `DefaultConfiguration` refactoring (reduced file size by about 400 lines):
  * Use null-coalescing assignment operators (`??=`) in as many properties as possible
  * Convert most `get`-only properties to use expression bodies for more concise code.
  * Refactors explicit collections in many places to use collection initializers.


tests/Agent/UnitTests/Core.UnitTest/Configuration/DefaultConfigurationTests.cs
```diff
@@ -62,8 +57,6 @@ public void SetUp()
             _agentHealthReporter = Mock.Create<IAgentHealthReporter>();
 
             _defaultConfig = new TestableDefaultConfiguration(_environment, _localConfig, _serverConfig, _runTimeConfig, _securityPoliciesConfiguration, _bootstrapConfiguration, _processStatic, _httpRuntimeStatic, _configurationManagerStatic, _dnsStatic, _agentHealthReporter);
-
-            TestableDefaultConfiguration.ResetStatics();
```
nrcventura, line 66: Did removing this cause the test coverage to drop?

tippmar-nr: The `ResetStatistics()` method didn't do anything useful -- it set the value of a couple of private fields that weren't referenced anywhere else. I don't know what we intended for it to do, but it doesn't do that...

nrcventura: It looks those fields can be removed. That TestableDefaultConfiguration was the only place left that they seemed to be used. But that can be done in a separate PR.

nrcventura (approved):

PR #3151, opened 2025-06-11
chore: ActivityBridge Client Http External support

tippmar-nr:
* Update `ActivityBridge` to handle Client activity kind for an external HTTP call
  * Creates the correct span name and attributes
  * Adds any OTel tags that weren't used when creating the ExternalSegmentData instance as User attributes on the segment
* Significant refactor in CompositeTests project to separate hybrid agent "smoke tests" from "general" tests
  * Added tests for external http activity to ensure code coverage  
* Add support for `http.statusText` attribute on spanEvent (exists in the spanEvent spec but was marked "optional") and refactor as needed. Did not modify existing http client instrumentation to include that attribute.


src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
```diff
@@ -506,15 +508,122 @@ private static void ActivityStopped(object originalActivity, IAgent agent, IErro
             }
         }
 
-        private static void AddActivityTagsToSegment(object originalActivity, ISegment segment)
+        private static void AddActivityTagsToSegment(object originalActivity, ISegment segment, IAgent agent)
```
nrcventura, line 511: This method is starting to get large, when we touch it again, we should refactor it so that it delegates work out to other methods.

tests/Agent/UnitTests/CompositeTests/HybridAgent/TestCaseDefinitions/HybridAgentTests.json
nrcventura, line 1: I like how this sets up the possibility for a cross agent test suite for the conversion rules.

PR #3169, opened 2025-06-26
test: Fix server name parsing logic for Oracle and Redis configuration

tippmar-nr:
This PR brings in some changes from my k8s PR that adjust how the Oracle and Redis tests parse the configured server name from configuration. The changes are necessary for tests to succeed when pointed at the Azure k8s based unbounded services. 

The previous version of the affected files worked only because the services were referenced by IP address instead of hostname.

### DIFF
--- a/tests/Agent/IntegrationTests/Shared/OracleConfiguration.cs
+++ b/tests/Agent/IntegrationTests/Shared/OracleConfiguration.cs
@@ -16,7 +16,7 @@ namespace NewRelic.Agent.IntegrationTests.Shared
         private static string _oraclePort;
         private static bool _parsedHostPort = false;
 
-        // example: "Data Source=1.2.3.4:4444/XE;User Id=SYSTEM;Password=oraclePassword;"
+        // example: "Data Source=some.server.name:4444/FREEPDB1;User Id=SYSTEM;Password=oraclePassword;"
         public static string OracleConnectionString
         {
             get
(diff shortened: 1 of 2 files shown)


tests/Agent/IntegrationTests/Shared/OracleConfiguration.cs
```diff
@@ -82,20 +72,43 @@ public static string OraclePort
...
                 }
+            }
+        }
 
-                return _oraclePort;
+        // Parses host and port from a data source string like host:port/service
+        private static void ParseHostAndPort(string dataSource, out string host, out string port)
+        {
+            // Regex matches host (hostname or IPv4), port, and ignores service name
+            var match = Regex.Match(dataSource, @"^(?<host>[^:/\[]+(?:\.[^:/\[]+)*)[:](?<port>\d+)");
```
nrcventura: I'm interested in hearing what other team members think, but I feel like using a regex here is more complicated than we need it to be. We could probably get away with something similar to what was done previously by doing something like `new UriBuilder("oracle://" + dataSource)`.

chynesNR: That's probably true, but it's only test code, and I think it's clear enough from the context what it's doing.

tippmar-nr: Refactored to eliminate regex

nrcventura (approved):

PR #3188, opened 2025-07-22
chore: Support Elastic.Clients.Elasticsearch v8.16.0+ using OpenTelemetry Bridge

tippmar-nr:
## Description

This PR includes support for `Elastic.Clients.Elasticsearch` v8.16.0 and later by way of the OpenTelemetry bridge. 

Note that the OpenTelemetry bridge is disabled by default and the `Elastic.Transport` activity source is not automatically included, so the changes in this PR will not have any impact on existing customers. 

We need to complete this PR as it forms that basis for some additional OTel Bridge work that is coming soon.


src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
```diff
@@ -217,7 +216,34 @@ private void ConfigureTraceIdGenerator(Type activityType, Type activityTraceIdTy
         var parseTraceIdMethod = activityTraceIdType.GetMethod("CreateFromString", BindingFlags.Public | BindingFlags.Static);
         var parseTraceIdExpression = Expression.Call(null, parseTraceIdMethod, currentTraceIdAsSpanExpression);
 
-        var getOrCreateTraceIdExpression = Expression.Condition(hasCurrentTraceIdExpression, parseTraceIdExpression, generateNewTraceIdExpression);
+        var logFinestMethod = typeof(Log).GetMethod("Finest", new[] { typeof(string), typeof(object[]) });
```
chynesNR, line 219: I realize this is just logging, but could you add it to the pseudocode comment above so it's easier to follow what it's doing?

src/Agent/NewRelic/Agent/Core/Core.csproj
```diff
@@ -76,9 +76,9 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Grpc.Net.Client" Version="2.71.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="[2.0.0]" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="[2.0.0]" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="[2.0.0]" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.0.0" />
```
nrcventura, line 81: What's the reason for unpinning these dependencies?

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/Instrumentation.xml
```diff
@@ -5,6 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 -->
 <extension xmlns="urn:newrelic-extension">
 	<instrumentation>
+    <!-- TODO: Add maxVersion="8.15.10" on all <match> elements when OtelBridge is enabled by default in the agent. -->
```
nrcventura, line 8: If the instrumentation does not work with those versions, should we add the maxVersion anyways?

src/Agent/NewRelic/Agent/Core/Core.csproj
tippmar-nr: Updates to the CompositeTests project required the addition of the OpenTelemetry package, which depends on a newer version of the MEC packages. Pinning the packages in Core was really aimed at stopping us from updating those versions via manual NuGet updates - but we're using Dependabot now to manage NuGet updates, which (for the moment, anyway) is excluding those packages.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/Instrumentation.xml
tippmar-nr: I was trying to preserve existing functionality -- method signatures in newer versions won't match, so we won't instrument Elastic anyway, with or without the maxVersion restriction.

src/Agent/NewRelic/Agent/Core/Core.csproj
nrcventura: We may want to double check that removing the pin did not result in the agent depending on a "newer" version of those assemblies.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Elasticsearch/Instrumentation.xml
nrcventura: So it seems like we're fine either way, because that means it would not be necessary for someone who wanted to try out the otel elastic instrumentation to also set the maxVersion on the new relic elastic instrumentation.

src/Agent/NewRelic/Agent/Core/Core.csproj
tippmar-nr: I think we're good here -- I originally [introduced the pinning last year ](https://github.com/newrelic/newrelic-dotnet-agent/commit/4c146cef52e79c1da9082a80c6713a164d0fbb8a#diff-11e473967014d04ced457fa2d39f3feadcd164c50b90efaa43d6292ecb5bea49R53-R72)specifically to address the package update issue; prior to that, it had been unpinned.

PR #3211, opened 2025-08-06
summary: Add support for TraceId Ratio-Based Sampler

tippmar-nr:
feat: Add a new Distributed Tracing sampler, the TraceId Ratio-Based Sampler, which is based on the [OpenTelemetry implementation](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Trace/Sampler/TraceIdRatioBasedSampler.cs). This sampler makes its sampling decision using the TraceId (a random GUID) and a configured sampling ratio (a target percentage of traces to be sampled).  For more information, see the [OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/).

With this feature, we are also introducing new configuration settings related to Distributed Tracing which will enable fine-grained control over which sampler is configured at various points in the Distributed Tracing flow. By default, the .NET agent will continue to use the original Adaptive Sampler for all sampling decisions. The new configuration allows for any of the 4 available samplers (Adaptive, TraceId Ratio-Based, Always On and Always Off) to be configured for the "root" of a trace or when the trace has (or does not have) a Remote Parent. Please refer to the [.NET Agent documentation](https://docs.newrelic.com/docs/apm/agents/net-agent/configuration/net-agent-configuration/#distributed_tracing) for a detailed explanation of this configuration option.


src/Agent/NewRelic/Agent/Core/Api/TraceMetadataFactory.cs
```diff
@@ -32,11 +33,11 @@ public interface ITraceMetadataFactory
 
     public class TraceMetadataFactory : ITraceMetadataFactory
     {
-        private readonly IAdaptiveSampler _adaptiveSampler;
+        private readonly ISamplerService _samplerService;
```
nrcventura: Is this used anywhere? Maybe we can remove it?

tippmar-nr: No it's not used. Missed that in my last round of refactoring.

PR #3212, opened 2025-08-08
Feature work/traceid sampler interface refactor

tippmar-nr:
Some significant refactoring:
* Added a `Samplers` folder under `DistributedTracing` which will contain the various samplers we implement. This may not be the final location if it turns out that moving the folder up a level makes more sense.
* Introduced `ISampler`, `ISamplingParameters` and `ISamplingResult` interfaces
* Refactored `AdaptiveSampler` to implement the new interfaces and remove the previous `IAdaptiveSampler` interface. 
* Implemented the new interfaces in `TraceIdRatioSampler`
* Added `AlwaysOnSampler` and `AlwaysOffSampler`, which are trivial implementations of the existing `RemoteParentSampledBehavior` sampling logic and refactored `TracingState` as needed. 
* Updated `TracingState` to implement use of `TraceIdRatioSampler` in `AcceptDistributedTraceHeaders`
* Added unit tests to `TracingStateTests` to exercise `TraceIdRatioSampler`
* Refactored unit tests as needed.

These changes are an intermediate step toward implementing a `SamplingService` which will manage the samplers at runtime.


src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/TraceIdRatioSampler.cs
```diff
@@ -0,0 +1,87 @@
...
+        };
+    }
+
+    public ISamplingResult ShouldSample(ISamplingParameters samplingParameters)
+    {
+        if (string.IsNullOrEmpty(samplingParameters.TraceId))
+        {
+            throw new ArgumentNullException(nameof(samplingParameters.TraceId), "Trace ID cannot be null or empty.");
+        }
+        if (samplingParameters.TraceId.Length < 16)
```
chynesNR: Maybe make the min length a constant? Looks like it's used in a few places

src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/AdaptiveSampler.cs
chynesNR, line 1: None of the logic here changed, right? It just moved? (And implements `ISampler` now)

tippmar-nr: `AdaptiveSampler.BoostPriority` was refactored significantly. Compare to the original at https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/DistributedTracing/AdaptiveSampler.cs#L255

src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/AdaptiveSampler.cs
```diff
@@ -0,0 +1,289 @@
...
+        _state = new AdaptiveSamplerState(targetSamplesPerInterval, TimeSpan.FromSeconds(targetSamplingIntervalInSeconds), rand, manualIntervalCheck);
+    }
+
+    public void StartTransaction()
+    {
+        _state.ManualCheckAndUpdateInterval();
+    }
+
+    /// <summary>
+    /// Atomically boost the priority
```
jaffinito, line 244: This looks fine to me.  I don't know how important it would be for this to be atomic, doesn't see to be.

src/Agent/NewRelic/Agent/Core/DistributedTracing/TracingState.cs
```diff
@@ -280,20 +308,18 @@ private void ApplyRemoteParentSampledBehavior(
...
-                    break;
-                case RemoteParentSampledBehavior.AlwaysOff:
-                    _priority = 0.0f; // set lowest possible priority
-                    _sampled = false;
-                    break;
-                // don't need a case for Default, as it is handled above
-                default:
-                    throw new ArgumentException($"Invalid {(_traceContext.Traceparent.Sampled ? "remoteParentSampledBehavior" : "remoteParentNotSampledBehavior")} value: {sampledBehavior}.");
-            }
+                RemoteParentSampledBehavior.AlwaysOn => new AlwaysOnSampler(),
```
chynesNR: Since these samplers have no internal state, maybe we don't need to create a new instance every time?

src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/AdaptiveSampler.cs
nrcventura: I think the main reason for the atomic update is if there are 2 external calls being made in parallel for the same transaction (assuming that one of those calls is triggering the sampling decision to be made). I believe that the priority and sampling decision are both part of the headers passed along in those calls.

src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/AlwaysOffSampler.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace NewRelic.Agent.Core.DistributedTracing.Samplers;
+
+public class AlwaysOffSampler: ISampler
+{
+    public ISamplingResult ShouldSample(ISamplingParameters samplingParameters) => new SamplingResult(false, 0.0f);
```
nrcventura: We should be able to reduce allocations if we store this AlwaysOff sampling result as a static field in this class, and then this method can just return a reference to that static instance.

src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/AlwaysOnSampler.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace NewRelic.Agent.Core.DistributedTracing.Samplers;
+
+public class AlwaysOnSampler : ISampler
+{
+    public ISamplingResult ShouldSample(ISamplingParameters samplingParameters) => new SamplingResult(true, 2.0f);
```
nrcventura: We should be able to apply the same optimization that I'm suggesting for the AlwaysOff sampler here too.

src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/AdaptiveSampler.cs
tippmar-nr: I did some analysis of the original implementation and all the gyrations around "atomically" boosting the priority are meaningless -- the BoostPriority method is called from within a `lock()` so it's already protected.

PR #3213, opened 2025-08-12
TraceIdSampler - Sampler Service

tippmar-nr:
* Add `SamplerService` to manage configuration and selection of samplers from one location. 
* Refactor `AdaptiveSampler` to remove the `ConfigurationBasedService` base class. 
* Update usages everywhere and refactor as required.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/TraceIdRatioSampler.cs
+++ b/src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/TraceIdRatioSampler.cs
@@ -11,6 +11,8 @@ public class TraceIdRatioSampler : ISampler
     private readonly long _idUpperBound;
     private const int TraceIdLength = 16;
 
+    private const float PriorityBoost = 1.0f;
+
     public TraceIdRatioSampler(float sampleRatio)
     {
         _idUpperBound = sampleRatio switch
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/DistributedTracing/Samplers/SamplerService.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    /// <param name="samplerType"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentOutOfRangeException"></exception>
+    ISampler GetSampler(SamplerType samplerType);
+}
+
+public class SamplerService : ConfigurationBasedService, ISamplerService
+{
+    private ISampler _rootSampler;
```
chynesNR: Rather than making these distinct variables, maybe this should be a Dictionary with the `SamplerType` as the key? That might make the code a little cleaner, and would make it easy to add any new sampler conditions later (Though admittedly there probably won't be any others).

src/Agent/NewRelic/Agent/Core/DistributedTracing/DistributedTracePayloadHandler.cs
```diff
@@ -78,7 +78,7 @@ public void InsertDistributedTraceHeaders<T>(IInternalTransaction transaction, T
                 }
                 else
                 {
-                    transaction.SetSampled(_sampler);
+                    transaction.SetSampled(_samplerService.GetSampler(SamplerType.Root)); // TODO: Is Root correct here?
```
chynesNR, line 81: I _think_ the way you're using it throughout is correct. While we don't strictly know if we want the "root" sampler here, if we didn't, the Sampled value would already be set, and this would do nothing.

PR #3219, opened 2025-08-14
Add configuration for sampler selection

tippmar-nr:
Significantly modifies the configuration XSD to add support for selecting from the 4 different sampler types (Default, TraceIdRatioBased, AlwaysOn, AlwaysOff) at the root and remote parent sampled / not sampled levels. 

The existing `remoteParentSampled` and `remoteParentNotSampled

Selecting Default at the root level uses the Adaptive sampler.
Selecting Default at the remote parent sampled / not sampled level uses the existing "default" behavior

Selecting TraceIdRatioBased sampler at any level, without also configuring a sampleRatio results in a warning message and falls back to using the Default sampler.

Refactored existing code as required; added and updated unit tests. 

Note that there is currently no test to verify behavior when replacing the root sampler with something other than default; that will follow in a separate PR.


src/Agent/NewRelic/Agent/Core/Api/TraceMetadataFactory.cs
```diff
@@ -57,7 +58,7 @@ private bool SetIsSampled(IInternalTransaction transaction)
             }
             else
             {
-                transaction.SetSampled(_samplerService.GetSampler(SamplerType.Root)); //TODO: Is Root correct here?
+                transaction.SetSampled(_samplerService.GetSampler(SamplerLevel.Root)); //TODO: Is Root correct here?
```
chynesNR: The name `SamplerType` did bother me a little but I couldn't think of a better one. I like `SamplerLevel`.
Do we still need the TODOs?

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -876,12 +876,88 @@ private bool IsDistributedTracingEnabled()
...
+                DefaultSamplerType => SamplerType.Default,
+                AlwaysOnSamplerType => SamplerType.AlwaysOn,
+                AlwaysOffSamplerType => SamplerType.AlwaysOff,
+                TraceIdRatioSamplerType => SamplerType.TraceIdRatioBased,
+                _ => throw new ArgumentOutOfRangeException(nameof(samplerItem), samplerItem, "Unknown sampler type in configuration.")
+            };
+
+        private float? TraceIdSamplerRatio(SamplerLevel samplerLevel)
+        {
+            // TODO: do we need to support environment variable override for these? Naming will be difficult.
```
chynesNR, line 942: I think customers are going to want that, but I think we can simplify it to the cases that make the most sense and just offer:
* Ratio for root sampling
* Honor the remote parent sampled flag (do we not have this already?)

We can do this as a separate work item, though

src/Agent/NewRelic/Agent/Core/Api/TraceMetadataFactory.cs
tippmar-nr: Yeah, I can get rid of the TODOs I think.

PR #3227, opened 2025-08-18
summary: Service Bus instrumentation fix / refactor

### LINKED ISSUE #3226: Service Bus instrumentation exception handling issue
**Description**
In certain circumstances during instrumentation of Azure Service Bus, an exception gets logged at FINEST level:

```
2025-08-18 15:25:21,770 NewRelic FINEST: [pid: 8504, tid: 38] Trx 083af67efa2fb9fe: Unexpected exception: 'System.Array' does not contain a definition for 'Count'
```
When this happens, the transaction that is in process doesn't get properly terminated and can lead to incorrect transaction duration reporting. 

The source of the issue is in the `ExtractDTHeadersIfAvailable()` and is caused by trying to access the `Count` property on a `dynamic` variable that represents a `System.Array` instance. The problem can be fixed by casting the object to `ICollection` before accessing the `Count` property.

Additionally, the `catch` block in the `GetAsyncDelegateFor()` method should be removed, as this was intended for use only during initial development and testing. For production use, exceptions like this should be allowed to bubble out and be handled by the agent. 

**Expected Behavior**
No exception should be thrown when accessing the message count property. Exceptions that do get thrown should not be caught by instrumentation wrappers.

P3
S

tippmar-nr:
fix: Exception incorrectly handled in Service Bus instrumentation wrapper.

chore: Refactor "no messages received" logic in Service Bus receive wrapper to ensure transactions are always ignored if `ReceiveMessagesAsync` returns 0 messages.

Fixes #3226

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -43,7 +43,7 @@ updates:
           - "*"
     ignore:
       - dependency-name: "Autofac" # v6+ doesn't support .NET Framework; we could upgerade to v5.x but there's no good reason for it
-      - dependency-name: "Microsoft.Extensions.Configuration.*"
+      - dependency-name: "Microsoft.Extensions.Configuration*"
       - dependency-name: "SharpZipLib" # we can't upgrade beyond 1.3.3
 
 


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiveWrapper.cs
```diff
@@ -106,31 +115,29 @@ public override AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMetho
...
                         {
-                            transaction.LogFinest("ReceiveMessagesAsync task was canceled in processor mode. Ignoring transaction.");
-                            transaction.Ignore();
+                            transaction.LogFinest($"{(responseTask.IsCanceled ? "ReceiveMessagesAsync task was canceled in processor mode" : "No messages received")}. Ignoring transaction.");
 
-                            // End the transaction here since end in the AzureServiceBusReceiverManagerWrapper is never called
+                            // Ignore and end the transaction here since end in the AzureServiceBusReceiverManagerWrapper is never called
                             // In FW this results in a transaction has been garbage collected message, probably due to different GC settings in FW vs. Core
-                            transaction.End(); 
+                            transaction.Ignore();
```
chynesNR, line 139: Is there any way this could discard a valid transaction?

tippmar-nr: Shouldn't be possible. This code is mostly just a refactor of existing logic, with the addition of handling the "no messages" case to ensure the transaction gets ended after being ignored.

chynesNR (approved):

PR #3230, opened 2025-08-19
chore: Fix ignore rule for Microsoft.Extensions.Configuration

nr-ahemsath:
Fix the ignore rule for Microsoft.Extensions.Configuration* to actually match Microsoft.Extensions.Configuration as well as its sub-namespaces.


tippmar-nr (approved):
It's worth a try, anyway.

PR #3239, opened 2025-08-20
ci: Refactor container integration tests, improve resiliency, add logging

tippmar-nr:
* Refactors the container integration tests workflow to use a matrix by architecture and distro name.
* Improves resiliency by adding retry logic for starting containers and ensuring that networks are not left dangling
* Additional logging sections added to test log to help diagnose future failures


.github/workflows/linux_container_tests.yml
```diff
@@ -78,77 +116,21 @@ jobs:
         uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4.3.1
         with:
           dotnet-version: '9.0.x'
-
```
tippmar-nr, line 81: It's a little hard to follow the changes here, but what's happening is that 2 separate jobs (amd64 / arm64) are now running as a single matrix job. So all these steps that look like they were dumped are just the steps from one of the 2 jobs that got coalesced into a single job.

chynesNR (approved):
Nice work, this looks a lot cleaner

PR #3245, opened 2025-09-03
chore: Add support for Producer / Consumer activity kinds in OpenTelemetry Bridge

tippmar-nr:
Add support for Producer / Consumer activity kinds in OTel bridge. Creates a MessageBrokerSegmentData instance using tags from the activity. 

Supports these activity kinds generically, with explicit support for vendor-specific tags used by RabbitMQ and Kafka. 

Does not modify the default set of activity sources that the .NET agent listens to and (still) does not enable OTel Bridge by default, so there is no impact on current users.


src/Agent/NewRelic/Agent/Core/Metrics/MetricNames.cs
```diff
@@ -346,6 +346,23 @@ public enum MessageBrokerDestinationType
             TempQueue,
             TempTopic
         }
+        public static MetricNames.MessageBrokerDestinationType AgentWrapperApiEnumToMetricNamesEnum(NewRelic.Agent.Extensions.Providers.Wrapper.MessageBrokerDestinationType wrapper)
```
nrcventura: Should the parameter name be something similar to `destinationType`? That parameter values does not represent a `wrapper`.

src/Agent/NewRelic/Agent/Core/Metrics/MetricNames.cs
```diff
@@ -358,6 +375,31 @@ public enum MessageBrokerAction
             Cancel
         }
 
+        public static MetricNames.MessageBrokerAction AgentWrapperApiEnumToMetricNamesEnum(NewRelic.Agent.Extensions.Providers.Wrapper.MessageBrokerAction wrapper)
```
nrcventura: This parameter is more like an `actionType` than a `wrapper`.

chynesNR (approved):
We're building up a pretty sizeable list of `TODO`s. Let's make sure we leave enough time for https://new-relic.atlassian.net/browse/NR-400936

PR #3246, opened 2025-09-11
chore: Support SERVER activity kind

tippmar-nr:
Adds preliminary support for Server activities in the Open Telemetry Bridge. Includes composite tests.

Does not change the default activity sources or enable the OTel bridge by default, so these changes will have no impact on existing customers.


tests/Agent/UnitTests/CompositeTests/HybridAgent/Helpers/HybridAgentTestsBase.cs
```diff
@@ -276,14 +276,36 @@ private void ValidateTransactions(IEnumerable<Transaction> transactionsToCheck)
...
+                Has.One.Matches<TransactionEventWireModel>(t => TransactionEventHasName(t, expectedTransaction.Name!)),
+                $"Expected transaction {expectedTransaction.Name} not found in {string.Join(", ", actualTransactionEvents.Select(GetTransactionNameFromTransactionEvent))}."
+            );
+
+            var actual = actualTransactionEvents.Single(t => TransactionEventHasName(t, expectedTransaction.Name!));
+
+            if (expectedTransaction.Attributes != null && expectedTransaction.Attributes.Any())
+            {
+                // Merge intrinsic, agent, and user attributes into a single lookup (we don't care where they came from for this validation)
+                var combined = new Dictionary<string, object>(StringComparer.Ordinal);
```
nrcventura: I think you can replace all of the foreach loops with a call to `GetAllAttributeValuesDic()` which should give you a dictionary of al of the attributes combined.

Issue #3240, opened 2025-08-21
Agent install on Debian 13 (Trixie) with APT fails with GPG key signature error


nr-ahemsath:
The following Dockerfile fails to build:
```
FROM debian:13

# Install the agent
RUN apt-get update && apt-get install -y wget ca-certificates gnupg
RUN echo 'deb [signed-by=/usr/share/keyrings/newrelic-apt.gpg] http://apt.newrelic.com/debian/ newrelic non-free' | tee /etc/apt/sources.list.d/newrelic.list
RUN wget -O- https://download.newrelic.com/NEWRELIC_APT_2DAD550E.public | gpg --import --batch --no-default-keyring --keyring /usr/share/keyrings/newrelic-apt.gpg
RUN apt-get update
RUN apt-get install -y newrelic-dotnet-agent
RUN rm -rf /var/lib/apt/lists/*

ENTRYPOINT ["bash"]
```
The error seen is:
```
15.57 W: OpenPGP signature verification failed: https://apt.newrelic.com/debian newrelic Release: Sub-process /usr/bin/sqv returned an error code (1), error message is: Error: Failed to parse keyring "/usr/share/keyrings/newrelic-apt.gpg"  Caused by:     0: Reading "/usr/share/keyrings/newrelic-apt.gpg": EOF     1: EOF
15.57 E: The repository 'http://apt.newrelic.com/debian newrelic Release' is not signed.
```

Debian 13 upgraded APT to version 3.0, which uses Sequoia PGP (`sv`) instead of GNU PGP (`gpg`) for repo signature verification.  GPG apparently uses a proprietary "keybox" format, depending on how you import keys, which sv will not understand.

Two possible solutions:
1. Come up with a different way of using GPG to import our public key that sv can work with (preferred solution)
2. Figure out how to tell APT 3.0 to use gpg instead of sv (okay for a temporary workaround)

References:
https://github.com/freedomofpress/dangerzone/issues/1052

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-454440

nr-ahemsath:
The following command successfully configures our signing key for our APT repo on both Debian 13 (APT 3/sqv) and Debian 11 (APT <3/gpgv):
`curl -fsSL https://download.newrelic.com/NEWRELIC_APT_2DAD550E.public | gpg --dearmor | sudo tee /usr/share/keyrings/newrelic-apt.gpg >/dev/null`

[...]

PR #3249, opened 2025-09-15
ci: Add more post-deploy Linux install tests

nr-ahemsath:
This PR updates the post-deploy APT/YUM repo install tests to:

1. Test Debian 13 (latest) and 11 (oldest-ish), using new repo signing key configuration logic that works equally well on both (see #3240 for context)
2. Use the new repo signing key config command for the `ubuntu-latest` case
3. Tweak the commands to get the version of the agent that was installed to output just the version, simplifying the version check logic
4. Made the names of the various validation jobs unique and descriptive


chynesNR (approved):
Very nice!

PR #3253, opened 2025-09-18
chore: Add preliminary OTel Bridge support for RPC server and client activities

tippmar-nr:
Adds initial OpenTelemetry Bridge support for RPC server and client activities, based largely on our otel tracing spec. 

Client implementation feels complete and produces the expected spans and attributes. Server, however, is a basic implementation that is largely untested due to the limited support for gRPC OTEL instrumentation in ASP.NET Core. 

As usual, this PR doesn't alter the default activity sources we listen to or enable the OTel Bridge feature, so there will be zero impact on existing customers.

Includes a set of cross-agent tests that exercise both client and server RPC activities with simulated spans.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridgeSegmentHelpers.cs
+++ b/src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridgeSegmentHelpers.cs
@@ -43,71 +43,63 @@ public static class ActivityBridgeSegmentHelpers
             return;
         }
 
-        // process tags only if the activity status is not Error
-        if ((int)activity.Status != (int)ActivityStatusCode.Error)
+        // based on activity kind, create the appropriate segment data
+        // remove all tags that are used in segment data creation
+        switch (activityKind)
         {
-            // based on activity kind, create the appropriate segment data
-            // remove all tags that are used in segment data creation
-            switch (activityKind)
-            {
-                case (int)ActivityKind.Client:
-                    // could be an http call or a database call, so need to look for specific tags to decide
-                    // order is important because some activities have both tags, e.g. a database call that is also an HTTP call, like Elasticsearch
-                    if (tags.TryGetAndRemoveTag<string>(["db.system.name", "db.system"], out var dbSystemName)) // it's a database call
-                    {
-                        ProcessClientDatabaseTags(segment, agent, activity, activityLogPrefix, tags, dbSystemName);
-                    }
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridgeSegmentHelpers.cs
```diff
@@ -18,520 +18,649 @@
...
+        string activityId = activity.Id;
+        string displayName = activity.DisplayName;
+
+        string activityLogPrefix = ActivityLogPrefixHelpers.ActivityLogPrefix(activityId, activityKind, displayName);
+
+        Log.Debug($"{activityLogPrefix} has stopped.");
 
-        public static void ProcessActivityTags(this ISegment segment, object originalActivity, IAgent agent)
+        // check status - stop processing if status is Error
+        if ((int)activity.Status == (int)ActivityStatusCode.Error)
```
nrcventura: The customer may still want the other tags added even if the span has an error.

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridgeSegmentHelpers.cs
```diff
@@ -18,38 +18,34 @@
...
-            }
+        var tags = ((IEnumerable<KeyValuePair<string, object>>)activity.TagObjects).ToDictionary(t => t.Key, t => t.Value);
+        if (tags.Count == 0)
+        {
+            Log.Finest($"{activityLogPrefix} has no tags. Not adding tags to segment.");
+            return;
+        }
 
+        // process tags only if the activity status is not Error
+        if ((int)activity.Status != (int)ActivityStatusCode.Error)
```
nrcventura: Why can't we run this code if the span is flagged as an error? I believe most instrumentation leaves the "error" determination to their users, and the instrumentation will set the other attributes as long as the data is available. The user of the library (that is instrumented) will then make a determination if a span should have its status set to "error", but when that happens they do not remove attributes. If an attribute is missing its related more to what type of problem occurred. If this was an external call, the error could have been that the status code returned was an error status code, and in that case I would still expect all of the expected attributes to be present.

tippmar-nr: I don't remember now why I added that, but it was to deal with some situation I observed in my test app. I can take it out but I'm not sure we should.

PR #3254, opened 2025-09-18
chore: Add supportability metric when accept DT headers is called after a segment exists.

tippmar-nr:


src/Agent/NewRelic/Agent/Core/AgentHealth/AgentHealthReporter.cs
```diff
@@ -391,6 +391,11 @@ public void CollectDistributedTraceSuccessMetrics()
             }
         }
 
+        public void ReportSupportabilityDistributedTraceHeadersAcceptedLate()
+        {
+            TrySend(_metricBuilder.TryBuildDistributedTraceHeadersAcceptedLate);
```
nrcventura: This method should probably just increment a counter, and then a separate Collect method should be added and called at harvest time to build the metric if the count is > 0. Doing that will keep the behavior consistent with the other distributed tracing metrics.

tippmar-nr: There's a mix of `TrySend()` and `_xxxCounter.Increment()` used throughout the DT related metrics (see lines 337-393) and I can't see any pattern to make it clear why one or the other is used.

nrcventura: It's an optimization thing, and it does look like it was not applied consistently. The idea is that metrics, that represent just a count of something, that need to be "aggregated" multiple times in a minute should call a method that just increments a counter. The `TrySend` call does a lot more work and allocations, because it creates a full timeslice metric with all of its values (that are not really used for simple counts), and results in a full metric aggregation process.

Because of how our custom instrumentation works where you cannot separate the creation of a transaction and the creation of the first segment, customers will not be able to use the DT APIs to accept the DT headers between when a transaction is started and the creation of the first segment, which may make this metric get incremented frequently enough that it should go through the "more optimized" path (just incrementing a integer).

PR #3257, opened 2025-09-23
fix: Kafka instrumentation doesn't play nice with custom instrumentation

### LINKED ISSUE #2829: Improve Kafka instrumentation interaction with custom instrumentation
The .net agent did not always support automatic instrumentation of Kafka. Many people previously relied on custom instrumentation of services to provide visibility into their kafka consumers. After adding support for kafka instrumentation, there have been scenarios where the kafka instrumentation can end the transaction before the custom instrumentation expected the transaction to end.

This happens because the Kafka instrumentation expected to create the transaction, and when it's call is done, it will end the transaction. 

There are two possible ways to handle this problem.

1. The agent instrumentation can check to see if a transaction already exists and only end the transaction if one didn't previously exist.
2. Try to leverage the unit of work parameter in the CreateTransaction method (this might need to be tied to a transaction existence check though).

The kafka instrumentation should not ignore a transaction that was not created by this instrumentation.

### Acceptance Criteria

- Custom transactions created before our kafka consumption instrumented method runs do not end until the custom instrumentation says it should.
- Custom transactions created before our kafka consumption instrumented method runs should not get marked as ignored if no messages were returned by kafka.

### Dependencies

None

### Estimates

Initial t-shirt size: M

tippmar-nr:
BEGIN_COMMIT_OVERRIDE
fix: Resolve issues with Kafka "Consume" instrumentation to ensure that automatic instrumentation works in conjunction with custom instrumentation. (#3257)
END_COMMIT_OVERRIDE

This PR changes the Kafka "Consume" instrumentation to ensure that automatic instrumentation works in conjunction with any custom instrumentation that might have been created. 

In particular, this change eliminates the automatic creation of transactions when the `Consume()` API is invoked, as those transactions wouldn't include any customer code and thus wouldn't provide any useful insights. We will now only instrument the `Consume()` API call when it is invoked from within a transaction (i.e., via custom instrumentation xml or the `[Transaction]` attribute in code). 

There were some additional issues with the specific overload of the `Consume()` API that we were instrumenting; we have adjusted that instrumentation to ensure that the segments we create are more meaningful.

This PR fixes #2829. 

Includes a significantly refactored Kafka integration test that includes a more realistic implementation of a Kafka consumer.


src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -308,7 +308,7 @@ public AbstractSegmentData CreateCustomSegmentData(string segmentName)
         public ISegment StartMessageBrokerSegment(MethodCall methodCall, MessageBrokerDestinationType destinationType,
             MessageBrokerAction operation, string brokerVendorName, string destinationName,
             string messagingSystemName = null, string cloudAccountId = null, string cloudRegion = null,
-            string serverAddress = null, int? serverPort = null, string routingKey = null)
+            string serverAddress = null, int? serverPort = null, string routingKey = null, bool isLeaf = false)
```
nrcventura: Instead of needing to plumb this parameter into the both Transaction.cs and NoOpTransaction.cs you can use the [MakeLeaf method on ISegmentExperimental](https://github.com/newrelic/newrelic-dotnet-agent/blob/d138517e9ab6d287d7f9f31badbd07de63583363/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Api/Experimental/ISegmentExperimental.cs#L31) from the instrumentation.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Kafka/KafkaConsumerWrapper.cs
```diff
@@ -37,20 +38,20 @@ public CanWrapResponse CanWrap(InstrumentedMethodInfo methodInfo)
...
-                destinationType: MessageBrokerDestinationType.Topic,
-                brokerVendorName: BrokerVendorName,
-                destination: "unknown"); // placeholder since the topic name is unknown at this point
+            // if the overload that takes a CancellationToken was used, then we want to make this a leaf segment
+            // because that overload calls the overload that takes an int (timeout) many times in a loop until the token is cancelled or a message is received
+            // and we don't want to create a segment for each of those calls.
+            var isCancellationTokenOverload = instrumentedMethodCall.MethodCall.MethodArguments[0] is CancellationToken;
 
-            var segment = transaction.StartMessageBrokerSegment(instrumentedMethodCall.MethodCall, MessageBrokerDestinationType.Topic, MessageBrokerAction.Consume, BrokerVendorName, "unknown");
+            var segment = transaction.StartMessageBrokerSegment(instrumentedMethodCall.MethodCall, MessageBrokerDestinationType.Topic, MessageBrokerAction.Consume, BrokerVendorName, "unknown", isLeaf: isCancellationTokenOverload);
```
nrcventura: Instead of passing an `isLeaf` parameter here, which would give that method even more responsibilities, we can simply do something like the following on the returned segment `((ISegmentExperimental)segment).MakeLeaf()` and just do that conditionally.

tests/Agent/IntegrationTests/dockerignore ContainerApplications/.dockerignore
nrcventura, line 1: Is this file supposed to exist? The path seems odd to me.

tests/Agent/UnitTests/NewRelic.Agent.Extensions.Tests/Helpers/SqsHelperTests.cs
nrcventura, line 1: These changes may not be needed, if the `isLeaf` parameter is removed.

PR #3263, opened 2025-09-29
fix: MSSQL connection string parsing can throw exceptions and disable Datastore instrumentation

### LINKED ISSUE #3255: Failure to parse connectionstring information should not cause Datastore instrumentation to be disabled
**Description**

#2948 resulted in the removal of some try-catch blocks in the connectionstring parsing code for the datastore instrumentation. This try-catch was apparently a safety net for the different types of connection strings that we encounter that we are unable to parse successfully. We need to update our datastore instrumentation so that failure to parse a connectionstring does not result in our datastore instrumentation raising an exception and causing the instrumentation to get disabled.

**Expected Behavior**

The datastore instrumentation should not get disabled and we should still see datastore segments being created.

The connection info for the datastore calls should either be reported as unknown, or left missing, whichever was the behavior prior to #2948.

**Steps to Reproduce**

Use a valid connection that our agent does not currently know how to parse.

*Suggested Priority (P1,P2,P3,P4,P5):* P4

tippmar-nr:
Refactors MSSQL (and MySql and ODBC) connection string parsing to be more resilient and ensure that exceptions don't leak.

Includes unit test updates.

Fixes #3255


chynesNR (dismissed):

tests/Agent/UnitTests/NewRelic.Agent.Extensions.Tests/Parsing/ConnectionStringParserTests.cs
```diff
@@ -99,13 +116,116 @@ public void ConnectionStringParser_DifferentConnectionStrings_NotMatched()
...
+                });
 
             Assert.That(connectionInfo.Host, Is.EqualTo("unknown"));
             Assert.That(connectionInfo.PortPathOrId, Is.EqualTo("unknown"));
             Assert.That(connectionInfo.DatabaseName, Is.EqualTo("unknown"));
             Assert.That(connectionInfo.InstanceName, Is.Null);
         }
+
+        [Test]
+        public void MsSql_MalformedConnectionStrings_DoNotThrow()
```
nrcventura: Should we add tests like this for the other datastore libraries that we added the try catch blocks to?

PR #3284, opened 2025-10-20
chore: Add entity guid to health check file per agent control spec.

tippmar-nr:
Adds entity guid at the beginning of the health check file as specified by the agent control spec. Includes updated tests.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentHealth/HealthCheck.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentHealth/HealthCheck.cs
@@ -49,12 +49,12 @@ namespace NewRelic.Agent.Core.AgentHealth
             }
         }
 
-        public string ToYaml()
+        public string ToYaml(string entityGuid)
         {
             lock (this)
             {
                 return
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Core/AgentHealth/HealthCheck.cs
```diff
@@ -49,12 +49,12 @@ public void TrySetHealth((bool IsHealthy, string Code, string Status) healthStat
             }
         }
 
-        public string ToYaml()
+        public string ToYaml(string entityGuid)
```
nrcventura, line 52: All the other health check information is part of the health check object. It might be better to make the entity guid part of the health check object itself.

tippmar-nr: This was intentional, as the entityGuid is a constant value that comes from configuration and is only needed when writing the yaml output. I considered adding a static initializer to the HealthCheck object, but that was clunky. This implementation works fine, as Configuration is readily available where the .ToYaml() method is invoked.

PR #3287, opened 2025-10-23
Chore: NR-341641 - Determine correct log level mapping for the otel sdk log events

ashishsinghnr:
…event

[HA] FY25Q4 - Spike: Determine correct log level mapping for the otel sdk log events

## Description

# Author Checklist
- [ Yes] Unit tests, Integration tests, and Unbounded tests completed
- [ Yes] Performance testing completed with satisfactory results (if required)

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/Logging/OpenTelemetrySDKLogger.cs
```diff
@@ -47,57 +47,88 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
...
-                EventLevel.Error => LogLevel.Error,
-                EventLevel.Warning => LogLevel.Warn,
-                EventLevel.Informational => LogLevel.Info,
-                EventLevel.LogAlways => LogLevel.Info,
-                EventLevel.Verbose => LogLevel.Debug,
-                _ => LogLevel.Finest
+                EventLevel.Critical => LogLevel.Error,     // Critical errors → ERROR
+                EventLevel.Error => LogLevel.Error,       // Errors → ERROR  
+                EventLevel.Warning => LogLevel.Warn,         // Warnings → WARN
+                EventLevel.Informational => LogLevel.Info,   // Informational → INFO
```
nrcventura: Do you think that EventLevel.Informational is too verbose for our agent to log at the info level which is the default log level for the agent?

ashishsinghnr: It depends, 
As "These messages provide real-time confirmation that the OTel bridge is functioning. Without them, user have no visibility into whether their OTel integration is working",  This mapping can provide valuable operational visibility.

If, We make something like 
EventLevel.Warning at INFO, in that case "One warning, but no context. Is OTel working? Are spans exporting?" So we have enable Debug for further understanding.

If we need to less logging at Info, then we can mark like
 EventLevel.LogAlways => LogLevel.Info,
 EventLevel.Informational => LogLevel.Debug
 For this, user needs to explicitly enable to Debug, for to understand operational visibility as well.
 
 So, it would be good if we keep 
 EventLevel.Informational => LogLevel.Info. and further exclude patterns something like below, to reduce the noise of logging ?
```
<exclude pattern="Microsoft.*" />
<exclude pattern="*.RequestStarted" />''  etc
```

nrcventura: Along those lines, is EventLevel.Informational required to give us some sort of heartbeat letting us know that things are successful? Or is EventLevel.LogAlways sufficient to give us that heartbeat information? 
Our agent at info level lets us know that we've successfully connected, but it does not log information every time we successfully harvest data, and instead writes some heartbeat information every so often to indicate that harvests have happened successfully.

ashishsinghnr: changed mapping from Info to Debug for less noise for logging

nrcventura:
Can we roll back the audit logging changes from this PR? I prefer to have a separate PR with those changes.

PR #3293, opened 2025-10-29
chore: NR-341644 - Determine how to add OTLP metrics data invocations to the audit logs

ashishsinghnr:

## Description

# Author Checklist
- [ Yes] Unit tests, Integration tests, and Unbounded tests completed
- [ Yes] Performance testing completed with satisfactory results (if required)

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
@@ -12,7 +12,7 @@ namespace NewRelic.Agent.Core.DataTransport
 {
     /// <summary>
     /// DelegatingHandler that provides audit logging for OTLP (OpenTelemetry Protocol) HTTP communications.
-    /// Captures requests and responses for compliance auditing while handling binary protobuf content appropriately.
+    /// Captures requests and responses for compliance auditing while handling binary protobuf content.
     /// </summary>
     public class OtlpAuditHandler : DelegatingHandler
     {
@@ -34,8 +34,12 @@ namespace NewRelic.Agent.Core.DataTransport
             }
             catch (Exception ex)
             {
-                // Audit log failed requests for compliance
-                LogOtlpException(request, ex);
+                // Log HTTP-level error at finest level for debugging
+                Log.Finest(ex, "OTLP metrics export failed: {ExceptionType} - {Message}. Endpoint: {RequestUri}, Method: {Method}", 
+                    ex.GetType().Name, ex.Message, request.RequestUri, request.Method);
+                
+                // Exception will propagate to OpenTelemetry SDK, which will log its own error
+                // via EventSource that our OpenTelemetrySDKLogger captures
                 throw;
             }
         }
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/DataTransport/Client/DataTransportAuditLogger.cs
```diff
@@ -27,7 +27,8 @@ public enum AuditLogSource
         {
             Collector = 1,
             Beacon = 2,
-            InstrumentedApp = 3
+            InstrumentedApp = 3,
+            OtlpExporter = 4
```
nrcventura: I think that we should reuse the `InstrumentedApp` and `Collector` options from this enum. Looking at the existing audit logs, we use `InstrumentedApp` for all data being sent from our agent, and `Collector` for all data received by our agent. This seems to make sense for the OTLP data too because the data is still coming from the instrumented app, and being sent to a New Relic "collector", so the response is still coming back from a New Relic "collector".

src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                // Never let audit logging break OTLP export
+                Log.Debug(ex, "Failed to audit log OTLP response");
+            }
+        }
+
+        private static void LogOtlpException(HttpRequestMessage request, Exception exception)
```
nrcventura: I don't think that we need to write these to the audit logs.

src/Agent/NewRelic/Agent/Core/DataTransport/Client/DataTransportAuditLogger.cs
ashishsinghnr: ok sure, understood, so there is no differentiation if we make it for OTEL or  not. It will be coming for instrumented app only.

src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
ashishsinghnr: It's just for safety for debugging purpose incase anything, although I think it will rarely produce logs for this. Can we make this finest level OR completely remove?

src/Agent/NewRelic/Agent/Core/DataTransport/Client/DataTransportAuditLogger.cs
ashishsinghnr: done

src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
nrcventura: Yes, we can just log the error using our normal logger (not the audit logger) and log it at the finest level. When an error occurs, I also expect the otel sdk logger to log an error too.

ashishsinghnr: done

PR #3296, opened 2025-10-31
fix: Update ReportedConfiguration to serialize certain enums as strings instead of integers for better clarity when viewing settings in the UI

tippmar-nr:
Certain enumerations used in `ReportedConfiguration` are serializing as integers in the connect payload when they should be string values instead. Our existing unit tests were expecting the string values and were passing, apparently because we didn't use the same serializer in the unit tests that we use when actually serializing payloads being sent to NR. 

This fix adds some extension methods for the affected enums and slightly refactors `ReportedConfiguration` to ensure the values are reported as strings.

Tests are updated as required.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiveWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiveWrapper.cs
@@ -17,10 +17,7 @@ public class AzureServiceBusReceiveWrapper : AzureServiceBusWrapperBase
 {
     private static readonly ConcurrentDictionary<Type, Func<object, object>> _getResultFromGenericTask = new();
 
-    private static Func<object, object> _innerReceiverAccessor;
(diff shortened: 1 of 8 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/ReportedConfiguration.cs
```diff
@@ -210,10 +211,18 @@ public ReportedConfiguration(IConfiguration configuration)
         public bool DistributedTracingEnabled => _configuration.DistributedTracingEnabled;
 
         [JsonProperty("distributed_tracing.sampler.root.sampler_type")]
+        public string RootSamplerName => RootSamplerType.ToSamplerTypeString();
+        [JsonIgnore]
         public SamplerType RootSamplerType => _configuration.RootSamplerType;
+
         [JsonProperty("distributed_tracing.sampler.remote_parent_sampled.sampler_type")]
+        public string RemoteParentSampledSamplerName => RemoteParentSampledSamplerType.ToSamplerTypeString();
```
tippmar-nr, line 219: Sorry for the noise in this file; this is where the relevant changes start. Here to line 225

chynesNR (approved):

PR #3301, opened 2025-11-11
ci: Update .NET quality from 'preview' to 'ga' across multiple workflows

tippmar-nr:
Installs .NET 10 GA version in all workflows


src/Agent/NewRelic/Profiler/NewRelic.Profiler.sln.DotSettings
nr-ahemsath, line 1: Did this sneak in accidentally, or is it related to .NET 10 somehow that I can't see?

tippmar-nr: No idea where the change to this file came from; it's a ReSharper-specific file so won't do anything if you're not using that tool. 

Since the file was already here and is just being appended to, I don't think there's any harm in letting this change go through, even though it's really not related to the rest of this PR. Or I can revert the change, either way.

nr-ahemsath: I'll let it slide this time 😄

PR #3305, opened 2025-11-12
fix: Azure Service Bus instrumentation fails due to null reference exception

### LINKED ISSUE #3303: Azure Service Bus instrumentation encountering a null reference error

**Description**
Beginning with the Azure Service bus instrumentation updates in 10.46.1 we are getting a null reference error in the instrumentation for ProcessOneMessage.

```
2025-11-12 11:35:21,007 NewRelic  ERROR: [pid: 1, tid: 8] Tracer invocation error
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: Cannot perform runtime binding on a null reference
   at CallSite.Target(Closure, CallSite, Object)
   at System.Dynamic.UpdateDelegates.UpdateAndExecute1[T0,TRet](CallSite site, T0 arg0)
   at NewRelic.Providers.Wrapper.AzureServiceBus.AzureServiceBusReceiverManagerWrapper.BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
   at NewRelic.Agent.Core.Wrapper.WrapperService.BeforeWrappedMethod(Type type, String methodName, String argumentSignature, Object invocationTarget, Object[] methodArguments, String tracerFactoryName, String metri
cName, UInt32 tracerArguments, UInt64 functionId)
   at NewRelic.Agent.Core.AgentManager.GetTracerImpl(String tracerFactoryName, UInt32 tracerArguments, String metricName, String assemblyName, Type type, String typeName, String methodName, String argumentSignature
, Object invocationTarget, Object[] arguments, UInt64 functionId)
```

This is most likely caused by trying to use the dynamic keyword on a type with non-public visibility, or a generic type with a non-public type parameter.

When this happens the azure service bus instrumentation gets disabled, and the customer's application should continue to function normally. However, they will not get any telemetry related to azure service bus usage.

**Expected Behavior**

The instrumentation should be able to function for both public and non-public types and handle null values appropriately.

**Steps to Reproduce**

TBD

**Additional context**

*Suggested Priority (P1,P2,P3,P4,P5):* P3
*Suggested T-Shirt size (S, M, L, XL, Unknown):* M

tippmar-nr:
Addresses a potential null reference exception that could occur in our instrumentation of `ServiceBusProcessor` by eliminating the use of `dynamic` and adding a null check.

Fixes #3303


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiverManagerWrapper.cs
```diff
@@ -25,11 +29,26 @@ public override CanWrapResponse CanWrap(InstrumentedMethodInfo instrumentedMetho
         public override AfterWrappedMethodDelegate BeforeWrappedMethod(InstrumentedMethodCall instrumentedMethodCall, IAgent agent, ITransaction transaction)
         {
             var receiverManager = instrumentedMethodCall.MethodCall.InvocationTarget;
-            _receiverAccessor ??= VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(receiverManager.GetType(), "Receiver");
-            dynamic receiver = _receiverAccessor(receiverManager);
 
-            string queueOrTopicName = receiver.EntityPath; // some-queue|topic-name
-            string fqns = receiver.FullyQualifiedNamespace; // some-service-bus-entity.servicebus.windows.net
+            var receiverManagerType = receiverManager.GetType();
+            _receiverAccessor = VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(receiverManagerType, "Receiver");
```
nrcventura: I don't think that we want to generate this on each invocation, and should probably cache this work.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiverManagerWrapper.cs
```diff
@@ -25,11 +29,26 @@ public override CanWrapResponse CanWrap(InstrumentedMethodInfo instrumentedMetho
...
+                    agent.Logger.Warn("AzureServiceBusReceiverManagerWrapper: Unable to access Receiver property on ReceiverManager instance of type {ReceiverManagerType}. Unable to instrument Azure ServiceBus Processor Mode.", receiverManagerType);
+                    _badReceiverWarningLogged = true;
+                }
+                return Delegates.NoOp;
+            }
+
+            _entityPathAccessor = _entityPathAccessor ?? VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(receiver.GetType(), "EntityPath");
+            _fullyQualifiedNamespaceAccessor = _fullyQualifiedNamespaceAccessor ?? VisibilityBypasser.Instance.GeneratePropertyAccessor<string>(receiver.GetType(), "FullyQualifiedNamespace");
+
+            string queueOrTopicName = _entityPathAccessor(receiver)?.ToString() ?? "unknown"; // some-queue|topic-name
```
nrcventura: It looks like we don't need a default value for this string, because the code previously assumed that the value could be null.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiverManagerWrapper.cs
tippmar-nr: Good catch, wasn't supposed to be written like that... ;-)

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiverManagerWrapper.cs
```diff
@@ -25,11 +29,26 @@ public override CanWrapResponse CanWrap(InstrumentedMethodInfo instrumentedMetho
...
+            string fqns = _fullyQualifiedNamespaceAccessor(receiver)?.ToString() ?? "unknown"; // some-service-bus-entity.servicebus.windows.net
```
nrcventura: It looks like the code that uses this is also able to handle values so we don't need to provide a default value here.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/AzureServiceBus/AzureServiceBusReceiverManagerWrapper.cs
```diff
@@ -25,11 +29,26 @@ public override CanWrapResponse CanWrap(InstrumentedMethodInfo instrumentedMetho
...
+            _receiverAccessor = VisibilityBypasser.Instance.GeneratePropertyAccessor<object>(receiverManagerType, "Receiver");
+            object receiver = _receiverAccessor(receiverManager);
+            if (receiver == null)
+            {
+                if (!_badReceiverWarningLogged)
+                {
+                    agent.Logger.Warn("AzureServiceBusReceiverManagerWrapper: Unable to access Receiver property on ReceiverManager instance of type {ReceiverManagerType}. Unable to instrument Azure ServiceBus Processor Mode.", receiverManagerType);
+                    _badReceiverWarningLogged = true;
+                }
+                return Delegates.NoOp;
```
nrcventura: Maybe we don't need to NoOp here, and can just let queueOTopicName and fqns be null.

PR #3322, opened 2025-11-19
fix: Cache all environment variable access to reduce possible thread contention

tippmar-nr:
A recently reported customer issue indicated significant thread contention that was eventually traced back to Agent code that reads environment variables in hot code paths. 

This PR introduces global caching for the environment variables referenced by the agent which should improve application performance and reduce thread contention.

Note that `GetEnvironmentVariable(string variable, EnvironmentVariableTarget environmentVariableTarget)` was removed because the only usages of that overload specified `EnvironmentVariableTarget.Process` -- which is the default value when calling `System.Environment.GetEnvironmentVariable(string variable)`, which made the method unnecessary and presented complexity in deriving a cache key.

Resolves: #3323

### DIFF
--- a/tests/Agent/UnitTests/CompositeTests/EnvironmentTests.cs
+++ b/tests/Agent/UnitTests/CompositeTests/EnvironmentTests.cs
@@ -15,9 +15,9 @@ namespace CompositeTests
         {
             System.Environment.SetEnvironmentVariable("NEWRELIC_HOME", null);
             System.Environment.SetEnvironmentVariable("NEW_RELIC_HOME", null);
+            NewRelic.Agent.Core.SharedInterfaces.Environment.ResetCache();
 
             _env = new NewRelic.Agent.Core.SharedInterfaces.Environment();
-            NewRelic.Agent.Core.SharedInterfaces.Environment.ResetCache(); // clear cache before each test
         }
 
         [TearDown]
@@ -25,6 +25,7 @@ namespace CompositeTests
         {
             System.Environment.SetEnvironmentVariable("NEWRELIC_HOME", null);
             System.Environment.SetEnvironmentVariable("NEW_RELIC_HOME", null);
+            NewRelic.Agent.Core.SharedInterfaces.Environment.ResetCache();
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/SharedInterfaces/Environment.cs
```diff
@@ -3,58 +3,94 @@
...
+    /// <exception cref="ArgumentNullException">Thrown if <paramref name="variable"/> is null.</exception>
+    public string GetEnvironmentVariable(string variable)
     {
-        public string[] GetCommandLineArgs()
-        {
-            return System.Environment.GetCommandLineArgs();
-        }
+        if (variable == null)
+            throw new ArgumentNullException(nameof(variable));
+        return _environmentVariableCache.GetOrAdd(variable, _ => System.Environment.GetEnvironmentVariable(variable));
```
nrcventura: You might be able to avoid using a lambda here if did something more like the following.
```suggestion
        return _environmentVariableCache.GetOrAdd(variable, System.Environment.GetEnvironmentVariable);
```
I don't know if the compiler will correctly disambiguate which overload to use though.

src/Agent/NewRelic/Agent/Core/AgentManager.cs
nrcventura, line 1: The changes in this file may not be necessary, because this code only during startup, and will cache values separately from the rest of the environment variable access in the agent.

tippmar-nr: `Environment` uses a static cache, so referencing a instance of the class here will pre-populate the cache for use everywhere else in the application. And this change helps ensure we're consistently using the environment abstraction everywhere in code that it possibly can be used.

tippmar-nr: ... Or, I should say, it's going to use a static cache shortly... ;-)  I thought I already implemented it that way but guess not...

tests/Agent/UnitTests/CompositeTests/EnvironmentTests.cs
nrcventura, line 27: We might want to call ResetCache in the TearDown method too. Especially if some other test runs after this one that deals with environment variables.

tests/Agent/UnitTests/Core.UnitTest/Utilization/VendorInfoTests.cs
```diff
@@ -45,6 +45,8 @@ public void Setup()
             _environment = Mock.Create<IEnvironment>();
             _vendorHttpApiRequestor = Mock.Create<VendorHttpApiRequestor>();
             _fileWrapper = Mock.Create<IFileWrapper>();
+
+            SharedInterfaces.Environment.ResetCache(); // clear cache before each test
```
nrcventura: Do we need to add a TearDown method and call ResetCache to ensure that things get restored to the previous state?

tippmar-nr: Actually, the cache reset here isn't necessary at all -- I failed to notice previously that these tests are using a mock `IEnvironment`

PR #3325, opened 2025-11-20
chore: Add instrumentation scope attributes to hybrid agent spans

tippmar-nr:
Adds 
`otel.scope.name`
`otel.library.name`
`otel.scope.version`
`otel.library.version`

as Agent attributes to spans created by the hybrid agent, per the otel tracing spec.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinition.cs
+++ b/src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinition.cs
@@ -110,7 +110,8 @@ namespace NewRelic.Agent.Core.Attributes
             return builder;
         }
 
-        private static object GenericConverter(object input)
+        // public to allow for testing
+        public static object GenericConverter(object input)
         {
             switch (input)
             {
--- a/tests/Agent/UnitTests/Core.UnitTest/Attributes/GenericConverterTests.cs
+++ b/tests/Agent/UnitTests/Core.UnitTest/Attributes/GenericConverterTests.cs
@@ -1,0 +1,126 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using NUnit.Framework;
+using System;
+
+namespace NewRelic.Agent.Core.Attributes.Tests
+{
+    [TestFixture]
+    public class GenericConverterTests
+    {
+        [Test]
+        public void ReturnsNull_ForNullInput()
+        {
+            var result = AttributeDefinitionBuilder.GenericConverter(null);
+            Assert.That(result, Is.Null);
+        }
+
+        [Test]
+        public void ConvertsTimeSpan_ToTotalSeconds()
+        {
+            var ts = TimeSpan.FromMilliseconds(12345);
+            var result = AttributeDefinitionBuilder.GenericConverter(ts);
+            Assert.That(result, Is.EqualTo(ts.TotalSeconds));
+            Assert.That(result, Is.TypeOf<double>());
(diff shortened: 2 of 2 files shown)


src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridgeSegmentHelpers.cs
```diff
@@ -111,6 +113,29 @@ public static void ProcessActivityTags(this ISegment segment, object originalAct
         }
     }
 
+    private static void GetInstrumentationScopeAttributes(ISegment segment, dynamic activity)
+    {
+        dynamic activitySource = activity.Source;
+        if (activitySource != null)
+        {
+            string scopeName = activitySource.Name;
+            if (scopeName != "NewRelic.Agent") // don't add New Relic's own activity source as a scope
```
nrcventura: Should we make this a constant and reference it where we create the ActivitySource and here?

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridgeSegmentHelpers.cs
```diff
@@ -111,6 +113,29 @@ public static void ProcessActivityTags(this ISegment segment, object originalAct
...
+                segment.AddCustomAttribute("otel.scope.name", scopeName);
+                // include the deprecated instrumentation library tags for backward compatibility
+                segment.AddCustomAttribute("otel.instrumentation_library.name", scopeName);
+
+                string scopeVersion = activitySource.Version;
+                if (!string.IsNullOrEmpty(scopeVersion))
+                {
+                    segment.AddCustomAttribute("otel.scope.version", scopeVersion);
+                    // include the deprecated instrumentation library tags for backward compatibility
+                    segment.AddCustomAttribute("otel.instrumentation_library.version", scopeVersion);
```
nrcventura: We should probably add these as `agent` attributes instead of `custom` attributes.

tippmar-nr: The spec doesn't say one way or the other. Maybe it should?

tippmar-nr: We also don't have a way via the ISegment API (or ISegmentExperimental) to add an arbitrary agent attribute. That would take a lot of plumbing for a one-off attribute.

nrcventura: That's a good idea. I'll get that clarified.

nrcventura: I think that adding something similar to the custom attribute method to the experimental api is worthwhile. This is something that we also ran into for the FAAS attributes.

tippmar-nr: I was able to implement the change without needing to expose anything in our public API.

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridgeSegmentHelpers.cs
tippmar-nr: Done.

PR #3336, opened 2025-12-02
chore: Update PR title checker regex to recognize more valid prefixes

tippmar-nr:
Converted to using regex instead of a static list of prefixes, primarily to allow things like `chore(hybrid-agent): blah blah blah` to be acceptable; Release Please already knows how to handle commit messages in that format.

### DIFF
--- a/.github/workflows/pr-title-checker-config.json
+++ b/.github/workflows/pr-title-checker-config.json
@@ -3,7 +3,7 @@
       "name": ""
     },
     "CHECKS": {
-      "regexp": "^(fix|feat|security|notice|chore|ci|docs|test|summary)(\\([^)]+\\))?: "
+      "regexp": "^(fix|feat|security|notice|chore|ci|docs|test|summary)(\([^)]+\))?: "
     },
     "MESSAGES": {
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-title-checker-config.json
```diff
@@ -3,7 +3,7 @@
       "name": ""
     },
     "CHECKS": {
-      "prefixes": ["fix: ", "feat: ", "security: ", "notice: ", "chore: ", "chore(deps): ", "chore(main): ", "ci: ", "docs: ", "test: ", "summary: "]
+      "regexp": "^(fix|feat|security|notice|chore|ci|docs|test|summary)(\\([^)]+\\))?: "
```
nr-ahemsath: I think what you actually want is:

```suggestion
      "regexp": "^(fix|feat|security|notice|chore|ci|docs|test|summary)(\([^)]+\))?: "
```

The difference being single instead of double backslashes for escaping the literal parentheses - a double backslash means a literal backslash.

FYI, I used regex101.com for testing a variety of titles with all the available flavors of regex.

PR #3345, opened 2025-12-08
[HA Metrics] Determine how to handle conflicts with the EventPipe

ashishsinghnr:
Determine how to handle conflicts with the EventPipe

## Description

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Config/BootstrapConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Config/BootstrapConfiguration.cs
@@ -51,6 +51,7 @@ namespace NewRelic.Agent.Core.Config
         private BootstrapConfiguration()
         {
             _agentEnabledWithProvenance = new ValueWithProvenance<bool>(true, "Default value");
+            GCSamplerV2Enabled = true;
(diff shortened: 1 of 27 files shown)


tippmar-nr:
It looks like this PR changes the defaults for `GCV2Samplers` and `EventListenerSamplers`, which I think might be more of a change than we really want. 

@nrcventura your thoughts?

nrcventura:
> It looks like this PR changes the defaults for `GCV2Samplers` and `EventListenerSamplers`, which I think might be more of a change than we really want.
> 
> @nrcventura your thoughts?

Correct, we still need to get the necessary approvals for a change like this. One of the open questions that I have that can also influence the decision here is getting an estimate about what the increase in ingest will be for apps that enable the collection of OTel metrics. That will influence whether OTel metrics capture can be enabled by default, or if we need to have it off by default. If it's off by default that also allows customers to opt-into the necessary sampler changes.

I updated this PR to be a draft for now, while we have the necessary discussions.

ashishsinghnr:
Closing the PR for now, will take in later and then raise new PR  for the review

PR #3172, opened 2025-07-01
fix: Profiler doesn't handle wide-character files correctly

tippmar-nr:
Thank you for submitting a pull request.  Please review our [contributing guidelines](/CONTRIBUTING.md) and [code of conduct](https://opensource.newrelic.com/code-of-conduct/).

## Description

Updates the Profiler to correctly handle reading of files with wide-character paths and file content. Fixes #3158. Fixes #3159.

Includes unit tests for various combinations of paths and file content.

# Author Checklist
- [ ] Unit tests, Integration tests, and Unbounded tests completed
- [ ] Performance testing completed with satisfactory results (if required)

# Reviewer Checklist
- [ ] Perform code review
- [ ] Pull request was adequately tested (new/existing tests, performance tests)

PR #3347, opened 2025-12-08
fix: Profiler doesn't handle UTF-8 files with BOM correctly.

jaffinito:
## Description

There was a regression caused by another fix, #3172, for handling wide characters properly.  This swaps from `wchar_t` to `xchar_t`, which handles the differences better.

This was tested on ubuntu, fedora, alpine, and Windows with newrelic.config files and custom instrumentation files contain with a BOM and non-english characters (Japanese).

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


nrcventura:
Should we add a regression test for this?

nrcventura (approved):

PR #3352, opened 2025-12-12
fix: Handle disposed streams in browser script injection

tippmar-nr:
Updates to improve stability and resilience of browser injection and ensure exceptions do not bubble out to the customer application:
* Adds robust error handling for `ObjectDisposedException` in ASP.NET 6 plus browser script injection logic. 
* Wraps all stream and context operations in try-catch blocks to prevent unhandled exceptions if the underlying stream or HTTP context is disposed.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Agent.cs
+++ b/src/Agent/NewRelic/Agent/Core/Agent.cs
@@ -325,7 +325,14 @@ namespace NewRelic.Agent.Core
             if (rumBytes == null)
             {
                 transaction?.LogFinest("Skipping RUM Injection: No script was available.");
-                await baseStream.WriteAsync(buffer, 0, buffer.Length);
+                try
+                {
+                    await baseStream.WriteAsync(buffer, 0, buffer.Length);
+                }
+                catch (ObjectDisposedException)
+                {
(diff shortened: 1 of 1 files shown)


chynesNR:
Do we really only want to catch `ObjectDisposedException`? Couldn't there be others?

tippmar-nr:
> Do we really only want to catch `ObjectDisposedException`? Couldn't there be others?

There could be, but this is the one that's most likely a non-fatal error, since it's known to happen during the normal course of operations like when a client browser navigates away from a page while the request is still in flight. Other exceptions may or may not be something we did wrong in the agent since we're interacting with the customers middleware pipeline and I'd be worried about catching every exception that could get thrown.

chynesNR:
> > Do we really only want to catch `ObjectDisposedException`? Couldn't there be others?
> 
> There could be, but this is the one that's most likely a non-fatal error, since it's known to happen during the normal course of operations like when a client browser navigates away from a page while the request is still in flight. Other exceptions may or may not be something we did wrong in the agent since we're interacting with the customers middleware pipeline and I'd be worried about catching every exception that could get thrown.

Oh because we're writing out the whole buffer not just our piece, I get it now.

chynesNR (approved):

PR #3355, opened 2025-12-15
chore: Hybrid Agent Tracing Config Update

tippmar-nr:
* Implements formal configuration support for hybrid agent tracing, conforming to the hybrid agent spec. 

By default, all activity sources are included. However, a static set of activity sources that would duplicate existing .NET agent instrumentation are excluded. The `include` config can be used to explicitly include an activity source that is on that list. Values for `include` and `exclude` are a comma-separated string of activity source names.

Example config:
```
  <opentelemetry enabled="true">
    <!-- Tracing: enable and control which ActivitySources (Tracers) are included/excluded.
         Env vars: NEW_RELIC_OPENTELEMETRY_TRACES_ENABLED, NEW_RELIC_OPENTELEMETRY_TRACES_INCLUDE, NEW_RELIC_OPENTELEMETRY_TRACES_EXCLUDE -->
    <traces
      enabled="true"
      include="SomeDefaultExcludedActivitySource"
      exclude="NoisySource,LegacySource" />
  </opentelemetry>
```
* Adds a list of activity source names that are excluded by default
* Implements updated logic for determining whether a particular activity source name should be listened to or not
* Updates internal configuration property names for consistency
* Updates unit tests as required.

(Ignore the enormous commit history -- the PR branch was originally based on `feature/hybrid-agent-metrics` so still reflects all the changes from that branch even though it's been merged to `main` and this PR was update to be based from `main`.)

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
+++ b/src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
@@ -45,16 +45,17 @@ public class ActivityBridge : IDisposable
     {
         if (!_agent.Configuration.OpenTelemetryTracingEnabled)
         {
+            Log.Debug("Open Telemetry Tracing Bridge is disabled via configuration.");
             return true;
         }
 
         if (_activityListener != null)
         {
-            Log.Debug("Activity listener has already been created. Not starting a new one.");
+            Log.Debug("OpenTelemetry Tracing Bridge has already been created. Not starting a new one.");
             return false;
         }
 
-        Log.Debug("OpenTelemetry Bridge Tracing is enabled. Starting the activity listener.");
+        Log.Debug("OpenTelemetry Tracing Bridge is enabled. Starting the activity listener.");
         return TryCreateActivityListener();
     }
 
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Config/Configuration.cs
```diff
@@ -1312,7 +1310,7 @@ public virtual AlwaysOnSamplerType Clone()
     
     /// <summary>
     /// Based on the OpenTelemetry TraceIdRatioBased sampler.
-    /// It samples traces with a probability defined by the ratio attribute, which must be a decimal value between 0.0 (no sampling) and 1.0 (sample all).
+    /// It samples a percentage of traces defined by the ratio attribute, which must be a decimal value between 0.0 (no sampling) and 1.0 (sample all).
```
tippmar-nr:
```suggestion
    /// It samples traces with a probability defined by the ratio attribute, which must be a decimal value between 0.0 (no sampling) and 1.0 (sample all).
```

src/Agent/NewRelic/Agent/Core/Config/Configuration.cs
```diff
@@ -4287,7 +4285,7 @@ public enum ItemChoiceType
         @default,
         
         /// <remarks/>
-        traceIdRatioBased
+        traceIdRatioBased,
```
tippmar-nr:
```suggestion
        traceIdRatioBased
```

PR #3362, opened 2025-12-16
chore: Add span status attributes for hybrid agent

tippmar-nr:
Adds `status.code` and `status.description` Agent attributes to hybrid agent segments as per the tracing spec.

Updates hybrid agent smoke tests to validate the attributes in normal and error conditions.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
+++ b/src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
@@ -516,13 +516,12 @@ public class ActivityBridge : IDisposable
             dynamic activity = originalActivity;
             int activityStatusCode = (int)activity.Status;
             segment.AddAgentAttribute("status.code", activityStatusCode.ToActivityStatusCodeString());
-            if (activityStatusCode == (int)ActivityStatusCode.Error)
+
+            // description is currently only available when status is Error
+            string activityStatusDescription = (string)activity.StatusDescription;
+            if (!string.IsNullOrWhiteSpace(activityStatusDescription))
             {
-                string activityStatusDescription = (string)activity.StatusDescription;
-                if (!string.IsNullOrWhiteSpace(activityStatusDescription))
-                {
-                    segment.AddAgentAttribute("status.description", activityStatusDescription);
-                }
+                segment.AddAgentAttribute("status.description", activityStatusDescription);
             }
 
             segment.ProcessActivityTags(originalActivity, agent, errorService);
@@ -538,23 +537,4 @@ public class ActivityBridge : IDisposable
     }
     #endregion
 
-}
-
-internal static class ActivityStatusCodeExtensions
-{
-    public static string ToActivityStatusCodeString(this int statusCode)
-    {
-        if (statusCode < 0)
-        {
-            throw new ArgumentOutOfRangeException(nameof(statusCode));
-        }
-
-        return statusCode switch
-        {
-            0 => "unset",
-            1 => "ok",
-            2 => "error",
-            _ => "unknown"
-        };
-    }
 }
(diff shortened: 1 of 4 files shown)


src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
```diff
@@ -512,6 +512,19 @@ private static void ActivityStopped(object originalActivity, IAgent agent, IErro
...
+            dynamic activity = originalActivity;
+            int activityStatusCode = (int)activity.Status;
+            segment.AddAgentAttribute("status.code", activityStatusCode.ToActivityStatusCodeString());
+            if (activityStatusCode == (int)ActivityStatusCode.Error)
+            {
+                string activityStatusDescription = (string)activity.StatusDescription;
+                if (!string.IsNullOrWhiteSpace(activityStatusDescription))
+                {
+                    segment.AddAgentAttribute("status.description", activityStatusDescription);
+                }
```
nrcventura: We should probably do this for more than just the `error` status. That will allow the agent to automatically capture the description of the otel api ever supports defining a description for more than just the error status.

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
```diff
@@ -526,3 +539,22 @@ private static void ActivityStopped(object originalActivity, IAgent agent, IErro
     #endregion
 
 }
+
+internal static class ActivityStatusCodeExtensions
+{
+    public static string ToActivityStatusCodeString(this int statusCode)
+    {
+        if (statusCode < 0)
```
nrcventura: Should we consider casting from the `int` to the enum, and then using the EnumNameCache (lower case) to handle the conversion? That would allow more flexibility if the enum ever changes.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
```diff
@@ -268,6 +268,9 @@ public ISegment StartActivitySegment(MethodCall methodCall, INewRelicActivity ac
             var simpleSegmentData = new SimpleSegmentData(activity.DisplayName);
             segment.SetSegmentData(simpleSegmentData);
 
+            // Add an agent attribute for span status code, defaulting to "unset"
+            segment.AddAgentAttribute("status.code", "unset");
```
nrcventura: Do we need to do this here? We probably only need to set these attributes for the OTel spans when they end. We do not need to set these attributes for spans created from New Relic segments.

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
tippmar-nr: We've ran into trouble trying to use the enums in `System.Diagnostics.DiagnosticSource` directly because of differences between customer and ILRepack version.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
tippmar-nr: I'm just following the spec. It says to set a value when the span is created. There's no guarantee the status will be set to anything else when the activity stops.

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
tippmar-nr: I'm just doing what the spec says...

src/Agent/NewRelic/Agent/Core/OpenTelemetryBridge/ActivityBridge.cs
nrcventura: Right, you can't cast from one to the other. But you can cast an int to the ilrepacked version of the enum.

src/Agent/NewRelic/Agent/Core/Transactions/Transaction.cs
nrcventura: The comment in the spec defines how otel works. For our agent we don't have to set it here, because it is the default value that is automatically set when you start an otel span.

PR #3365, opened 2025-12-17
feat: Increase the maximum length of `db.statement` span attribute to 4096 bytes

### LINKED ISSUE #3364: Increase Size of db.statement Attribute for Span Events
This is in line with a recent Feature Request, and an [Agent spec update](https://source.datanerd.us/agents/agent-specs/pull/701) from last year.

tippmar-nr:
Bumps the maximum length of the `db.statement` attribute on spans from 1999 to 4096 bytes and updates database statement truncation logic to match. 

Adds unit and integration tests to verify correct functionality.

Resolves #3364

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -487,10 +487,6 @@ jobs:
       CI_NEW_RELIC_APP_NAME: ${{ github.event_name == 'schedule' && 'DotNetIngestTracking' || '' }}
 
     steps:
-      - name: My IP
-        run: (Invoke-WebRequest -uri "http://ifconfig.me/ip").Content
-        shell: powershell
-
       - name: Checkout
         uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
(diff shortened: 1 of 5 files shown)


nr-ahemsath (approved):
Looks good, tests are thorough 👍

src/Agent/NewRelic/Agent/Core/Attributes/AttributeDefinition.cs
```diff
@@ -82,10 +82,10 @@ public static AttributeDefinitionBuilder<int, int> CreateInt(string name, Attrib
 
         public static AttributeDefinitionBuilder<string, string> CreateDBStatement(string name, AttributeClassification classification)
         {
-            const int dbStmtMaxLength = 1999;
+            const int dbStmtMaxLengthBytes = 4096; // max byte length, per agent spec
```
nrcventura, line 85: The previous limit was 2000, but we used 1999 probably because of the null terminating character. We might need to make this 4095 to account for that.

nrcventura: Alternatively, if the long queries are visible in NRDB, then we know that using 4096 here is fine.

tippmar-nr: Verified that the data is properly showing in the UI with 4096 as the limit.

PR #3379, opened 2026-01-05
ci: Update version of MongoDB used in integration tests

### LINKED ISSUE #3375: Update MongoDB server used in unbounded integration tests
See [this Slack thread](https://newrelic.slack.com/archives/C027NFCDZ2P/p1766986262473019) for details.

nr-ahemsath:
Resolves #3375

### DIFF
--- a/tests/Agent/IntegrationTests/UnboundedServices/mongodb60/Dockerfile
+++ b/tests/Agent/IntegrationTests/UnboundedServices/mongodb60/Dockerfile
@@ -1,1 +1,1 @@
-FROM mongo:6.0.2
+FROM mongo:6.0.27


tests/Agent/IntegrationTests/UnboundedServices/mongodb60/Dockerfile
```diff
@@ -1 +1 @@
-FROM mongo:6.0.2
\ No newline at end of file
+FROM mongo:6.0.27
```
tippmar-nr, line 1: I wonder if we shouldn't just change this to `mongo:6` and let it pick up whatever the latest release is? That would save us from having to update again and still keep us tied to v6.

nrcventura: Do we need to do anything so that the containers get rebuilt nightly or something like that?

nr-ahemsath: I prefer not to have our test infrastructure changing without our being aware of it, even for something as likely-to-have-zero-impact as this.

tippmar-nr: > Do we need to do anything so that the containers get rebuilt nightly or something like that?

There's really no reason to rebuild them on any particular cadence. If we're going to stay pinned to a specific version, we'll just have to respond to issues as they come up. Or add a recurring work item in the backlog to update the container services on some regular basis (quarterly would probably be more than enough).

tippmar-nr (approved):

PR #3385, opened 2026-01-09
chore: Add Span Link and Span Event support to Hybrid Agent

tippmar-nr:
Adds basic support for Span Link and Span Event to the Hybrid Agent, in compliance with the relevant agent specs.

Updates the `<opentelemetry>` configuration element to `<openTelemetry>` for consistency.

Includes unit and composite tests.

Co-authored-by: @nrcventura

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTestHelpers/AgentLogBase.cs
@@ -371,6 +371,14 @@ namespace NewRelic.Agent.IntegrationTestHelpers
 
         #endregion
 
+        #region Span Links
+        public IEnumerable<SpanEvent> GetSpanLinks()
+        {
+            return GetSpanEvents()
+                .Where(@event => @event?.IntrinsicAttributes?["type"]?.ToString() == "SpanLink");
+        }
+        #endregion
+
         #region ErrorEvents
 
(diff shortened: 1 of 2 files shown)


src/Agent/NewRelic/Agent/Core/Segments/SpanEventWireModel.cs
```diff
@@ -12,167 +12,183 @@
...
-                return false;
-            }
-
-            var lockObj = _lockObjects[attrDef.Classification];
-            lock (lockObj)
-            {
-                // Grumble... Different attributes can have the same name, so this is kind of problematic...
-                return dic.ContainsKey(attrDef.Name);
-            }
+            // Grumble... Different attributes can have the same name, so this is kind of problematic...
```
jaffinito, line 88: Should this be a todo?

tippmar-nr: That's old code. Not related to this PR. No idea how long that comment has been around... ;-)

nrcventura: If it's something that we want to address, I think it should be a tech debt ticket. We would also need to be consistent with other agents and whatever SED does when it merges all of the attributes together. That comment was added a while ago when dealing with some other problem that was solved by using locks for the different attribute classifications.

PR #3386, opened 2026-01-09
chore: code refactor for OpenTelemetryBridge folder files

ashishsinghnr:

## Description

code refactor for OpenTelemetryBridge folder files
Files are grouped in 
1.Metrics
2.Tracing
3.Common folders

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -62,35 +62,6 @@ namespace NewRelic.Agent.Core.Configuration
         public bool UseResourceBasedNamingForWCFEnabled { get; private set; }
         
         private bool _hasLoggedEventListenerSamplersDisabled;
-        private bool _eventListenerSamplersEnabledConfigured;
-
-        /// <summary>
-        /// Controls whether EventListener-based samplers (GCSamplerNetCore, ThreadStatsSampler) are enabled.
-        /// Automatically returns false when OpenTelemetry metrics are enabled to prevent EventListener conflicts.
-        /// </summary>
-        /// <remarks>
-        /// Uses explicit backing fields instead of 'field' keyword because this property needs two pieces of state:
-        /// 1. _eventListenerSamplersEnabledConfigured - the actual property value
-        /// 2. _hasLoggedEventListenerSamplersDisabled - tracks if the warning log has been written
-        /// The setter is required by IConfiguration interface, though it goes against the typical pattern in this class.
-        /// </remarks>
-        public bool EventListenerSamplersEnabled
-        {
-            get
-            {
-                if (OpenTelemetryMetricsEnabled)
-                {
-                    if (!_hasLoggedEventListenerSamplersDisabled)
-                    {
-                        Log.Info("EventListener-based samplers (GCSamplerNetCore, ThreadStatsSampler) have been automatically disabled because OpenTelemetry metrics are enabled.");
-                        _hasLoggedEventListenerSamplersDisabled = true;
-                    }
-                    return false;
-                }
-                return _eventListenerSamplersEnabledConfigured;
-            }
-            set => _eventListenerSamplersEnabledConfigured = value;
-        }
 
         public TimeSpan DefaultHarvestCycle => TimeSpan.FromMinutes(1);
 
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -2805,22 +2830,23 @@ private void LogDisabledPropertyUse(string disabledPropertyName, string newPrope
 
         public bool OpenTelemetryTracingEnabled => OpenTelemetryEnabled && EnvironmentOverrides(_localConfiguration.opentelemetry.traces.enabled, "NEW_RELIC_OPENTELEMETRY_TRACES_ENABLED");
 
+        private List<string> _openTelemetryTracingIncludedActivitySources;
         public List<string> OpenTelemetryTracingIncludedActivitySources
         {
             get
             {
-                if (field == null)
+                if (_openTelemetryTracingIncludedActivitySources == null)
```
tippmar-nr: This change (and similar changes through the rest of this file) shouldn't be necessary. 

The existing code leverages the (relatively new) implicit backing field property `field`. We should (evetually) update all our code to eliminate the use of explicit backing fields, as they're no longer necessary.

src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
```diff
@@ -125,7 +126,7 @@ private static string GetSafeContentInfo(HttpContent content)
 
                 // For protobuf content, add warning about binary data not being logged
                 if (contentType.StartsWith("application/x-protobuf", StringComparison.OrdinalIgnoreCase) || 
-                    contentType.Contains("protobuf", StringComparison.OrdinalIgnoreCase))
+                    contentType.IndexOf("protobuf", StringComparison.OrdinalIgnoreCase) >= 0)
```
tippmar-nr: Why is `IndexOf()` better than `Contains()`?

ashishsinghnr: Actually while review and running code, it was giving me some error , so tried this and it came in checkin, but yes, lets use contains which is good. Thanks!

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
ashishsinghnr: done, Thanks!

src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -61,7 +60,32 @@ public class DefaultConfiguration : IConfiguration
         private readonly Dictionary<string, string> _newRelicAppSettings;
 
         public bool UseResourceBasedNamingForWCFEnabled { get; private set; }
-        public bool EventListenerSamplersEnabled { get; set; }
+        
+        private bool _hasLoggedEventListenerSamplersDisabled;
```
tippmar-nr, line 64: I don't think you need this backing field anymore - but the getter below will have to be refactored to just reference `field` -- or, if the backing field is really necessary, the getter shouldn't reference `field` at all. 

I don't think we should implement a setter at all for properties in this class.

ashishsinghnr: With latest code, now.using explicit backing fields instead of 'field' keyword because this property needs two pieces of state, Also. the setter is required by IConfiguration interface, which it is inheriting. Please review and suggest

tippmar-nr: I see now what's going on - the original implementation of `EventListenerSamplersEnabled` was done poorly (or lazily, I'm not sure which) and the setter is required for testing. 

So for now, go ahead and leave the setter in place. 

But you still don't need a backing field -- update the constructor to set `EventListenerSamplersEnabled` like it used to and change your implementation of the getter to 
```
        public bool EventListenerSamplersEnabled
        {
            get
            {
                if (OpenTelemetryMetricsEnabled)
                {
                    if (field && !_hasLoggedEventListenerSamplersDisabled)
                    {
                        Log.Info("EventListener-based samplers (GCSamplerNetCore, ThreadStatsSampler) have been automatically disabled because OpenTelemetry metrics are enabled.");
                        _hasLoggedEventListenerSamplersDisabled = true;
                        field = false;
                    }
                }

                return field;
            }
            set => field = value;
        }
```

Also, this property should be defined down at the end of the `#region IConfiguration Properties` region (around line 2950 or so).

Issue #3190, opened 2025-07-24
Add configuration to ignore configured process names


nrcventura:
There are times, especially within cloud environments where environment variables are applied to more processes than desired. This can result in the dotnet agent instrumenting a helper process, when the customer does not want to collect data from that process. 

We should consider adding a mechanism to configure a list of process names that should not be instrumented. This configuration should be supported in both an environment variable and configuration file. 

In addition to being able to ignore instrumenting certain helper processes, it will also provide a workaround for scenarios where the dotnet agent needs to make changes to its built in list of processes to ignore.

### Acceptance Criteria

1. A process matching the exact process name (case insensitive) in the process exclude list should not be instrumented.
2. Wildcard characters or regex will not be supported.
3. A log line should be recorded if a process was not instrumented because of this setting.
4. The new setting does not need to be checked in the managed agent, and does not need to be reported in the settings payload.
5. This setting should be documented with the rest of our agent settings.

### Additional context

Issue #3189 is the motivation for this feature.

### Estimates
Please provide initial t-shirt size: M

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-448847

nr-ahemsath:
This was implemented in January in PR #3387

PR #3387, opened 2026-01-12
feat: Allow users to configure process names that should not be instrumented

nr-ahemsath:
## Description

Implements #3190.  A new optional attribute `include` is added to the `application` element in the agent config XML, with a default value of `true`.  This allows users to configure application names (case-insensitive) that should not be profiled/instrumented.

```
<configuration>
  <instrumentation>
    <applications>
      <application name="foo.exe" include="false" />
    </applications>
  </instrumentation>
</configuration>
```

Two new config environment variables are added: 
`NEW_RELIC_EXCLUDED_APPLICATION_NAMES="foo.exe,bar.exe"`
`NEW_RELIC_INCLUDED_APPLICATION_NAMES="bar.exe,baz.exe"`

Things to note:
* Exclude takes precedence over include, so `bar.exe` will not be instrumented given the above env var values.
* As per usual with the agent, environment variables take precedence over `newrelic.confg` values - completely (i.e., the presence of a non-empty `NEW_RELIC_EXCLUDED_APPLICATION_NAMES` variable means that the agent will completely ignore any excluded application names in a global or local `newrelic.config` file.
* As a side benefit of these changes, users can now include non-IIS .NET Framework apps to be instrumented using an environment variable rather than the config file.

Manually tested on Windows 11, Ubuntu 24.04, and Alpine 3.23.  Our team decided against integration tests for this feature.

Documentation update PR: https://github.com/newrelic/docs-website/pull/22715

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -10,6 +10,12 @@ on:
     types: [published]
 
   workflow_dispatch:
+    inputs:
+      aggregate_payload_data:
+        description: 'Collect and aggregate payload data'
+        required: false
+        type: boolean
+        default: false
 
   schedule:
     - cron: "0 9 * * 1-5"
@@ -154,6 +160,8 @@ jobs:
     needs:
       - build-fullagent-msi
     uses: ./.github/workflows/linux_container_tests.yml
+    with:
+      aggregate_payload_data: ${{ inputs.aggregate_payload_data == true || github.event_name == 'schedule' }}
     secrets: inherit
     permissions:
       actions: write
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Profiler/Configuration/Configuration.h
```diff
@@ -764,8 +841,8 @@ namespace NewRelic { namespace Profiler { namespace Configuration {
                 }
             }
 
-            if (IsProcessInProcessList(_processes, processName)) {
-                LogInfo(L"Enabling instrumentation for this process (", processName, ") due to existence of application node in newrelic.config.");
+            if (IsProcessInProcessList(_includedProcesses, processName)) {
```
tippmar-nr, line 844: Should we have a similar check and log message for excluded processes?

jaffinito: It would be helpful for support if things aren't working.

nr-ahemsath: It does log a message, just in a different part of the code (at/near the top of shouldInstrument).

PR #3392, opened 2026-01-13
ci: Collect / aggregate payload data on demand and on scheduled runs

tippmar-nr:
Updates the all_solutions workflow to add an option during a manual run for choosing whether to aggregate and collect payload data, defaulting to false.

Also updates the logic to always collect and aggregate payload data when the workflow is executed via the nightly schedule.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
     inputs:
       aggregate_payload_data:
-        description: 'Build and aggregate payload data'
+        description: 'Collect and aggregate payload data'
         required: false
         type: boolean
         default: false


.github/workflows/linux_container_tests.yml
```diff
@@ -130,7 +139,7 @@ jobs:
           --results-directory ${{ env.test_results_path }}
 
       - name: Extract payload bytes log from TRX
-        if: always()
+        if: inputs.aggregate_payload_data == true
```
nrcventura, line 142: Do these need the `schedule` check too?

tippmar-nr: This one doesn't -- the value is passed as an input from the all_solutions workflow that invokes the container test workflow.

PR #3398, opened 2026-01-14
chore: Update env var list to log environment variables related to feat 3387

neilpdx:
## Description

Added two new environment variables to log at `debug` level related to this new feature:

Allow users to configure process names that should not be instrumented (https://github.com/newrelic/newrelic-dotnet-agent/issues/3387) ([8dedf32](https://github.com/newrelic/newrelic-dotnet-agent/commit/8dedf324fed3559c5de8697a9728b31cf8dacf96))

NEW_RELIC_EXCLUDED_APPLICATION_NAMES
NEW_RELIC_INCLUDED_APPLICATION_NAMES

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/AgentManager.cs
```diff
@@ -290,6 +290,8 @@ private void LogInitialized()
                     "NEW_RELIC_DISABLE_SAMPLERS",
                     "NEW_RELIC_ERROR_COLLECTOR_IGNORE_ERROR_CODES",
                     "NEW_RELIC_ERROR_COLLECTOR_EXPECTED_ERROR_CODES",
+                    "NEW_RELIC_EXCLUDED_APPLICATION_NAMES",
+                    "NEW_RELIC_INCLUDED_APPLICATION_NAMES",
```
tippmar-nr: Can you update this to maintain alphabetical order please?

nr-ahemsath:
@neilpdx Thanks for suggesting this, Neil!  Note that these variables are only checked in the profiler, so logging them in the managed agent might not always be helpful to troubleshoot customers' "why isn't my application being profiled?" issues.

It might be better to add specific logging around these messages at DEBUG or FINEST level in the profiler.

src/Agent/NewRelic/Agent/Core/AgentManager.cs
neilpdx: @tippmar-nr My original thought was that variables related to the same feature could be grouped together, but I've made that change.

neilpdx:
@nr-ahemsath Good point! I figured it would be worth adding them here to log as many env vars as possible but I'll see if I can figure out how to add them to the profiler as well.

nr-ahemsath:
I just had Claude help me hammer out changes to avoid needing to add new env vars to log in this PR, which would supersede yours: https://github.com/newrelic/newrelic-dotnet-agent/pull/3401

Thanks for reminding us that we should log the new vars associated with #3387!

tippmar-nr:
Superseded by #3401

PR #3400, opened 2026-01-14
ci: Cleanup temporary payload archives

tippmar-nr:
Updates `all_solutions.yml` to ensure all temporary payload bytes archives are removed after the final aggregation step. This was supposed to be happening already, but the action I was using has a bug with wildcards.


nr-ahemsath (approved):
I verified that this is working correctly by comparing recent CI runs on `main` vs. this branch. 👍

PR #3401, opened 2026-01-14
chore: Simplify environment variable logging

nr-ahemsath:
Simplify the environment variable logging code by logging any env vars found beginning with `NEW_RELIC_`, while still respecting sensitive vars whose values should not be logged.  This should help reduce toil associated with adding new configuration env vars in the future.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/AgentManager.cs
+++ b/src/Agent/NewRelic/Agent/Core/AgentManager.cs
@@ -271,12 +271,12 @@ namespace NewRelic.Agent.Core
                     "NEW_RELIC_CLOUD_AWS_ACCOUNT_ID"
                 };
 
+                // This list purposefully does not include NEWRELIC_LICENSEKEY
                 List<(string,string)> envVarsDeprecated = new List<(string, string)>
                 {
                     ("CORECLR_NEWRELIC_HOME","CORECLR_NEW_RELIC_HOME"),
                     ("NEWRELIC_HOME", "NEW_RELIC_HOME"),
                     ("NEWRELIC_INSTALL_PATH", "NEW_RELIC_INSTALL_PATH"),
-                    ("NEWRELIC_LICENSE_KEY", "NEW_RELIC_LICENSE_KEY"),
                     ("NEWRELIC_LOG_DIRECTORY", "NEW_RELIC_LOG_DIRECTORY"),
                     ("NEWRELIC_LOG_LEVEL", "NEW_RELIC_LOG_LEVEL"),
                     ("NEWRELIC_PROFILER_LOG_DIRECTORY", "NEW_RELIC_PROFILER_LOG_DIRECTORY"),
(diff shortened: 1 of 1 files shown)


src/Agent/NewRelic/Agent/Core/AgentManager.cs
```diff
@@ -343,43 +271,44 @@ private void LogInitialized()
...
+                    ("NEWRELIC_LICENSE_KEY", "NEW_RELIC_LICENSE_KEY"),
                     ("NEWRELIC_LOG_DIRECTORY", "NEW_RELIC_LOG_DIRECTORY"),
                     ("NEWRELIC_LOG_LEVEL", "NEW_RELIC_LOG_LEVEL"),
                     ("NEWRELIC_PROFILER_LOG_DIRECTORY", "NEW_RELIC_PROFILER_LOG_DIRECTORY"),
                     ("NEWRELIC_FORCE_PROFILING", "NEW_RELIC_FORCE_PROFILING"),
                     ("NEWRELIC_AGENT_VERSION_OVERRIDE", "NEW_RELIC_AGENT_VERSION_OVERRIDE")
                 };
 
-                // so we can report the values as usual
-                environmentVariables.AddRange(environmentVariablesDeprecated.Select(tuple => tuple.Item1));
```
tippmar-nr, line 359: We should preserve this functionality -- deprecated environment variable names and values should be reported. `NEWRELIC_LICENSEKEY` should be handled separately to ensure the value doesn't get reported but the name does get reported as deprecated.

src/Agent/NewRelic/Agent/Core/AgentManager.cs
```diff
@@ -343,43 +271,44 @@ private void LogInitialized()
                     "NEW_RELIC_CLOUD_AWS_ACCOUNT_ID"
                 };
 
-                List<(string,string)> environmentVariablesDeprecated = new List<(string, string)>
+                List<(string,string)> envVarsDeprecated = new List<(string, string)>
                 {
                     ("CORECLR_NEWRELIC_HOME","CORECLR_NEW_RELIC_HOME"),
                     ("NEWRELIC_HOME", "NEW_RELIC_HOME"),
                     ("NEWRELIC_INSTALL_PATH", "NEW_RELIC_INSTALL_PATH"),
+                    ("NEWRELIC_LICENSE_KEY", "NEW_RELIC_LICENSE_KEY"),
```
tippmar-nr: The deprecated value should be `NEWRELIC_LICENSEKEY`

src/Agent/NewRelic/Agent/Core/AgentManager.cs
nr-ahemsath: Okay, thanks for explaining what was going on before - I couldn't quite understand it and I assumed it was Claude hallucinating.  Should be fixed now.

PR #3405, opened 2026-01-15
chore: Add supportability metrics for dropped span links and span events

tippmar-nr:
Adds supportability metrics for dropped span links and span events. Includes unit tests.


src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
```diff
@@ -104,6 +104,10 @@ public class Segment : IInternalSpan, ISegmentDataState, IHybridAgentSegment
 
     private const int MaxSpanEventsPerSegment = 100;
     private const int MaxSpanLinksPerSegment = 100;
+
+    public int SpanEventLinksDropped { get; private set; }
```
nr-ahemsath: Maybe these should be `InterlockedCounter()`s for thread safety?

tippmar-nr: Probably not strictly necessary but won't hurt anything. I'll update.

PR #3428, opened 2026-01-29
chore: Miscellaneous code cleanup, build warning change

tippmar-nr:
This PR includes some code cleanup and a couple small changes to build warnings. 

`IDE005: Using directive is not necessary` is now a warning rather than a suggestion and will fail the build. You could argue that it doesn't matter if we unnecessarily reference an unused namespace, but I'd argue that it's only minor housekeeping to keep in check, once the existing violations are addressed in this PR.

Enabling that, however, required that `<GenerateDocumentationFile>true</GenerateDocumentationFile>` be added to the `Directory.build.props` file, which in turn enabled a bunch of warnings related to incorrect formatting of XML comments. 

Rather than disable all of those warnings, I fixed the issues but ultimately decided to turn off `CS1591: Missing XML comment for publicly visible type or member` so we're not hounded endlessly to add more XML comments; the remaining warnings will ensure that all existing XML comments are properly formatted and parameter references in the comments match the parameters of the method they're documenting.

**NOTE**: These build warning changes (for now) apply only to files under the `src` folder - until I have time to refactor the rest of the repo (integration tests, build tools, etc.), we'll limit the scope. I added a new `.editorconfig` in the `src` folder and made some changes to the `Directory.build.props` file in that folder; later those changes will be moved up to the root-level files.

The bulk of the remaining changes in this PR are to address the warnings mentioned above -- enabling `IDE005` actually pointed out quite a few more unnecessary `using`s that weren't caught by my ReSharper refactoring. Note also that there are a couple of file-scope namespace updates, so you'll want to "ignore whitespace" to reduce the clutter.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -22,10 +22,11 @@ charset = utf-8 # supposed to get rid of the BOM issue
 dotnet_sort_system_directives_first = true
 dotnet_separate_import_directive_groups = false
 
+# save for later, only enabled for the `src` folder now
 # IDE0005: Using directive is unnecessary
-dotnet_diagnostic.IDE0005.severity = warning
+#dotnet_diagnostic.IDE0005.severity = warning
 # CS1591: Missing XML comment for publicly visible type or member
-dotnet_diagnostic.CS1591.severity = none
+#dotnet_diagnostic.CS1591.severity = none
 
 # this. preferences
 dotnet_style_qualification_for_field = false:silent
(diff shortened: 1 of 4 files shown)


chynesNR:
It's probably not a big deal, but I feel like escaping the XML characters makes some of these comments less readable.

tippmar-nr:
> It's probably not a big deal, but I feel like escaping the XML characters makes some of these comments less readable.

Well, I mean, XML isn't meant to be human-readable. But it renders nicely when you hover over a documented method name.

PR #3449, opened 2026-02-11
chore: Add agent metadata action step 1

mvicknr:
## Description

This PR adds the ability for the NR agent team to send agent metadata and binaries to New Relic for use in fleets and other features. This will be triggered normally during a release. but if it fails, it will not fail the release. It can also be run on-demand for backfilling older agent versions, to make corrections, or to re-run the job after a failure.

GH action info: https://github.com/newrelic/agent-metadata-action/blob/main/README.md

This PR is step 1 that will use a`TestAgent` agent name instead of `NRDotnetAgent`
This PR needs to be merged to main in order to run the new on-demand workflow
A later PR will add a step in the release and use the real agent name

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


.github/workflows/agent_metadata.yml
```diff
@@ -0,0 +1,57 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag (needs to exactly match the GH tag for the release)'
+        required: true
+        type: string
+      agent-type:
+        description: 'Agent type'
+        required: false
+        default: 'NRDotNetAgent'
```
nr-ahemsath, line 17: The description says that a "TestAgent" name would be used instead of "NRDotNetAgent"...or is that defined somewhere else?

.github/workflows/agent_metadata.yml
```diff
@@ -0,0 +1,57 @@
...
+  workflow_call:
+    inputs:
+      version:
+        description: 'Version tag (needs to exactly match the GH tag for the release)'
+        required: true
+        type: string
+      agent-type:
+        description: 'Agent type'
+        required: false
+        default: 'NRDotNetAgent'
```
nr-ahemsath, line 35: Same thing here.

.github/workflows/agent_metadata.yml
mvicknr: I plan to run the manual flow and input the `TestAgent` name to do the test

.github/workflows/agent_metadata.yml
```diff
@@ -0,0 +1,57 @@
...
+        type: boolean
+    secrets:
+      FC_SYS_ID_CLIENT_ID:
+        required: true
+      FC_SYS_ID_PR_KEY:
+        required: true
+
+jobs:
+  send-agent-metadata:
+    uses: newrelic/newrelic-agent-init-container/.github/workflows/agent-metadata.yml@main
```
mvicknr, line 50: When calling a reusable workflow, the `uses` cannot be within steps of the job so I cannot call harden from this job
https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#calling-a-reusable-workflow

Since the reusable workflow has the harden step, this should cover anything in that job

nr-ahemsath: Okay, thanks for looking into that for us!

Issue #3306, opened 2025-11-12
Perf testing in CI: Create test app runner


nr-ahemsath:
### Describe the story, please be clear on scope of the story.

Create a test app runner in GHA that is capable of running a test app (with or without the agent attached) and driving traffic to it.

This is the first part of implementing performance testing in our CI.  We are basing this work on what the Ruby agent team has recently done, which in turn was based on the PHP agent team's more elaborate perf/soak tests.

The test app should be a modern .NET ASP.NET Core WebApi.  The runner will need to be a Linux runner so we can use Docker to run the test app, as well as the traffic driver (Ruby is using Locust).

The workflow should use parameters to control things like whether or not the agent is attached, which agent version to use, how long to run the test, how much traffic to drive, etc.

See the Ruby team's parallel story: https://github.com/newrelic/newrelic-ruby-agent/issues/3199
See the sandbox repo where Ruby has been working on this before bringing it back into their main repo: https://github.com/tannalynn/agent_perf_testing

### Acceptance Criteria

A GHA workflow that can:

1. Launch an ASP.NET Core WebApi test app in Docker
2. Potentially attach an agent to the test app, reporting to our team's sandbox account in staging
3. Drive traffic to the test app from a different Docker container (to isolate the resource usage of the traffic driver from the test app)
4. Makes sure the test app runs without error and (if applicable) the agent successfully sent data to staging.

### Estimates
Please provide initial t-shirt size
L

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-487702

PR #3460, opened 2026-02-25
ci: Add performance test runner

nr-ahemsath:
Add performance test runner with ASP.NET Core WebAPI and Locust traffic driver.  Implements #3306 

Adds a GHA workflow and supporting infrastructure for running performance tests against the .NET agent in Docker on a Linux runner. The test app and traffic driver run in isolated containers on a shared bridge network.

Workflow inputs control agent attachment, agent source (GitHub release or build artifact), agent version, test duration, and Locust concurrency. When the agent is attached, the workflow checks the agent logs to make sure data was sent to the collector without error.

This is phase one of three for implementing performance testing in our CI.  The second will add Docker monitoring to the runner at the container level, and the third will add summarization and visualization.

### DIFF
--- a/.github/workflows/performance_tests.yml
+++ b/.github/workflows/performance_tests.yml
@@ -335,7 +335,7 @@ jobs:
           echo "Traffic driver completed successfully."
 
       # -----------------------------------------------------------------------
-      # Verify the agent sent data to New Relic staging (agent runs only)
+      # Verify the agent sent data to New Relic (agent runs only)
       # -----------------------------------------------------------------------
       - name: Verify agent sent data to New Relic
         if: steps.inputs.outputs.attach_agent == 'true'


.github/workflows/performance_tests.yml
```diff
@@ -1,13 +1,406 @@
 name: Performance Tests
 
+# Runs the .NET agent performance test suite.
+# Launches an ASP.NET Core WebAPI test app and a Locust traffic driver in
+# separate Docker containers, optionally with the agent attached.
+#
+# Required secrets (when attach_agent is true):
+#   STAGING_LICENSE_KEY  - New Relic staging license key (team sandbox account)
```
nrcventura: Should we just refer to license key? There is nothing preventing someone from using non-staging account to run these tests. I think we just need a combination of collector host and license key.

.github/workflows/performance_tests.yml
```diff
@@ -1,13 +1,406 @@
...
+
+      - name: Start test app and traffic driver
+        working-directory: ${{ env.PERF_TEST_DIR }}
+        env:
+          DOTNET_VERSION: ${{ steps.inputs.outputs.dotnet_version }}
+          TEST_DURATION: ${{ steps.inputs.outputs.test_duration }}
+          LOCUST_USERS: ${{ steps.inputs.outputs.locust_users }}
+          LOCUST_SPAWN_RATE: ${{ steps.inputs.outputs.locust_spawn_rate }}
+          NEW_RELIC_LICENSE_KEY: ${{ secrets.STAGING_LICENSE_KEY }}
+          NEW_RELIC_HOST: staging-collector.newrelic.com
```
nrcventura: I think we generally have this defined in a secret along with the license key for our integration tests.

.github/workflows/performance_tests.yml
```diff
@@ -1,13 +1,406 @@
...
+            if [ -f "$STATS_FILE" ]; then
+              echo "--- Locust stats ---"
+              cat "$STATS_FILE"
+            fi
+            exit 1
+          fi
+          echo "Traffic driver completed successfully."
+
+      # -----------------------------------------------------------------------
+      # Verify the agent sent data to New Relic staging (agent runs only)
```
nrcventura: If we clean up the staging restriction, we should take a pass through the comments to remove the references to staging.

.github/workflows/performance_tests.yml
nr-ahemsath: STAGING_LICENSE_KEY is the name of the secret in GHA.

There's nothing preventing us from adding the option to test against prod instead of staging, but why would we want to, given the use case of performance testing in CI?  As far as I know we don't have the ability to run the integration tests against prod.

.github/workflows/performance_tests.yml
nr-ahemsath: Yes, but it's in a big blob of JSON along with a bunch of other things.  Seems easier and more clear to just use the discrete license key secret.

.github/workflows/performance_tests.yml
nr-ahemsath: Again, not sure why we would.

nrcventura (dismissed):

PR #3470, opened 2026-03-06
feat: Extract structured log message arguments as context data for MEL

### LINKED ISSUE #2519: MicrosoftExtensionsLogging: Add each log argument/attribute to the context data
### Feature Description
Currently, with MEL, only log scopes are included in the context data (using `logger.BeginScope(...)`).  The log attributes that are passed as arguments to the standard `logger.Log*` methods are not in the context data.  They are included in the formatted message, but then they lose a lot of value because they need to be parsed back out during ingestion.

For example, explicit calls to the logger (or logger extension methods) like this do not have MyData as a context data value:
```csharp
logger.LogInformation("My log message with some data: {MyData}", myData);
```

or via the Source Generator (generated) calls like [this](https://github.com/dotnet/aspnetcore/blob/53db4d97d7c77d13e20e58a98f104e88d6af6040/src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.Log.cs#L70-L71) do not have the RouteData, MethodInfo, etc as context data values:
```csharp
[LoggerMessage(102, LogLevel.Information, "Route matched with {RouteData}. Executing controller action with signature {MethodInfo} on controller {Controller} ({AssemblyName}).", EventName = "ControllerActionExecuting", SkipEnabledCheck = true)]
private static partial void ControllerActionExecuting(ILogger logger, string routeData, MethodInfo methodInfo, string controller, string? assemblyName);
```

### Describe Alternatives
None considered

### Additional context
I believe this already works as described with other loggers (e.g. Serilog and NLog), but I could be wrong.

### Priority
Really Want

piekstra:
## Summary

Fixes #2519

When using Microsoft.Extensions.Logging with structured logging (e.g., `logger.LogInformation("User {UserId} did {Action}", userId, action)`), the named parameters are now extracted from the `TState` state parameter and merged into context data. This makes them available as individual log attributes in New Relic (e.g., `context.UserId`, `context.Action`), matching the behavior already available for Serilog and NLog.

### What changed

In `MicrosoftLoggingWrapper.cs`:

1. **`RecordLogMessage`** — captures the `state` object (`MethodArguments[2]`) and passes it to `GetContextData`
2. **`GetContextData`** — after harvesting scope-based context data (existing behavior), iterates the state as `IEnumerable<KeyValuePair<string, object>>` to extract individual structured log parameters. The `{OriginalFormat}` key (message template string) is excluded from the state extraction path.

**Note on `{OriginalFormat}`:** The existing scope harvesting code (unchanged by this PR) already leaks `{OriginalFormat}` as a context data attribute when MEL scopes contain `FormattedLogValues`. This PR filters it from the *state* extraction path only. Fixing the scope path leak is a separate pre-existing issue.

### Safety

- The state extraction is wrapped in its own inner `try/catch`, so if it fails for any reason, scope-based context data is still returned — existing behavior is not disrupted
- The `state` parameter defaults to `null`, maintaining backward compatibility
- Uses pattern matching (`state is IEnumerable<...>`) so non-enumerable state types are safely skipped

### Background

MEL's `FormattedLogValues` (the default `TState` for structured log calls) implements `IReadOnlyList<KeyValuePair<string, object?>>`. Each entry represents a named parameter from the message template. Currently the agent only uses `state.ToString()` for the rendered message but does not extract the individual parameters as context data. This means users who use structured logging with MEL can only access their log parameters via parsing rules, unlike Serilog/NLog users who get them as first-class attributes.

### Testing

Added an integration test (`StructuredLogArgContextDataTests.cs`) that verifies structured log arguments appear as context data attributes. The test:
- Configures MEL with context data enabled
- Logs a message with structured args: `logger.LogInformation("Person {Name} has id {Id}", name, id)`
- Asserts that `Name` and `Id` appear as individual attributes on the log event

Supporting changes:
- Added `InfoWithStructuredArgs` method to `MicrosoftLoggingLoggingAdapter`
- Added `CreateSingleLogMessageWithStructuredArgs` entry point to `LoggingTester`

We've also deployed this change in production against our services and confirmed structured log args appear as expected in New Relic.

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/Logging/StructuredLogArgContextDataTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/Logging/StructuredLogArgContextDataTests.cs
@@ -1,0 +1,96 @@
+// Copyright 2020 New Relic, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NewRelic.Agent.IntegrationTestHelpers;
+using NewRelic.Agent.IntegrationTestHelpers.RemoteServiceFixtures;
+using Xunit;
+
+namespace NewRelic.Agent.IntegrationTests.Logging.StructuredLogArgContextData;
+
+/// <summary>
+/// Verifies that structured log message arguments (e.g., logger.LogInformation("Person {Name} has id {Id}", name, id))
+/// are extracted as context data attributes when using Microsoft.Extensions.Logging.
+/// </summary>
+public abstract class StructuredLogArgContextDataTestsBase<TFixture> : NewRelicIntegrationTest<TFixture>
+    where TFixture : ConsoleDynamicMethodFixture
+{
+    private readonly TFixture _fixture;
+
+    private const string ExpectedName = "TestUser";
+    private const string ExpectedId = "12345";
+    private const string ExpectedMessage = "Person TestUser has id 12345";
+
+    public StructuredLogArgContextDataTestsBase(TFixture fixture, ITestOutputHelper output) : base(fixture)
+    {
+        _fixture = fixture;
+        _fixture.SetTimeout(TimeSpan.FromMinutes(2));
+        _fixture.TestLogger = output;
+
+        _fixture.AddCommand($"LoggingTester SetFramework MicrosoftLogging {RandomPortGenerator.NextPort()}");
+        _fixture.AddCommand($"LoggingTester Configure");
+        _fixture.AddCommand($"LoggingTester CreateSingleLogMessageWithStructuredArgs {ExpectedName} {ExpectedId}");
+
+        _fixture.AddActions
+        (
+            setupConfiguration: () =>
+            {
+                var configModifier = new NewRelicConfigModifier(fixture.DestinationNewRelicConfigFilePath);
+
+                configModifier
+                    .EnableContextData(true)
+                    .SetLogLevel("debug");
+            },
+            exerciseApplication: () =>
+            {
+                _fixture.AgentLog.WaitForLogLine(AgentLogBase.LogDataLogLineRegex, TimeSpan.FromSeconds(30));
+            }
+        );
+
+        _fixture.Initialize();
+    }
+
+    [Fact]
+    public void StructuredLogArgsAppearAsContextData()
+    {
+        var expectedLogLines = new List<Assertions.ExpectedLogLine>
+        {
+            new Assertions.ExpectedLogLine
+            {
+                Level = "INFO",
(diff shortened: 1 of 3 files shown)


piekstra:
## Testing Results

We built this change and deployed it to a production-like environment running .NET 10 with `Microsoft.Extensions.Logging` (no Serilog/NLog — pure MEL with the default console provider).

### What we verified

**Structured log message arguments now appear as individual `context.*` attributes in New Relic:**

For log calls like:
```csharp
_logger.LogInformation("Retrieved {Count} records in {ElapsedMilliseconds}ms", count, elapsed);
```

New Relic now receives these as first-class attributes:
- `context.Count` (numeric)
- `context.ElapsedMilliseconds` (numeric)

We also confirmed that ASP.NET Core's own structured log parameters (`ActionId`, `ActionName`, `RequestId`, `RequestPath`, `ConnectionId`, `ContentRoot`, `address`) are automatically extracted as well.

**Existing scope-based context data is unaffected:**

Attributes added via `ILogger.BeginScope()` (such as custom per-request and app-lifetime scopes) continue to work correctly alongside the new state-based extraction.

**Safety:**

The inner try/catch around state extraction worked as intended — no disruption to existing behavior. The `{OriginalFormat}` key is correctly filtered from the state extraction path and does not appear in context data from structured log arguments. Note: `{OriginalFormat}` can still leak from the existing scope harvesting path (e.g., when `BeginScope` receives `FormattedLogValues`) — this is a pre-existing issue unrelated to this PR.

### Build approach

We built the full `FullAgent.sln` on Windows (matching the upstream CI), extracted the `NewRelic.Providers.Wrapper.MicrosoftExtensionsLogging.dll` from the `newrelichome_arm64_coreclr_linux/Extensions/` output, and dropped it into the installed agent's extensions directory (`/usr/local/newrelic-dotnet-agent/extensions/`) in our Docker base image.

### Integration tests

Added `StructuredLogArgContextDataTests.cs` which configures MEL with context data enabled, logs a message with structured args (`"Person {Name} has id {Id}"`), and asserts that `Name` and `Id` appear as individual attributes on the forwarded log event.

nr-ahemsath:
Hello @piekstra, thanks for taking the time to submit this PR.  I will be taking a look at it this week.

nr-ahemsath:
@piekstra Unfortunately the way our CI works, we can't run our integration tests from forks.  We may also want to update our integration tests based on your code changes.  Can you please retarget your PR to this branch: feature/mel-structured-logging-as-context-data ?  We can then merge your changes into that branch and continue from there.  We will make sure you are listed as a co-author on the final PR into main.

piekstra:
Done — retargeted to `feature/mel-structured-logging-as-context-data`. Let me know if you need anything else from my end!

PR #3479, opened 2026-03-12
feat: Include structured log message arguments as context data for MEL

### LINKED ISSUE #2519: MicrosoftExtensionsLogging: Add each log argument/attribute to the context data

nr-ahemsath:
## Description

Fixes #2519

When using Microsoft.Extensions.Logging (MEL) with structured logging (e.g., logger.LogInformation("User {UserId} did {Action}", userId, action)), the named parameters are now extracted from the TState state parameter and merged into context data. This makes them available as individual log attributes in New Relic (e.g., context.UserId, context.Action), matching the behavior already available for Serilog and NLog.

The state extraction is wrapped in its own try/catch to ensure it cannot disrupt existing scope-based context data harvesting. The {OriginalFormat} key (the message template string) is excluded.

## Testing

Adds StructuredLogArgContextDataTests that verifies structured log message arguments (e.g., {Name}, {Id}) are extracted as context data attributes for supported use cases:
* MEL by itself
* MEL with Serilog
* MEL with NLog
* Serilog by itself
* NLog by itself

`Log4Net` is purposefully not tested because it doesn't support structured logging.

Additionally, some refactoring of the logging testing system is done for clarity, particularly around which testing frameworks are supported in which TFMs we build the ConsoleMF system for, and making the main code path for logging messages in LoggingTester easier to understand (there was a lot of entanglement with the context data dictionary logging path which is only used by one set of tests).

Co-authored by @piekstra 

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/tests/Agent/IntegrationTests/IntegrationTests/Logging/StructuredLogArgContextDataTests.cs
+++ b/tests/Agent/IntegrationTests/IntegrationTests/Logging/StructuredLogArgContextDataTests.cs
@@ -19,6 +19,7 @@ public abstract class StructuredLogArgContextDataTestsBase<TFixture> : NewRelicI
 {
     private readonly TFixture _fixture;
     private readonly LoggingFramework _loggingFramework;
+    private readonly bool _includeContext;
 
     private string _expectedName = "TestUser";
     private string _expectedId = "12345";
@@ -29,10 +30,23 @@ public abstract class StructuredLogArgContextDataTestsBase<TFixture> : NewRelicI
     private string _messageTemplate = "Person{Name}HasId={Id}";
     private string _expectedMessage  = "PersonTestUserHasId=12345";
 
-    public StructuredLogArgContextDataTestsBase(TFixture fixture, ITestOutputHelper output, LoggingFramework loggingFramework) : base(fixture)
+    private Dictionary<string, string> GetExpectedAttributes(LoggingFramework framework) => new Dictionary<string, string>()
+    {
+        { "framework", framework.ToString() },
+        { "mycontext1", "foo" },
+        { "mycontext2", "bar" },
+        { "mycontext3", "test" },
+        { "mycontext4", "value" },
+    };
+
+    private string FlattenExpectedAttributes(Dictionary<string, string> attributes) => string.Join(",", attributes.Select(x => x.Key + "=" + x.Value).ToArray());
+
+
+    public StructuredLogArgContextDataTestsBase(TFixture fixture, ITestOutputHelper output, LoggingFramework loggingFramework, bool includeContext) : base(fixture)
     {
         _fixture = fixture;
         _loggingFramework = loggingFramework;
+        _includeContext = includeContext;
         _fixture.SetTimeout(TimeSpan.FromMinutes(2));
         _fixture.TestLogger = output;
 
@@ -48,7 +62,15 @@ public abstract class StructuredLogArgContextDataTestsBase<TFixture> : NewRelicI
 
         _fixture.AddCommand($"LoggingTester SetFramework {loggingFramework} {RandomPortGenerator.NextPort()}");
         _fixture.AddCommand($"LoggingTester Configure");
-        _fixture.AddCommand($"LoggingTester CreateSingleLogMessageWithStructuredArgs {_messageTemplate} {string.Join(",", _expectedName, _expectedId)}");
+        if (_includeContext)
+        {
+            var expectedAttributes = GetExpectedAttributes(loggingFramework);
+            _fixture.AddCommand($"LoggingTester CreateSingleLogMessageWithStructuredArgsAndContext {_messageTemplate} {string.Join(",", _expectedName, _expectedId)} {FlattenExpectedAttributes(expectedAttributes)}");
+        }
+        else
+        {
+            _fixture.AddCommand($"LoggingTester CreateSingleLogMessageWithStructuredArgs {_messageTemplate} {string.Join(",", _expectedName, _expectedId)}");
+        }
         _fixture.AddActions
         (
             setupConfiguration: () =>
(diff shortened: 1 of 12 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
```diff
@@ -106,6 +110,35 @@ private static Dictionary<string, object> GetContextData(MEL.ILogger logger, IAg
...
+            // the individual named parameters. We merge these into context data so they appear as
+            // separate log attributes in New Relic (e.g., context.UserId, context.Action).
+            try
+            {
+                if (state is IEnumerable<KeyValuePair<string, object>> stateKvps)
+                {
+                    foreach (var kvp in stateKvps)
+                    {
+                        // Skip the "{OriginalFormat}" key — it's the message template string, not a parameter
+                        if (kvp.Key == "{OriginalFormat}")
```
tippmar-nr: Is there any kind of constant from MEL that we could reference here instead of a hard-coded string? If not, maybe add a comment with a reference to the MEL code that defines that string so we can find it when it breaks in the future?

tests/Agent/IntegrationTests/IntegrationTests/Logging/StructuredLogArgContextDataTests.cs
```diff
@@ -0,0 +1,223 @@
...
+/// </summary>
+public abstract class StructuredLogArgContextDataTestsBase<TFixture> : NewRelicIntegrationTest<TFixture>
+    where TFixture : ConsoleDynamicMethodFixture
+{
+    private readonly TFixture _fixture;
+    private readonly LoggingFramework _loggingFramework;
+
+    private string _expectedName = "TestUser";
+    private string _expectedId = "12345";
+    private string _messageTemplate = "Person{Name}HasId={Id}";
```
tippmar-nr, line 30: I'm guessing the goofy formatting (no spaces between words) here is a limitation of how we pass string parameters to the test harness? We've used Based64 encoding / decoding on the parameters for other tests to work around this. Maybe not a huge deal, but at least deserves a comment so someone doesn't "helpfully" try to change it in the future!

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/LogInstrumentation/DummyMELAdapter.cs
```diff
@@ -55,13 +43,24 @@ public void NoMessage()
         _logger.LogTrace(string.Empty);
     }
 
-    public void Configure() => CreateMelLogger();
+    public void InfoWithContextDictionary(string message, Dictionary<string, object> context)
```
tippmar-nr, line 46: Can we add a test that uses both the context dictionary and a message with structured arguments? That's going to be a very common use case - apps will frequently have (for example) a middleware that wraps execution with a `_logger.BeginScope()` and messages emitted in the execution will use structured arguments.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/MicrosoftExtensionsLogging/MicrosoftLoggingWrapper.cs
nr-ahemsath: It's not a constant in MEL, it's hardcoded there too.  I've made it a constant in our code and documented where (well, one of the places at least) that it's defined in the MEL code (https://github.com/dotnet/runtime/blob/54e4456b01060f54e5bb7d715e77e5f41bac558f/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LogValuesFormatter.cs#L233)

tests/Agent/IntegrationTests/IntegrationTests/Logging/StructuredLogArgContextDataTests.cs
nr-ahemsath: Yeah.  We use the base64 technique for the AI tests, where it's important that a realistic prompt ("In one sentence, what is a large language model?") be sent to the various AI clients.  But it requires that the exerciser methods know that the incoming argument is going to be a base64 string and decode it.  This would be totally possible for the logging tests, but would require touching a lot more files as part of this PR, and it's not really important for what the structured logging context data extraction tests are about.  I've gone the comment route.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/NetStandardLibraries/LogInstrumentation/DummyMELAdapter.cs
nr-ahemsath: Good idea, done

PR #3512, opened 2026-04-01
chore: Ensure consistent DT header insertion/replacement logic across all wrappers

tippmar-nr:
## Description
Validates and updates all distributed trace header insertion/replacement logic to ensure consistent behavior everywhere. Only SQS, RestSharp and the (very old) Wcf3 ChannelFactory wrapper were implemented inconsistently.

Adds unit and integration tests that ensure the replacement logic works correctly everywhere. 

See the [related Jira ticket](https://new-relic.atlassian.net/browse/NR-483373) and the Google doc referenced in it for a report on the entire inventory of DT header insertion across the .NET agent.

One change that's important to note is in the RestSharp `AppendHeaders` wrapper - header replacement was moved to the onComplete delegate to ensure anything added by the real `AppendHeaders` call would be correctly replaced if there was a collision. This was potentially a bug in the original implementation, but would have been very unlikely to occur.

Some wrapper implementations could not be tested easily or did not need a test because of the way the underlying library handles headers:
- NServiceBus — uses dictionary indexer headers[key] = value which inherently replaces. Headers are managed internally
   by the NServiceBus pipeline, so pre-populating DT headers before the agent's hook would require a custom pipeline
  behavior.
- MassTransit (current and legacy) — uses SendHeaders.Set(key, value) which is MassTransit's own API that replaces by
  design. Pre-populating would require a custom ISendObserver or IFilter.
- WCF NetTcp — the Message.Headers are not directly accessible before the agent's BeforeSendRequest inspector runs in
  the WCF client programming model, so there's no natural way for pre-existing DT headers to appear. The fix (FindHeader + RemoveAt + Add) was validated by code review.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/tests/Agent/IntegrationTests/ContainerIntegrationTests/Tests/AwsSdk/AwsSdkSQSTest.cs
+++ b/tests/Agent/IntegrationTests/ContainerIntegrationTests/Tests/AwsSdk/AwsSdkSQSTest.cs
@@ -161,21 +161,24 @@ public abstract class AwsSdkSQSTestBase : NewRelicIntegrationTest<AwsSdkContaine
         }
 
         // Verify DT propagation works for queue 4 (pre-existing DT headers should have been replaced by agent)
+        // Uses transaction events instead of spans since adaptive sampling may not sample later transactions
         if (_initCollections)
         {
-            var queueProduceDT = $"MessageBroker/SQS/Queue/Produce/Named/{_testQueueName4}";
-            var produceSpanDT = spans.LastOrDefault(s => s.IntrinsicAttributes["name"].Equals(queueProduceDT));
-            var processRequestSpanDT = spans.LastOrDefault(s =>
-                s.IntrinsicAttributes["name"].Equals("OtherTransaction/Custom/AwsSdkTestApp.SQSBackgroundService.SQSReceiverService/ProcessRequestAsync")
-                && s.IntrinsicAttributes.ContainsKey("parentId")
-                && s.IntrinsicAttributes["parentId"].Equals(produceSpanDT?.IntrinsicAttributes["guid"]));
+            var sendDTTransactionEvent = _fixture.AgentLog.TryGetTransactionEvent(_metricScope4);
+            var receiveDTTransactionEvents = _fixture.AgentLog.GetTransactionEvents()
+                .Where(e => e.IntrinsicAttributes["name"].Equals("OtherTransaction/Custom/AwsSdkTestApp.SQSBackgroundService.SQSReceiverService/ProcessRequestAsync")
+                    && e.IntrinsicAttributes.ContainsKey("parentId")
+                    && e.IntrinsicAttributes.ContainsKey("traceId"));
+            // Find the receive event that shares the same traceId as the send event
+            var receiveDTTransactionEvent = sendDTTransactionEvent != null
+                ? receiveDTTransactionEvents.FirstOrDefault(e => e.IntrinsicAttributes["traceId"].Equals(sendDTTransactionEvent.IntrinsicAttributes["traceId"]))
+                : null;
 
             NrAssert.Multiple(
-                () => Assert.True(produceSpanDT != null, "produceSpanDT should not be null"),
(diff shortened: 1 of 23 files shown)


tests/Agent/IntegrationTests/ContainerIntegrationTests/Tests/AwsSdk/AwsSdkSQSTest.cs
```diff
@@ -151,6 +160,25 @@ public void Test()
...
+                s.IntrinsicAttributes["name"].Equals("OtherTransaction/Custom/AwsSdkTestApp.SQSBackgroundService.SQSReceiverService/ProcessRequestAsync")
+                && s.IntrinsicAttributes.ContainsKey("parentId")
+                && s.IntrinsicAttributes["parentId"].Equals(produceSpanDT?.IntrinsicAttributes["guid"]));
+
+            NrAssert.Multiple(
+                () => Assert.True(produceSpanDT != null, "produceSpanDT should not be null"),
+                () => Assert.True(processRequestSpanDT != null, "processRequestSpanDT (with existing DT headers) should not be null — agent should have replaced stale headers"),
+                () => Assert.Equal(produceSpanDT!.IntrinsicAttributes["traceId"], processRequestSpanDT!.IntrinsicAttributes["traceId"]),
+                // Verify the stale traceId was NOT propagated
+                () => Assert.NotEqual("stale0000000000000000000000000", processRequestSpanDT!.IntrinsicAttributes["traceId"].ToString())
```
nr-ahemsath: I don't suppose there's any way of making this a constant defined somewhere accessible to both the test code here and the exerciser/test app code?  If you changed the string in the exerciser to be "stale1111111111111111111" this test would still pass, even if the header replacement logic we're testing wasn't working.  Perhaps at least document with a comment that this has to match exactly that same string in the other places.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/AzureServiceBus/AzureServiceBusW3CTests.cs
```diff
@@ -215,4 +214,88 @@ public AzureServiceBusW3CTopicTestsCoreLatest(ConsoleDynamicMethodFixtureCoreLat
...
+
+        Assertions.MetricsExist(expectedMetrics, metrics);
+
+        // Verify DT propagation — stale headers should have been replaced
+        var spanEvents = _fixture.AgentLog.GetSpanEvents().ToList();
+        foreach (var span in spanEvents)
+        {
+            if (span.IntrinsicAttributes.TryGetValue("traceId", out var traceId))
+            {
+                Assert.NotEqual("stale0000000000000000000000000", traceId.ToString());
```
nr-ahemsath, line 287: Same thing here

nr-ahemsath (commented):
Overall this looks great, I just have a minor concern about magic string values being hardcoded in both test code and exerciser code; this may be a fundamental limitation of how our tests are structured, but it least should be documented very clearly that those magic values have to match exactly in order for the tests to actually be verifying what we want them to.

tests/Agent/IntegrationTests/ContainerIntegrationTests/Tests/AwsSdk/AwsSdkSQSTest.cs
tippmar-nr: There's not an easy way to define that traceparent header globally that all apps and tests can see it. Currently the header (`00-stale0000000000000000000000000-stale000000000-01`) is used in about 10 places between test apps and tests - including some very old test apps that don't have access to a common project (like `Shared`). 

This is a pretty common pattern in our tests - lots of magic strings on both sides - all you have to do to break something is change one of them and not the other and it'll become pretty obvious what the issue is.

tests/Agent/IntegrationTests/UnboundedIntegrationTests/AzureServiceBus/AzureServiceBusW3CTests.cs
tippmar-nr: See above

nrcventura (approved):

PR #3520, opened 2026-04-03
feat: Add support for instrumenting Hangfire background jobs.

### LINKED ISSUE #2498: Add support for HangFire Jobs
It would be great if the newrelic agent had logic to automatically instrument hangfire jobs just like it does for NServiceBus. This would prevent having to always try and remember to add the Transaction Attribute along with setting the name of the transaction.

For Example, before my jobs execute, I have the following code.

```
       [Transaction(Web = false)] // 1. Create transaction
        public async Task MyHangfireJob()
        {
            NewRelic.Api.Agent.NewRelic.SetTransactionName("Hangfire", $"GetStoreJobs"); // 2. Change Transaction Name
            ...
        }
```

jaffinito:
## Description

Instruments Hangfire job execution, but does not capture queue related data (size, latency, etc).  Works with sync and async methods.  Supports versions 1.7.X and 1.8.X (latest).  Will capture the queue for 1.8.X jobs, but the in-memory storage we use for testing doesn't populate that field.

Includes new integration tests.  Note, timing on the integration tests is a bit difficult since the Hangfire server will process the jobs at it's own pace.  The current delays seem to work on my system, but may need adjusted upwards for CI.

Fixes #2498

Here is the OTel PR about the attribute names:
https://github.com/open-telemetry/semantic-conventions/issues/1688
And where Hangfire is linking to them:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Hangfire/README.md#metrics

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Hangfire/HangfirePerformerWrapper.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        // Required so that we can ensure that the transaction is active during the entire job execution, including any async continuations
+        transaction.AttachToAsync();
+        transaction.DetachFromPrimary(); //Remove from thread-local type storage
+
+        // If this is in a queue, we want to create a metric more like a message queue.  Based on how Ruby handles ActiveJob.
+        var segmentName = string.IsNullOrWhiteSpace(queueName) ? "Hangfire/" + taskName : "Hangfire/Queue/Consume/Named/" + queueName + "/" + taskName;
+        var segment = transaction.StartTransactionSegment(instrumentedMethodCall.MethodCall, segmentName);
+
+        segment.AddAgentAttribute("workflow.platform.name", "hangfire");
```
tippmar-nr, line 65: What's the source for these attribute names? I don't remember if there's a spec that applies to Hanfire or not.

tests/Agent/IntegrationTests/SharedApplications/Common/MultiFunctionApplicationHelpers/MultiFunctionApplicationHelpers.csproj
```diff
@@ -338,6 +338,26 @@
...
+    <PackageReference Include="Hangfire" Version="1.7.15" Condition="'$(TargetFramework)' == 'net462'" />
+    <PackageReference Include="Hangfire.InMemory" Version="0.11.0" Condition="'$(TargetFramework)' == 'net462'" />
+
+    <PackageReference Include="Hangfire" Version="1.7.15" Condition="'$(TargetFramework)' == 'net8.0'" />
+    <PackageReference Include="Hangfire.InMemory" Version="0.11.0" Condition="'$(TargetFramework)' == 'net8.0'" />
+
+    <PackageReference Include="Hangfire" Version="1.8.23" Condition="'$(TargetFramework)' == 'net481'" />
+    <PackageReference Include="Hangfire.InMemory" Version="1.0.0" Condition="'$(TargetFramework)' == 'net481'" />
+
+    <PackageReference Include="Hangfire" Version="1.8.23" Condition="'$(TargetFramework)' == 'net10.0'" />
```
tippmar-nr: We should update Dotty config to include Hangfire packages

jaffinito: Good idea, that makes sense.

src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Hangfire/HangfirePerformerWrapper.cs
jaffinito: There isn't a spec.  These are from a PR put in by the hangfire folks to OTEL.  They wanted to extend the AI attributes around workflows to cover background jobs (workflows). I aligned our metrics with theirs.  Metric names are based on how Ruby handles ActiveJob and DelayedJob.

tippmar-nr (dismissed):

tests/Agent/IntegrationTests/IntegrationTests/Hangfire/HangfireTests.cs
```diff
@@ -46,6 +43,8 @@ protected HangfireTestsBase(TFixture fixture, ITestOutputHelper output, Version
             exerciseApplication: () =>
             {
                 _fixture.AgentLog.WaitForLogLine(AgentLogBase.MetricDataLogLineRegex, TimeSpan.FromMinutes(2));
```
tippmar-nr, line 46: Can these waits be made shorter? In the case of a massive failure, it'll take 8 minutes for the exercise code to eventually time out and fail.

jaffinito: I would think so.  I noticed that the jobs sometime didn't get picked up right away or didn't end as quickly, even with in-memory storage, so i kept them a bit longer than I would have normally.

tippmar-nr (approved):

PR #3456, opened 2026-02-19
feat: add array support for custom attributes in .NET Agent API

tippmar-nr:
## Summary

This PR adds comprehensive array support for custom attributes in the New Relic .NET Agent API, allowing developers to pass array values to `AddCustomAttribute()` that are properly serialized as JSON arrays.

### Core Implementation
- **Enhanced JSON serialization** in `JsonSerializerHelpers.cs`:
  - Added support for `IEnumerable` types (arrays, lists, etc.)
  - Proper null filtering - null elements are excluded from serialized arrays  
  - Empty arrays and null-only arrays are treated as null (not serialized)
  - Maintains existing conversion logic for individual array elements
  - Refactored to use single `WriteValue` method for all value types

#### Unit Tests (`Core.UnitTest`)
- Added `JsonSerializerHelpersTests` covering all array scenarios
- Tests for string arrays, int arrays, bool arrays, and Lists
- Validates null filtering and empty array handling
- Ensures mixed null arrays properly exclude null elements

#### Integration Tests  
- **Complete end-to-end testing** through New Relic .NET Agent API
- **Dual platform support**: Both .NET Framework (WebAPI) and .NET Core applications
- **Test applications** with array attribute endpoints:
  - `CustomArrayAttributes` - Tests string, int, bool arrays
  - `CustomEmptyArrayAttributes` - Tests empty/null-only array handling  
  - `CustomArrayWithNulls` - Tests null filtering within arrays
- **Comprehensive validation**: Arrays appear correctly in both transaction traces and transaction events
- **Reliable test execution**: Uses transaction events for multi-endpoint test scenarios

## API Usage

Developers can now use array values with the New Relic API:

```csharp
// All of these now work correctly
NewRelic.Api.Agent.NewRelic.GetAgent().CurrentTransaction.AddCustomAttribute("colors", new[] { "red", "green", "blue" });
NewRelic.Api.Agent.NewRelic.GetAgent().CurrentTransaction.AddCustomAttribute("numbers", new[] { 1, 2, 3, 4, 5 });  
NewRelic.Api.Agent.NewRelic.GetAgent().CurrentTransaction.AddCustomAttribute("flags", new[] { true, false, true });
NewRelic.Api.Agent.NewRelic.GetAgent().CurrentTransaction.AddCustomAttribute("items", new List<string> { "item1", "item2" });

// Null handling
[...]

PR #3531, opened 2026-04-09
fix: Complete array attribute serialization for all payload types

### LINKED ISSUE #3530: Custom array attributes serialized incorrectly in span events and transaction traces
### Description

PR #3456 added support for serializing array-type custom attributes in `JsonSerializerHelpers`, but the implementation was incomplete — the original feature worked correctly for transaction events, but two additional code paths were missed:

1. **`GenericConverter` in `AttributeDefinition.cs`**: The converter that normalizes custom attribute values before storage had no `IEnumerable` case. Array values fell through to `ToString()`, producing garbage strings like `"System.Int32[]"` instead of actual array data. This affected transaction traces.

2. **`AttributeValue.SetValue` in `Segments/AttributeValue.cs`**: The protobuf-backed `AttributeValue` used by `SpanAttributeValueCollection` only handled scalar types (string, bool, double, long, TimeSpan, DateTimeOffset) and a TypeCode-based switch for other numerics. When a `List<object>` (the output of the fixed `GenericConverter`) was passed in, it fell through to `TypeCode.Object`, which called `ToString()` — again producing garbage strings. This affected span events.

### Expected Behavior

Custom attributes set via `NewRelic.GetAgent().CurrentTransaction.AddCustomAttribute()` with array or `IEnumerable` values should serialize as JSON arrays in all payload types:

- Transaction traces (user attributes)
- Transaction events (user attributes)
- Span events (user attributes)

For example, `AddCustomAttribute("tags", new[] { "red", "green", "blue" })` should produce `"tags": ["red", "green", "blue"]` in all three payloads, not `"tags": "System.String[]"`.

### Actual Behavior

- Array attributes appeared as garbage strings (e.g., `"System.Int32[]"`, `"System.String[]"`) in transaction traces, transaction events, and span events.
- Empty arrays and null-only arrays were not filtered out, potentially producing empty or unexpected attribute values.

P3
S

tippmar-nr:
## Description
PR #3456 added array serialization in `JsonSerializerHelpers` but missed two code paths:

1. **`GenericConverter`** (`AttributeDefinition.cs`): No `IEnumerable` case — arrays fell through to `ToString()`, producing garbage like `System.Int32[]` in transaction traces and events.
2. **`AttributeValue.SetValue`** (`Segments/AttributeValue.cs`): The protobuf-backed `AttributeValue` for span events had no `IEnumerable` handling — arrays fell through to `TypeCode.Object` → `ToString()`→ garbage strings.

## Changes
- **`GenericConverter`**: Added `IEnumerable` case with recursive element conversion and null filtering
- **`Segments/AttributeValue`**: Added `_arrayValue` field and `IEnumerable` handling in `SetValue`; updated `Value` getter to return array data
- **Unit tests**: `GenericConverterTests` (11 new), `EventWireModelSerializerTests` (5 new)
- **Composite tests**: Updated `Assertions.HaveSameValue` for `IEnumerable` element-wise comparison; added `SpanCustomAttributes_WithArrays` test
- **Integration tests**: Array attribute assertions for transaction traces, transaction events, and span events

Fixes #3530


src/Agent/NewRelic/Agent/Core/Segments/AttributeValue.cs
```diff
@@ -129,6 +142,14 @@ private void SetValue(object value)
             return;
         }
 
+        // Note: this check relies on the string check earlier in this method to have already
+        // returned, since string implements IEnumerable<char>.
+        if (value is IEnumerable enumerable)
+        {
+            _arrayValue = value as List<object> ?? enumerable.Cast<object>().ToList();
+            return;
+        }
```
nrcventura, line 156: This does not make array values work, and instead just silently drops array attributes (because the `_arrayValue` will not be a part of the serialized output). In the protobuf, an AttributeValue is restricted to be the following.
```
message AttributeValue {
  oneof value {
    string string_value = 1;
    bool bool_value = 2;
    int64 int_value = 3;
    double double_value = 4;
  }
}
```
There is not a newer version of the proto that supports additional attribute types.

nrcventura: That code is only exercised for infinite tracing, and not normal distributed tracing.

src/Agent/NewRelic/Agent/Core/Segments/AttributeValue.cs
```diff
@@ -129,6 +142,19 @@ private void SetValue(object value)
             return;
         }
 
+        // Note: this check relies on the string check earlier in this method to have already
+        // returned, since string implements IEnumerable<char>.
+        if (value is IEnumerable enumerable)
```
nrcventura, line 147: Should all of the changes in this file be rolled back? I don't remember all of the details but I think this class is only used for infinite tracing. I think we use a different IAttributeValue implementation for non-infinite tracing spans.

tippmar-nr: No, this fix was necessary to get array attributes to properly show up in segments and span events.

nrcventura (approved):

PR #3532, opened 2026-04-13
chore: Add defensive exception handling to MassTransitHelpers

tippmar-nr:
## Summary
- Make `MassTransitHelpers` a static class
- Wrap URI parsing in `GetQueueDataFromUri` with a try-catch so no exceptions propagate to callers (e.g. relative URIs, malformed percent-encoding). On failure, returns default `MassTransitQueueData` with `"Unknown"` queue name.

## Test plan
- [x] All 54 existing MassTransitHelperTests pass
- [x] MassTransit and MassTransitLegacy wrapper projects build cleanly


jaffinito (approved):
LGTM

PR #3545, opened 2026-04-22
fix: Pass custom vendor name through RecordDatastoreSegment API

### LINKED ISSUE #3536: `RecordDatastoreSegment()` API ignores the `vendor` parameter in metric and segment names
### Description

The `ITransaction.RecordDatastoreSegment()` API accepts a `vendor` string parameter, but the agent ignores it — all custom datastore segments are categorized under `Other` in metric names and the Databases UI, regardless of the value passed.

For example, calling:

```csharp
transaction.RecordDatastoreSegment(
    vendor: "CustomDocStore",
    model: "documents",
    operation: "select", ...);
```

Produces metrics like `Datastore/statement/Other/documents/select` instead of the expected `Datastore/statement/CustomDocStore/documents/select`.

### Expected Behavior

The `vendor` string should flow through to all metric names, segment names, and span attributes. Passing `"CustomDocStore"` should produce:

- Metrics: `Datastore/statement/CustomDocStore/documents/select`, `Datastore/operation/CustomDocStore/select`, `Datastore/CustomDocStore/all`, etc.
- Span attributes: `component: "CustomDocStore"`, `db.system: "customdocstore"`
- Databases UI: segments grouped under `CustomDocStore` rather than `Other`

Known vendors used by the agent's built-in instrumentation (MySQL, Postgres, MongoDB, etc.) should be unaffected.

### Root Cause

Two issues prevent the vendor string from flowing through to metrics:

1. In `TransactionBridgeApi.StartDatastoreSegment()`, the `ParsedSqlStatement` is constructed with `DatastoreVendor.Other` hardcoded, discarding the caller-provided vendor string:

   ```csharp
   var parsedSqlStatement = new ParsedSqlStatement(DatastoreVendor.Other, model, operation);
   ```

2. Downstream code in `DatastoreSegmentData` passes the `DatastoreVendor` enum (always `Other` for custom vendors) to `MetricNames` and `MetricWireModel` methods that convert the enum to the string `"Other"` via `EnumNameCache<DatastoreVendor>.GetName()`. Even if issue 1 is fixed, these code paths still resolve to `"Other"` because the enum value is `DatastoreVendor.Other` for any vendor not in the known enum.

   The affected code paths are all in `DatastoreSegmentData`:
   - `GetTransactionTraceName()` — transaction trace segment names
   - `AddMetricStats()` — all datastore metrics (statement, operation, rollup, instance)
   - `SetSpanTypeSpecificAttributes()` — `component` and `db.system` span attributes

### Suggested Fix

1. Add a string-based constructor to `ParsedSqlStatement` that accepts a vendor name string (setting `DatastoreVendor` to `Other` internally) and exposes it via a public `DatastoreVendorNameString` property. Use this constructor in `TransactionBridgeApi.StartDatastoreSegment()` so the caller-provided vendor string is preserved.
2. Add a `GetVendorName()` method to `DatastoreSegmentData` that returns the custom string when `DatastoreVendor.Other` and a custom name was provided, otherwise returns the enum name
3. Add string-based overloads to `MetricNames` and `MetricWireModel.MetricBuilder` for the datastore metric methods
4. Update `DatastoreSegmentData` to use `GetVendorName()` and the string overloads

This approach avoids changing any existing enum-based code paths used by auto-instrumentation.

### Affected Version

Present since `RecordDatastoreSegment()` was introduced in agent 10.22.0.

P3
M

tippmar-nr:
Closes #3536

## Summary

- Fixes a bug where custom vendor names passed to `RecordDatastoreSegment()` via the public API were discarded — all custom datastore segments incorrectly reported as `Other` instead of using the caller-supplied vendor string (e.g., `DynamoDB`).
- The root cause was a hardcoded `DatastoreVendor.Other` in `TransactionBridgeApi.StartDatastoreSegment()` that replaced the `vendor` parameter.
- The fix threads the vendor string through `ParsedSqlStatement`, `DatastoreSegmentData`, `MetricWireModel`, and `MetricNames` so custom names appear correctly in all metrics, transaction traces, and span events.

## Changes

- `TransactionBridgeApi.cs` — pass `vendor` instead of hardcoded `DatastoreVendor.Other`
- `ParsedSqlStatement` — new string-based constructor and `DatastoreVendorNameString` property; `FromOperation` string overload
- `DatastoreSegmentData` — new `GetVendorName()` method; updated metric generation and span attribute setting to use custom vendor name when present
- `MetricNames` — string-based overloads for all datastore metric name factory methods
- `MetricWireModel` — string-based overloads for all datastore metric builder methods
- Integration test updated with correct `MyVendor` metric names and new span event assertions
- Unit tests added for all new paths

## Test plan

- [ ] Unit tests: `ParsedSqlStatementTests`, `MetricNamesTests`, `DatastoreSegmentDataTests`, `SpanEventMakerTests`, `DatastoreSegmentTransformerTests` all pass
- [ ] Integration test: `RecordDatastoreSegmentTests` — `Metrics()` asserts `Datastore/statement/MyVendor/...` metrics; `SpanEvents()` asserts `component=MyVendor`, `db.system=myvendor`
- [ ] Verified enum-based vendor (non-Other) paths are unaffected
- [ ] Verified `DatastoreVendor.Other` (no custom string) still falls back to `"Other"` / `"other_sql"`


src/Agent/NewRelic/Agent/Core/Metrics/MetricNames.cs
```diff
@@ -493,6 +493,40 @@ public static MetricName GetDatastoreInstance(DatastoreVendor vendor, string hos
         return MetricName.Create(DatastoreInstance, EnumNameCache<DatastoreVendor>.GetName(vendor), host, portPathOrId);
     }
 
+    // String-based overloads for custom vendor names via RecordDatastoreSegment()
+
+    public static MetricName GetDatastoreVendorAll(string vendorName)
+    {
+        return MetricName.Create(Datastore, vendorName, All);
```
nrcventura, line 500: Should the enum-based overloads call these string overloads? It might be nice to have everything go through the same path.

src/Agent/NewRelic/Agent/Core/WireModels/MetricWireModel.cs
```diff
@@ -471,6 +471,61 @@ public static void TryBuildDatastoreInstanceMetric(DatastoreVendor vendor, strin
             txStats.MergeUnscopedStats(proposedName, data);
         }
 
+        // String-based overloads for custom vendor names via RecordDatastoreSegment()
+
+        public static void TryBuildDatastoreStatementMetric(ParsedSqlStatement sqlStatement,
```
nrcventura, line 443: It might be good to have all the overloads call the same method behind the scenes so that the logic will be in one method instead of 2.

nrcventura: Do we need both sets of overloads if the parsedSqlStatement always has the datastore name?

tippmar-nr: Good catches. Refactored to eliminate redundant code.

src/Agent/NewRelic/Agent/Core/Metrics/MetricNames.cs
tippmar-nr: Refactored to get rid of redundant code paths.

nrcventura (approved):

PR #3552, opened 2026-04-28
feat: Update CONTRIBUTING.md to use more modern language

nr-ahemsath:
Fake change just for the purposes of testing the PR title checker workflow.

Initially just testing the existing workflow to make sure it works like we think it will.  I am giving it a bad title at first, and then will edit it to be a valid title.  The title checker workflow should fail and then pass.


nr-ahemsath:
Closing test PR

PR #3553, opened 2026-04-28
Update CONTRIBUTING.md

nr-ahemsath:
Another test PR.  Again, creating with a bad title first and then edit it to be correct; the title check workflow should fail and then pass.  Since this PR targets the branch with the actually desired workflow changes it should hopefully be running that version of the workflow.


nr-ahemsath:
Never mind, the workflow only runs on PRs to main.

PR #3565, opened 2026-05-04
chore: Update license key configuration handling and logging

nr-ahemsath:
This PR brings the agent into compliance with a couple of agent specs, one new and one previously existing:

The new spec: Update ObfuscateLicenseKey (used whenever a license key is logged) to show the first ten characters in plaintext, as long as the key is more than ten characters long (currently agent license keys are 40 chars long).  

The old spec: while agent license keys are currently 40 (printable, ASCII) chars long, the agent should not assume or constrain this to be the case.

Additionally, redact the license key in the MSI installer's log, which was not previously happening.

### DIFF
--- a/src/Agent/MsiInstaller/InstallerActions/CustomActions.cs
+++ b/src/Agent/MsiInstaller/InstallerActions/CustomActions.cs
@@ -388,12 +388,13 @@ public class CustomActions
             return licenseKey;
         }
 
-        if (licenseKey.Length == 40)
+        const int PlaintextPrefixLength = 10;
+        if (licenseKey.Length <= PlaintextPrefixLength)
         {
-            return licenseKey.Substring(0, 10) + new string('*', 30);
+            return new string('*', licenseKey.Length);
         }
 
-        return new string('*', licenseKey.Length);
+        return licenseKey.Substring(0, PlaintextPrefixLength) + new string('*', licenseKey.Length - PlaintextPrefixLength);
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/Agent/MsiInstaller/InstallerActions/CustomActions.cs
```diff
@@ -380,6 +380,21 @@ public static void Log(Session session, string message, params object[] argument
...
+    {
+        if (string.IsNullOrEmpty(licenseKey))
+        {
+            return licenseKey;
+        }
+
+        if (licenseKey.Length == 40)
+        {
+            return licenseKey.Substring(0, 10) + new string('*', 30);
+        }
```
nrcventura: We're not supposed to make assumptions about license key length. This can probably be adjusted to handle other lengths.

src/Agent/NewRelic/Agent/Core/Utilities/Strings.cs
```diff
@@ -194,11 +194,11 @@ public static string ObfuscateLicenseKeyInAuditLog(string text, string licenseKe
 
     public static string ObfuscateLicenseKey(string licenseKey)
     {
-        // We can log up to 8 characters of a 40-character license key, the rest must be obfuscated
+        // We can log up to 10 characters of a 40-character license key, the rest must be obfuscated
         // For our agent, the license key should always be 40 characters. For safety, if it isn't, just obfuscate the whole thing
         if (licenseKey.Length == 40)
         {
-            return licenseKey.Substring(0, 8) + new string('*', 32);
+            return licenseKey.Substring(0, 10) + new string('*', 30);
```
nrcventura: While we are making this change we should probably fix our compliance issue that assumes that the key length is 40.

nrcventura (approved):

PR #3567, opened 2026-05-05
ci: Fix Windows profiler build error caused by missing ATL components

nr-ahemsath:
The `windows-latest` GHA runner image recently switched to new Visual Studio tooling and no longer includes the ATL components needed by the profiler by default.  This PR adds a step to the Windows profiler build job to explicitly install the ATL components needed.

Note: the initial list of ATL components being installed is probably overkill; I'll try to get them down to the minimum required set before marking this ready for review.


nr-ahemsath:
This does not appear to be working.  For the time being we've fixed the problem by pinning the profiler build to `windows-2022`.

PR #3576, opened 2026-05-12
chore: Vendor CoreCLR profiling-API headers and eliminate upstream clone

tippmar-nr:
## Summary

Previously the Windows and Linux profiler builds cloned `dotnet/coreclr` at `release/3.1` on every build to resolve the CLR profiling-API headers. That repository was archived after .NET Core 3.1 reached EOL in December 2022; the clone adds network and supply-chain fragility for a header set that has not changed since.

This PR vendors the exact transitive closure of headers (and `corhlpr.cpp`, which the Linux build compiles into the profiler) into `src/Agent/NewRelic/Profiler/externals/coreclr-headers/`, preserving upstream path layout so `#include` directives continue to resolve without source changes. Provenance is pinned to upstream commit [`1b04187`](https://github.com/dotnet/coreclr/commit/1b04187394cd1de5b316c40a3ad9d9189f4cb5b0) — the final commit on `release/3.1` before archival.

This is step P1.1 of a broader profiler modernization effort. Later steps (Windows build-tooling preparation, Linux toolchain modernization, packaging hygiene) are out of scope here and will land in separate PRs.

## Changes

| Area | File | Change |
|---|---|---|
| Build (Windows) | `Profiler/Profiler.vcxproj` | Drop `CheckCoreClrPath` target, `CoreClrRepo` / `CoreClrBranch` / `CloneCoreClr` / `CORECLR_PATH` properties, the `Choose/When/Otherwise` block, and the `BuildDependsOn` wiring. Include paths now point at `$(SolutionDir)externals\coreclr-headers`. |
| Build (Linux) | `CMakeLists.txt` | Replace the `CORECLR_PATH` cache variable with a `CORECLR_HEADERS` path pointing at the vendored tree. |
| Build (Linux) | `linux/Dockerfile` | Drop the multistage `ClrRepoCloner` builder. |
| Build (Linux) | `linux/Arm64Dockerfile`, `linux/Dockerfile.new` | Drop the coreclr clone. |
| Build (Linux) | `linux/build_profiler.sh` | Drop `-DCORECLR_PATH=` from the cmake invocation. |
| CI | `.github/workflows/build_profiler.yml` | Drop the coreclr clone from the arm64 inline install block. |
| Licensing | `licenses/THIRD_PARTY_NOTICES.txt` | Add a CoreCLR (MIT) entry with the upstream commit SHA and vendored-location pointer. |
| Vendored | `src/Agent/NewRelic/Profiler/externals/coreclr-headers/` | 102 headers + `corhlpr.cpp` (~3.1 MB), upstream layout preserved, with `LICENSE.TXT` (MIT) and a `README.md` documenting provenance and refresh procedure. |

**Intentionally untouched:** `src/Agent/NewRelic/Profiler/linux/DebugDockerfile` still clones and builds coreclr — for SOS/lldb debug tooling, a different use case from header sourcing. Out of scope for this PR.

## Verification

The plan for P1.1 specified byte-for-byte binary parity: because this is a pure header-sourcing change (the vendored content is byte-identical to upstream), the rebuilt binaries should produce identical code.

I rebuilt the profiler on all four targets and compared against the pre-change artifacts locally. Results:

| Platform | Exported symbols | Undefined symbols | Disasm diff | Max GLIBC | DT_NEEDED / exports |
|---|---|---|---|---|---|
| **Windows x64** (VS 2026, v143) | identical | identical | **0 lines** | — | identical |
| **Windows x86** (VS 2026, v143) | identical | identical | **0 lines** | — | identical |
| **Linux x64** (Ubuntu 14.04 / clang-3.9, via existing Dockerfile) | 4,551, identical | 155, identical | **0 lines** | 2.14 (unchanged) | `libm.so.6`, `libgcc_s.so.1`, `libpthread.so.0`, `libc.so.6` — identical |
| **Linux arm64** (Ubuntu 18.04 / clang-3.9, via existing Arm64Dockerfile + QEMU) | 4,183, identical | 198, identical | **0 lines** | 2.17 (unchanged) | same + `ld-linux-aarch64.so.1` — identical |

`objdump -d --no-show-raw-insn` on Linux, `dumpbin /disasm` on Windows. Disassembly diffs were zero lines after filtering out embedded filename references in debug-info sections. The only binary-level deltas are:

- **Linux:** +218 bytes x64, +224 bytes arm64 — entirely in `.debug_*` sections. The absolute source path of `corhlpr.cpp.o` now reads `/profiler/externals/coreclr-headers/src/inc/` instead of `/root/git/coreclr/src/inc/`, which changes the string length embedded in debug info.
- **Windows:** identical file sizes (1,161,216 bytes x64; 1,037,824 bytes x86).

### Alpine runtime smoke test

The rebuilt Linux x64 `libNewRelicProfiler.so` was loaded into `mcr.microsoft.com/dotnet/sdk:10.0-alpine` via the standard `CORECLR_*` environment variables. The profiler log emits `Profiler initialized`, `ICorProfilerInfo10 available`, and identifies **372 instrumentation points** — identical to the pre-change binary.

### Native unit tests

Left for CI to exercise via the existing `build_profiler.yml` workflow: `CommonTest`, `ConfigurationTest`, `LoggingTest`, `MethodRewriterTest`, `SignatureParserTest`, `SicilyTest` (run via `vstest.console.exe` under `OpenCppCoverage`).

## Risk

Low. This is a sourcing change with no functional intent. Binary parity across all four build targets (zero-line disassembly diff, identical symbol tables, identical DT_NEEDED, unchanged max GLIBC baseline) is the strongest evidence available that nothing about the emitted binary has changed. The Alpine smoke test confirms runtime behavior on the musl-based distro that is most sensitive to header drift.

## Test plan

- [x] CI: Windows x64 + x86 build, native unit tests pass (`build_profiler.yml` → `build-windows-profiler` job)
- [x] CI: Linux x64 build (`build-linux-x64-profiler` job) — should no longer clone coreclr
- [x] CI: Linux arm64 build (`build-linux-arm64-profiler` job) — same
- [x] Code coverage upload succeeds (no test-side changes in this PR)
- [x] Confirm the produced Linux `libNewRelicProfiler.so` still initializes under `dotnet` on Alpine via an existing `ContainerIntegrationTests.AlpineX64ContainerTestFixture` run

## Out of scope / follow-ups

- `src/Agent/NewRelic/Profiler/linux/DebugDockerfile` — still clones and builds coreclr for SOS/lldb debugging. Retaining its clone is intentional; reworking that file is unrelated to header vendoring.
- `Dockerfile.new` was retained (with its clone stripped) because the plan calls for a later consolidation step that will delete it outright; keeping it in sync now avoids merge churn.
- Any refresh of the vendored headers must re-verify binary parity; refresh procedure is documented in `src/Agent/NewRelic/Profiler/externals/coreclr-headers/README.md`.


src/Agent/NewRelic/Profiler/externals/coreclr-headers/README.md
```diff
@@ -0,0 +1,61 @@
...
+## How to refresh
+
+If a change in the profiler later requires a header that is not in the vendored set, or a bug fix lands in a still-maintained CoreCLR branch that we need, refresh by re-running the audit against a fresh checkout of `dotnet/coreclr`:
+
+```powershell
+git clone https://github.com/dotnet/coreclr.git C:\source\repos\coreclr
+cd C:\source\repos\coreclr
+git checkout release/3.1    # or whatever commit is being tracked
+
+# Re-run the audit Python logic from the PR that introduced vendoring
```
nr-ahemsath: I'm not sure what Python logic this is referring to.  It doesn't seem included in any of the files in this PR.

nr-ahemsath (commented):
I just have the one comment about "Python audit logic" wording in the new README; otherwise this looks good.

src/Agent/NewRelic/Profiler/externals/coreclr-headers/README.md
tippmar-nr: Good catch. Replaced the "How to refresh" section entirely: the archived `coreclr` repo is a dead end for any real future need, so the section now points to `dotnet/runtime` (`src/coreclr/inc/` and `src/coreclr/pal/inc/`) as the current source of truth.

nr-ahemsath (approved):

chynesNR (approved):
Seems reasonable

jaffinito (approved):

PR #3579, opened 2026-05-13
chore: Parametrize PlatformToolset across profiler vcxproj files

tippmar-nr:
## Summary

- Adds `src/Agent/NewRelic/Profiler/Directory.Build.props` with `<NativeToolset Condition="'$(NativeToolset)'==''">v143</NativeToolset>`
- Replaces the hardcoded `<PlatformToolset>v143</PlatformToolset>` literal in all 16 native profiler vcxproj files with `<PlatformToolset>$(NativeToolset)</PlatformToolset>`
- Part of the profiler modernization plan (P1.2) — preparation for the eventual migration off `windows-2022` CI runners

## Motivation

The `windows-2022` GitHub Actions runner image will eventually be retired. The immediate blocker for migrating to `windows-2025-vs2026` is that the VS2026 image carries v145 as its primary ATL toolset with only v143 compilers (no paired ATL) in a sidecar. This PR does **not** perform the cutover — it is the prerequisite step that enables switching toolsets without touching any project files, so the cutover can be done as a single CI-workflow change and its binary output independently verified.

## What changed

| File | Change |
|------|--------|
| `Directory.Build.props` (new) | Defines `NativeToolset=v143` as the default MSBuild property |
| 16 × `*.vcxproj` | `<PlatformToolset>v143</PlatformToolset>` → `<PlatformToolset>$(NativeToolset)</PlatformToolset>` (4 configs each = 64 replacements) |

## Binary parity verification

Default build (`NativeToolset=v143`, no override) was verified to produce functionally identical output on both Windows targets.

**Method:** Built before and after the change using `build.ps1 -Platform windows`; compared PE sections directly with Python.

| Section | x64 | x86 |
|---------|-----|-----|
| `.text` (all code) | **IDENTICAL** (836,608 bytes) | **IDENTICAL** (759,808 bytes) |
| `.data` | **IDENTICAL** (33,280 bytes) | **IDENTICAL** (28,672 bytes) |
| `.rdata` | **IDENTICAL** (246,272 bytes) | **IDENTICAL** (200,704 bytes) |

The only raw-byte differences in the DLL files are:
- 4 × `TimeDateStamp` fields in `IMAGE_DEBUG_DIRECTORY` entries (linker-stamped on every build)
- 1 × PDB age counter in the CodeView record (increments on every link)

Both are expected on any rebuild regardless of source changes and do not indicate any difference in code or data.

## How to use the override (future cutover)

```
MSBuild.exe -restore -m -p:Platform=x64 -p:Configuration=Release -p:NativeToolset=v145 NewRelic.Profiler.sln
```

The actual v143→v145 cutover will be a separate PR that:
1. Verifies ATL availability on `windows-2025-vs2026`
2. Updates the CI runner image
3. Passes `/p:NativeToolset=v145` (or sets it as the new default in `Directory.Build.props`)
4. Performs the full binary-diff verification gate against the v143 baseline

## Test plan

- [x] Both x64 and x86 Release builds succeed (0 errors, 0 warnings) with default `NativeToolset=v143`
- [x] `.text`, `.data`, `.rdata` sections byte-identical to pre-change build
- [x] CI: Windows profiler build + unit tests pass
- [x] CI: Linux x64 and arm64 profiler builds pass (no changes to Linux build path)

### DIFF
--- a/src/Agent/NewRelic/Profiler/Common/Common.vcxproj
+++ b/src/Agent/NewRelic/Profiler/Common/Common.vcxproj
@@ -27,26 +27,26 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>$(NativeToolset)</PlatformToolset>
     <CharacterSet>MultiByte</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>$(NativeToolset)</PlatformToolset>
     <CharacterSet>MultiByte</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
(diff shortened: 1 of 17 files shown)


nr-ahemsath (approved):
Very straightforward, looks good to me. 👍

PR #3591, opened 2026-05-15
chore: replace deprecated apt-key add with keyring-file pattern

tippmar-nr:
## Summary

Migrates the LLVM apt repo from the deprecated `apt-key add` pattern to the modern `signed-by=` keyring-file pattern in all three places it appears in the profiler Linux build.

## Why

`apt-key` is deprecated and emits a warning on every invocation:

```
Warning: apt-key output should not be parsed (stdout is not a terminal)
```

apt will eventually drop `apt-key` entirely, at which point the LLVM repo line breaks and the Linux profiler builds fail. Migrating now removes a future breakage and silences the warning.

## What changed

The same migration applied identically in three places:

- `linux/Dockerfile` (x64 build, Ubuntu 14.04)
- `linux/Arm64Dockerfile` (orphaned but kept aligned for future CI consolidation)
- `.github/workflows/build_profiler.yml` arm64 inline install block (active code path today)

Old:
```
echo "deb https://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.9 main" | tee /etc/apt/sources.list.d/llvm.list
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
```

New:
```
install -d -m 0755 /etc/apt/keyrings
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /etc/apt/keyrings/llvm.gpg
echo "deb [signed-by=/etc/apt/keyrings/llvm.gpg] https://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.9 main" | tee /etc/apt/sources.list.d/llvm.list
```

`gnupg` (which provides `gpg --dearmor`) is already pulled in transitively by `software-properties-common` in all three places, so no new package was needed.

## Test plan

- [x] Built both Dockerfiles locally via Docker Desktop + buildx (linux/arm64 via QEMU). Verified the keyring file is created at `/etc/apt/keyrings/llvm.gpg`, the apt source line correctly references it via `signed-by=`, and `apt-key list` no longer contains the LLVM key.
- [x] CI fully green on this branch (run 25927307076): Windows, Linux x64, and Linux ARM64 all pass — meaning the LLVM apt repo resolves correctly and `clang-3.9` / `lldb-3.9` install successfully via the new keyring pattern.

### DIFF
--- a/src/Agent/NewRelic/Profiler/linux/Dockerfile
+++ b/src/Agent/NewRelic/Profiler/linux/Dockerfile
@@ -15,7 +15,7 @@ RUN sed -i 's/mozilla\/DST_Root_CA_X3.crt/!mozilla\/DST_Root_CA_X3.crt/g' /etc/c
 RUN update-ca-certificates
 
 RUN install -d -m 0755 /etc/apt/keyrings
-RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /etc/apt/keyrings/llvm.gpg
+RUN wget --no-cache --no-cookies -O - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /etc/apt/keyrings/llvm.gpg
 RUN echo "deb [signed-by=/etc/apt/keyrings/llvm.gpg] https://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.9 main" | tee /etc/apt/sources.list.d/llvm.list
 
 # Install the build tools that the profiler requires
--- a/src/Agent/NewRelic/Profiler/linux/Dockerfile
+++ b/src/Agent/NewRelic/Profiler/linux/Dockerfile
@@ -14,8 +14,9 @@ RUN apt-get update && apt-get install -y \
 RUN sed -i 's/mozilla\/DST_Root_CA_X3.crt/!mozilla\/DST_Root_CA_X3.crt/g' /etc/ca-certificates.conf
 RUN update-ca-certificates
 
-RUN echo "deb https://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.9 main" | tee /etc/apt/sources.list.d/llvm.list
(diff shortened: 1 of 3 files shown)


nr-ahemsath (dismissed):
Looks good.  Before looking at this I was concerned that this would somehow impact the signing of our own shipped code, but this is only for configuring the containers that build the Linux profiler binaries and should have zero impact on our own code signing.

nrcventura (dismissed):

nrcventura (dismissed):

src/Agent/NewRelic/Profiler/linux/Dockerfile
```diff
@@ -14,8 +14,9 @@ RUN apt-get update && apt-get install -y \
 RUN sed -i 's/mozilla\/DST_Root_CA_X3.crt/!mozilla\/DST_Root_CA_X3.crt/g' /etc/ca-certificates.conf
 RUN update-ca-certificates
 
-RUN echo "deb https://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.9 main" | tee /etc/apt/sources.list.d/llvm.list
-RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+RUN install -d -m 0755 /etc/apt/keyrings
+RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /etc/apt/keyrings/llvm.gpg
```
nr-ahemsath: I'm not sure if it matters, but this usage of `wget` doesn't match the others - it's missing the `--no-cache --no-cookies` bit.  Just for consistency it would be nice to have them all identical.

tippmar-nr: Good catch. I'll take a look.

tippmar-nr: Fixed in 5b2c34913 — added `--no-cache --no-cookies` to the x64 Dockerfile wget so all three places match.

nr-ahemsath (approved):

PR #3593, opened 2026-05-15
chore: consolidate arm64 profiler build onto docker compose + Arm64Dockerfile

tippmar-nr:
## Summary

Replaces the `uraimo/run-on-arch-action` step in the arm64 profiler build job with a native arm64 GitHub-hosted runner (`ubuntu-24.04-arm`) running `docker compose run build_arm64`. The `Arm64Dockerfile` becomes the single source of truth for the arm64 build environment, matching the pattern already used by the x64 build.

## Why

Two configurations described the arm64 build environment:

1. `src/Agent/NewRelic/Profiler/linux/Arm64Dockerfile` — referenced by `docker-compose.yml`'s `build_arm64` service, but **not used by CI**.
2. `.github/workflows/build_profiler.yml`'s arm64 inline `install:` block under `uraimo/run-on-arch-action` — the actual code path CI ran.

The two were kept in sync only by manual diligence — touching one and not the other meant silent drift between the orphaned Dockerfile and the live config. Recent PRs (#3590 cmake, #3591 apt-key) had to make matching edits in both places to keep them aligned.

This PR collapses the two onto `Arm64Dockerfile` and runs the build on a native arm64 runner, the same overall pattern that the OpenTelemetry .NET instrumentation project uses for its multi-architecture native builds.

## What changed

In `.github/workflows/build_profiler.yml`'s `build-linux-arm64-profiler` job:

- Removed the `uraimo/run-on-arch-action` step and its 18-line inline `install:` block (the duplicate of `Arm64Dockerfile`).
- Switched the runner from `ubuntu-22.04` to `ubuntu-24.04-arm` (native arm64 runner).
- Added a `docker compose build build_arm64 && docker compose run build_arm64` step. This matches exactly the pattern used by the existing x64 job two jobs above.
- Set `CORECLR_NEW_RELIC_HOME` env (mirror of x64 pattern; defensive against `docker-compose.yml`'s `build_debug` service `:-` reference).
- Removed `permissions: packages: write` (was needed for `run-on-arch-action`'s ghcr image cache; not needed here).

The `build_arm64` service in `docker-compose.yml` already existed (`platform: linux/arm64/v8`, `dockerfile: Arm64Dockerfile`) — this PR just wires the CI job to use it.

The first iteration of this PR used `docker/setup-qemu-action` to register binfmt for aarch64 emulation on an x64 runner, then proved binary-shape parity. After that landed, review feedback from @nrcventura suggested using a native arm runner instead. That switch dropped CI time from 9m18s to 1m39s (5.6× speedup) while remaining byte-identical to the QEMU output (see test plan).

## Why this matters for future dual-build (glibc + musl)

Today, adding a new arm64 build variant means writing another inline-install block in the workflow YAML. After this PR, adding a variant means: write a new Dockerfile, add a service to `docker-compose.yml`, add a CI job that calls `docker compose run <service>`. That mirrors how OTel's `docker/` directory organizes their per-target builds (alpine, debian, debian-arm64, centos-stream9, ubuntu1604).

## Test plan

- [x] CI green on this branch — Windows, Linux x64, Linux ARM64 all pass.
- [x] **Binary-shape parity check (the actual gate for this PR).** Downloaded `libNewRelicProfiler.so` from this run and from the most recent post-#3590 main run and compared:

  | Property | Baseline (run-on-arch-action / QEMU on ubuntu-22.04) | New (docker compose on ubuntu-24.04-arm native) |
  |---|---|---|
  | SHA-256 | `7fa2827fbfd6316530c47871ccda8db48d32ef38d252783d2ccb357954eff4a4` | identical |
  | DT_NEEDED set | `libm.so.6, libgcc_s.so.1, libpthread.so.0, libc.so.6, ld-linux-aarch64.so.1` | identical |
  | Max GLIBC referenced | 2.17 | identical |
  | libc++/libstdc++ runtime dep | none | none |
  | Exported symbol count | 4183 | identical |

  The arm64 binary is **bit-for-bit identical** between the old run-on-arch-action build and the new native-arm-runner build. The plan's gate was parity on the four binary properties; the actual result is byte-identity, which is the strongest possible parity outcome.

- [x] **CI duration: 9m18s (QEMU baseline) → 1m39s (native arm runner), a 5.6× speedup on the arm64 job.**

### DIFF
--- a/.github/workflows/build_profiler.yml
+++ b/.github/workflows/build_profiler.yml
@@ -213,10 +213,10 @@ jobs:
     if: ${{ needs.check-modified-files.outputs.source-files-changed == 'true' || github.event_name == 'workflow_dispatch' }}
     permissions:
       contents: read # for actions/checkout to fetch code
-      packages: write # for uraimo/run-on-arch-action to cache docker images
 
     env:
       profiler_path: ${{ github.workspace }}/src/Agent/NewRelic/Profiler
+      CORECLR_NEW_RELIC_HOME: ${{ github.workspace }}/src/Agent/NewRelic/newrelichome_arm64_coreclr_linux # not used but required by Profiler/docker-compose.yml
 
     steps:
       - name: Harden Runner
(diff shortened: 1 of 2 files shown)


.github/workflows/build_profiler.yml
```diff
@@ -236,41 +236,17 @@ jobs:
           rm -rf ${{ github.workspace }}/src/Agent/_profilerBuild/x86-Release || true
         shell: bash
 
-      - uses: uraimo/run-on-arch-action@f9b26e3a1a408d5fd530d20c17b9f3f4428ff8d9 # v3.1.0
-        name: Run commands
-        id: runcmd
+      - name: Set up QEMU for arm64 emulation
+        uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4.0.0
```
nrcventura: Instead of using QEMU, could we use an arm runner instead?

nrcventura (dismissed):

.github/workflows/build_profiler.yml
tippmar-nr: Good call — switched to `ubuntu-24.04-arm` (native arm64 runner, GA + free for public repos since Jan 2025) and dropped the `setup-qemu-action` step. Picked 24.04 over 22.04 since the x64 job's runner is itself approaching its sunset window; aligns the next x64 bump.

Verified the change is binary-shape-neutral. Native runner output is **byte-identical** to the prior QEMU build:

| Property | QEMU baseline | Native `ubuntu-24.04-arm` |
|---|---|---|
| SHA-256 | `7fa2827fbfd6316530c47871ccda8db48d32ef38d252783d2ccb357954eff4a4` | identical |
| DT_NEEDED | libm, libgcc_s, libpthread, libc, ld-linux-aarch64 | identical |
| Max GLIBC referenced | 2.17 | identical |
| libc++/libstdc++ runtime dep | none | none |
| Exported symbol count | 4183 | identical |

CI duration: 9m18s (QEMU) → 1m39s (native), a 5.6× speedup on the arm64 job.

nrcventura (approved):

Issue #817, opened 2021-11-29
Fix existing NSB 5 integration tests


jaffinito:
* Test works locally with locally installed NServiceBus.
* Fails in GHA
* Unable to test in GHA until we land the work to re-enable unbounded tests in GHA (can run tests in different repo as needed.)

nr-ahemsath:
Also see issue #361 for context and history of my (failed) attempts to fix the problem by making this test use console applications only.

jaffinito:
I have checked out the latest runs and the problem no longer occurs.
Closing out the issue.

PR #3608, opened 2026-06-01
feat(kafka): add MessageBroker/Kafka/Cluster/{id}/Produce|Consume/{topic} metrics

shashank-reddy-nr:
## What
Adds per-cluster, per-topic metrics for Confluent.Kafka.

Both produce and consume use `agent.RecordCountMetric()` (internal API, no
`Custom/` prefix) consistently — same as the existing `Nodes` metric:
```
  MessageBroker/Kafka/Cluster/{id}/Produce/{topic}
  MessageBroker/Kafka/Cluster/{id}/Consume/{topic}
```

## KafkaBuilderWrapper.cs
Captures the bootstrap-servers string and calls `KafkaHelper.ScheduleClusterIdFetch()`
with the built producer/consumer instance itself — no longer captures/passes the full
config dictionary (see optimization below).

## KafkaHelper.cs (optimized)
- `ScheduleClusterIdFetch()`: reuses the producer/consumer's own already-connected
  native handle via `((IClient)instance).Handle` — `Producer<K,V>`/`Consumer<K,V>` both
  implement the non-generic `IClient` interface directly, no reflection needed. Builds
  the admin client via `DependentAdminClientBuilder(handle)` (Confluent.Kafka 2.3+)
  instead of a standalone `AdminClientBuilder`, so no second connection is opened just
  to resolve the cluster ID. Calls `admin.DescribeClusterAsync()` once and caches the
  result forever — the cluster ID is a stable identifier for the lifetime of the
  cluster, so there's no TTL or retry-backoff loop; a fresh producer/consumer built
  against the same bootstrap-servers string naturally retries if the first attempt
  didn't resolve.
- `RecordKafkaNodeMetrics()`: uses agent.RecordCountMetric() (internal API,
  no Custom/ prefix) — preserves existing MessageBroker/Kafka/Nodes/... names.
- Generic fix: the per-instance bootstrap-servers lookup used a one-field
  wrapper class purely to satisfy `ConditionalWeakTable`'s reference-type
  constraint; since `string` already satisfies that constraint, the wrapper
  was dropped in favor of storing the string directly.
- Bug fix: the cluster-id fetch runs on a deferred background task, so the
  producer/consumer it was scheduled for can be disposed before that task
  runs. Building a dependent admin client against an already-released native
  handle isn't a catchable managed exception; added a handle-validity check
  immediately before use to close that window.

## KafkaConsumerWrapper.cs
- Cluster metric: `agent.RecordCountMetric(name)`.
- kafka.cluster.id custom attribute on transaction.
- Preserves DT header acceptance, producer identity extraction, size metrics.

## KafkaProducerWrapper.cs
- Auto-creates background transaction for sync Produce() only (not ProduceAsync)
  — prevents transaction leak when async produce throws.
- Cluster metric uses same API as consumer for consistency.

## Metric name format + why no aggregation rework was needed

Per [agent-specs #817](https://source.datanerd.us/agents/agent-specs/pull/817)
review, the metric name now matches the `Nodes` shape (verb before topic, not
after): `Cluster/{id}/Produce/{topic}`, not `Cluster/{id}/Topic/{topic}/Produce`.

Separately, the spec no longer requires per-message emission — these metrics
exist to power entity relationships, not to report exact throughput, so counts
just need to be aggregated over the same interval as `Nodes`. No code rework
was needed for that: `RecordCountMetric` already merges same-named metrics
into an in-memory queue (`MetricAggregator.Collect`) and only flushes once per
60s harvest cycle — the exact same mechanism the existing `Nodes` metric already
uses. Calling it once per message only costs a string allocation + dictionary
merge on the hot path; the data New Relic receives was already one aggregated
count per harvest interval, not one per message.

## Tests (integration only — per CLAUDE.md, wrapper projects have no unit tests)
Validated end-to-end against a live Kafka cluster: the consumer wrapper reported
`MessageBroker/Kafka/Cluster/.../Consume` metrics correctly (pre-rename; format
updated since — see note above).

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


tippmar-nr:
@shashank-reddy-nr We just merged the Kafka internal metrics PR to main; there are going to be some significant conflicts in your branch that need to be resolved. If you need help, reach out and we'll see what we can do.

shashank-reddy-nr:
Hey @tippmar-nr. Sure, I will try to fix the conflicts. If I need help, I will let you know. I will also try to refactor the code changes I generated and update the PR description.

PR #3610, opened 2026-06-01
ci: Fix markdown check failure

nr-ahemsath:
Dependabot updated the version of Linkspector to a newer version that doesn't include its own Chrome dependency, and so we need to add a step to install Chrome for node with `npx puppeteer`.  Note that the specific version of Chrome being installed is the one that the pinned version of Linkspector requires, so if the Linkspector version gets bumped again, I suspect that the Chrome version will need to be updated as well.


tippmar-nr:
Closed in favor of #3611

PR #3627, opened 2026-06-09
chore: Add compatibility doc generator and generated doc

tippmar-nr:
> **TL;DR**: Adds `build/CompatibilityDocs`, a tool that generates the customer-facing `docs/net-agent-compatibility.md` from curated YAML (`compatibility.yaml`), deriving each package's latest tested version from the integration-test `.csproj` files. Seeds it with the current library list plus 6 previously-undocumented libraries (Sitecore.Logging, OpenRasta, OpenSearch, gRPC, Microsoft.Extensions.AI, AWSSDK.Lambda). Wires the generator into the Dotty workflow so version bumps and the doc update land in the same PR (toggle: `DOTTY_GENERATE_COMPAT_DOC`, default on). **Seeking feedback on the generated doc's content/format.**

Adds `build/CompatibilityDocs`, a net10.0 tool that generates `docs/net-agent-compatibility.md` â€” the customer-facing list of libraries the .NET agent instruments. Curated data lives in `compatibility.yaml` (validated by JSON schema + `SchemaValidator`); the latest tested version of each package is derived from the integration-test `.csproj` files, while minimums are curated. The generated doc is included for review. Output is deterministic - regenerate with `dotnet run --project build/CompatibilityDocs`.

The doc renders separate **.NET Core** and **.NET Framework** sections, and the data model lets a library differ between the two where reality does:

- **Per-platform version ranges** - `minVersion`/`latestVersion` accept either a scalar (all tabs) or a `{core, framework}` map.
- **Per-tab and per-package `minAgentVersion`** - also scalar-or-map, and settable on an individual package to override the library value for that row (e.g. the ODBC NuGet package requires a newer agent than the built-in `System.Data` assembly).
- **Per-note `tabs:` scoping** - a note can be restricted to a subset of the tabs its library/package appears in.
- **Notes-only libraries** - libraries instrumented generically with no NuGet package or version (e.g. IBM DB2) render as a single table row with dashes plus their notes.
- **`requiresHybridAgent` note type** - flags libraries/versions supported only when [OpenTelemetry API support](https://docs.newrelic.com/docs/apm/agents/manage-apm-agents/opentelemetry-api-support/) is enabled — either an entire library (gRPC, Microsoft.Extensions.AI) or versions above a threshold (Elasticsearch > 8.15.10, RabbitMQ > 6.8.1).

`build/CompatibilityDocs/README.md` documents the schema, the version/min-agent shapes, note scoping, and the four library shapes.

## Newly documented libraries

A gap audit found six instrumented-and-tested libraries missing from the doc; all are now included:

- **Sitecore.Logging** (logging, framework) - repackaged log4net, instrumented via the log4net instrumentation.
- **OpenRasta** (app frameworks, framework) - `openrasta-core`.
- **OpenSearch** (datastores, core + framework) - `OpenSearch.Client`.
- **gRPC (grpc-dotnet)** (external calls, core) - instrumented as external calls via the OpenTelemetry bridge; requires hybrid agent.
- **Microsoft.Extensions.AI** (LLM, core) - instrumented via the OpenTelemetry bridge; requires hybrid agent.
- **AWSSDK.Lambda** (external calls, core + framework) - instruments *invoking* a Lambda from the AWS SDK (distinct from running as a Lambda; links to the serverless docs).

Plus a note on the existing StackExchange.Redis entry covering the strong-named `StackExchange.Redis.StrongName` package.

## Keeping the doc in sync via Dotty

The [Dotty workflow](.github/workflows/dotty.yml) periodically bumps tested NuGet versions in the integration-test `.csproj` files and opens a PR. Since the doc's "latest tested version" column is derived from those same `.csproj` files, Dotty now **regenerates the compatibility doc and includes it in the same PR**, so the doc never drifts from the versions under test.

- Controlled by `DOTTY_GENERATE_COMPAT_DOC` (default **true**) and a matching `generateCompatDoc` `workflow_dispatch` input. When disabled, the doc is neither regenerated nor added to the PR.
- The generator runs only when Dotty is actually creating a PR (real updates, token present, not test mode), and a generation failure is non-fatal - the version-bump PR still proceeds without the doc.
**Seeking feedback on the generated doc content/format.**


jaffinito (approved):
Very nice!

PR #3629, opened 2026-06-10
chore(otel-bridge): Enable gzip compression for OTLP metric exports (NR-574461)

tippmar-nr:
## Summary

Enables gzip compression on OTLP metric export requests (NR-574461). The Hybrid Agent spec requires OTLP metric exports to be compressed (REQ-022, MUST).

## Changes

- Set `Compression = OtlpExportCompression.GZip` on the OTLP exporter options in `OtlpExporterConfigurationService`.

The SDK compresses the payload before it enters the HttpClient pipeline, so the existing audit and retry handlers still observe the compressed content, matching prior behavior.

## Testing

- FullAgent.sln builds clean.
- Existing OTLP metric integration tests provide end-to-end coverage; the mock OTLP collector auto-decompresses gzip, exercising the compression path.

PR #3631, opened 2026-06-11
chore(otel-bridge): Enforce NR OTLP data collection limits (NR-574462)

tippmar-nr:
## Summary

Implements Phases 1-3 of NR-574462. Prunes and truncates metric attributes at NR OTLP ingest limits (REQ-015 MUST / REQ-016 SHOULD) and adds a payload size guard to prevent 413 responses.

**Phase 1 & 2 - Attribute limits (MeterBridgingService)**

| Limit | Value |
|---|---|
| Max data point attributes | 64 |
| Max attribute key length | 255 chars |
| Max string attribute value length | 4,095 chars |
| Max array attribute entries | 64 |
| Max byte array attribute size | 128,000 bytes |
| Max scope (meter) attributes | 8 |

Enforcement is split across two code paths:
- `FilterValidTags` - regular instruments (`ReadOnlySpan<KVP>` measurement tags)
- `ApplyTagLimits` - observable instruments (`IEnumerable<KVP>` measurements)
- `ApplyScopeTagLimits` - meter-level tags applied in `CreateBridgedMeter`

Resource attribute enforcement was evaluated; current usage is ~5 known attributes, well within the 64-attribute limit, so no pruning is needed.

**Phase 3 - Payload size guard (OtlpAuditHandler)**

Drops export cycles whose serialized payload exceeds 1 MB before they reach the wire, fires `ReportSupportabilityPayloadsDroppeDueToMaxPayloadSizeLimit`, and returns a synthetic 413 to the SDK (which retries on the next interval).

`OtlpAuditHandler` sits inner to `GzipCompressionHandler` (PR #3629). Once that PR merges, `bytesSent` will reflect the compressed wire payload - the limit NR enforces - with no changes needed here.

## Testing

- 11 new unit tests in `MeterBridgingServiceTagLimitTests` covering all attribute limits across both measurement paths
- 2 new unit tests in `OtlpAuditHandlerAdditionalTests` covering oversized-payload drop and at-limit passthrough
- All existing `MeterBridgingService` and `OtlpAuditHandler` tests pass

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/OtlpAuditHandler.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
+using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


jaffinito (approved):
Looks fine, but I was confused by the use of 1,000,000 for the max bytes value instead of 1,048,576.  I guess I always think in mibibytes, but write megabytes.

PR #3655, opened 2026-06-24
chore: Log New Relic collector response headers at debug level.

### LINKED ISSUE #112: Logging response headers from New Relic endpoint to audit logs.
### Feature Description
When we send data up to new relic, the response headers that we get back from New Relic have been useful for debugging issues like missing data due to 503s, it's useful to know if the issues were isolated to a particular geo / colo. Additionally, if the failure is cloudflare related, cloudflare is going to need a ray ID and request ID to debug these issues.

jaffinito:
## Description

**What**
Log the collector's HTTP response headers to the agent debug log, on both the success and error paths.

**Why** (NR-132421)
Only the response body was logged. Headers (cloudflare, request id, geo markers) are needed by support to diagnose 503s and Cloudflare failures.

**How**
- Added `IHttpResponse.GetHeaders()`; implemented on `HttpResponse` (.NET Core) and `WebRequestClientResponse` (.NET Framework).
- Exposed `Headers` on `IHttpResponseMessageWrapper` for the Core mock seam.
- Added `HttpResponseHeaderFormatter` (pure helper) with `IEnumerable<KeyValuePair<...>>` and `#if NETFRAMEWORK WebHeaderCollection` overloads. Output: `cf-ray=[abc-DFW]; x-request-id=[id-1]`, or `(none)`.
- `HttpCollectorWire.SendData` logs headers at debug on success and error paths, guarded by `Log.IsDebugEnabled` (GetHeaders builds its string eagerly).
- Has some protection to prevent headers containing "auth", "token", or "key" from outputting their values to the logs.

**Tests**
- `HttpResponseHeaderFormatterTests`, `HttpResponseTests`, `WebRequestClientResponseTests`, `HttpCollectorWireTests` (guard verified both ways).
- Pass on `net10.0` and `net481`.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

Fixes #112

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
+++ b/src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
@@ -69,6 +69,9 @@ public class HttpCollectorWire : ICollectorWire
 
             if (!response.IsSuccessStatusCode)
             {
+                if (Log.IsDebugEnabled)
+                    Log.Debug("Request({0}): Invocation of \"{1}\" returned response headers : {2}", requestGuid, method, response.GetHeaders());
+
                 ThrowExceptionFromHttpResponseMessage(serializedData, response.StatusCode, responseContent, requestGuid);
             }
 
@@ -80,6 +83,8 @@ public class HttpCollectorWire : ICollectorWire
             // Possibly combine these logs? makes parsing harder in tests...
             Log.Debug("Request({0}): Invoked \"{1}\" with : {2}", requestGuid, method, serializedData);
             Log.Debug("Request({0}): Invocation of \"{1}\" yielded response : {2}", requestGuid, method, responseContent);
+            if (Log.IsDebugEnabled)
+                Log.Debug("Request({0}): Invocation of \"{1}\" returned response headers : {2}", requestGuid, method, response.GetHeaders());
 
             DataTransportAuditLogger.Log(DataTransportAuditLogger.AuditLogDirection.Received, DataTransportAuditLogger.AuditLogSource.Collector, responseContent);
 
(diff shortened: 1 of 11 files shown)


src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
```diff
@@ -80,6 +83,8 @@ public string SendData(string method, ConnectionInfo connectionInfo, string seri
             // Possibly combine these logs? makes parsing harder in tests...
             Log.Debug("Request({0}): Invoked \"{1}\" with : {2}", requestGuid, method, serializedData);
             Log.Debug("Request({0}): Invocation of \"{1}\" yielded response : {2}", requestGuid, method, responseContent);
+            if (Log.IsDebugEnabled)
```
nr-ahemsath, line 86: Why not also put the two Log.Debug statements above this line inside the `if` block?

tippmar-nr:
Are there likely to be any headers carrying sensitive data that we should avoid logging or obfuscate if we do log?

jaffinito:
In the request headers, definitely.  In the response headers, I would not expect there to be sensitive data.  I will do a quick google survey to see if anything comes up.

jaffinito:
I added some basic protection for headers that might have data we don't want in logs.

tippmar-nr (approved):
Nicely done! LGTM

src/Agent/NewRelic/Agent/Core/DataTransport/HttpCollectorWire.cs
jaffinito: I could have, but they don't need to be. My header line calls response.GetHeaders() — that executes regardless of log level (iterate headers, string.Join, allocate). Serilog/NoOp can drop the message, but it can't un-run the method that built the argument. The guard skips that work when Debug is off (the production default).

The existing lines pass serializedData / responseContent — strings that already exist. When Debug is off, the logger drops the message and those strings cost nothing extra.

PR #3659, opened 2026-06-25
feat: Report request queue time from X-Request-Start/X-Queue-Start headers across all web hosts

tippmar-nr:
## Summary

Reports request queue time from the `X-Request-Start` / `X-Queue-Start` headers across every web host the .NET agent instruments -- ASP.NET Core, OWIN self-host, and classic ASP.NET/IIS -- bringing the agent to parity with the Ruby and Node agents (from NR-581936; spike NR-528762).

Previously the agent only computed queue time under IIS, using `HttpWorkerRequest` internal timing, and never consulted the `X-*` headers. As a result:
- Non-IIS hosts (Kestrel/Kubernetes, OWIN self-host) reported no queue time at all, even when an upstream load balancer/ingress was already emitting these headers.
- IIS apps behind a load balancer only saw the IIS-internal queue, not the full front-end wait (LB + network + IIS).

## Changes

- **Shared parser** `QueueTimeHeaderParser` (in `NewRelic.Agent.Extensions`): reads `X-Request-Start` and `X-Queue-Start` (earliest valid wins), handles `t=`/bare/leading-token forms, auto-detects ns/us/ms/s, omits on clock skew (future-dated start), and omits when the computed queue time exceeds a sanity cap (10 minutes) -- which rejects misdetected, stale, or garbage timestamps regardless of how far in the past they decode to. Regex-free manual parsing; the header source is threaded through a non-capturing `static` delegate, so the absent-header path (the common case) allocates nothing. Returns `null` when no usable header is present.
- **Shared call site** `ITransaction.TrySetQueueTimeFromHeaders(...)` extension used by all hosts, generic over the header source so each call site passes a non-capturing static delegate (no per-request closure).
- **Wired into** ASP.NET Core (1.x-5.x and 6+), OWIN self-host, and classic ASP.NET/IIS.
- **Classic ASP.NET/IIS** now prefers the header (full front-end queue) and falls back to the existing `HttpWorkerRequest` timing when no header is present (no regression). Gated by `newrelic.config` `<appSettings>` switch `UseHeaderBasedRequestQueueTimeForClassicAspNet` (default `true`)  - to revert the classic ASP.NET/IIS path to legacy IIS-internal timing, set to `false`. The switch is specific to classic ASP.NET -- the non-IIS hosts have no legacy `HttpWorkerRequest` timing to fall back to, so it does not apply to them.

## Testing

- Unit tests for the parser (units, earliest-wins, max-queue-time cap (just-over and just-under boundary), stale-timestamp omit, clock-skew omit, malformed input, plus real-world nginx `t=${msec}000` and Heroku bare-millisecond formats) and the config switch.
- Integration tests for ASP.NET Core, IIS (default + switch-off), and OWIN, asserting the `WebFrontend/QueueTime` metric and `queueDuration` attribute, plus a clock-skew/omit case.
- Full solution builds clean; unit tests and the ASP.NET Core integration tests pass locally. IIS/OWIN integration tests run in CI.


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Helpers/QueueTimeHeaderParser.cs
```diff
@@ -0,0 +1,171 @@
...
+/// Parsing avoids regex and is allocation-free when the headers are absent (the common
+/// case); a header value is threaded in via a state object and a static delegate so the
+/// hot path allocates no closure. This matters because it runs on every web request.
+/// </summary>
+public static class QueueTimeHeaderParser
+{
+    private const string HeaderRequestStart = "X-Request-Start";
+    private const string HeaderQueueStart = "X-Queue-Start";
+
+    private static readonly DateTime _epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
jaffinito: We have a version of this in Core DateTimeExtensions.cs.  Could we update Core to use the Extensions value instead?

tippmar-nr: Yeah, that might be doable -- I was looking at it from the other direction and knew I couldn't reference Core from Extensions.

nrcventura (approved):

PR #3669, opened 2026-06-30
test: Fix flaky Azure Service Bus integration tests

tippmar-nr:
Unscoped rollup metric assertions used `callCount`, which only checks the first harvest payload. When the metrics harvest fires mid-test (e.g. during a slow Azure Service Bus send), the rollup count splits across two harvests and the assertion sees a partial count, causing intermittent failures across branches.

Switch the unscoped Produce/Consume/Peek/Settle/Cancel assertions to `CallCountAllHarvests` so they aggregate across all harvests. Scoped metrics each belong to a single transaction and harvest, so they remain on `callCount`.


chynesNR (approved):
I could have sworn we got rid of all the usages of `callCount`. Maybe we should deprecate it? I'm not sure there's anywhere we actually want that rather than all harvests.

PR #3673, opened 2026-07-01
test: Replace callCount with CallCountAllHarvests in integration test metric assertions

tippmar-nr:
- Switch integration-test metric-count assertions from `callCount` (validated only the first harvest, a recurring source of single-harvest timing flakes) to `CallCountAllHarvests` (sums CallCount across all harvests), and remove the now-unused `callCount` field from `ExpectedMetric`/`Assertions.cs`.
- Assert existence only for per-harvest Supportability metrics (`OS/Linux`, `MetricHarvest/transmit`, `CustomInst/Count`), whose re-emission every harvest cycle makes any cross-harvest sum non-deterministic in CI.


chynesNR (approved):
For the next bit of cleanup, maybe we could default `callCountAllHarvests` to 1 so we don't have to specify it everywhere. 🙂

PR #3694, opened 2026-07-13
chore(build): adopt Central Package Management for FullAgent.sln projects

tippmar-nr:
## Summary

Adopts [.NET Central Package Management (CPM)](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management) for the projects `FullAgent.sln` builds, so shared package versions have a single source of truth and a dependabot bump can no longer leave one project behind (the NU1605 drift that broke #3692).

## Boundary

- Root `Directory.Packages.props` holds every centrally-managed version. Enable flag lives in root `Directory.Build.props` as a conditional default so subtree opt-outs can override it (the SDK imports `Directory.Packages.props` after the `Directory.Build.props` chain).
- Opt-outs: 
  - wrapper providers
  - MSI/Wix
  - the profiler solution's managed projects
  - integration/MFA compat apps
  - performance tests
  - build tools.
- `Home.csproj` is a project-level opt-out and keeps `NewRelic.Agent.Internal.Profiler` pinned inline so the existing `build_profiler.yml` workflow continues to work without modification.
- `StyleCop.Analyzers` is imported across the CPM boundary via `build/StyleCop.props`. It is centrally versioned; `StyleCop.props` sets a `_UsesStyleCop` marker and the root `Directory.Build.targets` (auto-imported last) adds the reference gated on that marker -- versionless under CPM, inline `1.1.118` for opt-outs.
- One `VersionOverride` exception: `SharpZipLib=1.4.2` in `TestSerializationHelpers` (Core caps at 1.3.3).
- `build/build_functions.ps1`'s `Get-GrpcPackagePath` now reads `Grpc.Core` from `Directory.Packages.props` (it previously scraped the stripped `Core.csproj` line).
- Dependabot config keeps `nuget-agent` narrowly scoped to the in-scope src project directories so it does not sweep the wrapper subtree's intentional old pins; the `nuget-tests` allow-list plus its `/` scan handles centrally-versioned test packages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,7 +24,13 @@ updates:
 #        patterns:
 #          - "*"
 
-  # Update all NuGet packages used by core agent projects
+  # Update all NuGet packages used by core agent projects. Under CPM, dependabot
+  # updates the central Directory.Packages.props via any in-scope project that
+  # references its packages, so the narrow directory list below is sufficient
+  # and, more importantly, keeps wrapper/integration/perf/MSI/profiler pins
+  # (which intentionally reference old library versions) out of scope. Root '/'
+  # is NOT scanned here because that would pull FullAgent.sln's full project
+  # graph including the Providers/Wrapper/* projects.
   - package-ecosystem: nuget
     directories: # not recursive - only the specified directory will be scanned for .csproj or .sln files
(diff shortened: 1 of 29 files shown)


nr-ahemsath (approved):
Looks good to me. 👍

PR #3705, opened 2026-07-15
fix: Instrument new method for local log decoration in Serilog 4.4.0+

### LINKED ISSUE #3704: Local log decoration (NR-LINKING) breaks with Serilog 4.4.0+ (instrumented method renamed Dispatch -> PostLevelCheckEmit)
## Description

The .NET agent's local log decoration feature (`application_logging.local_decorating.enabled`) stops working when an application uses **Serilog 4.4.0 or later**. Log lines are no longer decorated with the `NR-LINKING` metadata marker, so logs-in-context linking via local decoration does not function.

Log forwarding (`application_logging.forwarding`) is **not** affected - only local decoration.

## Affected versions

- **Serilog:** 4.4.0 and later. Serilog 4.3.1 and earlier work correctly.
- **.NET agent:** all current agent versions, on both .NET Framework and .NET (Core). The relevant instrumentation is not agent-version-specific, so any app that upgrades Serilog to 4.4.0+ loses local decoration regardless of agent version.

## Root cause

The agent decorates Serilog logs by instrumenting the private method `Serilog.Core.Logger.Dispatch(LogEvent)` (tracer factory `SerilogDispatchWrapper` in `src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/SerilogLogging/Instrumentation.xml`). The wrapper injects an `NR_LINKING` property into the `LogEvent`, which a user's `{NR_LINKING}` output-template token then renders as the linking metadata.

In Serilog 4.4.0 this method was **renamed from `Dispatch` to `PostLevelCheckEmit`**, as part of the upstream change "Capture the most important internal logger health metrics using System.Diagnostics.Metrics" (the 4.4.0 "new metrics source" minor bump). `Serilog.Core.Logger` no longer contains any method named `Dispatch`.

Because the agent uses an `exactMethodMatcher` for `methodName="Dispatch"`, the profiler no longer matches any method in Serilog 4.4.0+, the decoration wrapper never runs, and the `NR_LINKING` property is never added. The underlying `LogEvent` property storage the wrapper writes to (`_properties`) is unchanged between 4.3.1 and 4.4.0 - only the method name changed - so the feature becomes a complete no-op (no partial decoration).

Upstream reference: https://github.com/serilog/serilog/compare/v4.3.1...v4.4.0
- v4.3.1 `src/Serilog/Core/Logger.cs`: `void Dispatch(LogEvent logEvent)`
- v4.4.0 `src/Serilog/Core/Logger.cs`: renamed to `void PostLevelCheckEmit(LogEvent logEvent)`

## Reproduction

1. An ASP.NET Core or console app using Serilog 4.4.0+ with a sink whose output template includes the `{NR_LINKING}` token.
2. Enable local decoration: `NEW_RELIC_APPLICATION_LOGGING_LOCAL_DECORATING_ENABLED=true`.
3. Emit a log line inside a transaction.
4. Expected: the emitted log line contains an `NR-LINKING|...|...|` marker. Actual on Serilog 4.4.0+: no marker is present.

Reproduced internally with the `SerilogWebPatternLayoutDecorationEnabledTestsNetCoreLatestTests` integration test, which passes on Serilog 4.3.1 and fails on 4.4.0.

## Workaround

Pin Serilog to a 4.3.x version until a fix ships.

## Suggested fix

Add a matcher for the renamed method in `SerilogLogging/Instrumentation.xml` so both method names are instrumented (keep `Dispatch` for Serilog < 4.4.0, add `PostLevelCheckEmit` for 4.4.0+). Both methods take a single `LogEvent` argument, so `SerilogDispatchWrapper` itself needs no code change.

nr-ahemsath:
## Description

Fix #3704 by adding a new instrumentation point for `PostCheckLevelEmit`, in `Serilog.Core.Logger`, which replaces `Dispatch` in version 4.4.0+.

Verified that the local log decoration integration tests for Serilog all passed on my dev system.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/SerilogLogging/Instrumentation.xml
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/SerilogLogging/Instrumentation.xml
@@ -8,12 +8,17 @@ SPDX-License-Identifier: Apache-2.0
     <tracerFactory name="SerilogDispatchWrapper">
       <!-- For Serilog 1.x -->
       <match assemblyName="Serilog" className="Serilog.Core.Pipeline.Logger">
-        <exactMethodMatcher methodName="Dispatch" />
+        <exactMethodMatcher methodName="Dispatch" maxVersion="2.0.0"/>
       </match>
 
       <!-- For Serilog 2.x and above -->
       <match assemblyName="Serilog" className="Serilog.Core.Logger">
-        <exactMethodMatcher methodName="Dispatch" />
+        <exactMethodMatcher methodName="Dispatch" minVersion="2.0.0" maxVersion="4.4.0" />
+      </match>
+
+      <!-- For Serilog 4.4.0 and above -->
+      <match assemblyName="Serilog" className="Serilog.Core.Logger">
+        <exactMethodMatcher methodName="PostLevelCheckEmit" minVersion="4.4.0" />
       </match>
     </tracerFactory>
     
--- a/docs/net-agent-compatibility.md
+++ b/docs/net-agent-compatibility.md
@@ -77,7 +77,7 @@ The .NET agent [can be configured](https://docs.newrelic.com/docs/apm/agents/net
 | Library | NuGet package | Supported versions | Min agent version | Notes |
 | --- | --- | --- | --- | --- |
 | Log4Net | [log4net](https://www.nuget.org/packages/log4net/) | 2.0.10 – 3.3.2 | 9.7.0 |  |
-| Serilog | [Serilog](https://www.nuget.org/packages/Serilog/) | 2.5.0 – 4.3.1 | 9.7.0 |  |
+| Serilog | [Serilog](https://www.nuget.org/packages/Serilog/) | 2.5.0 – 4.4.0 | 9.7.0 |  |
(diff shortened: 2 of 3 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/SerilogLogging/Instrumentation.xml
```diff
@@ -8,12 +8,17 @@ SPDX-License-Identifier: Apache-2.0
     <tracerFactory name="SerilogDispatchWrapper">
       <!-- For Serilog 1.x -->
       <match assemblyName="Serilog" className="Serilog.Core.Pipeline.Logger">
-        <exactMethodMatcher methodName="Dispatch" />
+        <exactMethodMatcher methodName="Dispatch" maxVersion="2.0.0"/>
```
tippmar-nr, line 11: Was this change necessary? It's a different class/method from anything else we instrument in that assembly, so probably doesn't need a cap (even though it is, I suppose, technically correct).

nr-ahemsath: Since I was adding version boundaries to the other matchers in that section, I wanted to complete the pattern for that one as well.

tippmar-nr (approved):

nrcventura (approved):

PR #3709, opened 2026-07-16
feat: Unified injected bytecode method cache

tippmar-nr:
BEGIN_COMMIT_OVERRIDE
feat: The code that the profiler injects into your application now caches its lookups into the agent with a lock-free cache, instead of resolving each call by reflection, resulting in lower agent overhead for .NET applications. (#3709)
END_COMMIT_OVERRIDE

Thank you for submitting a pull request.  Please review our [contributing guidelines](../CONTRIBUTING.md) and [code of conduct](https://opensource.newrelic.com/code-of-conduct/).

## Description
Unifies the cache that the profiler's injected bytecode uses to reach the managed agent (NR-184027).

**Before:** every injected tracer-dispatch call, and every Agent API call, resolved its target through `AppDomain.GetData`/`SetData`. On .NET Core that maps to `AppContext.GetData`, which serializes every lookup behind a single process-wide monitor.

**After:** the profiler defines a helper type (`__NRInitializer__`) into the core library, whose static object fields act as a lock-free fast-path cache and fall back to the old AppDomain storage on a miss. Agent API calls additionally dispatch through a new managed-side `ProfilerAgentMethodInvoker`, which caches compiled expression delegates instead of re-resolving `MethodInfo` by reflection on every call.

Delivered in three slices:

- **Phase 1** (#3695) - the `AgentCallStyle` strategy abstraction and runtime-agnostic core-library-name plumbing.
- **Phase 2** (#3698) - the static-field cache and its helper methods, .NET Framework only at that point.
- **Milestone B** (#3707) - core-library injection on CoreCLR. This is net-new capability rather than a refactor; CoreCLR never ran the injection path before.

The two runtimes carry separate assembly-reference tables, public key tokens and versions for `mscorlib` versus `System.Private.CoreLib`, selected by an `isCoreClr` flag threaded through `Module`, `ModuleInjector`, `InstructionSet` and `FunctionManipulator`.

### Operational notes

- No new configuration. `NEW_RELIC_DISABLE_APPDOMAIN_CACHING` is pre-existing and unchanged; setting it selects the reflection strategy.
- The profiler now always logs the resolved calling strategy at startup (`Calls to the managed agent will use the calling strategy - ...`), where that line was previously conditional.
- If helper-type injection into the core library cannot be verified, the process automatically downgrades to the reflection strategy and logs it. Both of these are native profiler log lines only, not supportability metrics.

### Test coverage

Profiler C++ unit tests for the strategy resolution, bytecode generation and instruction-set changes; agent unit tests for `ProfilerAgentMethodInvoker`; and integration plus container coverage of both caching arms. Also adds a .NET Framework integration test that runs two ASP.NET applications in one Hosted Web Core process (one application pool, two AppDomains, two agents) and verifies per-application agent reporting under both arms, which nothing previously exercised across AppDomains. The container AppDomain-caching tests moved from the `Distro` trait axis to `TestArea`.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review


src/Agent/NewRelic/Agent/Core/AgentShim.cs
```diff
@@ -94,6 +94,33 @@ static bool DeferInitialization(string method)
         return DeferInitializationOnTheseMethods.Contains(method);
     }
 
+    public static object GetFinishTracerDelegateFunc()
+    {
+        return (Func<object[], Action<object, Exception>>)GetFinishTracerDelegateParameterWrapper;
+    }
+
+    /// <summary>
+    /// This method is used to work around .net framework and .net &lt; 6.0 not supporting generic delegates
```
nr-ahemsath:
```suggestion
    /// This method is used to work around .NET Framework and .NET < 6.0 not supporting generic delegates
```

jaffinito (approved):
LGTM

PR #3715, opened 2026-07-22
chore(build): replace StyleCop with a package-free MSBuild copyright-header check

nr-ahemsath:
### Summary

StyleCop was used in this repo for exactly one thing: enforcing the two-line New Relic license header on source files (only the SA1633/1634/1635/1636/1640/1641 header rules were enabled; every other rule was disabled). This PR removes StyleCop entirely and replaces that enforcement with a small MSBuild target.

### Motivation

After Central Package Management was adopted repo-wide (#3694), the shared StyleCop.Analyzers reference had to be versionless under CPM but inline-versioned for the CPM opt-out subtrees. Visual Studio's IDE restore mishandled that split for the legacy (non-SDK) .NET Framework integration-test apps: it emitted the versionless reference during item collection but then resolved those projects as non-CPM, so restore failed with:

`NU1015: The following PackageReference item(s) do not have a version specified: StyleCop.Analyzers`

This broke IntegrationTests.sln in Visual Studio (CI, which restores via MSBuild.exe, was unaffected — which is why it stayed green). Rather than keep fighting the CPM/analyzer/restore interaction for a tool we only use for a header check, this removes the dependency.

### Changes

- New: EnforceFileHeaders target in the root Directory.Build.targets — runs before CoreCompile on every C# project (SDK-style and legacy non-SDK), and fails the build with NR1001 if a compiled file is missing the required header:
- Generated files are skipped by name (*.AssemblyInfo.cs, *.AssemblyAttributes.cs, *.g.cs, *.Designer.cs, GlobalUsings.g.cs, TemporaryGeneratedFile_*) and by <auto-generated> marker. A project can opt out with <EnforceFileHeaders>false</EnforceFileHeaders>.
- Removed: build/StyleCop.props, build/stylecop.json, build/Common.ruleset; the StyleCop.Analyzers central package version; the StyleCop.props imports in 4 Directory.Build.props files and 3 profiler csprojs; the StyleCop.Analyzers reference in AspNetCoreFeatures.csproj; and the CodeAnalysisRuleSet in InstallerActions.csproj.
- Docs: updated the StyleCop/CPM section in src/CLAUDE.md.

### Why an MSBuild target (not IDE0073)

The built-in IDE0073 file-header analyzer only runs at build for SDK-style projects, so it would silently drop enforcement on the ~16 legacy .NET Framework integration-test apps. An MSBuild target covers every project type uniformly and has no NuGet/CPM/restore entanglement, so the whole class of restore failures is gone.

### Testing

- Verified the target passes on both a legacy and an SDK project, and fails with NR1001 (naming the offending file) when a header-less file is present.
- FullAgent.sln (Debug): builds clean, 0 errors.
- IntegrationTests.sln via MSBuild.exe -restore (CI-equivalent): builds clean, 0 errors — no NU1015, no NR1001; the previously-failing legacy projects build.


chynesNR:
Do you have an example of what the error will look like if you leave out a header, with this change?

nr-ahemsath:
> Do you have an example of what the error will look like if you leave out a header, with this change?

Yes!  Because these changes uncovered the fact that almost all of the files in the build tools solution were missing the copyright header, and the first CI run on this PR failed for that reason: https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/29940441056

Directory.Build.targets
```diff
@@ -1,11 +1,101 @@
...
+       versionless under Central Package Management but inline for the CPM opt-out subtrees, and
+       VS's IDE restore mishandled that split for legacy non-SDK projects (emitting a versionless
+       reference it then couldn't resolve -> NU1015). An MSBuild target has no such entanglement
+       and covers legacy projects, which analyzer-based options (IDE0073) do not. -->
+  <PropertyGroup>
+    <RequiredFileHeaderLine1>// Copyright 2020 New Relic, Inc. All rights reserved.</RequiredFileHeaderLine1>
+    <RequiredFileHeaderLine2>// SPDX-License-Identifier: Apache-2.0</RequiredFileHeaderLine2>
+  </PropertyGroup>
+
+  <UsingTask TaskName="CheckFileHeaders" TaskFactory="RoslynCodeTaskFactory"
```
nrcventura, line 19: This doesn't feel like the right solution. Even OpenTelemetry is still using stylecop to enforce the copyright headers.

nr-ahemsath:
Per team pairing discussion, the custom MSBuild logic is overly complex and will be difficult to maintain.  Closing this and will work on a simpler solution.

PR #3727, opened 2026-07-28
fix: Add on-insert check to SimpleCache to prevent runaway growth.

jaffinito:
## Description

`SimpleCache` cleared its entire contents whenever it grew past capacity, so a busy cache (e.g. SQL parse cache) thrashed constantly and `ConcurrentDictionary.Count` locked every internal bucket on every check/clear, blocking all writers ~2x/sec per cache instance across ~15+ live instances.

- Drop new inserts on the floor once at capacity; add a `Dropped` metric (`CountDropped`) alongside the existing Hits/Misses/Ejections, plumbed through `ResetStats` the same way.
- `MaintainCache` (500ms timer) now clears only when something was actually dropped, or when tracked size is over capacity (handles `Capacity` being lowered) — otherwise persists the cache even sitting at capacity.
- Replaced `ConcurrentDictionary.Count` with a self-tracked `Interlocked` counter for size/capacity checks — removes the full-lock-acquisition cost from every insert and from `Size`.
- Fixed root cause of a null-value bug: a cached `null` was indistinguishable from an absent key, causing `GetOrAdd` to re-run the factory on every call and `Contains`/`TryAdd` to disagree about whether the key existed. `Contains`/`Get`/`GetOrAdd` now key off `ConcurrentDictionary`'s own presence signal, not `value != null`.
- Added a constructor overload (`maintainCacheIntervalMs`, default unchanged) so tests can disable the internal timer for determinism instead of racing it.

## Test plan
- [x] Unit tests added/updated in `SimpleCacheTests.cs` for drop-on-insert, capacity increase/decrease, maintenance persist-vs-clear, null-value handling, and lock-free size tracking (32 tests, all passing on net10.0/net481)
- [x] `Core.UnitTest` regression check on all real `SimpleCache` consumers (`DatabaseStatementParser`, `CacheByDatastoreVendor`, `DatastoreSegmentData`, `DatabaseService`, `JsonArrayConverter`) — all passing

## Performance

**Scenario 1** — below capacity (1000 keys, capacity 2000, 3s, 20 threads):
Old: 17.2M ops/sec   New: 22.9M ops/sec   (+33%)
Even under light load with zero clears either side, New wins — Old's 500ms timer still calls _cacheMap.Count every tick, which locks every internal dictionary segment (ConcurrentDictionary's default concurrency level scales with core count — 80 locks on this box) and briefly stalls all 20 writer threads each time. New reads a plain int.

**Scenario 2** — well above capacity (capacity 500, every key unique, 3s, 20 threads):
Old: peak size 2,047,235 entries (!) — grows unbounded for up to 500ms between clears
New: peak size       518 entries    — bounded at ~capacity + in-flight writers, always
Old: 3.5M ops/sec        New: 8.8M ops/sec   (2.5x)
This is the actual bug: Old lets the dictionary grow completely unchecked between ticks, then eats a full-lock Clear() on a multi-million-entry table. New drops on insert, so it never has more than ~capacity to hold or clear.

**Scenario 3** — null-value "negative cache" pattern (if (!Contains(k)) TryAdd(k, factory), factory returns null, 5 keys, 2s):
Old: 32,846,943 factory calls (every single op re-runs it — the poison-slot bug)
New:          5 factory calls (once per key, ever — correct)
Old: 16.3M ops/sec       New: 456.4M ops/sec   (28x)
Old's Contains treats a cached null as absent forever, so every call re-triggers TryAdd, and TryAdd's factory argument evaluates before its own internal existence check — so the "wasted" work runs unconditionally, every time, forever. New's fix eliminates it entirely after the first insert.

# Author Checklist

# Reviewer Checklist
- [ ] Perform code review

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Caching/SimpleCache.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Caching/SimpleCache.cs
@@ -91,52 +91,41 @@ public class SimpleCache<TKey, TValue> : ICacheStats, IDisposable where TValue :
     /// </summary>
     public TValue Peek(TKey key)
     {
-        var node = PeekInternal(key);
-        return node;
+        _cacheMap.TryGetValue(key, out var value);
+        return value;
     }
 
     /// <summary>
-    /// Checks whether the specified key exists in the cache without updating stats.
+    /// Checks whether the specified key exists in the cache without updating stats. Unlike <see cref="Peek"/>,
+    /// this is correct for a key whose cached value is itself null.
     /// </summary>
-    public bool Contains(TKey key) => PeekInternal(key) != null;
+    public bool Contains(TKey key) => _cacheMap.ContainsKey(key);
 
     /// <summary>
-    /// Allows searching of the cache.  If found, returns the existing item and updates the statistics.
-    /// If not found, returns NULL.
+    /// Allows searching of the cache. If found, returns the existing item (which may itself be null) and
+    /// records a hit. If not found, records a miss and returns null.
     /// </summary>
     /// <param name="key"></param>
     /// <returns></returns>
     public TValue Get(TKey key)
     {
-
-        var node = PeekInternal(key);
-
-        if (node != null)
-        {
-            Interlocked.Increment(ref _countHits);
-        }
-        else
-        {
-            Interlocked.Increment(ref _countMisses);
-        }
-
-        return node;
+        TryGetTrackingStats(key, out var value);
+        return value;
     }
 
     /// <summary>
-    /// Attempts to find an item in the cache.  If found, returns the existing item and updates the statistics.
-    /// If not found, will add the item to the cache, unless the cache is at capacity, in which case the item
-    /// is dropped (not cached) but is still computed and returned.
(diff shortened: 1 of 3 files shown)


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Caching/SimpleCache.cs
```diff
@@ -145,43 +209,64 @@ private void SetCapacity(int newCapacity)
...
     /// </summary>
-    public int Size => _cacheMap.Count;
+    public int Size => _count;
 
-    private TValue PeekInternal(TKey key)
+    /// <summary>
+    /// Looks up <paramref name="key"/> and records a hit or miss based on whether it is actually present -
+    /// not on whether <paramref name="value"/> is null, since a cached null value is a valid hit.
+    /// </summary>
+    private bool TryGetTrackingStats(TKey key, out TValue value)
```
nrcventura: This name confused me at first, we should consider renaming it. Some possibilities are `TryGetWithTrackingStats` or `TryGetInternal`.

jaffinito: Fixed!

nrcventura (approved):

PR #3730, opened 2026-07-30
fix: Guard null OperationContext in WCF MethodInvokerWrapper

### LINKED ISSUE #3728: Wcf3 MethodInvokerWrapper throws NullReferenceException when OperationContext.Current is null, disabling WCF instrumentation for the process
## Problem

`MethodInvokerWrapper.CaptureHttpRequestHeadersAndMethod` dereferences
`OperationContext.Current` without a null check
(`src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/MethodInvokerWrapper.cs#L323-L327`).

It is reached only on the `!transactionAlreadyExists` branch -- when the agent found no
transaction in context storage. Those two conditions are not independent: WCF transaction
storage is backed by the same ambient objects the wrapper dereferences.
`OperationContextStorage` (priority 5) reads `OperationContext.Current`;
`HttpContextStorage` (priority 10) reads `HttpContext.Current`. A thread with neither fails
the transaction lookup *and* the dereference.

Per affected call:

- `WrapperService` catches the NRE, so the request succeeds but no transaction or segment is
  recorded -- the operation is invisible.
- The transaction created at line 159 is left orphaned in `ThreadLocalStorage`. The next
  invocation on that thread adopts it, so a reported transaction can be named and populated
  by a different call than the one it was created for.
- That adoption path calls `TrackedWrapper.NoticeSuccess()`, which zeroes the
  consecutive-failure counter. Failures therefore self-limit on a warm thread pool, and a
  process can log thousands of these over days without tripping anything.
- When 5 consecutive failures do occur, `WrapperService` swaps in the NoOp wrapper keyed by
  `functionId`. Each instrumented invoker method is JIT'd once per process and shared by
  every operation of that invocation style, so WCF transaction naming is lost process-wide
  until restart and transactions fall back to the generic host-level name.

Three further sites take the same context unguarded, so fixing only the first moves the
throw rather than removing it:

- `ExtractHeaderValue` (lines 93, 102), reached via `AcceptDistributedTraceHeaders` at line
  201. The DT path catches and swallows it (mislogged as `CreateExecuteEveryTimer() failed`),
  but `Agent.TryProcessSyntheticsData` has no try/catch, so that one propagates.
- `ProcessResponse` line 384 (`context.OutgoingMessageHeaders`), reached whenever
  `GetResponseMetadata()` returns a header. The method it calls immediately below already
  guards `context == null`.

Not a regression -- present since #593 (2021). Reproduced on .NET Framework 4.7.2, agent 10.51.

## Steps to reproduce

1. Host a WCF service with a synchronous operation and instrument it with the agent.
2. Install an `IOperationBehavior` that replaces `DispatchOperation.Invoker` with a wrapper
   delegating to the original invoker from another thread (`Task.Run`, a new `Thread`, or a
   long-lived worker). Attach it during `IServiceBehavior.Validate`, so `Invoker` is already
   populated when the operation behavior runs.
3. Call the operation repeatedly.

Each call logs `Tracer invocation error` / `System.NullReferenceException` at
`CaptureHttpRequestHeadersAndMethod`. After 5 consecutive failures the log shows
`Wrapper ... MethodInvokerWrapper is being disabled for Invoke`, and `WebTransaction/WCF/*`
naming stops for the life of the process.

Verified self-hosted (net.tcp) and IIS-hosted (`.svc` with
`aspNetCompatibilityEnabled="true"`). A fresh thread per call makes it deterministic; a
reused thread makes it intermittent via the adoption behavior above.

## Recommended fix

1. Guard all four sites.
2. When `OperationContext.Current` is null, return before `agent.CreateTransaction` rather
   than guarding inside `CaptureHttpRequestHeadersAndMethod`. Guarding alone still creates a
   transaction that *is* stored (in `ThreadLocalStorage`) and harvested as
   `WebTransaction/WCF/<Type>.<Method>` with no URI, request method, request headers, or
   accepted DT headers.
3. Record a supportability metric for the null-context case so its frequency is measurable
   instead of silent.

Separately: correct the `CreateExecuteEveryTimer() failed` message in
`DistributedTracePayloadHandler`, which misattributes every DT-accept failure, and consider
giving `TryProcessSyntheticsData` the try/catch its CAT counterpart has.

tippmar-nr:
Fixes #3728. 

`MethodInvokerWrapper` dereferenced `OperationContext.Current` with no null check, which throws when a custom `IOperationInvoker` dispatches the operation off the WCF request thread -- after `WrapperExceptionLimit` consecutive failures the agent swaps in the NoOp wrapper per `functionId`, losing `WebTransaction/WCF` naming process-wide until restart.

All four dereference sites are guarded, and the wrapper now returns before creating a transaction when there is no ambient context, recording `Supportability/WCFService/NullOperationContext/<method>`. Bailing out matters: guarding alone still leaves a contentless `WebTransaction/WCF` entry stored in `ThreadLocalStorage` and harvested.

Verified with a new integration test (8 off-thread invocations, so it exceeds the disable threshold) and a standalone repro harness across 7 dispatch modes -- zero NREs, no wrapper disable, and no fabricated transactions, with the inline control unaffected.

Two related defects are deliberately left out of scope: the mislabeled `CreateExecuteEveryTimer() failed` message in `DistributedTracePayloadHandler`, and the missing try/catch on `Agent.TryProcessSyntheticsData`.

### DIFF
--- a/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/MethodInvokerWrapper.cs
+++ b/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/MethodInvokerWrapper.cs
@@ -63,6 +63,22 @@ public class MethodInvokerWrapper : IWrapper
     private const string InvokeEndMethodName = "InvokeEnd";
     private const string InvokeAsyncMethodName = "InvokeAsync";
     private const string WrapperName = "MethodInvokerWrapper";
+    private const string NullOperationContextMetricPrefix = "WCFService/NullOperationContext";
+    private const string NullOperationContextMetricNameOther = NullOperationContextMetricPrefix + "/Other";
+
+    // Precomputed per-invoker-method metric names for the null-OperationContext bail-out.
+    // This is the granularity at which the wrapper actually gets disabled (the NoOp swap is
+    // keyed per functionId, i.e. one JIT'd method), so it is deliberately not the invocation
+    // style (Sync/APM/TAP) - InvokeEnd's style is ambiguous and is omitted from
+    // _methodNameInvocationTypesDic below, but InvokeEnd can still reach the bail-out.
+    private static readonly Dictionary<string, string> _nullOperationContextMetricNamesByMethod = new Dictionary<string, string>()
+    {
+        { SyncMethodName, NullOperationContextMetricPrefix + "/" + SyncMethodName },
+        { InvokeBeginMethodName, NullOperationContextMetricPrefix + "/" + InvokeBeginMethodName },
+        { InvokeEndMethodName, NullOperationContextMetricPrefix + "/" + InvokeEndMethodName },
+        { InvokeAsyncMethodName, NullOperationContextMetricPrefix + "/" + InvokeAsyncMethodName }
+    };
+
     /// <summary>
     /// Translates the method name to the type of invocation
     /// </summary>
@@ -88,6 +104,11 @@ public class MethodInvokerWrapper : IWrapper
 
     private static IEnumerable<string> ExtractHeaderValue(OperationContext context, string key)
     {
+        if (context == null)
+        {
+            return null;
+        }
+
         string headerValue = null;
 
         if(context.IncomingMessageProperties != null && context.IncomingMessageProperties.TryGetValue(HttpRequestMessageProperty.Name, out var httpRequestMessageAsObject))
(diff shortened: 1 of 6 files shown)


src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/MethodInvokerWrapper.cs
```diff
@@ -130,14 +151,33 @@ private string GetHeaderValueFromWebHeaderCollection(System.Collections.Speciali
...
+        // is still stored (in ThreadLocalStorage) and harvested, and would be left
+        // orphaned for the next invocation on this thread to adopt. Bail out instead.
+        var operationContext = OperationContext.Current;
+        if (!transactionAlreadyExists && operationContext == null)
+        {
+            var nullOperationContextMetricName = _nullOperationContextMetricNamesByMethod.TryGetValue(instrumentedMethodName, out var perMethodMetricName)
+                ? perMethodMetricName
+                : NullOperationContextMetricNameOther;
+
+            agent.GetExperimentalApi().RecordSupportabilityMetric(nullOperationContextMetricName);
```
nrcventura, line 171: Should we also write something to the agent log when this happens?

nrcventura (approved):

PR #3740, opened 2026-08-04
test: authenticate Bedrock integration tests via OIDC instead of a static key

tippmar-nr:
Removes the long-lived AWS access key from the test configuration surface, from NR-601301. The key itself was already deleted in AWS on 2026-08-01 under NR-601293, so the Bedrock tests could not authenticate by any means before this change.

- CI assumes a repo-scoped role via GitHub OIDC, gated to the `LLM` matrix job so no other job holds AWS credentials.
- Locally, `AwsTestCredentials` asks the AWS CLI for the current credentials at test startup and passes them to the test application. The AWS SDK for .NET cannot read an SSO session without the `AWSSDK.SSO`/`AWSSDK.SSOOIDC` packages, which the test apps deliberately do not reference, so this routes local runs through the same environment-variable path CI uses.
- `AwsAccessKeyId`, `AwsSecretAccessKey`, `AwsAccountNumber` and `AwsTestRoleArn` are gone from `IntegrationTestConfiguration`; `AwsRegion` stays.

The AWS role and its least-privilege policy are already provisioned. The policy grants `bedrock:InvokeModel` on exactly the five models the tests name, and carries an explicit `Deny` on `meta.llama2-70b-chat-v1` because `LLMErrorTests` uses it as its access-denied case and asserts the call is refused -- do not widen that.

Verified that the LLM namespace tests are successful in CI with this change.

### DIFF
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -567,6 +567,9 @@ jobs:
     needs: [build-integration-tests, get-test-namespaces]
     name: Run IntegrationTests
     runs-on: windows-2025-vs2026
+    permissions:
+      contents: read   # required: a job-level block replaces the workflow-level one
+      id-token: write  # required for OIDC role assumption (LLM namespace only)
     strategy:
       matrix:
         namespace: ${{ fromJson(needs.get-test-namespaces.outputs.integration-namespaces) }}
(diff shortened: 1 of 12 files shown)


nrcventura (approved):

nr-ahemsath (approved):
Verified that the LLM tests failed locally before doing the `aws login` step, and passed locally after performing that step.

PR #3752, opened 2026-08-11
fix: Report the database a SQL command actually ran against

tippmar-nr:
`ConnectionInfo.DatabaseName` came only from the connection string. That string does not change when an application calls `ChangeDatabase`, calls `ChangeDatabaseAsync`, or runs `USE` on an open connection. Commands after a switch reported the original database.

The SQL and ODBC wrappers now read the live `IDbConnection.Database`. If it differs from the parsed name, the wrappers use the live value. A shared helper, `ConnectionInfoResolver`, does the comparison. Each wrapper also reads the connection string once per execution instead of two or three times. From NR-576099.

New tests:

- Unit tests for `ConnectionInfo.WithDatabaseName` (7 cases) and for `ConnectionInfoResolver` (8 cases). They cover the port and path-or-id forms, normalization of a null or empty name, immutability of the receiver, the case-insensitive compare, the composite cache key, and a cache hit.
- Integration tests for a database switch on an open connection, in `DatabaseSwitchTestsBase`. Concrete classes cover MSSQL (`ChangeDatabase`, `ChangeDatabaseAsync`, and a raw `USE`), ODBC (`ChangeDatabase` and a raw `USE`), and MySQL (`ChangeDatabase`). Each test asserts that two segments in one transaction report different `database_name` values.
- New exerciser methods on `MicrosoftDataSqlClientExerciser`, `SystemDataOdbcExerciser`, and `MySqlExerciser` drive those tests.

Two changes to emitted values are intended:

- `database_name` and `db.instance` now show a real database name when the connection string omits one. They showed `"unknown"` before.
- MSSQL can report server casing instead of connection-string casing after a switch.


nr-ahemsath:
A few observations based on Claude's code review and subsequent grilling me to make sure that I understand what is going on:

1. These changes use a composite `connectionString + "|" + liveDatabaseName` key for the ConnectionInfo cache.  My understanding is that this helps performance in the case where an app switches from database A to database B and then *back* to database A within the same connection.  However, this use case is not tested in unit or integration tests.  I think that at least unit testing this would be a good idea.
2. The review flagged that, while this is an extremely unlikely scenario, a vertical bar/pipe character could be used in either a connection string or a database name (which I sort of confirmed by searching online or "can a | char be used in a database name or connection string" and got several results which said "yes"), and therefore a cache key collision is theoretically possible if `A + B|C` collided with `A|B + C`.  This could be avoided by using a (connectionString, liveDatabaseName) tuple as the cache key instead of the string concatenation.  However, it seems like such a remote possibility that I will not mind if you ignore this feedback.

jaffinito (approved):
Real bug, correct fix, thin wrappers, logic pushed into testable helper per repo convention, thorough tests. LGTM.

PR #3759, opened 2026-08-13
chore: Remove dead Linux package tests and modernize what remains

nr-ahemsath:
A CodeQL alert pointed out that there were some .NET 6 apps in this part of the repo.  Most of what was under build/Linux/test was completely unused either in CI or by anybody locally.  This PR:

1. Removes most of what was in build/Linux/test
2. Modernizes what remains (a simple package install smoke test, appropriate for local DEB/RPM packaging development testing)
3. Removes the commented-out and unused OSHT bash testing library.


build/Linux/test/scripts/test.sh
```diff
@@ -8,6 +8,11 @@ install_agent
 
 verify_no_logs
 
-dotnet run
+# run the sample mvc app in the background and make a request to it to generate logs
+dotnet ./bin/Debug/net10.0/data.dll --urls "http://+:8080" >/dev/null 2>&1 &
```
nrcventura: It seems like this could be easily missed the next time we need to update the .net version used by the test app. Should we update the project to use a version independent build output folder?

nrcventura (approved):

PR #3771, opened 2026-08-21
chore: Add agent-log analysis and test-failure triage skills

tippmar-nr:
Adds two agent skills used while triaging a WCF integration test failure from NR-609087:

- `analyze-dotnet-agent-logs` - parses agent and profiler logs without pulling huge payload lines into context.
- `triage-integration-test-failures` - classifies an intermittent CI failure (infra, timing, tooling, upstream, or a real product defect) before any fix is proposed.

### DIFF
--- a/.claude/skills/analyze-dotnet-agent-logs/SKILL.md
+++ b/.claude/skills/analyze-dotnet-agent-logs/SKILL.md
@@ -1,0 +1,109 @@
+---
+name: analyze-dotnet-agent-logs
+description: Parse and diagnose New Relic .NET agent logs (newrelic_agent_*.log) and profiler logs (NewRelic.Profiler.<pid>.log) from a support ticket. Use when handed a log file or a logs directory, or when asked why the agent is not reporting, not connecting, not instrumenting a library, or why custom instrumentation XML has no effect.
+---
+
+# Analyze .NET agent logs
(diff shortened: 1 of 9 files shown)


.claude/skills/analyze-dotnet-agent-logs/SKILL.md
```diff
@@ -0,0 +1,109 @@
...
+| `instrumented` | List the methods the profiler rewrote, for checking custom instrumentation |
+
+Launcher: `python` on Windows, `python3` elsewhere.
+
+Derived files land in `nrlog-work/` beside the source log unless `--out` says
+otherwise. The source log is never modified.
+
+## Changing this skill
+
+Regenerate the fixtures and re-run the checks before trusting an edit to
```
chynesNR, line 88: Is there a way we could automate this? e.g. a workflow that runs if `nrlog.py` has changed?

tippmar-nr: I don't think it's necessary - the "regenerate" takes a couple of seconds and this blurb in the skill should catch the need to rebuild; I don't anticipate anyone manually editing that script so any changes would be AI-generated.

nrcventura (approved):

PR #3780, opened 2026-08-26
test: Various fixes for MongoDB and other tests

tippmar-nr:
The MongoDB unbounded job was spending most of its time waiting on a log line that could not appear, and hiding the failure. Measured on one fixture: 120.0s of the exercise phase became 27.8s, and the test duration went from 3.7 min to 1.3 min.

- **FINEST log level in the 6 MongoDB classes and 2 container classes.** They waited on `TransactionTransformCompletedLogLineRegex`, which requires the FINEST prefix, while the agent logged at DEBUG. Each wait spent its full 2-minute timeout.
- **Removed the `exerciseApplication` waits from the MongoDB classes.** The app runs to completion and the agent flushes metrics on shutdown.
- **Fixed the swallowed exception in `RemoteApplicationFixture`.** The catch set `retryMessage` but never `retryTest`, so a timed-out wait passed as success with partial data. One line, restoring the evident intent: retry, then fail after `MaxTries`.
- **2-minute fixture timeout in the 6 MongoDB classes.** With the waits gone, the 30-second default became the total budget for the app's work, and `RemoteConsoleApplication.Shutdown` kills the process silently on expiry. Matches the CosmosDB and Elasticsearch precedent.
- **New `DropTestDatabase` command, called last in the 5 driver classes.** The per-run Guid database was never dropped, so every run leaked one database on the shared server.
- **Stopped installing IIS-HostableWebCore for the MongoDB unbounded job.** Every MongoDB fixture is a `ConsoleDynamicMethodFixture`, so the feature is unused.

## Waits the retry fix exposed

Making the swallow loud broke 11 test namespacs. None was a flake: each wait was unsatisfiable in that configuration, and the old code hid it. Every wait below is now keyed on evidence the test actually asserts.

- **AwsLambda** Ses/CloudWatch/DynamoDb/Kinesis/KinesisFirehose waited for 2 serverless payloads while asserting `Assert.Single`. Now 1. The four tests that assert 2 are unchanged.
- **Api** `RecordDatastoreSegment` waited for `sql_trace_data`; the RequiredOnly variant passes no commandText, so no SQL trace is emitted. Guarded with `_allOptions`.
- **ReJit** `RejitAddFile` waited for the file-watcher refresh with the watcher disabled. Guarded with `_disableFileSystemWatcher`.
- **AuditLogTests** waited for the agent shutdown line, which cannot appear during exercise. Now waits for an audit received line, and asserts both audit collections are non-empty -- the old ratio also held at `0 == 2 * 0`.
- **Unbounded Oracle** Async/Sync now wait on the asserted `Datastore/all` aggregate call count, MetadataComment on `sql_trace_data`. The shutdown-line wait had been supplying dwell for a periodic harvest; without it the only `metric_data` snapshot was the one forced at process exit, taken before the metrics transform contributed, so every Oracle datastore metric was missing.
- **`OracleExerciser`** is now idempotent: `CreateTable` drops first and swallows ORA-00942, and the procedure uses `CREATE OR REPLACE`. A retry over state left by a killed try was hitting ORA-00955 22 times.

## Docs

`CLAUDE.md` and the `build-dotnet-agent` skill both described `dotnet build` and omitted that `IntegrationTests.sln` and `UnboundedIntegrationTests.sln` need VS MSBuild -- `dotnet build` fails them with `MSB4019`. Both now cross-reference `tests/CLAUDE.md`, which owns the invocation and its flags.

### DIFF
--- a/.claude/skills/build-dotnet-agent/SKILL.md
+++ b/.claude/skills/build-dotnet-agent/SKILL.md
@@ -17,6 +17,7 @@ Debug is default -- never pass `-c Release`. It's slow; run in background or tee
 
 - **Never build a single `.csproj`.** `Core.csproj` and its dependents have a post-build step (ILRepack + `AssemblyModifier.exe`) whose ordering only the solution provides. Building one project alone yields `AssemblyModifier.exe not found`, exit code 3, or `$(SolutionDir)`/`*Undefined*` errors.
 - **Debug, never Release** (Release is for CI/releases).
+- **`IntegrationTests.sln` / `UnboundedIntegrationTests.sln` build with VS MSBuild**, not the `dotnet build` above; the `dotnet` SDK cannot build their legacy ASP.NET FW web apps. Invocation and required flags: [tests/CLAUDE.md](../../../tests/CLAUDE.md#building-the-solution).
 - **`Core.UnitTest` from the CLI needs `SolutionDir`** (VS sets it, `dotnet` doesn't), trailing backslash required. Both forms below pass the same value (`<repo-root>\`):
   ```bash
   # bash (Git Bash)
--- a/.github/workflows/all_solutions.yml
+++ b/.github/workflows/all_solutions.yml
@@ -892,7 +892,7 @@ jobs:
 
       - name: Install HostableWebCore Feature
         if: | # only install for the required namespaces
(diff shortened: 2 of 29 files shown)


nr-ahemsath (approved):
Looks good to me.

PR #3794, opened 2026-09-02
feat: Allow setting the application name per IIS application pool in `newrelic.config`

tippmar-nr:
Adds an optional `appName` attribute to the `applicationPool` element in `newrelic.config`. It resolves as a new step directly below the `application/name` element, so unmapped pools still fall through to their own pool name.

```xml
<configuration xmlns="urn:newrelic-config">
  <!-- the mapping only applies when this element carries no name -->
  <application />

  <applicationPools>
    <applicationPool name="StorefrontPool" instrument="true" appName="Storefront" />
    <applicationPool name="CheckoutPool" instrument="true" appName="Storefront" />
    <applicationPool name="ReportingPool" instrument="true" appName="Reporting" />
    <!-- no appName, so this pool reports under its own name -->
    <applicationPool name="LegacyPool" instrument="true" />
  </applicationPools>
</configuration>
```

`StorefrontPool` and `CheckoutPool` roll up into one application. `instrument` is still required, and `appName` does not change whether a pool is instrumented. Matching is case-insensitive on the pool name, and `appName` is a single name rather than a comma-separated list.

Covered by unit tests for resolution and precedence, and by an integration test that maps a pool through the real config file on .NET Framework and .NET Core. From NR-605854.

### DIFF
--- a/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
+++ b/src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
@@ -340,16 +340,12 @@ public class DefaultConfiguration : IConfiguration
             return true;
         }
 
-        // Resolved before step 8 so that step 8 can report when it shadows a mapping.
-        var appPoolId = GetAppPoolId();
-        var pooledAppName = TryGetApplicationNameForApplicationPool(appPoolId);
-
         // 8. newrelic.config
         if (_localConfiguration.application.name.Count > 0)
         {
-            if (pooledAppName != null)
+            if (HasApplicationPoolMapping())
             {
-                Log.Warn("Application pool '{0}' is mapped to application name '{1}' in newrelic.config, but the application/name element takes precedence. Remove that element to use the application pool mapping.", appPoolId, pooledAppName);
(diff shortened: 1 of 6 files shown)


src/Agent/NewRelic/Agent/Core/Configuration/DefaultConfiguration.cs
```diff
@@ -373,6 +389,45 @@ public bool TryGetApplicationNames(out IEnumerable<string> names)  // CHANGED: w
         return false;
     }
 
+    private bool HasApplicationPoolMapping()
+    {
+        var pools = _localConfiguration.applicationPools?.applicationPool;
+        return pools != null && pools.Any(pool => !string.IsNullOrWhiteSpace(pool.appName));
+    }
+
+    private string TryGetApplicationNameForApplicationPool(string appPoolId)
```
nrcventura: Should we rename this method to something like `TryGetMappedApplicationNameForApplicationPool` so that we don't confuse it with normal application pool based naming (which just uses the name of the app pool)?

tippmar-nr: Good call - renamed to `TryGetMappedApplicationNameForApplicationPool` in 2e4704b.

nrcventura (approved):

PR #3803, opened 2026-09-10
feat: Add Kafka cluster metrics

tippmar-nr:
- Kafka produce and consume metrics are now reported per cluster (when this feature is enabled; cluster metrics are off by default), keyed by the Kafka cluster id. An application that talks to more than one cluster no longer has that traffic merged into a single view.
- Produce and consume spans carry a `kafka.cluster.id` attribute, so a trace shows which cluster handled the message.
- The cluster id is resolved once per set of bootstrap servers on a background interval, so produce and consume calls are not slowed down.

Enable cluster metrics via `newrelic.config` or environment variable:
```xml
<kafka>
  <metrics clusterMetricsEnabled="true" />
</kafka>
```
or
```
NEW_RELIC_KAFKA_METRICS_CLUSTER_METRICS_ENABLED=true
```

From NR-614794.


src/Agent/NewRelic/Agent/Extensions/NewRelic.Agent.Extensions/Helpers/KafkaStatisticsHelper.cs
```diff
@@ -700,4 +701,27 @@ private static string NormalizeNodeId(KafkaBrokerStats broker)
...
+    /// </summary>
+    public static List<KeyValuePair<string, string>> WithStatisticsInterval(IEnumerable existingConfig, int intervalMs)
+    {
+        var result = new List<KeyValuePair<string, string>>();
+
+        if (existingConfig != null)
+        {
+            foreach (var item in existingConfig)
+            {
+                if (item is KeyValuePair<string, string> kvp && kvp.Key != "statistics.interval.ms")
```
chynesNR: This should probably be a constant?

tippmar-nr: @copilot Fix this issue

nrcventura (approved):

PR #3804, opened 2026-09-10
feat: Instrument runtime-async methods

nr-ahemsath:
Adds support for instrumenting .NET 11 **runtime-async** methods — methods the compiler lowers to the runtime's async calling convention instead of a generated state machine. Covers the native profiler (detect and rewrite them correctly) and the managed agent (normalize their results so existing wrappers work unchanged).

Ticket: NR-610232

## Why this can merge before .NET 11 GA

**The change is inert unless runtime-async is explicitly opted into.** Everything is gated on `MethodImplAttributes.Async` (`0x2000`), which the compiler emits only when a project sets `<Features>runtime-async=on</Features>`, and which additionally requires `DOTNET_RuntimeAsync=1` at run time. No application built today carries it. On such a build the chain is:

`Function::IsRuntimeAsync()` false → tracer flag bit 19 never set → `TracerArgument.IsRuntimeAsync` false → `resultNormalizer` null → `isAsync` unchanged → `IsRuntimeAsync` false → no `DetachFromPrimary()`.

So for every currently shipping app the code paths added here are not reachable. That is the main thing worth checking in review.

## How it works

1. **Profiler** detects runtime-async methods and instruments them by substituting the *effective* return type (`MethodRewriter/RuntimeAsyncReturnType.h`), since a runtime-async method's raw return is the unwrapped `T` rather than `Task<T>`.
2. **Profiler** emits a distinct `RuntimeAsyncMethod` tracer flag (bit 19), deliberately **not** reusing `AsyncMethod` (bit 23).
3. **Managed** reads bit 19, and `WrapperService` synthesizes an already-completed `Task`/`Task<T>` from the real result, *then* sets `IsAsync = true`. Making the result-slot promise true before anything downstream reads it is what lets all ~47 existing `GetAsyncDelegateFor` call sites work **with zero wrapper files touched**.
4. `OtherTransactionWrapper` pairs its `AttachToAsync()` with `DetachFromPrimary()` for runtime-async methods. Without this a runtime-async transaction stays in the creating thread's primary (thread-local) storage for the method's whole life, and completing on a different thread strands it there — silently dropping the segments of later continuations that land on that thread.

Open generics use `Task<object>`; the `onComplete` overload gates on `result is T`, so a mismatch fails closed rather than surfacing to customer code.

## Profiler pairing

`Home.csproj` is pinned to `NewRelic.Agent.Internal.Profiler` **10.54.0.35**, built from this branch, so the managed and native halves in this PR are the matching pair. Verified locally: the package's profiler contains the runtime-async code paths, and the integration test passes against it with no local overlay.

## Verification

| Layer | Result |
|---|---|
| `Core.UnitTest` | 8118 passed, 0 failed (local; CI runs it on this PR) |
| Profiler unit tests | 591 / 591 |
| `all_solutions.yml` on this branch | **104 success / 0 failed / 5 skipped** of 109 jobs ([run](https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/34500106903)) |

The 5 skipped are payload-log aggregation (gated off by default) and a Dependabot-only job — no test job was skipped.

**The integration test was confirmed to actually catch the bug it guards**, not merely to pass: with the `DetachFromPrimary()` fix reverted it reports 19 nested segments instead of 20 and 11 root transactions instead of 10. Both assertion families fire.

### About the new integration test

`RuntimeAsyncApplication` is a standalone console app rather than an addition to the shared `MultiFunctionApplication` helpers, because the `<Features>` and `NoWarn` properties it needs cannot be scoped to one file — putting them in the shared helpers would force `DOTNET_RuntimeAsync=1` onto the large share of the suite that depends on them.

It targets **net10.0** with runtime-async as a preview feature, so **no .NET 11 SDK is needed anywhere in CI**. Two easy-to-get-wrong preconditions both fail loudly rather than quietly:

- the fixture sets `DOTNET_RuntimeAsync=1`; without it the app dies at type load, before any assertion;
- the app reflects over its own methods for `0x2000` and fails with an explicit message if the SDK compiled ordinary state-machine async instead — which would otherwise be a silent pass that exercised nothing.

## Deliberately out of scope

- **`ValueTask` support.** State-machine async `ValueTask` methods never end their segments (21 starts, 0 ends), which strands their transaction. This is **pre-existing on main and unrelated to runtime-async**; it is being approached separately rather than folded in here.
- **`thread.id` parity with state-machine async is argued, not measured.** Runtime-async spans are verified to omit `thread.id`, against a *synchronous* positive control. A state-machine async control cannot be written in that app because runtime-async is enabled assembly-wide. That the two async forms agree follows from their sharing the single `!IsAsync` gate in `Segment.cs`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nr-ahemsath:
Pushed bc5aa5b44 to address a review finding on the managed side.

**What was wrong:** `RuntimeAsyncResultNormalizer.TryCreate` did its reflection (`GetMethods` / `ParameterType` / `ReturnType` / `MakeGenericMethod` / `CreateDelegate`) unguarded, and its call site sits *outside* the only `try` in `WrapperService.BeforeWrappedMethod`. Reflecting over a customer type whose member signature references an assembly that can't be loaded throws `TypeLoadException`/`FileNotFoundException` — the classic `GetMethods()` failure with optional dependencies.

**Why it mattered:** the throw escaped to `AgentShim.GetTracer`, which swallows it and returns a null tracer — so `_functionIdToWrapper` was never populated, and every subsequent call to that method repeated the full enumeration *plus* the throw. An unbounded per-call cost on a potentially hot path, with only a Debug log line as evidence.

**Fix:** `TryCreate` is now a thin guard around the reflection, which moved into a private `Classify`. Any exception degrades to `null` — the outcome the caller already handles ("keep synchronous completion semantics") — and is logged at Debug with the exception. Because control now reaches the cache write, the failure is recorded once instead of retried per call.

**Tests:** 3 new, using a `TypeDelegator` subclass that throws from `GetMethods` (a real `Type` failing exactly where a customer type with a missing optional dependency would). Two cover the degrade-to-null path for both exception types; one calls `BeforeWrappedMethod` twice with the same functionId and asserts `GetMethods` ran **once**, pinning the caching behavior against future refactors. Verified red before green. Full `Core.UnitTest` suite green on net10.0 (8121 tests); the new tests pass on net481 as well.

nr-ahemsath:
Pushed c616c781a for a second review finding, this one in the native profiler.

**What was wrong:** `TryGetInstrumentationPoint` returns the `InstrumentationPoint` stored in the configuration's map rather than a copy, and one point serves *every* function that matches it — a parameterless `<exactMethodMatcher methodName="Handle" />` matches all overloads of the name. `DefaultInstrumentor` then OR-ed each function's own tracer flags into that shared object, so they stayed set for every sibling instrumented afterwards. The flags were also applied *above* the runtime-async return-shape guard, so the leak outlived even methods the profiler declined to rewrite.

**Why it mattered:** for `RuntimeAsyncMethod` this is not just a mislabel. Given a runtime-async `async Task<int> Handle(int)` alongside an ordinary `Task<int> Handle(string)`, once the first is JIT'd the second is reported as runtime-async too, so `WrapperService` promotes it to `IsAsync` and hands the wrapper a synthesized completed `Task` in place of its real pending one.

**Fix:** copy the configuration's instrumentation point before applying per-function flags. The synthesized `ShouldTrace` point is already private so it skips the copy; `InstrumentationPoint` already has a deep copy constructor and `InstrumentDefault` only reads through the pointer. This also fixes the same pre-existing leak for `AsyncMethod`. Cost is one copy per JIT/ReJIT of a matched method — not a per-call path.

**Unit tests:** 3 new in `InstrumentorsTest` — the shared point stays unmutated after instrumenting a runtime-async method; it also stays unmutated on the declined path; and, over two overloads on one shared point, the runtime-async method's own flags still reach its emitted `ldc.i4 <args>` while the sibling's IL does not carry `RuntimeAsyncMethod`. All three fail without the fix, reporting `Actual:<524288>` (`1<<19`). Full 7-DLL profiler suite: 594 passed.

**Also verified against a real application, not just unit tests.** I extended `RuntimeAsyncApplication` with exactly the shape above — a runtime-async `Handle(int)` and a deliberately non-`async` `Handle(string)` returning a genuinely pending task, both covered by one parameterless custom-instrumentation matcher, with the runtime-async overload called first. A runtime precondition check asserts the pair really is mixed (one carries `MethodImplAttributes.Async`, the other does not), so the scenario can't silently degenerate. Running that against two local profiler builds:

- **pre-fix profiler:** fails with *"No `HandleOverload` span carried `thread.id`"* — the ordinary overload had inherited the runtime-async flag and was wrongly treated as async. The companion assertion that those spans exist passed, so this is a real wrong-value result, not missing instrumentation.
- **fixed profiler:** passes.

In both runs the `Custom/HandleOverload` metric came in at the full 10 calls, confirming both overloads really were instrumented through the single shared point.

Those test-application changes are **not** in this PR. The pinned `NewRelic.Agent.Internal.Profiler` NuGet is built from `91d5c260a`, which predates this fix, so the assertion would be red in CI until a profiler package is published and the reference bumped. I'll land it after that bump rather than merge a knowingly-failing test.

One caveat on what the end-to-end run proves: it confirms the flag leak and the resulting incorrect `IsAsync`, but not the further "segment ends before the work finishes" consequence. Demonstrating that needs a state-machine-async sibling, and `runtime-async=on` is assembly-wide in that app, so every `async` method there compiles runtime-async. The unit tests cover the mechanism directly.

nr-ahemsath:
Pushed a914956c8 for a third review finding, in the managed normalizer.

**What was wrong:** `BuildParameterTypeNames` claimed to mirror `MethodSignature::ToString` from `SignatureParser/Types.h`, but used `Type.FullName`. Those agree only for primitives, arrays and by-ref types. For anything generic they diverge:

| Parameter | Profiler (`Types.h`) | `Type.FullName` |
|---|---|---|
| `List<int>` | ``…List`1[System.Int32]`` | ``…List`1[[System.Int32, <assembly>, Version=…]]`` |
| method's `T` | `!!0` | `null` |
| type's `T` | `!0` | `null` |

The profiler emits its own rendering into the instrumented IL as `argumentSignature` (`FunctionManipulator.h:228`), so that string is the contract.

**Why it mattered:** an overloaded runtime-async method with a generic-typed parameter matched no candidate, `TryResolveMethod` returned null, and the method was instrumented with **synchronous** completion semantics — `thread.id` recorded on its span, and for an attribute-instrumented one the sync `OtherTransactionWrapper`, which doesn't `AttachToAsync`. Two details made it nasty: only the overload *with* the generic parameter was affected (its plainly-rendered sibling resolved fine, so the same method name behaved two ways), and the sole signal was a `Log.Finest` line, off at default log levels. Non-overloaded methods were never affected — they take the single-candidate early return without comparing strings, which is why this is Low severity.

**Fix, three parts:**

1. A recursive renderer that actually mirrors the `Types.h` subclasses: `ByRefType` `&`, `SzArrayType` `[]`, `VarType` `!N`, `MvarType` `!!N`, and `GenericType` as the definition name plus *unqualified* arguments in brackets. It **refuses rather than approximates** for pointers, bare generic type definitions and null `FullName` — a wrong rendering could coincidentally equal a different candidate's correct one and silently select the wrong overload.

2. Resolution no longer treats "cannot disambiguate" as "no normalization." Only the *result shape* matters, so classification is split into `DescribeResultShape` and `Build`, and when no single signature match exists the candidates' shapes are compared — all agreeing means the ambiguity cannot change the normalizer. Genuine disagreement still returns null rather than guessing. This is the part that makes future drift between the C# renderer and `Types.h` non-fatal instead of silently losing async-ness, and it alone would have fixed the motivating example.

3. Corrected the comment claiming unmatched generic parameters "take the `Task<object>` tier anyway" — true only on the single-candidate path; in the multi-candidate path the result was no normalizer at all.

**Tests:** 10 new covering each rendering and both fallback outcomes. Against the old resolver exactly seven fail — the five generic-rendering cases and the two agreement cases — while the `System.String`, `System.Int32[]` and `System.Int32&` controls pass, which independently confirms only generics diverged. Full `Core.UnitTest`: 8131 passed, 0 failed; normalizer tests green on net481 too.

**One deliberate behavior change to review, not skim past:** I removed `TryCreate_ReturnsNull_WhenTwoCandidatesShareTheSameSignatureString`. Its premise ("refuse rather than pick one arbitrarily") no longer holds — both `AmbiguousGeneric` overloads return `Task<int>` and are now *verified* to normalize identically rather than guessed at. It's replaced by a test asserting the normalizer is produced. That widens when a runtime-async method gets async treatment, on purpose, and it's the one judgement call here worth a second opinion.

Rebased onto a237029b0 (the profiler NuGet bump to v10.54.0.38, built from the finding-#2 fix in c616c781a) — no file overlap. Findings #4 and #5 from the review are still open.

nr-ahemsath:
Pushed 277a3dca8 for the fourth review finding.

**What was wrong:** `Delegates.GetAsyncDelegateFor`'s five-argument overload passed the caller's concrete `Task<TResult>` to `GetDelegateFor<T>`, while its three sibling overloads all pass `Task`. `GetDelegateFor` skips `onSuccess` outright when the result is not of its type argument, and `onFailure` requires an exception — so a result that was some *other* `Task` ran **nothing at all**: the segment was never ended, never removed from the call stack (leaving a stale frame to misparent later segments), and `onComplete` never fired.

The runtime-async normalizer can produce exactly that shape: when a method's declared result type is an open generic it substitutes `Task<object>`, which is not the `Task<TResult>` a wrapper pinning a typed `onComplete` asks for. This corrects the PR description's claim that the fallback "fails closed" — it did the opposite.

**Latent rather than live today.** The only call site of this overload is `HttpClient/SendAsync.cs:111` with a closed `Task<HttpResponseMessage>`. The two other wrappers that name a concrete task type (`WebApi1`/`WebApi2` `InvokeActionAsyncWrapper`) route through the `TaskContinueWithOption` overload, which already gates on `Task`.

**Fix:** gate on `Task` in that overload too. `OnSuccess` only ever needs a `Task` — its parameter is already `Task`, and it narrows to `T` with `as` before invoking `onComplete` — so this required no change to the shared `GetDelegateFor<T>` that fifty-odd wrappers depend on.

I deliberately did **not** take the other available route, having the normalizer return null instead of substituting `Task<object>`. That would trade a latent leak for *guaranteed* synchronous instrumentation of every generic runtime-async method — the same `thread.id`-and-sync-wrapper degradation as finding #3. Worth disagreeing with me if you see it differently, since it's the one judgement call here.

Also raised the open-generic notice in `WrapperService` from `Finest` to `Debug`. It's logged once per `functionId` and is the only signal the substitution happened, so `Finest` (off in practice) made it invisible.

**Tests:** 6 new in `DelegatesTests`. Against the unfixed overload exactly four fail — segment not ended, frame not removed from the call stack, hold/release cycle absent, `onComplete` not fired — while the matching-type and exception-path controls pass, so they're detecting this specific defect rather than any change. Full `NewRelic.Agent.Extensions.Tests`: 1017 passed on net10.0, 1041 on net481. `Core.UnitTest` 8131 passed, `CompositeTests` 622 passed.

**One correction to my own reasoning, since it affects how the bug reads:** my first draft of the code comment said the bug left "the transaction never released." The red run disproved it — `Hold()` was never called either, so nothing was outstanding to release. The committed comment describes the real consequences instead. The stale call-stack frame is the part with the widest blast radius, since it can misparent unrelated later segments.

Only finding #5 is left, and it's a doc-comment correction in `InstrumentedMethodInfo`.

nr-ahemsath:
Pushed c9b03dd0e for the fifth and last review finding. Comment-only.

**What was wrong:** the `IsRuntimeAsync` doc comment said `OtherTransactionWrapper` uses the flag to decide whether to call `DetachFromPrimary()` *"in the after-delegate"*. The call is in `BeforeWrappedMethod` (`OtherTransactionWrapper.cs:84`), immediately after `AttachToAsync()`.

That distinction is the whole correctness argument, not a nitpick: a runtime-async method's after-delegate does not fire until true completion, so deferring the clear to it would leave the transaction sitting in the creating thread's thread-local slot for the method's entire life — exactly the bug the early clear exists to prevent. The comment pointed the next maintainer at the one place the logic cannot live.

**A second inaccuracy the review didn't catch:** the comment named `OtherTransactionWrapper` as the only consumer. `MultithreadedTrackingWrapper` reads the flag too, in `CanWrap` (line 22), to refuse runtime-async methods outright because it calls `AttachToAsync()` with no paired `DetachFromPrimary()` — the guard added in c9a3798a4. Naming one of two readers is presumably how the comment drifted to begin with.

**The corrected comment** names both consumers and the phase each reads in, notes the `newTransactionCreatedByWrapper` gate, and distinguishes this flag from `IsAsync` (which `WrapperService` also sets for these methods once it can restore the Task). It defers to the two wrappers for the full reasoning rather than duplicating the explanation, so there stays one authoritative copy.

**Verification:** `FullAgent.sln` builds with 0 errors, which also confirms the `<see cref="IsAsync"/>` resolves — an unresolved cref is CS1574, an error here under `TreatWarningsAsErrors`. I also checked mechanically that every changed line is a doc comment, so there is no behavior change and no new tests: the coverage rules apply to changed code, and there is nothing to assert here that would not just be testing the compiler.

---

That closes out all five findings from the review of this PR:

| # | Fix | Commit |
|---|---|---|
| 1 | Guard the runtime-async classification reflection; cache the failure | bc5aa5b44 |
| 2 | Stop per-function tracer flags leaking across overloads (profiler) | c616c781a |
| 3 | Resolve overloads by the profiler's own type rendering, plus a result-shape fallback | a914956c8 |
| 4 | End the segment when an async result is not the declared task type | 277a3dca8 |
| 5 | Correct the `IsRuntimeAsync` doc comment | c9b03dd0e |

Two things reviewers should look at rather than skim, both already called out on their own commits: the removed `TryCreate_ReturnsNull_WhenTwoCandidatesShareTheSameSignatureString` test in #3, whose premise no longer holds, and the choice in #4 not to "fail closed" by dropping the `Task<object>` substitution.

Finding #2 was additionally verified end-to-end against a real instrumented application with a mixed runtime-async/ordinary overload pair, red against a pre-fix profiler and green against the fixed one. Those test-application changes are held back deliberately and are not in this PR.

nr-ahemsath:
I responded to PR feedback with commit 6cc64b88750b1ca299e7b9a971a92e0a4fb1592e:

1. Remove references to .NET 11 in favor of generic "runtime async" wording, as it is available in .NET 10 as a preview feature.
2. Replace British spelling in comments with American spelling (e.g. "honor" instead of "honour")
3. Use `alwaysOn` sampler configuration to guarantee that a key test method gets sampled to provide data to test assertions rather than relying on test method ordering.
4. Add Linux testing of runtime async as a container integration test.

src/Agent/NewRelic/Agent/Core/Wrapper/RuntimeAsyncResultNormalizer.cs
```diff
@@ -0,0 +1,387 @@
...
+/// Converts the value a runtime-async method's IL body actually returns into the
+/// already-completed Task the agent's after-delegate machinery expects.
+///
+/// Per the ECMA-335 augment (I.8.4.5) a runtime-async body pushes nothing before `ret` for
+/// Task/ValueTask and an unwrapped T for Task&lt;T&gt;/ValueTask&lt;T&gt;, so the profiler hands
+/// FinishTracer a null or a boxed T where every wrapper expects a Task. Restoring the Task here --
+/// once, at the single choke point in WrapperService -- is what lets existing GetAsyncDelegateFor
+/// call sites keep working unchanged.
+/// </summary>
+public static class RuntimeAsyncResultNormalizer
```
nrcventura, line 28: I ran the following idea past claude and I think it could significantly simplify the code and remove many edge cases.

1. Update the injected bytecode for the instrumented method to pass the `Type` of the allocated `local` value (plus possibly some other flag indicating an open generic type) to the AgentShim method call so that instrumentation can have a type instance to use to create a completed Task<T> or Task instance from.
2. Simplify the RuntimeAsyncResultNormalizer to use the Type parameter passed in the agentshim call so that it doesn't have to enumerate methods on class being instrumented to try to find the matching method overload.

nr-ahemsath: Good idea, done.

nr-ahemsath:
### Profiler now hands the effective return type to managed code

Commits `b6ee0c5`, `6cdb57b`, `a9d640b`. Net −533 lines.

**Profiler.** `CallGetTracer`'s argument array grows a 12th slot carrying the effective return type as a `System.Type`, emitted through the existing `InstructionSet::AppendTypeOfArgument`. Gated on `IFunction::IsRuntimeAsync()`; every other method pushes `ldnull`. The gate isn't only about per-call cost — `GetTypeTokenForReturnType` throws for by-ref returns and `GetTypeTokenForType` for pointers, so tokenizing unconditionally would newly fail instrumentation for methods that work today.

**Managed.** Threaded through `AgentShim` → `IAgentManager` → `WrapperService`. `RuntimeAsyncResultNormalizer` goes from 387 lines to ~100: `FindCandidates`, `DescribeResultShape`, `ShapesAgree`, and the block of `TryAppendTypeName` that mirrored `MethodSignature::ToString` just to match a signature string are gone, along with `RuntimeAsyncNormalization` and `IsExactlyTyped`.

**Generic methods are now exactly typed.** `ldtoken` on a `VAR`/`MVAR` `TypeSpec` resolves in the method's generic context, so the concrete closed type arrives and the `Task<object>` substitution is gone. Consequence worth noting: a `functionId` is per method *definition*, so normalizers are cached by result type and resolved per call rather than stored in the `functionId` cache entry — otherwise the first instantiation's delegate would be reused and later ones would silently yield `default(T)`.

**Verification.** Profiler 597/597; Core.UnitTest 8108/8108 (net10.0) and 8153/8153 (net481); CompositeTests 622/622; Extensions 1017/1041; PublicApiChangeTests green. End-to-end against a locally built profiler on the .NET 11 runtime-async repro: no `InvalidProgramException`, 189/189 segments ended, nested segments 42/42, and the open-generic fallback log line dropped from 2 occurrences to 0.

**Two things for review attention:**

1. The open-generic case is refused by an explicit `ContainsGenericParameters` check rather than by letting the bind fail, because the targets disagree — .NET Framework binds an open generic and returns a delegate over a still-open method, while .NET 10 throws. Only the net481 test leg caught this.
2. `GetFinishTracerDelegateParameterWrapper` reads the new slot only when `parameters.Length > 11`. That guard is transitional — it kept the default call path working before the profiler reference was bumped — and its comment states the condition for removing it.

nrcventura (approved):

PR #3813, opened 2026-09-14
test: fix integration test and harness defects exposed by a Linux test lane

tippmar-nr:
Nine test-harness and test defects, split out of a larger branch that partitions the integration suites into a Windows-Framework lane and a Linux-Core lane. No product code changes, and no test is skipped.

**This PR is independent and targets `main`.** It can be reviewed and merged on its own merits whatever happens to the partition effort. Three further PRs build on this branch (#3814 traits, #3815 lane resolver and guards, #3816 the CI lane partition), so their changes cannot reach `main` before this one, but none of them is needed for this PR to stand up. **When merging, please delete the branch** -- that triggers GitHub to retarget #3814 onto `main` automatically.

Most of these defects are invisible on Windows. A case-insensitive filesystem hides a wrong directory case, and a Windows-only synchronization primitive only throws once the same application runs elsewhere. Two are plain harness bugs that have been failing runs on Windows all along: the Azure Function agent-log selection, which was the root cause of five failed CI runs, and the xunit output-helper crash after a test ends.

### Highlights

- **Bounded process exit waits**, replacing an unbounded `WaitForExit()`, and `Kill(entireProcessTree: true)` so a killed host leaves no orphaned children to compete with the next test.
- **The Azure Function fixture now names the agent log it reads.** `AgentLogFileName` became `protected virtual` and the fixture overrides it. An Azure Function test produces two agent logs, because the profiler attaches to `func` as well as to the worker, and `func` outlives the worker by about 300 ms, so the assertions were reading a log that never held the worker's transactions.
- **`Path.Combine`** in place of a hardcoded backslash separator across the MsSql, MySql, Oracle and Postgres tests.
- **Two Otlp fixtures asked for the wrong directory case**, `OTelMetricsApplication` against a tracked `OtelMetricsApplication`. NTFS resolved it either way, so `dotnet publish` only failed on Linux, with `MSB1009`. This is a class of bug rather than one instance: any fixture string literal naming an application directory is unchecked by the compiler and silently correct on NTFS.
- **The two Otlp applications used a named `EventWaitHandle`**, which is Windows-only on .NET, so they threw `PlatformNotSupportedException` at startup on Linux and every test in their fixture failed before reaching an assertion. They now use `AppLifecycleManager`, which already switches to a named pipe where named handles are unavailable. A sweep confirmed these two were the only applications that both hand-rolled the handle and target a .NET Core framework.
- **`AppLifecycleManager` gained an arm-then-wait form**, and this is the part most worth a reviewer's attention. The harness signals shutdown with `EventWaitHandle.OpenExisting`, which throws when the channel does not exist, and its catch block then kills the process tree; a killed application never writes the agent's graceful shutdown line, so any test waiting for that line times out. The contract, stated at `RemoteApplication.cs:460`, is that **the application creates the channel and the runner only opens it**. An application that does work before waiting must therefore arm the channel at startup, not at the moment it waits. `ArmTestCompletion(port)` returns a `TestCompletionSignal` with a `Wait()`; the two Otlp applications arm before writing their pid file and wait at the end. `WaitForTestCompletion` keeps its signature and behaviour as arm-then-wait, so the 16 applications that start a server and wait almost immediately are unchanged. This also closes the same window on Linux, where the pipe server was likewise created only at wait time while the runner's connect gives up after one second.
- **The Lambda event tests** asserted attribute values the agent renders with the machine's culture and timezone. They assert presence now.

### Note for anyone touching a shutdown path later

The 16 existing callers of `WaitForTestCompletion` are misleading precedent. They are web applications that start a server and wait at once, so their channel always exists before a signal can arrive, which makes creating it at wait time look safe. It is only safe when nothing happens between startup and the wait.

### Verification

Full CI is green on this branch, including the `OpenTelemetry` shard on Windows. The changed shards are `AwsLambda`, `Configuration`, `OpenTelemetry`, and the unbounded `MsSql`, `MySql`, `Postgres`, `Oracle` and `AzureServiceBus` namespaces.


tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/AzureFuncTool.cs
```diff
@@ -110,8 +110,10 @@ protected override void WaitForProcessToStartListening(bool _)
         {
             try
             {
-                //We need to attempt to clean up the process that did not successfully start.
-                RemoteProcess.Kill();
+                // The npm-installed func on Linux is a node shim that spawns the real
```
nrcventura: I thought that we are using a dotnet tool and not an npm-installed func?

tippmar-nr: Good catch, the comment was stale and wrong. Nothing installs func via npm: Windows uses chocolatey (`choco install azure-functions-core-tools`), and the Linux lane in the follow-on partition PR uses a sha256-verified native zip, specifically because the npm package was a node shim. Corrected in 3912b1ca2 to state only what justifies the tree kill, which holds on both platforms: func spawns a host which spawns the worker, so killing only func leaves them holding the redirected output pipe. The tree kill itself is warranted on Windows regardless of install method, from the AzureFunction ServiceBus failure where an orphaned func.exe child under plain Kill() caused shared-queue message theft.

nrcventura (approved):

tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/AzureFuncTool.cs
```diff
@@ -157,7 +157,7 @@ public override void Shutdown(bool _)
         try
         {
             if (RemoteProcess is { HasExited: false })
-                RemoteProcess.Kill();
+                RemoteProcess.Kill(entireProcessTree: true);
```
chynesNR, line 160: We're fine here, but it's worth remembering that `Process.Kill(entireProcessTree)` isn't supported in .NET Standard or Framework (That's bitten me in the past; it's a pretty handy overload).

tippmar-nr: This code gets executed by the test runner, which targets .NET 10, so it should never be an issue.

PR #3815, opened 2026-09-14
test: add the Platform trait guards and the lane resolver

tippmar-nr:
Adds the resolver behind the `Platform` trait and the guards that keep the trait correct as the suites change. The trait lines themselves are already on `main` from #3814, generated once by a tool that is not retained in the repo; from here they are maintained by hand, one attribute per class, and the guards are what keep them honest.

Seven files, no code generation and no regex. `RuntimeLaneResolver` derives a test class's runtime lane from the fixture it uses, which is a verifiable fact about the fixture. `PlatformTrait.RequiresWindows(RuntimeLane)` maps that fact to the platform requirement, and it is the single place where one becomes the other. The per-suite `RuntimeTraitPolicy` carries the exempt list and the overrides.

### The guards

One `PlatformTraitTests` per suite, four facts each, all by reflection over the built assembly:

- **`NoPortableClassDeclaresThePlatformTrait`** is the load-bearing one. A portable class wrongly carrying the trait runs on no lane at all and CI still reports green, because the Linux and Windows-Core lanes exclude the trait while the Windows-Framework lane skips any shard listed in `linux_only`. That is the one failure mode a test run cannot surface, since there is no failing test to see.
- **`EveryTestClassIsClassifiableByTheResolver`** fails the build when the resolver cannot classify a class, which is what keeps a hand-added trait checkable at all.
- **`EveryDeclaredPlatformTraitValueIsRecognised`** fails on any value other than `WindowsOnly`.
- **`EveryExemptionNamesAClassThatStillExists`** keeps the exempt list from rotting.

`RuntimeLaneResolverTests` covers the resolver itself.

Because `TraitAttribute` is inherited and `-trait-` is absolute, an inherited `WindowsOnly` cannot be cancelled further down the hierarchy. A base class shared by portable and Windows-only derived classes therefore carries no trait, and `NoPortableClassDeclaresThePlatformTrait` is what catches a violation.

A class that needs Windows but carries no trait is left to the test run: it lands on the Linux lane, its fixture cannot start, and the lane fails.

### Verified, not assumed

- Both test solutions build in Debug with 0 errors.
- Guards pass: integration 18 tests, unbounded 4 tests, `Failed: 0, Errors: 0, Skipped: 0` in both.
- **`NoPortableClassDeclaresThePlatformTrait` was demonstrated to fail**, by adding one trait line to the portable class `AgentFeatures.InstrumentationLoaderTestsCore` and confirming the guard failed and named that class exactly, then reverting and confirming it passed again. A guard nobody has watched fail is not evidence of anything.

### One known gap, deliberately left open

These guards reflect over class attributes, so they verify what a class *declares*, not what the runner's trait filter actually *matches*. Those can differ. `build/Scripts/check-trait-filter.sh` in the PR above narrows the gap at assembly level by proving the filter is live before a lane runs, but not per shard.


nrcventura:
Is the logic and new test in this PR necessary? If the problem we're trying to prevent is running windows only tests on linux, we still need to wait for a test run to determine if we didn't flag a test correctly. If we didn't have this additional test, a windows only test that we try to run on linux will also just fail. The only scenario that isn't covered is a linux compatible test being flagged as windows only, but I don't think that is a big concern.

tippmar-nr:
@nrcventura --
> Is the logic and new test in this PR necessary? If the problem we're trying to prevent is running windows only tests on linux, we still need to wait for a test run to determine if we didn't flag a test correctly. If we didn't have this additional test, a windows only test that we try to run on linux will also just fail. The only scenario that isn't covered is a linux compatible test being flagged as windows only, but I don't think that is a big concern.

You're right about one of them. `EveryWindowsOnlyClassDeclaresThePlatformTrait` adds no detection: a Framework class that forgets its trait lands on the Linux lane, its fixture can't start, and the lane fails. It only buys a build failure naming the class instead of a lane failure that looks like a product bug.

The reverse case is different, and it comes down to how `test_selection.yml` builds the three lanes from one canonical shard list:

| Lane | Shards it runs | Trait filter |
| --- | --- | --- |
| Linux-Core | all, minus `windows_only` | excludes `Platform=WindowsOnly` |
| Windows-Framework | all, minus `linux_only` | includes `Platform=WindowsOnly` |
| Windows-Core smoke | only the `core_smoke` allowlist (5 shards) | excludes `Platform=WindowsOnly` |

So each lane gates twice: once on shard membership, once on the trait. `linux_only` means "this shard has no Framework tests", and it's subtracted from the Windows-Framework lane so that lane doesn't match zero tests and trip the zero-total check. It currently holds 17 integration shards and 2 unbounded ones -- the eleven `AwsLambda.*`, plus `Blazor`, `Grpc`, `Logging.AuditLog`, `ReJit.NetCore`, `RequestHeadersCapture.AspNetCore`, `RabbitMq` and `TestInfrastructure`.

Now flag a portable class in one of those shards as `WindowsOnly` by mistake. Linux-Core has the shard but the trait filter drops the class. Windows-Framework doesn't have the shard. The smoke lane doesn't have the shard either, and excludes the trait anyway. The class runs on no lane and CI is green. Those shards hold 165 test classes, 155 of them portable, so that's the exposed surface.

That's why `NoPortableClassDeclaresThePlatformTrait` is the one I'd keep: it's the only failure mode a test run can't find, because there's no failing test to see. It's also easy to hit by accident, since `TraitAttribute` is `Inherited=True` and `-trait-` is absolute -- one trait on a shared base class flags every portable class under it, and a derived class can't cancel it.

`EveryTestClassIsClassifiableByTheResolver` is the same shape: it fails the build when the resolver can't classify a class, which is what keeps the generated trait lines in #3814 trustworthy after someone changes a fixture.

`NoLegacyRuntimeTraitRemains` is transitional and only useful until this stack merges. Happy to drop that one as a follow-up.

PR #3826, opened 2026-09-17
feat: continuous profiling native sampling engine (1/6)

jaffinito:
## Summary
Native C++ engine that does the actual CPU/thread sampling for continuous profiling. Runs inside the profiler's stop-the-world suspend, walks a per-tick window of managed threads' stacks (round-robin over a fixed capacity when live thread count exceeds it — see below), and writes samples into a bounded ring buffer for the managed side to drain.

- `ContinuousProfiler.h` — the sampling loop: suspend all threads (CoreCLR `ICorProfilerInfo10`), walk the tick's thread window, resolve frame names (cached in `namecache.h`), encode into `SampleBufferWriter`, resume. `PlanCaptureWindow` computes that window: when live thread count exceeds the fixed per-tick capacity (100 slots), each tick visits a rotating slice rather than the same subset every time, so every thread is still covered over `ceil(threadCount/capacity)` ticks — but each sample still carries the full configured period, so per-thread/aggregate CPU totals are proportionally under-counted on hosts well over capacity. Native logs an overflow counter each tick so this is observable.
- `AgentWorkMap.h` / `TraceContextMap.h` — per-thread bookkeeping the profiler reads without taking a lock on the sampled thread (agent-work depth, current trace/span id for correlation). `AgentWorkMap` slots are tombstoned on `ThreadDestroyed` (a dead thread's identity can be recycled by the CLR) and the whole table's generation is bumped on `Start`, so a fresh session never inherits stale depth counts from a prior one.
- `PendingPushMap.h` / `RetiredSpanRegistry.h` — close two correlation edge cases: `RetiredSpanRegistry` records span ids whose owning transaction has already ended so the sampler won't link a live sample to a dead span (an idle thread could otherwise link to a transaction that finished up to 272s earlier); it's reclaimed only once `CompactAgainst` proves no live correlation slot still references the value, never on a size/time budget. `PendingPushMap` narrows (does not fully close) a related race for a thread that's about to push a new span but isn't yet visible in `TraceContextMap`.
- `SampleBufferQueue.h` / `SampleBufferWriter.h` — the fixed-capacity capture buffer and its wire encoding (`BatchVersion` must stay in lockstep with the managed-side `BufferParser.cs` in slice 3 — no shared constant enforces this, just a comment on both sides).
- `SuspendMutex.h` — serializes stop-the-world between this and the legacy Thread Profiler so the two never suspend concurrently.
- `ThreadDescriptionResolver.h` — OS thread name/id resolution for sample attribution.

Native-only; the C++ unit test suite (`ContinuousProfilerTest/*`) is included in this slice, along with the profiler's build/CI/Docker wiring. Builds and tests green standalone.

## What to look at closely
- **Backpressure/truncation diagnostics exist but aren't proven under real load.** When the capture buffer fills mid-batch, `EncodeAndPublish` truncates the remaining threads for that tick, logs a one-time `Warn` ("THIS WILL RESULT IN LOSS OF PROFILING DATA"), and increments `TruncatedThreads`/`DroppedTicks` counters. No native test currently forces the buffer full and asserts these counters accumulate/reset correctly across ticks — what's tested is the wire format and the managed-side surfacing of whatever value it's handed, not that native produces the right value under real backpressure. Not a blocker, but a known open test gap if you're looking for one.
- **The round-robin capacity model (`PlanCaptureWindow`) trades sampling frequency for coverage on high-thread-count apps.** It replaced an earlier "every thread every tick" design that had no bound on suspend-window cost as thread count grew. The accepted tradeoff is: no thread is ever permanently skipped, but on a process with more live threads than the 100-slot capacity, each thread's effective sampling cadence degrades proportionally, and the emitted profile has no way to signal per-sample "this thread was actually sampled less often" — every sample still carries the full configured period. Worth knowing before extrapolating absolute CPU totals from a profile on a large host.
- **.NET Framework is intentionally out of scope for CP.** Some native code paths (e.g. per-thread `Thread*` lifetime in `GetThreadInfo`) would be a real issue on Framework, but `ContinuousProfilingServiceFactory` (slice 3) gates CP off entirely there — this native code is unreachable on the shipped target. Don't be surprised the FW path isn't hardened; it's deliberately inert, not untested by oversight.
- **`SuspendMutex` ordering**: lock order is always `_mtx_lifecycle` → `SuspendMutex`, never reversed, across both this profiler and the Thread Profiler — verified no deadlock potential across multiple independent review passes. If you touch either profiler's start/stop path, re-verify this still holds.

## Test plan
- [ ] `Profiler.sln` builds Debug, x86 + x64, Windows
- [ ] `build.ps1 -Platform linux -Configuration Debug` builds via Docker
- [ ] `ContinuousProfilerTest`/`SignatureParserTest` native test suites pass, including `ContinuousProfilerSuspendResumeTest`


nrcventura (approved):

tippmar-nr:
### Code review

Found 1 issue:

1. The new `catch (...)` in `ProfilerThreadStart` calls `LogError` while the runtime is still suspended: `SuspendRuntime()` has already succeeded and `ResumeRuntime()` only runs afterwards. The same file already states the rule at [ThreadProfiler.h#L756-L758](https://github.com/newrelic/newrelic-dotnet-agent/blob/46cdc9d18bc4f62ec69adffaac15423abcdd2937/src/Agent/NewRelic/Profiler/ThreadProfiler/ThreadProfiler.h#L756-L758) ("do not log here because of deadlock (the suspend thread issue)"), and explains the mechanism at [L627-L631](https://github.com/newrelic/newrelic-dotnet-agent/blob/46cdc9d18bc4f62ec69adffaac15423abcdd2937/src/Agent/NewRelic/Profiler/ThreadProfiler/ThreadProfiler.h#L627-L631): a suspended thread can hold the loader or heap lock that the logging thread then tries to take. `Logging/Logger.h` takes a `std::mutex` and builds a `std::wostringstream` per call, so both the lock and the allocation are in play. This PR's own `ContinuousProfiler::CaptureAllThreads` handles the identical case correctly by setting a flag inside the suspend window and logging only after `ResumeRuntime()` returns. Suggest the same flag-and-defer pattern here.

https://github.com/newrelic/newrelic-dotnet-agent/blob/46cdc9d18bc4f62ec69adffaac15423abcdd2937/src/Agent/NewRelic/Profiler/ThreadProfiler/ThreadProfiler.h#L669-L680

nr-ahemsath:
One very minor bikeshedding comment: CLAUDE.md in our repo's root says that we prefer to use compact namespaces in the profiler (e.g. `namespace Foo::Bar::Baz {` instead of `namespace Foo { namespace Bar { namespace Baz {`), and none of the new source files in this PR are compliant with that.  However, none of the pre-existing profiler source files are compliant with that either, and I feel like it would be easiest to change the Claude style guide statement than touch every single profiler source file just to make an arbitrary change (I never even would have thought about it if Claude's code review process hadn't flagged it).

nr-ahemsath (approved):

tippmar-nr (approved):

PR #3827, opened 2026-09-17
feat: thread-profiler, transaction, and wrapper integration for continuous profiling (2/6)

jaffinito:
## Summary
Changes to *existing*, always-on agent infrastructure that continuous profiling needed in order to coexist safely with the legacy Thread Profiler and the transaction/wrapper hot path. None of this is CP-gated — it runs for every customer, whether or not they ever enable CP.

- `ProfilingMutualExclusionGate.cs` — new process-wide lock so the Thread Profiler and CP can never both be stop-the-world-suspending at once. Both real call sites hold the gate across their native-start call, not just the check-then-start, so there's no TOCTOU window.
- `Transaction.cs` / `TransactionFinalizer.cs` / `TransactionService.cs` — push/reset the current trace context into CP's native side on transaction start/end, and retire a terminating transaction's ended spans into CP's retired-span registry so idle threads stop linking samples to spans whose transaction already finished (the registry itself stays bounded via `RetiredSpanRegistry::CompactAgainst`, not via the retire action).
- `WrapperService.cs` — pushes the CP trace context on wrapped-method entry/exit, gated behind `ContinuousProfilingContext.AnyEnabled` so it's a single volatile read + not-taken branch when CP is off.
- `Scheduler.cs` — tags CP's own background timers (drain, retune) as agent work so they don't get mistaken for instrumented app work.
- `ThreadProfilingService.cs` / `ThreadProfilingSampler.cs` — two behavior changes that affect **all** Thread Profiler users, not just CP adopters (see below).

**Does not build standalone** — references `ContinuousProfilingContext`/`IContinuousProfilingSessionControl` (slice 3). Expected; this whole feature is a circular compile unit across slices 2-5, it only builds green once all of them are merged.

## What to look at closely
- **Two real behavior changes for non-CP customers, both intentional bugfixes, called out so they don't look like scope creep:**
  - `ThreadProfilingSampler.Stop()` — the old guard made an explicit `stop_profiler` command a no-op whenever a worker was actually running (the code checked the opposite of what the comment claimed). Now an operator-issued stop actually stops the session. This changes legacy Thread Profiler behavior for every customer, CP or not.
  - `ThreadProfilingService.Stop()` on agent shutdown now passes `reportData: false`, discarding an in-flight Thread Profiler session's already-collected data instead of POSTing it — deliberate, to avoid a synchronous collector POST stalling CLR exit. This is a real, accepted data-loss tradeoff (bounded to one in-flight session, at shutdown only) that the team has explicitly decided to ship as-is; it's tested (`ThreadProfilingServiceTests.cs`) and documented in-code. Flagging it here because it's the kind of thing a reviewer should knowingly accept, not discover later.
- **`Transaction.cs`'s `ResetTraceContext()` call is unconditional**, unlike `WrapperService`'s `AnyEnabled`-gated hot-path pushes. Negligible given transaction-end frequency, but worth a second look if you're auditing for CP-when-disabled overhead.
- **`Scheduler.cs`'s agent-work tagging is "correct only because nothing throws" today** — `SetAgentWork()` runs before the `try` whose `finally` calls `ResetAgentWork()`. Safe now because the current `IContinuousProfilingContext` implementation swallows its own exceptions, but it's a latent fragility if that ever changes.


tippmar-nr:
### Code review

Found 1 issue:

1. `StopThreadProfilingSession` returns `true` unconditionally, so the new join-timeout signal never reaches its caller. This PR changed `IThreadProfilingSampler.Stop()` from `void` to `bool` so callers can tell whether the bounded 5-second `Join` completed, and the method correctly branches on it to skip resetting `_profileSessionId` and the cache on timeout -- but the `return true;` at line 307 sits outside both branches. The one consumer, [`StopThreadProfilerCommand` line 43](https://github.com/newrelic/newrelic-dotnet-agent/blob/1bbcb6e266e279305ddf035c4b2609f01ac00eaa/src/Agent/NewRelic/Agent/Core/Commands/StopThreadProfilerCommand.cs#L42-L44), branches on the return value and will therefore always report the `stop_profiler` command as successful, even on a real timeout where the worker is still running and profile state was deliberately left dirty. `return workerStopped;` wires the new signal through.

https://github.com/newrelic/newrelic-dotnet-agent/blob/1bbcb6e266e279305ddf035c4b2609f01ac00eaa/src/Agent/NewRelic/Agent/Core/ThreadProfiling/ThreadProfilingService.cs#L296-L308

src/Agent/NewRelic/Agent/Core/Segments/Segment.cs
```diff
@@ -213,7 +213,41 @@ public string SpanId
...
     }
 
-    public string TryGetActivityTraceId() => _activity?.TraceId;
+    /// <summary>
+    /// Returns this segment's span id only if it has already been materialized -- generated on a first
+    /// <see cref="SpanId"/> read, or explicitly assigned. Returns null otherwise and NEVER generates one.
+    ///
+    /// <para>Exists so continuous profiling can retire a finished transaction's spans without minting ids
+    /// that were never used: <see cref="SpanId"/>'s getter is a lazy generator, and for an unsampled
+    /// transaction (or one with span events disabled) no segment's id is ever generated today. An id that
```
nr-ahemsath, line 222: This seems to be contradicted by line 296 in WrapperService.cs, inside `PushContinuousProfilingContext`:
`var spanId = (currentSegment != null && currentSegment.IsValid) ? currentSegment.SpanId : null;`

src/Agent/NewRelic/Agent/Core/Transactions/TransactionFinalizer.cs
```diff
@@ -44,12 +47,111 @@ public bool Finish(IInternalTransaction transaction)
...
+    /// transaction, and a stall would stall every finalizer in the process. The <c>Segments</c> walk below
+    /// does take one lock -- <c>ConcurrentList&lt;T&gt;</c>'s <c>ReaderWriterLockSlim</c> READ lock -- and
+    /// that is deliberately accepted, not an oversight: the identical lock is already taken on this same
+    /// finalizer path by <c>ConvertToImmutableTransaction</c> (<c>ImmutableTransaction</c>'s ctor copies
+    /// <c>Segments</c>), and a finalizer runs with the world RESUMED, so no writer can be suspended while
+    /// holding it. The rule being stated is about the CP-suspend-window reader, which never runs here.</para>
+    ///
+    /// <para>Deliberately probes for ALREADY-MATERIALIZED span ids rather than reading
+    /// <c>Segment.SpanId</c>: that getter is a lazy generator, so reading it would mint ids for spans that
+    /// are never reported (an unsampled transaction generates none today). An id that was never
```
nr-ahemsath, line 79: Same thing here: contraindicated by
src/Agent/NewRelic/Agent/Core/Wrapper/WrapperService.cs:296, inside PushContinuousProfilingContext — var spanId = (currentSegment != null && currentSegment.IsValid) ? currentSegment.SpanId : null;

PR #3841, opened 2026-09-18
test: run integration and unbounded tests on Linux

tippmar-nr:
Adds one Linux job per namespace to `integration_tests.yml` and `unbounded_tests.yml`. Each job runs the suite assembly with `-trait- Platform=WindowsOnly`, which excludes tests marked `WindowsOnly`. The Windows jobs are unchanged: both workflow diffs are additions only, with zero removed lines.

A discovery step runs `-list classes` before the tests. It skips the rest of the job when a namespace has no Linux-eligible class, and fails the job when the trait is missing from the assembly or the exclusion filter matches nothing. There is no list of which namespaces run where -- the `[Trait("Platform", "WindowsOnly")]` markings added in #3814 are the only declaration.

Also in this PR:

- Bumps the oldest-tested log4net on `net8.0` in `MultiFunctionApplicationHelpers.csproj` from 2.0.10 to 2.0.12. 2.0.10 carries LOG4NET-652: rendering a logging event reads `LoggingEvent.UserName`, which calls `WindowsIdentity.GetCurrent()` and throws `PlatformNotSupportedException` on Linux. That failed five tests across `Logging.LocalDecoration` and `Logging.ContextData` in the new Linux jobs. 2.0.12 is the first release with the fix, and it is still older than the oldest version the Windows jobs exercise, so oldest-version coverage is preserved.
- Builds the test `.csproj` directly on Linux and globs the test apps, so no Linux solution file is needed.
- Selects three namespaces that never ran: `DistributedTracing.W3CInstrumentationTests`, `HttpClientInstrumentation.NetFramework`, and `RabbitMq.Legacy`. The `-namespace` filter matches exactly, not by prefix, so a sub-namespace needs its own entry.
- Keeps host test evidence when a run fails. Inside a composite action `failure()` sees only that action's own steps, so the archive step gated on the caller's step outcome instead.


.github/actions/build-test-solution-linux/action.yml
```diff
@@ -0,0 +1,118 @@
...
+        name: ${{ inputs.artifact-name }}
+        path: |
+          test.runsettings
+          tests/Agent/IntegrationTests/**/bin/**/*
+          !tests/Agent/IntegrationTests/**/obj/**/*
+          !tests/Agent/IntegrationTests/**/bin/**/net4*/**
+        # An isolated function app publishes a .azurefunctions folder, and this
+        # action drops dot-prefixed entries by default. Without it func starts
+        # and then reports no job functions.
+        include-hidden-files: true
```
nrcventura, line 117: Is this going to accidentally capture secrets?

.github/actions/report-host-tests/action.yml
```diff
@@ -0,0 +1,244 @@
...
+    description: 'License key the flaky-retry telemetry reports with.'
+    required: true
+  nr-ci-account-id:
+    description: 'Account the flaky-retry telemetry reports to.'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Report flaky retry telemetry (Windows)
```
nrcventura: We may want to see if there is a way to use the same script/tool on both platforms so that the logic can be in one place. That doesn't need to be part of this PR.

tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/AzureFuncTool.cs
```diff
@@ -35,7 +35,13 @@ public override void CopyToRemote()
 
         // copy local.settings.json from the AzureFunctionApplication project to destination app folder
         var deployPath = Path.Combine(DestinationRootDirectoryPath, ApplicationDirectoryName, "local.settings.json");
-        var localSettingsPath = Path.Combine(SourceApplicationDirectoryPath, "local.settings.json");
+
+        // The file is CopyToPublishDirectory=Never, so it exists only in build output.
```
nrcventura: Should we just update that settings file so that it copies on publish? Or will that cause any other problems if we try to deploy the test func app to azure?

.github/actions/build-test-solution-linux/action.yml
tippmar-nr: No risk. This job never holds any secrets, only the run jobs.

.github/actions/report-host-tests/action.yml
tippmar-nr: I'll go ahead and address it now so we don't have to come back.

tippmar-nr: Actually, after review, there was a different fix: There were a number of (unreachable) `runner.os='Windows'` gated steps that shouldn't have been in the workflow. Cleared those out but will defer unification across integration/unbounded/container workflows until later.

tests/Agent/IntegrationTests/IntegrationTestHelpers/RemoteServiceFixtures/AzureFuncTool.cs
tippmar-nr: Yes -- done. Both `local.settings.json` items now use `CopyToPublishDirectory=PreserveNewest`, and the `CopyToRemote` override in `AzureFuncTool` is gone entirely: the publish output carries the file, so the fixture no longer has to find it in build output or guess at the RID subdirectory.

nrcventura (approved):

